PR #5, opened 2020-02-21
Add a GitHub Actions workflow for PR and master

jjtang1985:
The pipeline is triggered when:
- pushing to master
- opening/reopening/synchronizing pull request

For the 2 scenarios mentioned, the pipeline will then run the stages like below:

#### master build:
1. test stage
1. check stage
1. release canary stage

#### pr build:
1. test stage
1. check stage


.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,36 @@
+name: CI/CD Pipeline
```
marikaner: Can we call this `build`. As far as I can see this is a little bit like a convention and will make a nice badge 😊

marikaner: And maybe also call the file build.yml

PR #9, opened 2020-02-27
Specify the package version of the generated VDMs from the generator CLI parameters.

jjtang1985:
Currently, a VDM package that contains a `package.json` with the same version as the `generator`. However, as a user, if the generated VDMs are the same, I don't really care whether I'm using `generator` `v1.0.0` or `v3.0.0`.

Also, without this change, it will force the `generator` and the `VDM` repo keep the same version forever.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ blog:
 
 ## improvements
 
-- The VDM package version can be specified now from the parameter of the generator CLI, while the generator version is used before.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -25,7 +25,7 @@ blog:
 
 ## improvements
 
--
+- The VDM package version can be specified now from the parameter of the generator CLI, while the generator version is used before.
```
mr-flannery:
```suggestion
- When generating OData client code using the `generator` package, the version that is put in the generated `package.json` can now be supplied as argument (`versionInPackageJson`). Call `generate-odata-client help` for more details.
```

PR #21, opened 2020-03-04
Fix: Wrong URL for Complex Types

FrankEssenberger:
### Context

When calling the `fieldPath()` of the Complex extensions of the primitive properties e.g. `ComplexTypeStringPropertyField` then method did not return the correct path to the field. If the was something like:
```
Entitiy
   |--ComplexProperty(SomeComplexType)
        |--SomeProperty(ComplexTypeStringPropertyField)
        |--ComplexPropertyNested(SomeOtherComplexType)
               |--SomeNestedProperty(ComplexTypeStringPropertyField)
        |--SomeNextProperty(ComplexTypeBooleanPropertyField)
        ....
   |--SomeOtherProperty(StringType)
  ....
```
The fieldPath() on `SomeProperty` did not return `ComplexProperty/SomeProperty` but `SomeComplexType/SomeProperty`. So the type was used in the path and not the property name. In addition for Nested complex type the implementation was not working at all. So if you call findPath() on `SomeNestedProperty` the expected result is `ComplexProperty/ComplexPropertyNested/SomeNestedProperty` which should work also now.

### Changes

The string based creation in `fieldTypeClass` has been reduced so that only one `this` pointer of the parent field is passed. The base constructor `ComplexTypeField` which is the abstract base class for complex types is adjusted to handle two cases:
- On the root level the constructor passes the entity
- In the nested case the constructor takes the parent field.

The second big change is the following: The type name is removed from the signatures of the complex fields not holding data. It seemed like redundant information since the type is also given by the constructor. For the properties actually holding data the type remains since a `ComplexStringProperty` can related to a `Edm.String` or `Edm.Guid` or `ComplexNumberProperty` can be `Edm.Int16`, `Edm.Float` etc.


packages/core/test/test-util/filter-factory.ts
```diff
@@ -22,7 +22,7 @@ export const testFilterBoolean = {
 export const testFilterComplexType = {
   filter: TestEntity.COMPLEX_TYPE_PROPERTY.stringProperty.equals('test'),
   // odataStr: "ComplexTypeProperty/StringProperty eq 'test'"
-  odataStr: "A_TestComplexType/StringProperty eq 'test'"
+  odataStr: "ComplexTypeProperty/StringProperty eq 'test'"
```
marikaner: You can remove the comment.

packages/core/src/selectable/complex-type-field.ts
```diff
@@ -21,14 +21,18 @@ import { Field } from './field';
...
  */
 export abstract class ComplexTypeField<EntityT extends Entity> extends Field<EntityT> {
+  readonly pathToRootComplex;
+
   /**
    * Creates an instance of ComplexTypeField.
    *
    * @param fieldName Actual name of the field used in the OData request
-   * @param entityConstructor Constructor type of the entity the field belongs to
+   * @param rootEntityOrParentComplexField If the complex field is on root level of entity it is the entity otherwise the parent complex field
```
marikaner:
```suggestion
   * @param fieldOf The constructor of the entity or the complex type this field belongs to
```

packages/core/src/selectable/complex-type-field.ts
```diff
@@ -21,14 +21,18 @@ import { Field } from './field';
  * @typeparam EntityT Type of the entity the field belongs to
  */
 export abstract class ComplexTypeField<EntityT extends Entity> extends Field<EntityT> {
+  readonly pathToRootComplex;
```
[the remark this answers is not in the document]
FrankEssenberger: yes, recursion makes this obsolete.

packages/core/src/selectable/complex-type-field.ts
FrankEssenberger: Like it. I wille use this notion.

packages/core/src/selectable/complex-type-field.ts
```diff
@@ -21,14 +21,18 @@ import { Field } from './field';
...
+   * @param rootEntityOrParentComplexField If the complex field is on root level of entity it is the entity otherwise the parent complex field
    * @param complexTypeName Type of the field according to the metadata description
    */
-  constructor(fieldName: string, entityConstructor: Constructable<EntityT>, readonly complexTypeName: string) {
-    super(fieldName, entityConstructor);
+  constructor(fieldName: string, rootEntityOrParentComplexField: Constructable<Entity> | ComplexTypeField<EntityT>) {
+    super(fieldName, (rootEntityOrParentComplexField as ComplexTypeField<EntityT>)._entityConstructor || rootEntityOrParentComplexField);
+    const pathToRoot = (rootEntityOrParentComplexField as ComplexTypeField<EntityT>).pathToRootComplex;
+    this.pathToRootComplex = pathToRoot ? `${pathToRoot}/${fieldName}` : fieldName;
   }
```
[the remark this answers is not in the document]
FrankEssenberger: I like using recursion instead of class state to solve the problem.

packages/core/src/selectable/big-number-field.ts
```diff
@@ -77,12 +78,11 @@ export class ComplexTypeBigNumberPropertyField<EntityT extends Entity> extends B
...
    *
    * @param fieldName Actual name of the field used in the OData request
-   * @param entityConstructor Constructor type of the entity the field belongs to
-   * @param parentTypeName Name of the parent complex type
+   *  @param parentComplexField  Parent complex field
    * @param edmType Type of the field according to the metadata description
    */
-  constructor(fieldName: string, entityConstructor: Constructable<EntityT>, readonly parentTypeName: string, edmType: EdmType) {
-    super(fieldName, entityConstructor, edmType);
+  constructor(fieldName: string, readonly parentComplexField: ComplexTypeField<EntityT>, edmType: EdmType) {
```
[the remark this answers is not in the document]
FrankEssenberger: I added a overloaded constructor to keep the old method in parallel. Also renamed the field.

packages/core/test/test-util/filter-factory.ts
FrankEssenberger: yes

FrankEssenberger:
I get a failling type test. The test expects that the `TestEntity.COMPLEX_TYPE_PROPERTY.stringProperty` is not selectable. However, the changes of this PR break this test. The test was still working, before I overloaded the constructors, but I can not reproduce the change, which did break it, because I do not know the steps to update the `TestEntity` properly. I only realized that after some reinstall the error appeared again.

package.json
```diff
@@ -17,6 +17,7 @@
     "test:integration": "npx lerna run test --scope \"@sap-cloud-sdk/integration-tests\"",
     "test:type": "npx lerna run test --scope \"@sap-cloud-sdk/type-tests\"",
     "generate:test-services": "npx ts-node generate-test-services.ts && npm run lint:fix && lerna link --force-local",
+    "megacleaninstall": "rm -f package-lock.json && find . -name \"node_modules\" -exec rm -rf '{}' + && find . -name \"dist\" -exec rm -rf '{}' + && find . -name \"tsconfig.tsbuildinfo\" -exec rm -rf '{}' + && npm i",
```
marikaner: Why do you need this?

FrankEssenberger: Because I am hunting the ghost of the failing type test.....

packages/generator/src/entity/namespace.ts
```diff
@@ -31,14 +31,17 @@ function properties(entity: VdmEntity): VariableStatementStructure[] {
 
 function property(prop: VdmProperty, entity: VdmEntity): VariableStatementStructure {
   const type = `'${prop.edmType.startsWith('Edm') ? prop.edmType : prop.edmType.split('.').pop()}'`;
+  const initilizer = prop.isComplex
```
florian-richter: `initilizer` is misspelled. Should be `initializer`
```suggestion
  const initializer = prop.isComplex
```

packages/generator/src/entity/namespace.ts
```diff
@@ -31,14 +31,17 @@ function properties(entity: VdmEntity): VariableStatementStructure[] {
...
+    : `new ${prop.fieldType}('${prop.originalName}', ${entity.className}, ${type})`;
   return {
     kind: StructureKind.VariableStatement,
     declarationKind: VariableDeclarationKind.Const,
     declarations: [
       {
         name: prop.staticPropertyName,
         type: `${prop.fieldType}<${entity.className}>`,
-        initializer: `new ${prop.fieldType}('${prop.originalName}', ${entity.className}, ${type})`
+        initializer: initilizer
```
florian-richter: If the suggestion above is accepted, this can be re-written as:
```suggestion
        initializer
```

test-packages/type-tests/test/complex-types.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { and, asc, desc, or } from '@sap-cloud-sdk/core';
-import { TestComplexType, TestEntity } from '@sap-cloud-sdk/test-services/test-service';
+// import { TestComplexType, TestEntity } from '@sap-cloud-sdk/test-services/test-service';
+import { TestComplexType, TestEntity } from '../../test-services/srv/test-service';
```
florian-richter: Is this intended?

FrankEssenberger: No, I thought this could make a difference but lerna puts a symbolic link so that this is equivalent.

FrankEssenberger:
The tests are green so perhaps somebody can have a final look?

package.json
marikaner: Haha, but then this is probably does not have to be merged, right?

packages/core/src/selectable/big-number-field.ts
```diff
@@ -73,16 +74,34 @@ export class BigNumberField<EntityT extends Entity> extends BigNumberFieldBase<E
...
    * @param fieldName Actual name of the field used in the OData request
    * @param entityConstructor Constructor type of the entity the field belongs to
    * @param parentTypeName Name of the parent complex type
    * @param edmType Type of the field according to the metadata description
    */
-  constructor(fieldName: string, entityConstructor: Constructable<EntityT>, readonly parentTypeName: string, edmType: EdmType) {
-    super(fieldName, entityConstructor, edmType);
+  constructor(fieldName: string, entityConstructor: Constructable<EntityT>, parentTypeName: string, edmType: EdmType);
+
+  constructor(fieldName: string, fieldOf: ConstructorOrField<EntityT>, arg3: string | EdmType, arg4?: EdmType) {
```
marikaner: Should we also deprecate this (although it is new) and add documentation? Or maybe just a short documentation, that this is the union of the two constructors before?

marikaner:
> I get a failling type test. The test expects that the `TestEntity.COMPLEX_TYPE_PROPERTY.stringProperty` is not selectable. However, the changes of this PR break this test. The test was still working, before I overloaded the constructors, but I can not reproduce the change, which did break it, because I do not know the steps to update the `TestEntity` properly. I only realized that after some reinstall the error appeared again.

Do you still encounter this? I also had issues with the type tests, but only when updating typescript.

FrankEssenberger:
No, after understanding the issue related to `structural-typing` I kept a property to keep the classes different. Lesson learned: Do not use classes and inheritance in TS to allow for functionality.

package.json
FrankEssenberger: yes I removed this

packages/core/src/selectable/big-number-field.ts
```diff
@@ -73,16 +74,34 @@ export class BigNumberField<EntityT extends Entity> extends BigNumberFieldBase<E
  * @typeparam EntityT Type of the entity the field belongs to
  */
 export class ComplexTypeBigNumberPropertyField<EntityT extends Entity> extends BigNumberFieldBase<EntityT> {
+  /**
+   * The constructor of the entity or the complex type this field belongs to
+   */
+  readonly fieldOf: ConstructorOrField<EntityT>;
```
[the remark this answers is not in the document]
FrankEssenberger: the constructor overloading does not work with `readonly` keyword.

packages/core/src/selectable/big-number-field.ts
FrankEssenberger: ok I add a little sentence

PR #31, opened 2020-03-09
Fix: Dependency updates broke the generate test entity 

FrankEssenberger:
## Context

For some reason the `generate-test-service` had an issue with the promisified `fs` methods. This issue is most likely related to the large dependency updated being merged this morning.

When I merged master I came across the issue. So to reproduce the problem just checkout master and run the `generate:test-services` task.

I added the explicit types and the compiler was happy again. Also we should consider running the `npm run generate:test-services` in the pipeline so that we always have the latest version in the repo and find this kind of errors in the future.

### DIFF
--- a/generate-test-services.ts
+++ b/generate-test-services.ts
@@ -7,7 +7,8 @@ import path from 'path';
 import util from 'util';
 import { generate } from './packages/generator/src';
 
-const [readdir, readFile, writeFile] = [fs.readdir, fs.readFile, fs.writeFile].map(fsModule => util.promisify(fsModule));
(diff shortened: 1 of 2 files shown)


FrankEssenberger:
Regarding your questions:
- If I checkout master and make a `rm -rf package-lock.json  node_modules/ && npm i`. I get the same huge changes in the package-lock.json, so I guess the lock file was not updated after the recent dependency updates. 
- I created an issue so that we do not forget to put the generate step somewhere in the cd cycle. 
- Yes on master it also fails with the same error in the generate step.

FrankEssenberger:
yes the `npm i` vs `npm ci` was the issue. I have to remember this in the future.

PR #38, opened 2020-03-12
chore: Migrate from tslint to eslint

marikaner:
## Context

Tslint is being discontinued. This PR replaces it with an according configuration in eslint.


docs/classes/sap_cloud_sdk_core.booleanfilterfunction.html
```diff
@@ -70,7 +70,7 @@ <h1>Class BooleanFilterFunction&lt;EntityT&gt;</h1>
 			<section class="tsd-panel tsd-comment">
 				<div class="tsd-comment tsd-typography">
 					<div class="lead">
-						<p>Representation of a filter function, that has returns a value of type boolean.</p>
+						<p>Representation of a filter function, that returns a value of type boolean.</p>
```
jjtang1985: I guess you regenerated the docs.
But from my point view, the docs should reflect the stable version instead of canary versions.
Therefore, only release pipeline should update the docs.

package.json
```diff
@@ -18,13 +18,10 @@
     "test:type": "npx lerna run test --scope \"@sap-cloud-sdk/type-tests\"",
     "generate:test-services": "npx ts-node generate-test-services.ts && npm run lint:fix && lerna link --force-local",
     "compile": "npx lerna run compile",
-    "lint": "npm run tslint && npm run prettier",
-    "lint:fix": "npm run tslint:fix && npm run prettier:fix",
+    "lint": "npx eslint --ext .ts .",
+    "lint:fix": "npm run lint -- --fix",
+    "lint:fix-cached": "npm run lint:fix -- --cache",
     "postinstall": "npx lerna bootstrap --hoist --force-local --strict",
-    "prettier": "prettier -l \"**/*.ts\"",
```
jjtang1985: why `prettier` is also removed?
if this is intended, then it should also be removed form the `devDependencies`

packages/analytics/test/analytics-data.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
jjtang1985: I guess this is generated from the eslint, or the eslint shows error/warning?

jjtang1985:
I left some minor comments.
I did check the affected files, but not every line of the `.eslintrc.js`.
I guess you used some examples and applied some additional customised configurations.
One **important** question, have your tried to run the eslint with some negative tests?
When enabling a new test, negative tests help convince you that the whole testing toolkit can protect your projects.

docs/classes/sap_cloud_sdk_core.booleanfilterfunction.html
marikaner: Good catch, I will remove that.

package.json
marikaner: This is intended, but we need the dev dependency. Eslint is now taking over the check, so we don't need any extra command for that.

packages/analytics/test/analytics-data.spec.ts
marikaner: Yes, I condensed it to one line, because it was easier to implement and actually nicer to loose only one line of code on that.

docs/classes/sap_cloud_sdk_core.booleanfilterfunction.html
marikaner: Actually, I just remembered, that I did not regenerate, I just fixed the grammar. Do you think it is fine to leave it? I could remove it otherwise.

marikaner:
> I left some minor comments.
> I did check the affected files, but not every line of the `.eslintrc.js`.
> I guess you used some examples and applied some additional customised configurations.
> One **important** question, have your tried to run the eslint with some negative tests?
> When enabling a new test, negative tests help convince you that the whole testing toolkit can protect your projects.

I use a migration tool to migrate the config and had to adjust some specifics manually afterwards. It is possible that some things will not be perfect right from the get go, but I am quite happy with the configuration as of now. Also, yes, I tested some corner cases 😊

PR #46, opened 2020-03-16
Remove the Promise<void> and use `undefined`

FrankEssenberger:
## Context

@marikaner pointed out that the newly created return type for the function imports uses a Promise<void> as type. However none of the other transformers uses that and the reason behind it, is that the function import call gets wrapped in an `executor` providing the async way of live. So I changed it to `undefiened` in this PR.


packages/core/src/request-builder/response-transformers.ts
```diff
@@ -6,7 +6,7 @@ import { Constructable } from '../constructable';
 import { Entity } from '../entity';
 import { deserializeEntity } from '../entity-deserializer';
 
-export async function transformReturnValueForPromiseVoid<ReturnT>(data: any, builderFn: (data: any) => ReturnT) {
+export function transformReturnValueForPromiseVoid<ReturnT>(data: any, builderFn: (data: any) => ReturnT) {
```
marikaner: I think the function should be renamed to `transformReturnValueForUndefined`

FrankEssenberger: yes.

PR #48, opened 2020-03-17
chore: Update package-lock.json and remove sub package.jsons

marikaner:
Update package-lock.json files to not fail on master and pass security checks by only failing pipeline when there are high vulnerabilities. Ratio: With github we are constantly aware of issues and will fix them accordingly as soon as possible.


marikaner:
Does not work I will start from scratch

PR #52, opened 2020-03-20
Remove Hoisting

FrankEssenberger:
## Context

The hoisting from lerna caused problems with changes done to the package-lock.json. In this PR I tried to remove the hoisting. Besides a few minor things (jest needs a tsconfig and noch is not multi module enabled) it work quiete well. 
 
I also needed to adjust some dependencies. I also "fixed" the minimist audit error. I added the minmist in version 1.2.5 to the root dependencies and removed adjusted the package-lock.json by removing the minimist in the `optmist` dependecy used by handlebar. This removed the audit error. Run `npm audit` with or without the entry to reproduce the issue.

Today a fix was merged: https://github.com/wycats/handlebars.js/issues/1658 
So we should check if this hack is still necessary once this gets published. For doing it one has to remove minimist and rerun `npm i` to regenerate the package lock. 

General:
- In the investigation I found that `yarn` provides more tools for fixing such issues. Especially if we can get rid of lerne by using it.
- I am not sure about the package-lock.json on the module level. These were generated but do we need them? I simply do not know lerna well enough.


package.json
```diff
@@ -21,16 +21,14 @@
     "lint": "npx eslint --ext .ts .",
     "lint:fix": "npm run lint -- --fix",
     "lint:fix-cached": "npm run lint:fix -- --cache",
-    "postinstall": "npx lerna bootstrap --hoist --force-local --strict",
+    "postinstall": "npx lerna bootstrap  --strict",
```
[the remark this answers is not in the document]
marikaner: Was `--force-local` causing issues too?

packages/core/test/test-util/destination-service-mocks.ts
```diff
@@ -12,8 +20,14 @@ export function mockInstanceDestinationsCall(response: any, responseCode: number
...
-export function mockSubaccountDestinationsCall(response: any, responseCode: number, accessToken: string, uri: string = destinationServiceUri) {
-  return nock(uri, {
+export function mockSubaccountDestinationsCall(
+  abc: nockFunction,
+  response: any,
+  responseCode: number,
+  accessToken: string,
+  uri: string = destinationServiceUri
+) {
+  return abc(uri, {
```
marikaner: `abc` is not the best of names. What is it?

packages/core/test/scp-cf/destination-accessor.spec.ts
```diff
@@ -79,11 +79,11 @@ describe('destination-accessor', () => {
       mockUserApprovedServiceToken();
 
       const httpMocks = [
-        mockInstanceDestinationsCall([], 200, subscriberServiceToken),
-        mockSubaccountDestinationsCall(oauthMultipleResponse, 200, subscriberServiceToken),
-        mockInstanceDestinationsCall([], 200, providerServiceToken),
-        mockSubaccountDestinationsCall([], 200, providerServiceToken),
-        mockSingleDestinationCall(oauthSingleResponse, 200, destinationName, userApprovedSubscriberServiceToken)
+        mockInstanceDestinationsCall(nock, [], 200, subscriberServiceToken),
```
marikaner: Why was this change necessary. It seems to me like a sensible default...

test-packages/integration-tests/package.json
```diff
@@ -20,11 +20,14 @@
     "@sap-cloud-sdk/util": "^1.18.1"
   },
   "devDependencies": {
+    "axios": "0.19.2",
     "@types/jest": "^25.1.4",
+    "jsonwebtoken": "^8.5.1",
+    "bignumber.js": "8.1.1",
+    "nock": "^12.0.2",
```
marikaner: Is this really necessary? As far as I understand if this is available on root level we don't need to add it to the packages, except for executables.

test-packages/integration-tests/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.json",
```
marikaner: Is this needed? Or is it just a left-over from trying things out?

test-packages/test-services/package.json
```diff
@@ -16,6 +16,8 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@sap-cloud-sdk/core": "^1.18.1"
+    "@sap-cloud-sdk/core": "^1.18.1",
```
marikaner: Is this necessary? (Same as above)

test-packages/integration-tests/tsconfig.json
FrankEssenberger: it is necessary. It turns out that ts-jest needs a tsconfig.json implcitly.

test-packages/test-services/package.json
FrankEssenberger: I tried to remove it and got compilation errors.

package.json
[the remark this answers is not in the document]
FrankEssenberger: no put it back

packages/core/test/test-util/destination-service-mocks.ts
FrankEssenberger: was a temp name renamed after the draft stage was left.

test-packages/integration-tests/package.json
FrankEssenberger: I tried to remove it and got compilation errors.

FrankEssenberger:
Update: Hadlebars is updated with version 4.7.4 to not use the old minimist. So a npm audit fix was able to resolve the issue.

packages/core/test/scp-cf/destination-accessor.spec.ts
mr-flannery: Huh. Can you share more details here? Is that related to a version upgrade? I never noticed this problem before.

FrankEssenberger: No it is not related to an update. It is related to the previous hoisting. Situation before:
root
|---node_modules
|      |---nock
|---package_1
      |---node_modules
            |---ref_to_nock_root
|---package_2
      |---node_modules
            |---ref_to_nock_root

So when you did a `require(nock)` in package_1 or package_2 it was taking the same sources. Withtout the hoisting this has changed to:
root
|---package_1
      |---node_modules
            |---nock
|---package_2
      |---node_modules
            |---nock
So when you do a `require(nock)` in package_1 for example in the `destination-service-mocks.ts` of core/test-util and then in the `request-builder.spec.ts ` of the test-packages/integration-tests you load different source files. In principle this should not make a difference, but I found that there must be some encapsulation involved which breaks the matching on the first loaded nock. In other words the method `nock.activeMocks()` in package_1 do not show active mocks created by the second required nock in the  package_2 and when calls are executed the mocks are not triggered. 

Thats is why I pass

mr-flannery: Wow, that's evil... With that context this change makes a lot more sense.

marikaner: But couldn't that be solved by only having nock on the root level? According to (my understanding of) the node resolution algorithm, the root level node_module would be used and it would always be the same.

mr-flannery: technically yes, but I'm not a fan. You're essentially coupling the package to the project root. I am afraid that this a decision that has consequences that are not immediately obvious. Also it's something that you need to know when working with the project. If someone who's new to the project tries to run tests, then gets an error at runtime because nock isn't found, then looks at the package json and sees that nock isn't even listed as dependency and then adds it, I wouldn't blame them.

FrankEssenberger: I already discussed with Marika on this topic and I have no strong opinions. In principle it would be strict to have only linting etc. things on root and all package related stuff on the sub-packages. However this is also kind of inconvinient because something is needed everywhere so why putting it N times? 

My real Problem overall is: It should be consistent. Either put everything on root, or only the minimal stuff. Which one we chose I do not really care.

FrankEssenberger:
I just made a test. I removed the all node_modules folders and then the whole `npm i` took around 50 secs including the compile steps:

```
added 948 packages from 764 contributors and audited 42165 packages in 50.488s
```

If I use `npm ci` it is a bit less:

```
added 948 packages in 45.761s
```

In the VPN it was around 10sec longer. On master using the hoisting it took my machine 

```
added 969 packages from 773 contributors and audited 42060 packages in 61.443s
``` 
but it was already hot from the previous build so I guess the effect of heating up is more than the additional download time?  In the pipeline I see times between 50 and 70 sec for the build with [hoisting](https://github.com/SAP/cloud-sdk/pull/73/checks?check_run_id=563418953). 

So it seems like the times are not really increased by the change. I do not know, perhaps also other colleagues can try it out.

packages/core/test/scp-cf/destination-accessor.spec.ts
marikaner: I am in favor of putting it in the root except for dependencies that are run from the command line (there it has to be in the same context). I think dragging the dependency through the tests is worse than someone running `npm install` in a subpackage only. That can be covered by documentation and reviews.
I am also ok with doing this one we move to yarn.

PR #57, opened 2020-03-23
Update CHANGELOG.md

marikaner:
Update changelog to reflect compatibility notes.


sacnl:
looks good, you could give an example of how an error may look like to guide readers that are surprised by this

PR #65, opened 2020-03-27
Fix: Adjust Generated Docs

FrankEssenberger:
## Context

Due to the changed linting also the generated docs of the entities are changed. In the past the doc was created not in block form but the linter took accidentally care.

In this PR, empty lines are added so that the entities keep their nice block style as before.

In order to test, please run `npm run generate:tes-sources` and see if no changes are induced. I tried it, but you never know.

### DIFF
--- a/packages/generator/src/typedoc.ts
+++ b/packages/generator/src/typedoc.ts
@@ -37,7 +37,7 @@ export function getFunctionDoc(
       description += tagToText('returns', `${tags.returns.description}`);
     }
   }
-  return addLeadingNewline(description);
+  return description;
 }
 
 export function getComplexTypeFieldDescription(
--- a/packages/generator/src/typedoc.ts
+++ b/packages/generator/src/typedoc.ts
@@ -37,7 +37,7 @@ export function getFunctionDoc(
       description += tagToText('returns', `${tags.returns.description}`);
     }
   }
-  return description;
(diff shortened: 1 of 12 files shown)


packages/generator/src/typedoc.ts
```diff
@@ -50,21 +50,32 @@ export function getPropertyDescription(
   property: VdmProperty,
   constraints: VdmPropertyValueConstraints = { nullable: false }
 ): string {
-  return addConstraints(
-    property.description ||
-      endWithDot(toTitleFormat(property.instancePropertyName).trim()),
-    constraints
+  return makeBlockComment(
```
marikaner: I like the approach of having one function that creates valid input for block comments. In this case I initially thought that `makeBlockComment` actually creates something that looks like this:
```
/**
 * desc
 */
```
which it doesn't. It actually creates a block comment ready description. Maybe this can be reflected in the name of the function. Realistically what it does is preparing a block comment description or prepending `\n`.

marikaner: Second minor point: consider whether `makeBlockComment` shouldn't be called outside of `getPropertyDescription`. It seems as if this was a little bit more consistent that way. But as I said, minor.

FrankEssenberger:
For consistency I move the `add leading new line` call always to the point where the `doc` property is assigned.

PR #70, opened 2020-04-01
chore: Reenable all proxy tests

marikaner:
As we disabled those tests previously due to issues with the test setup, here I implemented all of the missing cases and cleaned up left overs from the old solution.


test-packages/integration-tests/test/proxy.spec.ts
```diff
@@ -43,8 +44,98 @@ describe('proxy', () => {
...
+      setTestDestination({ ...destination, name: destination.name });
+      await TestEntity.requestBuilder()
+        .getAll()
+        .execute({ destinationName: destination.name! });
+      expect(requestSpy).toHaveBeenCalledWith(objectContainingProxyAgent());
+    });
+
+    it('for destinations in env variable with proxy env variable', async () => {
+      setTestDestination({ url: destination.url, name: destination.name });
+      process.env.http_proxy = `${destination.proxyConfiguration?.protocol}://${destination.proxyConfiguration?.host}:${destination.proxyConfiguration?.port}`;
```
jjtang1985, line 96: I like the optional chaining, but since the `destination` is a const, why you have the null check here?

marikaner: Because in the destination type the proxyConfiguration is optional.

PR #71, opened 2020-04-02
feat: Allow setting log level globally

denidaja:
## Context

Fixed issue where the logger utility would show undefined if passed an Error object. 
This issue was coming from Winston library, more specifically the formatter, which was not accepting stack errors.  The fix was done on both local and Kibana format.

Fixed issue that allows setting log level globally. This was done by setting the container option to our global choice and prioritizing it when loggers get created. This solution was also tested by the associated unit test.

## Labels
`please review` and `please merge`


packages/util/src/logger/cloud-sdk-logger.ts
```diff
@@ -81,6 +81,7 @@ export function createLogger(
       : { ...messageContext };
   return container.get(customFields.messageContext, {
     level:
+      container.options.level ||
       customLogLevels[customFields.messageContext] ||
       customFields.level ||
       'info',
```
florian-richter, line 87: I think more specific log levels (e.g. `customLogLevels[customFields.messageContext]`) should take precendence over the global level. Maybe there could be an overwrite (like `.overwriteAllLogLevels()` or similar). @marikaner what do you think.

Also the precedence order should be tested to prevent accidentally breaking it.

mr-flannery: Good observation, I didn't even think about this when opening the ticket. If I understand you correctly you're proposing that:
``` 
setGlobalLogLevel('info')
setLogLevel('HelloWorldService', 'info')
```
and 

``` 
setLogLevel('HelloWorldService', 'info')
setGlobalLogLevel('info')
```
should behave the same, i.e. independent of the order in which they're called, which I'd agree with.
I think something like `overwriteAllLogLevels` or something like a  `force: boolean` parameter would then be a good addition.

FrankEssenberger: I do not really get the point from your example @mr-flannery if the log level is `info` in both cases. They are currently not behaving the same? I do not see this from the `cloud-sdk-logger.ts` snopped above. In general the order shoul matter, right

```
setGlobalLogLevel('warn')
setLogLevel('HelloWorldService', 'error')
=> only errors are logged in HelloWorldService.
```

versus 

```
setLogLevel('HelloWorldService', 'error')
setGlobalLogLevel('warn')
=> also war is logged in HelloWorldService
```

denidaja: At the moment the level that is set through setLogLevel will be always kept at higher precedence. Even if a global level is declared below a specific one, it will not overwrite it.

So in your example above: 
`
setLogLevel('HelloWorldService', 'error')
`
`setGlobalLogLevel('warn')
`
The HelloWorldService would still only log errors, but any other loggers created without a specific level will be using the global level.

I think this approach might be a little safer, as otherwise setting a global level anywhere in our code would overwrite all the loggers, which is a side-effect that I think will cause more harm than good in the long run and is bound to unexpected behavior.

Still, either design choice can be implemented, it just depends on what do we want from the logger and what is more intuitive to the developers.

marikaner: I think the idea to add a force parameter here might be useful in case someone wants to overwrite all custom log levels, but this is a nice to have and we don't need it if nobody asks for it. Realistically, if you set a specific log level for something you might really want this to be set.

packages/util/test/logger/cloud-sdk-logger.spec.ts
```diff
@@ -141,6 +144,23 @@ describe('Cloud SDK Logger', () => {
     });
   });
 
+  describe('get message when passing error objects', () => {
```
marikaner, line 147: 👍

packages/util/test/logger/cloud-sdk-logger.spec.ts
```diff
@@ -16,6 +18,7 @@ describe('Cloud SDK Logger', () => {
 
   afterEach(() => {
     logger.close();
+    setLogLevel('', messageContext);
```
marikaner, line 21: I think that is a fine approach. This could also be a `unsetLogLevel` but most likely this won't be needed.

packages/util/test/logger/cloud-sdk-logger.spec.ts
```diff
@@ -162,6 +182,38 @@ describe('Cloud SDK Logger', () => {
...
+      expect(level).toEqual(getGlobalLogLevel());
+    });
+
+    it('should have the global log level, if not applied a more specific level', () => {
+      logger = createLogger(messageContext);
+
+      expect(logger.level).toEqual(getGlobalLogLevel());
+    });
+
+    it('should have the context level, if applied a more specific level after creation', () => {
```
marikaner, line 203: "log level"

packages/util/test/logger/cloud-sdk-logger.spec.ts
```diff
@@ -162,6 +182,38 @@ describe('Cloud SDK Logger', () => {
...
+      logger = createLogger(messageContext);
+      setLogLevel('warn', messageContext);
+
+      expect(logger.level).toEqual('warn');
+    });
+
+    it('should have the context level, if applied a more specific level before creation', () => {
```
marikaner, line 210: "log level"

packages/util/src/logger/cloud-sdk-logger.ts
FrankEssenberger: Now I got the problem. I agree some force flag and good documentation is also my preferred option.

PR #78, opened 2020-04-14
fix: Serializing of Edm.Int64

marikaner:
I just stumbled upon this and was wondering what went wrong here...

### DIFF
--- a/packages/core/test/payload-value-converter.spec.ts
+++ b/packages/core/test/payload-value-converter.spec.ts
@@ -150,7 +150,7 @@ describe('tsToEdm()', () => {
   });
 
   it('should parse BigNumber to Edm.Int64', () => {
-    const expected = new BigNumber(16);
+    const expected = '16';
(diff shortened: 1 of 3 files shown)


packages/core/test/payload-value-converter.spec.ts
```diff
@@ -150,7 +150,7 @@ describe('tsToEdm()', () => {
   });
 
   it('should parse BigNumber to Edm.Int64', () => {
```
mr-flannery, line 152: Is there a test for the reverse direction, i.e. `Edm.Int64` to `BigNumber`?

marikaner: Yes, but this didn't change: https://github.com/SAP/cloud-sdk/blob/3828e4661894fffbe618fd9f1f85560cea27c9c5/packages/core/test/payload-value-converter.spec.ts#L37

PR #79, opened 2020-04-18
chore: Cleanup

marikaner:
During my daily work I found that we no longer remove unused imports. I also found, that the tsconfig-files for our tests were either missing or misconfigured, so that vscode could not find jest type information. I cleaned this up in this PR.


FrankEssenberger:
When I wask checking out your PR #80 I got this message: 
image

So perhaps there is some mixup between the two PRs?

marikaner:
> When I wask checking out your PR #80 I got this message:
> image
> 
> So perhaps there is some mixup between the two PRs?

I assume you have to run `npm i`?

PR #80, opened 2020-04-18
chore: Refactor header builder

marikaner:
In order to reuse the generic http builder I refactored the header builder to use the same approach of building header for odata requests and generic http requests.


packages/core/src/http-client/http-client.ts
```diff
@@ -24,18 +24,20 @@ import {
  * If a destination name (and a JWT) are provided, it will try to resolve the destination.
  *
  * @param destination - A destination or a destination name and a JWT.
+ * @param customHeaders - Custom default headers for the resulting HTTP request.
  * @returns A [[DestinationHttpRequestConfig]].
  */
 export async function buildHttpRequest(
-  destination: Destination | DestinationNameAndJwt
+  destination: Destination | DestinationNameAndJwt,
+  customHeaders?: MapType<any>
```
FrankEssenberger, line 32: just a general remark for version 2.0. If we have a build method taking multiple potentially optional arguments the good old builder pattern is always an option instead of having too many arguments. Here still ok wit 3 ones. Destination and DestiantionNameAndJwt are actually two arguments.

packages/core/src/request-builder/header-builder/auth-headers.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  return basicHeader(destination.username, destination.password);
+}
+
+export function basicHeader(username: string, password: string): string {
+  return `Basic ${Buffer.from(`${username}:${password}`).toString('base64')}`;
+}
+
+function headerForPrincipalPropagation(destination: Destination): MapType<any> {
```
FrankEssenberger, line 71: Minor: a few of the `headerForXXX` methods return a string and then a `toAuthHeader` or so is used to build the key-value header map. Others like this one return the map directly. Perhaps you can unify this. I know that I added this inconsistency partly.

packages/core/test/request-builder/header-builder/header-builder.spec.ts
```diff
@@ -234,6 +236,47 @@ describe('Header-builder:', () => {
     expect(headers['x-csrf-token']).toBe('mocked-x-csrf-token');
   });
 
+  it('Skips csrf token retrieval for existing csrf header', async () => {
```
jjtang1985, line 239: 👍 for more tests.

packages/core/src/request-builder/header-builder/auth-headers.ts
```diff
@@ -0,0 +1,162 @@
...
+  );
+
+  let principalPropagationHeader;
+  try {
+    principalPropagationHeader = headerForPrincipalPropagation(destination);
+  } catch (e) {
+    logger.warn('No principal propagation header found.');
+  }
+
+  // TODO the proxy header are for OnPrem auth and are now handled correctly in the authorization-header.ts and should be removed here
```
FrankEssenberger, line 105: Could you move this TODO to the `getProxyRelatedAuthHeaders` method and say that the whole `legacyNoAutOnPremiseProxy`  method should be removed in version 2.0.

packages/core/src/request-builder/header-builder/auth-headers.ts
```diff
@@ -0,0 +1,162 @@
...
+    destination.authentication === 'NoAuthentication'
+  ) {
+    return legacyNoAuthOnPremiseProxy(destination);
+  }
+
+  // The connectivity service will raise an exception if it can not obtain the 'Proxy-Authorization' and the destination lookup will fail early
+  return headerForProxy(destination);
+}
+
+async function getAuthenticationRelatedAuthHeaders(
```
FrankEssenberger, line 125: I find the name a bit tautologish (I do not know if this is the adjective for Tautology).  You have auth headers for the Proxy and auth headers for the target system. Perhaps call it like this: `DestinationRelatedAuthHeaders` or `TargetSystemRelatedAuthHeaders`....

packages/core/src/request-builder/header-builder/authorization-header.ts
```diff
@@ -36,178 +23,34 @@ export async function addAuthorizationHeader<
...
-  if (isNil(destination.username) || isNil(destination.password)) {
-    throw new Error(
-      'AuthenticationType is "BasicAuthentication", but "username" and/or "password" are missing!'
-    );
-  }
-
-  return basicHeader(destination.username, destination.password);
-};
-
-export function basicHeader(username: string, password: string): string {
```
FrankEssenberger, line 211: Question: When this method is moved to the auth-header.ts file. Isn't it a breaking change?

FrankEssenberger: If so also keep it and depricate it and move a copy to auth-header?

packages/core/src/request-builder/header-builder/csrf-headers.ts
```diff
@@ -0,0 +1,84 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: General: Why did you create a new file `csrf-header` and `auth-headers` to contain the refactored/adjusted code. Why not put the new exported method `getCsrfHeaders` in the old file and deprecate the `addCsrfTokenAndCookies`.

packages/core/src/request-builder/header-builder/auth-headers.ts
marikaner: I thought of this and then forgot. Will do.

packages/core/src/request-builder/header-builder/headers-util.ts
```diff
@@ -0,0 +1,57 @@
...
+): any | undefined {
+  return Object.entries(headers).find(
+    ([entryKey]) => entryKey.toLowerCase() === key.toLowerCase()
+  )?.[1];
+}
+
+export function filterNullishValues(headers: MapType<any> = {}): MapType<any> {
+  return Object.entries(headers)
+    .filter(([_, value]) => !isNullish(value))
+    .reduce((filtered, [key, value]) => ({ ...filtered, [key]: value }), {});
```
FrankEssenberger, line 34: what is the reduce doing after the filter? Could you not just return the filtered? And why do you use `_` as name for the key in the filter?

packages/core/src/request-builder/header-builder/authorization-header.ts
marikaner: Unfortunately, yes. But I cannot export the same function twice (because it is exported from the index.ts). Not sure what the best option would be here... I could also keep in the old file and rename the other, but I don't love this.

packages/core/src/request-builder/header-builder/header-builder.ts
```diff
@@ -21,89 +21,80 @@ import { addCsrfTokenAndCookies } from './csrf-token-header';
...
-    addProxyHeaders(request.destination.proxyConfiguration),
-    addLocationIdHeader(request.destination.cloudConnectorLocationId)
+
+  const defaultHeaders = replaceDuplicateKeys(
+    filterNullishValues({
+      Accept: 'application/json',
+      'Content-Type': request.config.contentType,
+      'if-match': getETagHeaderValue(request.config)
+    }),
+    request.config.customHeaders
```
FrankEssenberger, line 37: Is it possible to use the spread `...` to replace the  keys from the first object with the ones from the customHeaders object?

packages/core/src/request-builder/header-builder/header-builder.ts
```diff
@@ -21,89 +21,80 @@ import { addCsrfTokenAndCookies } from './csrf-token-header';
...
+
+  const destinationRelatedHeaders = await buildHeadersForDestination(
+    request.destination,
+    request.config.customHeaders
   );
+
+  const csrfHeaders =
+    request.config.method === 'get'
+      ? {}
+      : await getHeaderByKeyOrExecute(
```
FrankEssenberger, line 48: I see that this method is used in three places, but I am not a big fan. Could you just use the nullish coalescing: const myValue = header['yourKey'] ?? fn(). Would be more transparent then the getHeaderByKeyOrExecute().

packages/core/src/request-builder/header-builder/csrf-headers.ts
marikaner: Good question, I started by really going function by function and removing everything, and I didn't want everything else to break. I slightly prefer the new names and thought it might be easier to delete this later on also in combination with the according tests, but I see your point - it might be confusing when you want to work on one of those files. What do you think we should do?

packages/core/src/request-builder/header-builder/headers-util.ts
marikaner: The secret here is: Object.entries(headers).
1. The reduce transforms the array of entries [key, value][] to an object.
2. no, because of 1.
3. I don't need the key here and want to make it clear by basically ignoring the value (I need it in the reduce though)

packages/core/src/request-builder/header-builder/header-builder.ts
```diff
@@ -21,89 +21,80 @@ import { addCsrfTokenAndCookies } from './csrf-token-header';
...
+  );
 
-const addAcceptHeader = (headers: MapType<string>) =>
-  assoc('Accept', 'application/json')(headers);
+  const sapHeaders = replaceDuplicateKeys(
+    filterNullishValues({
+      'sap-client': destination.sapClient,
+      'SAP-Connectivity-SCC-Location_ID': destination.cloudConnectorLocationId
+    }),
+    customHeaders
```
FrankEssenberger, line 87: also here spread?

packages/core/src/request-builder/header-builder/header-builder.ts
marikaner: No because this does not take into consideration upper and lower case, e. g. if customHeaders contains a 'accept' header.

packages/core/src/request-builder/header-builder/headers-util.ts
```diff
@@ -0,0 +1,57 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+import { MapType } from '@sap-cloud-sdk/util';
+
+export function isNullish(x: any): x is null | undefined {
+  return x === null || x === undefined;
+}
+
+export function toHeaderObject(key: string, value: any): MapType<any> {
+  return isNullish(value) ? {} : { [key]: value };
```
FrankEssenberger, line 9: what happens if the key is nullish?

packages/core/src/request-builder/header-builder/headers-util.ts
```diff
@@ -0,0 +1,57 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+import { MapType } from '@sap-cloud-sdk/util';
+
+export function isNullish(x: any): x is null | undefined {
```
FrankEssenberger, line 4: It seems that there should be a build in method option to do that?

packages/core/src/request-builder/header-builder/headers-util.ts
```diff
@@ -0,0 +1,57 @@
...
+  headers: MapType<any> = {}
+): MapType<any> {
+  const entry = Object.entries(headers).find(
+    ([entryKey]) => entryKey.toLowerCase() === key.toLowerCase()
+  );
+  return entry ? { [entry[0]]: entry[1] } : {};
+}
+
+export function getHeaderValue(
+  key: string,
```
FrankEssenberger, line 23: could you use the getHeader() method inside here?

packages/core/src/request-builder/header-builder/headers-util.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export function isNullish(x: any): x is null | undefined {
+  return x === null || x === undefined;
+}
+
+export function toHeaderObject(key: string, value: any): MapType<any> {
+  return isNullish(value) ? {} : { [key]: value };
+}
+
+export function getHeader(
```
FrankEssenberger, line 12: This does actually more then the name suggests, case is ignored and a {} is returned in case it is not the an empty object is given. Perhaps you can reflect in the name or check if the ignoring case etc. is really needed.

packages/core/test/http-client/http-client.spec.ts
```diff
@@ -67,8 +67,7 @@ describe('generic http client', () => {
         headers: {
           authorization: 'Basic VVNFUk5BTUU6UEFTU1dPUkQ=',
           'sap-client': '001',
-          'Proxy-Authorization': 'jajaja',
-          'SAP-Connectivity-Authentication': 'yeahyeah'
+          'Proxy-Authorization': 'jajaja'
```
FrankEssenberger, line 70: perhaps we should use more mature dummy values or more funny ones like `Testination` or `BernieAndCert`. `jajaja` is a bit weak.

packages/core/test/request-builder/header-builder/auth-headers.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    await expect(
+      buildAuthorizationHeader({
+        url: 'https://example.com',
+        authentication: 'ClientCertificateAuthentication'
+      })
+    ).resolves.not.toThrow();
+  });
+
+  it('defaults to NoAuthentication', async () => {
+    await expect(
```
FrankEssenberger, line 31: shouldn't there be an assertion that there is no auth header?

packages/core/test/request-builder/header-builder/auth-headers.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+      buildAuthorizationHeader({
+        url: 'https://example.com',
+        authentication: 'NoAuthentication'
+      })
+    ).resolves.not.toThrow();
+  });
+
+  it('does not throw on ClientCertificateAuthentication', async () => {
+    await expect(
```
FrankEssenberger, line 22: Is there an other test checking the content expected as the header field if auth type is `ClientCertificateAuthentication`? If not I would add an assertion and not only chcek that it does not throw.

packages/core/test/request-builder/header-builder/header-builder.spec.ts
FrankEssenberger, line 239: General: What is the strategy to keep track on which tests relate to the new or old depricated methods. In case we clean up the deprecated ones it could become difficult to understand if thre is already a test covering it or you need to migrate. If see the `Prioritizes custom Authorization headers (upper case A)` kind of doulbled but some other test are only in the old spec.

packages/core/src/request-builder/header-builder/header-builder.ts
marikaner: Great suggestion. I added that (in slightly different form).

packages/core/src/request-builder/header-builder/headers-util.ts
marikaner: I now also check for keys. Although this is not used anywhere, you are right, if this is used differently in the future, we are safe.

packages/core/src/request-builder/header-builder/header-builder.ts
marikaner: not possible because of lower case and upper case

packages/core/src/request-builder/header-builder/headers-util.ts
marikaner: There is the nullish coalscing operator but this is not helpful in many cases. Rambda has a function that does that, but I would like to get rid of rambda in the long run (at least for everything related to destination handling) if there is not good reason to keep it. I agree, that this doesn't belong here though and I will move it to util. We have similar functionality there.

packages/core/src/request-builder/header-builder/headers-util.ts
marikaner: sure

packages/core/src/request-builder/header-builder/headers-util.ts
marikaner: I don't quite understand. In case a matching header is found an object containing this header is returned (using its original name). In case it is not, an empty object is returned.

marikaner: Ok, I see, you don't mean `getHeader' but `toHeaderObject`. I partially agree, I just took it over, this was there before. What do you think of `toSanitizedHeaderObject`?

packages/core/test/http-client/http-client.spec.ts
marikaner: wasn't me 🤓but I also thought of this...

packages/core/test/request-builder/header-builder/auth-headers.spec.ts
marikaner: 100% agree. I think we have a lot of potential to improve our test strategy. In the end this tests what we want implicitly, but is way more complicated than it should be in my opinion. I just didn't want to touch this more than I already did.

packages/core/test/request-builder/header-builder/auth-headers.spec.ts
marikaner: I agree on this, but for all tests. I only adjusted the tests (and added some for code that I added), but I didn't want to also refactor our tests. This implicitly tests that headers are set and therefore built correctly, so it is testing the correct thing at least. As mentioned before, I created a ticket for formalizing and improving our test strategy.

packages/core/test/request-builder/header-builder/header-builder.spec.ts
marikaner: Let's discuss this in person. I agree that this is not optimal.

packages/core/src/request-builder/header-builder/authorization-header.ts
FrankEssenberger: I would keep it in the old file. You never know who uses it and gets annoyed be a change.

packages/core/src/request-builder/header-builder/csrf-headers.ts
FrankEssenberger: The argument of removing the files I agree. So for me the approach is fine just harder to see differences.

packages/core/src/request-builder/header-builder/headers-util.ts
FrankEssenberger: No I meant to getHeader() if I call it for getHeader('KEY_IN_CAPITAL',header) I can get as a return {'key_in_capital':'someValue;} which is sort of surprising. So you can also add some JS doc in the util just telling the user that case is ignored and {} is returned in case no key is found in contrast to an exception beeing thrown.

packages/core/src/request-builder/header-builder/authorization-header.ts
marikaner: I put everything back in the old files for auth and csrf.

packages/core/src/request-builder/header-builder/headers-util.ts
marikaner: Added documentation

packages/core/test/request-builder/header-builder/auth-headers.spec.ts
marikaner: Created a ticket for this

packages/core/src/request-builder/header-builder/headers-util.ts
```diff
@@ -0,0 +1,71 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: Does it make sense to have tests for the utils methods?

marikaner: yes, will do.

marikaner: Done

PR #110, opened 2020-05-03
Split commit into 2 files, improve naming, fix typos

artemkovalyov:
## Context

Addressing comments to the previous pipeline setup for docs:
- Split into 2 files
- Rename `checks` to `build`
- Hydrate everything needless
- Improve naming for irrelevant things
- Reflect similar structure to master
- Permission is still managed via key
- Tested locally on my fork and proven to work

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,17 @@
+name: Build Cloud SDK Documentation
+
+on:
+  pull_request:
+    branches: [documentation]
+
+jobs:
+  build:
+   runs-on: ubuntu-latest
+   steps:
+     - uses: actions/checkout@v1
+     - uses: actions/setup-node@v1
+       with:
+         node-version: '12.x'
+         run: |
+          npm ci
+          npm run build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,25 +1,11 @@
-name: documentation
+name: Release Cloud SDK documentation
 
 on:
-  pull_request:
-    branches: [documentation]
   push:
     branches: [documentation]
 
 jobs:
- checks:
-   if: github.event_name != 'push'
-   runs-on: ubuntu-latest
-   steps:
-     - uses: actions/checkout@v1
-     - uses: actions/setup-node@v1
-       with:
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,17 @@
...
+  build:
+   runs-on: ubuntu-latest
+   steps:
+     - uses: actions/checkout@v1
+     - uses: actions/setup-node@v1
+       with:
+         node-version: '12.x'
+         run: |
+          npm ci
+          npm run build
```
jjtang1985, line 17: Could you please share more insights here?
What does this script do here?
I would expect the `build` (or the old name `check`) tests something to ensure that merging the PR does not break anything + brings the value.

.github/workflows/release.yml
```diff
@@ -39,12 +25,12 @@ jobs:
...
+    - name: Release Documentation to GitHub Pages
       env:
         USE_SSH: true
-        GIT_USER: git
+        GIT_USER: irrelevant
       run: |
-        git config --global user.email "actions@gihub.com"
-        git config --global user.name "gh-actions"
+        git config --global user.email "irrelevant@completely.irrelevant"
+        git config --global user.name "irrelevant"
```
jjtang1985, line 34: If the username/email are not relevant, is it possible to remove some of them?

artemkovalyov: @jjtang1985, sadly both are needed to preconfigure git so that it doesn't complain and fails with non-zero exit code. A similar approach is used in Docusaurus docs for Circle CI. I don't know a more beautiful workaround for it.

.github/workflows/build.yml
artemkovalyov: This has the sole purpose of ensuring the PR can be built which in most cases means it is deployable after it's merged. It will catch inconsistencies in config files like `sidebar.js`, `docusaurus.config.js`, and some syntax errors for MD and JS files. Some runtime bugs will certainly slip through, but that is good enough test for now. If we will identify some sources of failure for docs we'll make sure to add more tests like checking links to work, linting Markdown, etc.

PR #118, opened 2020-05-05
chore: Add fs-extra types

marikaner:
Adding types to the generator


FrankEssenberger:
Good that you make a separate PR also for such small changes and do not include them in some other PR which is currently open.

PR #120, opened 2020-05-06
feat: Internal split between v2 / v4

marikaner:
This is the full internal split between v2 / v4


packages/core/src/v4/entity.ts
```diff
@@ -0,0 +1,307 @@
...
+    value: EntityTypeForceMandatoryT[property]
+  ) => EntityBuilderType<EntityT, EntityTypeForceMandatoryT>;
+} &
+  EntityBuilder<EntityT, EntityTypeForceMandatoryT>;
+
+/**
+ * Super class for all representations of OData entity types.
+ */
+export abstract class Entity {
+  static _serviceName: string;
```
FrankEssenberger, line 24: Perhaps you can add an interface or parent class which holds the common fields of the Entity in v2 and v4 so that one has less duplication

FrankEssenberger:
I do not see the `common` folder yet. I guess this will come soon since in the meeting I have seen something.

notes.md
```diff
@@ -0,0 +1,30 @@
+# Notes OData POC
```
FrankEssenberger, line 1: Depening on the decision if we put the version in the request or on the entity we should polish this document to a proper decision document and put it to the `knowledge-base` folder.

packages/core/src/scp-cf/env-destination-accessor.ts
```diff
@@ -2,9 +2,9 @@
 
 import { createLogger } from '@sap-cloud-sdk/util';
 import {
-  addProxyConfigurationInternet,
+  proxyStrategy,
   ProxyStrategy,
-  proxyStrategy
+  addProxyConfigurationInternet
```
FrankEssenberger, line 7: run linter for alphabetic ordering.

packages/core/src/v2/constructable.ts
```diff
@@ -14,9 +14,9 @@ export interface Constructable<
   _serviceName: string;
   _entityName: string;
   _defaultServicePath: string;
-  _allFields: Selectable<EntityT>[];
-  _keyFields: Selectable<EntityT>[];
-  _keys: { [keys: string]: Selectable<EntityT> };
+  _allFields: (Field<EntityT> | Link<EntityT>)[];
```
FrankEssenberger, line 17: This seems like a quick fix to some type issue?

packages/core/src/v2/entity-deserializer.ts
```diff
@@ -79,13 +79,15 @@ function extractODataETag(json: MapType<any>): string | undefined {
 
 function getFieldValue<EntityT extends Entity, JsonT>(
   json: Partial<JsonT>,
-  selectable: Selectable<EntityT>
+  selectable: Field<EntityT> | Link<EntityT>
```
FrankEssenberger, line 82: I am confused now. The original selectable type contained:```  | SimpleTypeFields<EntityT>
  | Link<EntityT, any>
  | ComplexTypeField<EntityT>
  | CustomField<EntityT>
  | AllFields<EntityT>;
```
So quite a lot. In this PR this is reduced. Also from a logical point of view it seems like the implementation from getFieldValue  considers only EdmTypeField, Link and ComplexTypeField. So the type above should indicate this?

File: packages/core/src/v2/entity.ts [hunk 6]
```diff
@@ -275,9 +278,9 @@ export interface EntityIdentifiable<T extends Entity> {
  */
 export function isSelectedProperty<EntityT extends Entity>(
   json,
-  selectable: Selectable<EntityT>
+  field: Field<EntityT> | Link<EntityT>
```
FrankEssenberger, line 281: see comments above on the selectable type.

packages/core/src/v2/filter/filter-function.ts
```diff
@@ -0,0 +1,95 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: for some reason the git move did not work for this file.

packages/core/src/v2/request-builder/create-request-builder.ts
```diff
@@ -0,0 +1,100 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: also here try to use `git mv`

packages/core/src/v2/request-builder/delete-request-builder.ts
```diff
@@ -0,0 +1,89 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: also here try to use `git mv`

packages/core/src/v2/request-builder/request/odata-function-import-request-config.ts
```diff
@@ -0,0 +1,65 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: also here try to use `git mv`

packages/core/src/v2/filter/filter-function.ts
FrankEssenberger: There are some files in the v2 which are marked as new files and not as moved. But I get all content there is just a move from the old location. So perhaps you can check this, because this makes the review easier.

FrankEssenberger:
Here is the section from the meeting notes regarding file strucutre:
- In src: odata/v2, odata/v4 and odata/common (core package)
- In test odata/v2, odata/v4 and odata/common (core package) 
- Export the v2 to root. Export the v4 to v4 (import v4 from '@sap-cloud-sdk/core').
- Request builders get duplicated for v4. 
- (De)Serializers get duplicated for v4.
- Classes with the specific v2/v4 implementation get a v4 extension (v2 extension would be a breaking change). Files no v2/v4 extension
- Marika & Dennis discuss: RequestBuilder vs. Entity version aware. 
- Postpone decision on generator. First we learn what we want to generate.

mr-flannery:
> Marika & Dennis discuss: RequestBuilder vs. Entity version aware

something I need to be aware of?

packages/core/src/v2/constructable.ts
marikaner: No, actually I noticed that along the way. The type before was just wrong.

packages/core/src/v2/entity-deserializer.ts
marikaner: This is related to your comment before. When checking this, I realized, that Selectable is not the correct type here. This function deserializes a field of an entity, which can either be a "property" or a "navigation property" (we call them "Field" and "Link" - my bad). What the function does is correct, but the type it took was too broad.

packages/core/src/v2/entity.ts
marikaner: Same here. In this case we check whether a field is being deserialized (so only fields or links) is also selected.

notes.md
marikaner: Yeah, this was more for the reviewer. I think Junjie took proper notes in the meeting.

packages/core/src/scp-cf/env-destination-accessor.ts
marikaner: It seems that our linter does not do that ;)

FrankEssenberger:
FunctionImportParameters => potential common
ODataRequest => remove
followUpTicket => chain from deserializeEntity is generic quiet a lot.
response-transformers => potential common /reuse
get-keys => potential common check also other uri converters
get-XXX => get-select, get-expand be close to the names

regarding the `v4` suffix. I changed my initial vote to do it. The reason is that we will move quiete some things to the `common` folder in the next days. In this context the suffix would make it more difficult. After we are really converged I would like to revisit this point. So for the moment we stick to the v2 v4 folders to distinguish the two versions.

FrankEssenberger:
- [ ] New Request configs mark as experimental
- [ ] export * as v4 from './odata/v4'; or something similar
- [ ] extractCustomFields  TODO only exported for tests. 
- [ ] add the Field<EntityT> | Link<EntityT> TODO as sdk 2.0 issue

packages/core/src/header-builder/header-builder.ts
```diff
@@ -1,15 +1,16 @@
...
+import { ODataRequestConfig, isWithETag } from '../odata/common/request';
+import { ODataRequest } from '../odata/common/request/odata-request';
+import { buildAuthorizationHeaders } from './authorization-header';
 import { buildCsrfHeaders } from './csrf-token-header';
 import {
-  filterNullishValues,
+  getHeader,
   replaceDuplicateKeys,
-  getHeader
+  filterNullishValues
```
FrankEssenberger, line 12: minor: use linter to sort imports alphabetically.

marikaner: I did, our linter currently does not do that. ;)

FrankEssenberger:
Points for final discussion:
- [ ] Sorting imports alphabetically example `header-builder.ts`
- [ ] In `common\filter-function.ts` a few methods are depricated. Give sentence what to use. I guess the new `v2\filter-function.ts`. 
- [ ] The other methods like `equals`, `notEqual` are not deprecated. But they are not the same for v2 and v4, rigth?
- [ ] `request-builder-base.ts the @abstract annotation is removed, why?
- [ ] I have not really understood the sdk 2.0 remark in the `odata-request.ts` regarding the circular dependency and did you create a 2.0 issue for that?
-[ ] In the `common\selectable\link.ts` the static clone() is replace by a memeber clone() for the other links the static method is deprecated and kept. Breaking change?
- [ ] In the links the static clone is depricate say what to use instead.
- [ ] `common/uri-conversion/get-filter.ts ` , `uri-conversion/get-resource-path.ts`is shown as new file, is copied?
- [ ] `request-builder/function-import-request-builder.ts` is in v2 v4 and not in common?
- [ ] I am confused by the three `get-filter.ts` in common,v2 and v4.

marikaner:
E2E tests are green!
Removing `get-filter.ts` for v4 was not possible without introducing another circular dependency, so I left it.

PR #121, opened 2020-05-06
Add Documentation for SCP Workflow REST API Client Lib

Emdee89:
Introducing documentation on the SCP Workflow REST API Client Lib


docs/java/features/rest/clients/scp-workflow-rest-api.md
```diff
@@ -0,0 +1,229 @@
...
+cf create-service workflow standard my-workflow-service -c <path-to-json-file>
+```
+This command creates an instance of the SCP Workflow Service in the CF space that your CLI points to. More specifically, it uses the service plan "standard" and takes the JSON configuration into account. Note that we named the service instance ``my-workflow``. You can name it as you want, it is just important to remember the name as you need it in your deployment descriptor `manifest.yml` later on.
+
+Once the service instance creation is finished, you can see the service instance in your CF space under ``Services`` and `Service Instances` in the left-hand side menu.
+
+
+##### Bind your App to Service Instance
+Open the file ``manifest.yml`` in your project and mention your service instance under `services`. Refer to this example:
+```
```
MatKuhr, line 80: yaml

docs/java/features/rest/clients/scp-workflow-rest-api.md
```diff
@@ -0,0 +1,229 @@
...
+  services:
+  - my-destination
+  - my-workflow
+  routes:
+  - route: <omitted-on-purpose>
+```
+Redeploy your app with ``cf push``.
+
+##### Take Note of API endpoint and OAuth Credentials
+After the deployment of the app happened, go to your CF space and navigate to ``Services`` and thereafter to `Service Instances`. You should see the service instance you created along with the information that is is bound to your application.
```
MatKuhr, line 102: I recommend going through the document with strg+f and replacing all the double quotes with single ones, so "`"  instead of "``" 

It looks to me like when you type you don't press space after hitting the "`" key. That is needed since how it is printed depends on the following character. Just FYI in case you don't know. Or is there a reason for the double ones?

Emdee89: good hint, thanks

docs/java/features/rest/clients/scp-workflow-rest-api.md
```diff
@@ -0,0 +1,228 @@
...
+
+##### Create Service Instance
+
+Open the command line and authenticate at your Cloud Foundry organization by invoking `cf login`.
+
+Consider specifying the respective subaccount, organization and space with `cf target` if necessary.
+
+Use `cd` to navigate to the directory that contains the JSON file created beforehand.
+Being in that directory, create the service instance as follows:
+```
```
MatKuhr, line 70:
```sugestion
```bash
```

PR #126, opened 2020-05-08
Fix syntax highlighting

MatKuhr:
Regex to find these things in the future: ` ```[A-Z]`

Should I create a GH action for it?


fwilhe:
> Should I create a GH action for it?

sure, why not? seems like a low hanging fruit and it helps with quality.

PR #134, opened 2020-05-13
Add forEntity reference to the documentation (java)

newtork:
## Context

This PR adds documentation details to the latest feature for OData v4 CRUD operations in Cloud SDK for Java: the API extension `forEntity`.


docs/java/features/odata/use-vdm.md
```diff
@@ -101,6 +101,25 @@ Request parameters:
    - `ignoreAnyVersionIdentifier()` will instead always send a `*` which acts as a wildcard to match all ETags.
 - All operations allow for adding custom headers via `withheader(...)`
 
+Navigable entities:
+- By using the `forEntity` method on service classes, you can recursively traverse navigation properties in entity sets. It's a generic API extension with type-safe methods to allow for requests along chained entities, according to their navigation properties.
```
MatKuhr, line 105:
```suggestion
- By using the `forEntity` method on service classes you can directly access navigation properties in entity sets. It's a generic API extension with type-safe methods to allow for requests along chained entities, according to their navigation properties.
```

MatKuhr: Since this topic is a bit complicated I would try to word it a bit more, lets say, "direct"...

docs/java/features/odata/use-vdm.md
```diff
@@ -101,6 +101,25 @@ Request parameters:
...
+- Consider the following example:
+  ```java
+  // Create a new Trip in Vacation (id 2020) of Person (username "John")
+  // POST /ODataService/People('John')/Vacations(2020)/Trips
```
MatKuhr, line 109: For consistency I would suggest to separate the Java code and the Http request it produces.

docs/java/features/odata/use-vdm.md
```diff
@@ -101,6 +101,25 @@ Request parameters:
...
+  
+  Trip tripItem;
+  Person personById = Person.builder().username("John").build();
```
MatKuhr, line 112: Did you consider using the BusinessPartner service? I think it would make things a bit easier to understand since there the datamodel is already known to the reader (at least through the examples contained already).

docs/java/features/odata/use-vdm.md
```diff
@@ -101,6 +101,25 @@ Request parameters:
    - `ignoreAnyVersionIdentifier()` will instead always send a `*` which acts as a wildcard to match all ETags.
 - All operations allow for adding custom headers via `withheader(...)`
 
+Navigable entities:
```
MatKuhr, line 104: We actually thought of having a new section (or even new document) `Advanced OData Features`. because we still have batch, function import, (actions) etc.

And this feels like a  more advanced feature to me, so putting it directly under `Basic CRUD` is maybe not the right place. But we can move it also later on.

docs/java/features/odata/use-vdm.md
newtork: The HTTP request is not mentioned anywhere else. There is no "consistency" (?)

docs/java/features/odata/use-vdm.md
newtork: Hmm.. Your proposal will hide the recursive nature of the API.

docs/java/features/odata/use-vdm.md
newtork: Yeah.. would be better.

docs/java/features/odata/use-vdm.md
newtork: I'm also in favor of _advanced feature_ section. Although "advanced" would be an exaggeration, we simply mean _non-CRUD_.

However this feature just allows CRUD operations, but on navigation properties instead of root entity-sets. That's why I felt okay with leaving it there.

docs/java/features/odata/use-vdm.md
MatKuhr: True, then maybe add another sentence. Otherwise it's too unclear what we are talking about here IMO.

docs/java/features/odata/use-vdm.md
MatKuhr: Yes, I meant rather the URI as in [the example for select](https://sap.github.io/cloud-sdk/docs/java/features/odata/use-generated-odata-vdm-v2-v4#select).

docs/java/features/odata/use-vdm.md
MatKuhr: Mhm yes, so from the CRUD point of view it fits there. Maybe we just name the current section `Basic Operations` and then another `Advanced Operations`?

docs/java/features/odata/use-vdm.md
newtork: I restructured the content. I'm no longer focusing on the recursive nature.

docs/java/features/odata/use-vdm.md
newtork: Changed to _BusinessPartner_, although this is not a good example service since all entities are accessible from service root.

docs/java/features/odata/use-vdm.md
newtork: Opposing to the "example for select" this example focuses on URI path and not URI query.
Maybe my updated content better fits the page.

docs/java/features/odata/use-vdm.md
newtork: I moved the content to a dedicated section within the page.
I wouldn't feel comfortable with introducing a generic "Advanced Operations" headline.

PR #140, opened 2020-05-14
Pipeline to handle API reference update to documentation branch

artemkovalyov:
The pipeline was tested manually. What it does:
- triggered only on changes in the `docs` folder
- checks out master
- clones `documentation` branch into a separate folder
- syncs API docs inside of the container
- pushes to `documentation` branch which in turn triggers the release pipeline on it.

I manually tested it end to end

image
image
image

I also added the snippet below to `build.yml` to avoid running full builds on every docs change. 

```
paths-ignore:
- 'docs/**'
```
API docs should be now fully automated whenever anything new is added there which should make releasing less of a burden.

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -1,0 +1,42 @@
+name: update-api-documentation
+
+on:
+  push:
+    branches: [master]
+    tags: ['v*']
+    paths:
+    - 'docs/**'
+
+jobs:
+  push-to-documentation-branch:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Add key
+        env:
+          SSH_AUTH_SOCK: /tmp/ssh_agent.sock
+        run: |
+          mkdir -p ~/.ssh
+          ssh-keyscan github.com >> ~/.ssh/known_hosts
+          echo "${{ secrets.SERVER_DEPLOY_KEY }}" > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+          cat <<EOT >> ~/.ssh/config
+          Host github.com
+          HostName github.com
+          IdentityFile ~/.ssh/id_rsa
+          EOT
+      - name: Push generated API documentation to documentation branch
+        env:
+          USE_SSH: true
+          GIT_USER: irrelevant
+        run: |
+          git config --global user.email "irrelevant@completely.irrelevant"
+          git config --global user.name "irrelevant"
+          set -x
+          cd ..
+          git clone --depth 1 -b documentation git@github.com:SAP/cloud-sdk.git documentation
+          rsync -avz cloud-sdk/docs/api/ documentation/static/api/
+          cd documentation
+          git add -A
+          git commit -a -m "JS API documentation update via pipeline" || exit 0
+          git push
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,8 @@ on:
   push:
     branches: [master]
     tags: ['v*']
+    paths-ignore:
+    - 'docs/**'
 
 jobs:
   tests:


artemkovalyov:
@FrankEssenberger, you're right, the trigger for this is any change in `/docs` folder on `master` branch including when release is done. Then it pushes this to the `documentation` branch `/static/api` folder. And then the release job from the `documentation` branch is triggered. Not the release job for `master` if you asked about it.

FrankEssenberger:
> @FrankEssenberger, you're right, the trigger for this is any change in `/docs` folder on `master` branch including when release is done. Then it pushes this to the `documentation` branch `/static/api` folder. And then the release job from the `documentation` branch is triggered. Not the release job for `master` if you asked about it.

Ah ok. I was afraid it is the release job for master. Then it is fine.

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,42 @@
...
+    tags: ['v*']
+    paths:
+    - 'docs/**'
+
+jobs:
+  push-to-documentation-branch:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Add key
```
marikaner, line 15: Is there a reason, why we are not using the `GITHUB_TOKEN` here? If I am not mistaken it has the correct permission.
https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#using-the-github_token-in-a-workflow

You can find an example of how we use it to create releases here: https://github.com/SAP/cloud-sdk/blob/eb89e31e5ab42a77f8a00b997cfa404d504f46c1/.github/workflows/build.yml#L70

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,42 @@
...
+          IdentityFile ~/.ssh/id_rsa
+          EOT
+      - name: Push generated API documentation to documentation branch
+        env:
+          USE_SSH: true
+          GIT_USER: irrelevant
+        run: |
+          git config --global user.email "irrelevant@completely.irrelevant"
+          git config --global user.name "irrelevant"
+          set -x
```
marikaner, line 35: Can you add a comment what this is necessary for?

artemkovalyov: Without preconfiguring git with a username and email, it fails with a non-zero exit code which in turn fails the pipeline. I saw my PR towards facebook/docusaurus merged today with this same approach. A bit verbose but works just fine.

`set -x` configures a container's shell to outputs bash command before executing it. It helps very much to understand the pipeline logs, especially in case smth fails.

It's not a `copy/paste` coding just in case. Or do you mean adding a comment in a pipeline code?

marikaner: As this is not an everyday command (`set -x`) a comment would be helpful for the next person working on it.
Also: Isn't that the default behavior of github actions?

.github/workflows/api-docs.yml
artemkovalyov: I don't exactly recall every small detail, but it seems a `GITHUB_TOKEN` enables auth only for GitHub actions that are provided by GitHub. For CLI inside of a container it doesn't work. For that reason, you need either an ssh key or a token from a developer who has merge rights to the repo. To me, a key seems like a more universal and independent way for repositories used by teams instead of individuals.

I have nothing against updating this pipeline by someone more experienced with the GH actions API and container configuration. I haven't found a good manual on this topic or definitive best practices. For example, I believe checkout action can be probably used instead of my clone CLI command, but I don't know if it makes a shallow checkout or not and what directory structure gonna be to sync everything. Also, for push, we'll have to provide our push token or key to a 3rd party, which might be a security concern.

.github/workflows/api-docs.yml
artemkovalyov: Ah, my bad, it's ubiquitously used when you create docker containers and it sticks in my mind since then. I'll add a comment.

I don't think it's the case by default. Check this job: https://github.com/SAP/cloud-sdk/runs/672213308?check_suite_focus=true. The only first one on the list is printed.

Whereas after `set -x` you can see every command output with `+` sign here: https://github.com/SAP/cloud-sdk/runs/672505625?check_suite_focus=true. Helps when debugging and reading job's logs.

marikaner: Cool!

.github/workflows/api-docs.yml
marikaner: Unless it is a verified action I guess. But, ok, got it. And yes, we can also improve later on.

PR #145, opened 2020-05-19
feat: V4 converters

FrankEssenberger:
## Context

Included the payload and uri converters. I tried to put stuff used by v2 and v4 un the `common` folder. I had some issues with the imports and used sometimes very specific paths for the import instead of the `src` root which should in principle work but lead me some errors for example the `parsNumber()` method was not found. Please review this critically.

Regarding the tests I just put to files with an `v4` extension next to the v2 payload and uri converter tests. Perhaps it makes sense to also make a split there in v2, v4 and common? But this should be done in a seperate ticket I would say.

## Definition of Done

~~Release notes updated.~~
~~If applicable: Properly documented (JSDoc of public API)~~

## Labels
We use labels to indicate the status of PRs. 
Please select `please review`, `please merge` or `don't merge` for your PR.


packages/core/src/odata/v4/payload-value-converter.ts
```diff
@@ -40,117 +43,74 @@ type EdmTypeMapping = {
...
 }
 
-/**
- * @hidden
- */
-export function parseNumber(value: string | number): number {
-  if (typeof value === 'number') {
-    return value;
+function edmDurationToMoment(value: string): Duration {
+  const durationPattern = /([\+,\-]{1,1})?P(\d+D)?T(\d{1,2}H)?(\d{1,2}M)?(\d{1,2}S)?(\d{2,2}\.\d+S)?/;
```
marikaner, line 70:
```suggestion
  const durationPattern = /([\+,\-]{1,1})?P(\d{1,2})?(T(\d{1,2}H)?(\d{1,2}M)?(\d{1,2}S)?(\d{2,2}\.\d+S)?)?/;
```

Two points here:
* The D block can also have only two digits.
* The whole block including T is optional, if I read it correctly.

packages/core/src/odata/v4/payload-value-converter.ts
```diff
@@ -40,117 +43,74 @@ type EdmTypeMapping = {
...
+  const prefix = 'HH:mm:ss';
+  const validFormats = [`${prefix}`, `${prefix}.SSS`];
+  const parsed = moment.utc(value, validFormats, true);
+  if (!parsed.isValid()) {
+    throw new Error(
+      `Provided time-of-day value ${value} does not follow the Edm.TimeOfDay pattern: HH:mm:ss(.SSS)`
+    );
   }
+  return moment.utc({
+    y: 1970,
```
marikaner, line 90: Ouf. Isn't that why we have `Time`? Shouldn't we rather use that?

packages/core/src/odata/v4/payload-value-converter.ts
```diff
@@ -40,117 +43,74 @@ type EdmTypeMapping = {
...
-  const str = value.toString();
-  if (str.length >= targetLength) {
-    return str;
-  }
-  return '0'.repeat(targetLength - str.length) + str;
+function momentToEdmTimeOfDay(value: Moment): string {
+  return value.format('HH:mm:ss.SSS');
 }
 
 // Prettier-ignore
```
marikaner, line 116: Also not part of this PR, but can this be removed?

packages/core/src/odata/v4/payload-value-converter.ts
```diff
@@ -40,117 +43,74 @@ type EdmTypeMapping = {
...
 
-  if (value.toLowerCase() === 'inf') {
-    return Number.POSITIVE_INFINITY;
-  }
-  if (value.toLowerCase() === '-inf') {
-    return Number.NEGATIVE_INFINITY;
-  }
-  if (value.toLowerCase() === 'nan') {
-    return Number.NaN;
+function edmTimeOfDay(value: string): Moment {
```
marikaner, line 80: `edmTimeOfDay` should be `edmTimeOfDayToTime` if we change moment to time here.

packages/core/src/odata/v4/payload-value-converter.ts
marikaner: Maybe the optional T block is irrelevant?

packages/core/src/odata/v4/request-builder/request/uri-conversion/get-expansion.ts
```diff
@@ -0,0 +1,58 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+
+import { Entity } from '../../../entity';
```
marikaner, line 3: Where does this file come from and is it necessary?

packages/core/src/odata/v4/uri-conversion/uri-value-converter.ts
```diff
@@ -1,65 +1,35 @@
...
+  'Edm.Duration': value => `duration'${value}'`,
+  'Edm.Guid': identity
+};
 /**
  * @hidden
  */
 export function convertToUriFormat(
   value: any,
-  edmType: EdmTypeShared<'v4'>
+  edmType: EdmTypeShared<'v2'>
```
marikaner, line 27:
```suggestion
  edmType: EdmTypeShared<'v4'>
```

marikaner:
Two minor additions:
* `convertToUriForEdmString` does not need to be passed to v2 / v4 and then back to common
* Change the PR title (feat: ...)

packages/core/src/odata/common/edm-types.ts
```diff
@@ -12,17 +13,28 @@ export type EdmTypeCommon =
...
 export type EdmTypeShared<VersionT extends 'v2' | 'v4' | 'any'> =
   | EdmTypeCommon
   | EdmTypeV2
   | EdmTypeV4;
+
+/**
+ * for literal unions
+ * @example Sub<'Y' | 'X', 'X'> // === 'Y'
+ */
+export type Sub<O extends string, D extends string> = {
```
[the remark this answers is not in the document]
FrankEssenberger: yes. Strange that the first hit via google and stack overflow was not the correct solution. If

packages/core/src/odata/common/payload-value-converter.ts
```diff
@@ -0,0 +1,106 @@
...
+  const num = Number(value);
+
+  if (Number.isNaN(num)) {
+    throw new Error(`Cannot create number from input "${value}"`);
+  }
+
+  return num;
+}
+
+// Prettier-ignore
```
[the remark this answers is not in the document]
FrankEssenberger: this can all go away. The prettier ignore was there before, but I think it is ignored by eslint.

packages/core/src/odata/common/service.ts
```diff
@@ -0,0 +1,9 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+
+import { EntityBase } from './entity';
+
+export interface ServiceIdentifiable {
```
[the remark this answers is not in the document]
FrankEssenberger: hm, I do not know. I removed the file

packages/core/src/odata/v4/payload-value-converter.ts
FrankEssenberger: oh, yes good point. I created a extension to time containing fractional seconds.

packages/core/src/odata/v4/payload-value-converter.ts
FrankEssenberger: removed it.

packages/core/src/odata/v4/payload-value-converter.ts
FrankEssenberger: yes

packages/core/src/odata/v4/uri-conversion/uri-value-converter.ts
FrankEssenberger: copy paste error

packages/core/src/odata/v4/request-builder/request/uri-conversion/get-expansion.ts
FrankEssenberger: hm, no idea. Deleted

packages/core/src/odata/v2/uri-conversion/uri-value-converter.ts
```diff
@@ -1,9 +1,26 @@
...
+import { edmToTs, tsToEdm } from '../payload-value-converter';
 import { EdmTypeShared } from '../../common';
+import { uriConvertersCommon } from '../../common/uri-conversion/uri-value-converter';
+import { EdmType } from '../edm-types';
 
+type UriConverterMapping = { [key in EdmType]: (value: any) => string };
+/**
+ * @hidden
+ */
+export const uriConverters: UriConverterMapping = {
```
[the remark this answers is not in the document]
FrankEssenberger: 🥳

packages/core/src/odata/common/payload-value-converter.ts
```diff
@@ -0,0 +1,106 @@
...
+  'Edm.Single': parseNumber,
+  'Edm.String': identity
+};
+
+export const serializersCommom: EdmTypeMapping = {
+  'Edm.Binary': identity,
+  'Edm.Boolean': identity,
+  'Edm.Byte': toNumber,
+  'Edm.Decimal': fromBigNumber,
+  'Edm.Double': parseNumber,
```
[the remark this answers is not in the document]
FrankEssenberger: I agree. I will create a ticket. The reason is most likely that the round trip tests edm -> ts -> edm do not consider the +-inf cases and also the parser only checks the parseNumber for theses inputs never the `tsToEdm` or `edmToTs`.

packages/core/src/odata/v4/payload-value-converter.ts
FrankEssenberger: I did not understand the spec placeholders so well. If you say it can only have two digits and the whole T block is irrelevant I adjust it and add also tests.

PR #148, opened 2020-05-20
feat: support V4 collection 

jjtang1985:
## Context

- [x] add collection types to test vdm
    - [x] collection of basic edm types
    - [x] collection of complex types
- [x] adjust serializeEntity/deserializeEntity

## Definition of Done

- [ ] Release notes updated.

## Labels
We use labels to indicate the status of PRs. 
Please select `please review`, `please merge` or `don't merge` for your PR.


FrankEssenberger:
Sorry Junjie. I created a conflict with my converter PR being merged. Could you resolve the conflicts before I review.

jjtang1985:
Resolved;)

FrankEssenberger:
I was so busy writing some documentation today that I did not made it to the review. I realized that you resolved the conflicts yesterday which was a public holiday so I feel even more sorry.

packages/core/src/odata/v4/entity-deserializer.ts
```diff
@@ -1,8 +1,42 @@
...
+function getFieldValueV4<EntityT extends EntityBase, JsonT>(
+  json: Partial<JsonT>,
+  field: Field<EntityT> | Link<EntityT>
+) {
+  if (deserializer.isODataV2Field(field)) {
+    return deserializer.getFieldValue(json, field);
+  } else if (field instanceof CollectionField) {
+    return deserializeCollectionType(json[field._fieldName], field);
+  }
+}
```
marikaner, line 27: This function confuses me a lot:

1. First of all, I would expect `deserializer.isODataV2Field(field)` to always return false in the context of a function called `getFieldValueV4` - The field should never be a v2 field. I understand that this could probably be improved by different naming.
2. This is meant to add deserialization for collection types. I think we already discussed this and decided to add it to the original `getFieldValue` function https://github.com/SAP/cloud-sdk/blob/319be49d7dfe80f92811423375417fd7844bdab2/packages/core/src/odata/common/entity-deserializer.ts#L102. This would make the code simpler (no need to pass functions back to the common converter) and it does not interfere with the v2 implementation. In addition this can be used for wrong v2 implementations, that allow to have collections (e. g. SFSF).

packages/core/src/odata/v4/entity-serializer.ts
```diff
@@ -1,9 +1,34 @@
...
 import { tsToEdm } from './payload-value-converter';
-const serializer = entitySerializer(tsToEdm);
+import { CollectionField } from './selectable/collection-field';
+const serializer = entitySerializer(tsToEdm, updateSerialized);
 
 export const serializeEntity = serializer.serializeEntity;
 export const serializeEntityNonCustomFields =
   serializer.serializeEntityNonCustomFields;
+
+function updateSerialized(fieldValue, field, serialized) {
```
marikaner, line 13: Same ratio as for deserialization.

packages/core/src/odata/v4/entity-deserializer.ts
```diff
@@ -1,8 +1,42 @@
...
+
+function deserializeCollectionType<EntityT extends EntityBase>(
+  json: any[],
+  selectable: CollectionField<EntityT>
+) {
+  if (selectable._elementType instanceof EdmTypeField) {
+    const edmType = selectable._elementType.edmType;
+    return json.map(v => edmToTs(v, edmType));
+  } else if (selectable._elementType instanceof ComplexTypeField) {
```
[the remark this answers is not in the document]
marikaner: I found out that our linter was not configured correctly and created a PR for this. When this is merged you can fix these findings with `npm run lint:fix`.

packages/core/src/odata/v4/selectable/collection-field.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Entity } from '../entity';
+import {
+  ComplexTypeField,
+  Field,
+  SelectableEdmTypeField,
+  SimpleTypeFields
+} from '../../common/selectable';
+import { Constructable } from '../../common';
+
+export class CollectionField<EntityT extends Entity> extends Field<EntityT>
```
marikaner, line 11: Maybe we could also implement the [`in`](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_In) function (and add a test for it). This would help understand whether we have thought collections through for filters as well. This can also be done in a follow up, but as far as I can see this should be very simple and if it is not it would be good to understand this early on.

packages/core/test/test-util/test-services/v4/test-service/TestEntity.ts
```diff
@@ -115,6 +115,16 @@ export class TestEntity extends Entity implements TestEntityType {
...
+  /**
+   * Collection Property with string inside.
+   * @nullable
+   */
+  collectionPropertyWithString?: string[];
+  /**
+   * Collection Property with complex type inside.
+   * @nullable
+   */
+  collectionPropertyWithComplexType?: TestComplexType[];
```
marikaner, line 127: Is this a manual addition? If yes, please add it to either the generate-test-services script or the VDM somehow, otherwise it will disappear on regeneration and cause problems in the future.

packages/core/src/odata/v4/selectable/collection-field.ts
jjtang1985: - `in` function should be called from simple field e.g., `City` as string field -> `Address/City in ('Redmond', 'London')`
- `eq`/`ne` don't support collection field, since the reference service says:`The operand for a binary operator 'Equal' is not a single value. Binary operators require both operands to be single values.`

packages/core/src/odata/common/entity-deserializer.ts
```diff
@@ -13,12 +13,13 @@ import {
...
   EntityBase
 } from '../common';
+import { CollectionField } from '../v4/selectable/collection-field';
 
 // eslint-disable-next-line valid-jsdoc
 /**
  * @experimental This is experimental and is subject to change. Use with caution.
  */
-export function enityDeserializer(edmToTs) {
+export function entityDeserializer(edmToTs) {
```
marikaner, line 22: Oh, oops.

packages/core/src/odata/v4/selectable/collection-field.ts
marikaner: As discussed ;)

packages/core/src/odata/v4/selectable/collection-field.ts
```diff
@@ -0,0 +1,22 @@
...
```
marikaner, line 11: Can you add the note that this is experimental?

PR #149, opened 2020-05-22
Proxy documentation

FrankEssenberger:
## Context

This adds some technical "what is happening inside the sdk" documentation on two topic:
-Proxy handling
-Destination handling
I did this today since it was my last day before a two week holiday. So I will be not available for a long review process. I hope you can polish and merge this also without me :-).

I added @mr-flannery since he knows the details very well especially regarding the destination lookup.

@artemkovalyov I wanted to add the picture in a subfolder `img` in the JS part. But I was not able to do that with the `useBaseUrl` property. Perhaps you can adjust this, since you know the docusaurus much better then I do.


docs/js/features/connectivity/destination.md
```diff
@@ -0,0 +1,147 @@
...
+For simplicity an optional argument of the destination lookup has been neglected in the beginning:
+```
+.execute({"destinationName": 'myDestination', jwt: 'yourJWT})
+```  
+The `jwt` argument takes the JSON web token (JWT) issued by an XSUAA as an input.  
+If the JWT was issued by a XSUAA of a subscriber account it is possible to fetch the destination of this account.
+For fetching destinations of the provider accounts no JWT is needed because the application has a service binding to the destination service instance. 
+
+There are three selection strategies defined in the SDK: `alwaysSubscriber`, `alwaysProvider` and `subscriberFirst`.
+The selection strategy can be passed as an optional option to the `.execute()` method.
```
mr-flannery, line 129:
```suggestion
The selection strategy can be passed as an optional argument to the `.execute()` method. 
```

mr-flannery: A code example would probably also be helpful here.

docs/js/features/connectivity/destination.md
```diff
@@ -0,0 +1,147 @@
...
+On SAP Cloud Platform (SCP) this abstraction is provided by a so called destination object. 
+This object can be obtained at runtime of the application and contains information like:
+- URL
+- Authentication
+- Proxy Settings (see the [proxy documentation](./proxy.md) for more details)   
+...
+The Cloud SDK helps you receiving this object and provides also options for local testing outside the SCP.
+
+The SDK provides a generator to create a client or data model based on a service definition.
+For S/4 HANA this client has been created and published to npm.
```
artemkovalyov, line 34:
```suggestion
For S/4HANA this client has been created and published to npm. 
```

PR #150, opened 2020-05-25
fix: Generation of aggregator package

marikaner:
## Context

Fixes versions in the generator

## Definition of Done

- [x] Release notes updated.


packages/generator/src/aggregator-package/package-json.ts
```diff
@@ -22,11 +22,11 @@ export function packageJson(
           version: 'node ../../../after-version-update.js'
         },
         dependencies: dependencies.reduce((deps, service) => {
-          deps[service] = `^${generatorVersion}`;
+          deps[service] = `^${versionInPackageJson || generatorVersion}`;
```
jjtang1985, line 25: I think the type of this parameter: `versionInPackageJson` should be a `version 2 topic`.
It was introduced when going open source.
In order to avoid the breaking changes, we made it optional.
However, in real world, the vdm packages should always be different from the generator version.
Having a fall back value is good, but can hide potential wrong usage.
I want to make it mandatory parameter in version 2.

marikaner: Hmm, yes if you create a ticket this should be discussed. You only need this parameter when generating a package.json and then it can be mandatory. We should check whether we can achieve this.

PR #161, opened 2020-05-27
Adding Eslint, minor fixes to JS files

artemkovalyov:
## Context

Things are growing little by little and we have to observe a good standard. Next steps:
- Adding pre-commit GitHub action
- Adding a linting to Pipeline

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,26 @@
+module.exports = {
+  env: {
+    browser: true,
+    es6: true
+  },
+  extends: [
+    'plugin:react/recommended',
+    'standard'
+  ],
+  globals: {
+    Atomics: 'readonly',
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,26 @@
...
+  },
+  parserOptions: {
+    ecmaFeatures: {
+      jsx: true
+    },
+    ecmaVersion: 11,
+    sourceType: 'module'
+  },
+  plugins: [
+    'react'
```
marikaner, line 22: Is this for the JS SDK api docs?

artemkovalyov: Yep, the whole thing is rendered in React and without this plugin, it complains about imported but not used variables.

artemkovalyov: You're correct, it affects only docs.

PR #164, opened 2020-05-28
feat: V4 getall

jjtang1985:
## Context

Open discussion:
- [x] Assumption: the following items are common so they don't have to be tested in v4
    - destination
    - authentication
    - http related, e.g., csrf token
    - existing error handling

## Definition of Done

- [ ] Release notes updated.


marikaner:
And to your question in the description. I don't think this is necessary. Destination handling etc. is separate from OData.

packages/core/src/odata/common/entity-deserializer.ts
```diff
@@ -99,7 +99,9 @@ export function entityDeserializer(edmToTs) {
       return getLinkFromJson(json, field);
     }
     if (field instanceof ComplexTypeField) {
-      return deserializeComplexType(json[field._fieldName], field);
+      return json[field._fieldName]
```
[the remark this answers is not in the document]
jjtang1985: Actually, this is a bug fix and why the real response from the travel service makes sense.
The response contains
```
HomeAddress: null
```
where the home address is a complex type.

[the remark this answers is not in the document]
marikaner: Good catch

test-packages/integration-tests/test/request-builder-v4.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  it('should resolve for all field', async () => {
+    nock(destination.url, {
+      reqheaders: {
+        authorization: basicCredentials(destination),
+        accept: 'application/json',
+        'content-type': 'application/json'
+      }
+    })
+      .get(`${servicePath}/${entityName}?$format=json&$select=*`)
```
marikaner, line 68: Minor: I think it is generally a better approach to use the object style notation for query strings (https://github.com/nock/nock#specifying-request-query-string), because it makes easier to account for order and encoding. I'll leave this up to you ;)

test-packages/integration-tests/test/request-builder-v4.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    const request = TestEntity.requestBuilder()
+      .getAll()
+      .select(TestEntity.ALL_FIELDS)
+      .execute(destination);
+
+    await expect(request).resolves.not.toThrow();
+  });
+
+  it('should allow setting custom headers', async () => {
```
marikaner, line 79: Should that be part of the getAll request test? This is in the base class for all requestbuilders and nobody is overwriting this. I assume this was in the v2 getall spec because the super class is abstract. This should probably be in a separate file (request-builder-base.spec.ts) and be removed from the v2 / v4 specific tests.

test-packages/integration-tests/test/request-builder-v4.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+        authorization: 'customcustom'
+      })
+      .execute({
+        url: destinationServiceUri
+      });
+
+    await expect(request).resolves.not.toThrow();
+  });
+
+  it('should allow setting custom query parameters', async () => {
```
marikaner, line 102: Same as above

PR #169, opened 2020-06-02
chore: Skip canary release on tag

marikaner:
Fix pipeline to only create one release

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,7 +41,7 @@ jobs:
       - run: npm run check:license
         name: License Check
   canary-release:
-    if: github.event_name != 'pull_request'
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - run: npm run check:license
         name: License Check
   canary-release:
-    if: github.event_name != 'pull_request'
+    if: github.event_name != 'pull_request' || !startsWith(github.ref, 'refs/tags/v')
```
jjtang1985, line 44:
```suggestion
    if: github.event_name != 'pull_request' && !startsWith(github.ref, 'refs/tags/v')
```

marikaner: Oops

PR #170, opened 2020-06-03
feat: V4 get by key

jjtang1985:
## Context

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/core/src/odata/v4/request-builder/get-by-key-request-builder.ts
+++ b/packages/core/src/odata/v4/request-builder/get-by-key-request-builder.ts
@@ -18,6 +18,7 @@ import { MethodRequestBuilderBase } from '../../common/request-builder/request-b
 import { ODataGetByKeyRequestConfig } from '../../common/request/odata-get-by-key-request-config';
 import { Expandable } from '../../common/expandable';
 import { oDataUri } from '../uri-conversion';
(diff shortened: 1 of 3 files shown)


packages/core/src/odata/v4/request-builder/get-by-key-request-builder.ts
```diff
@@ -87,11 +86,3 @@ export class GetByKeyRequestBuilder<EntityT extends Entity>
       );
   }
 }
-
-/*
-C4C response to getByKey requests with the collection response format instead of the single element one
-To account for this, we test for this and use the normal format if `.result` return undefined.
-*/
-function extractData(response: HttpReponse): MapType<any> {
-  return response.data.d.results || response.data.d;
```
marikaner, line 96: Why did you remove this? I think this was there because some APIs (e. g. C4C) do that wrong. I think it would probably be best so have one file where we collect these extractions, because this is used in multiple places, if I am not mistaken.

jjtang1985: Discussed, we don't know whether c4c has the same way (using collection instead of single item) like v2.
But keeping the extractData function helps make abstract class and a single file for collecting all these extraction in one location.

PR #175, opened 2020-06-05
feat: V4 create

jjtang1985:
## Context

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/core/src/odata/v4/request-builder/create-request-builder.ts
+++ b/packages/core/src/odata/v4/request-builder/create-request-builder.ts
@@ -1,6 +1,6 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 
-import { errorWithCause } from '@sap-cloud-sdk/util';
+import { errorWithCause, MapType } from '@sap-cloud-sdk/util';
 import { Constructable, EntityIdentifiable, Link } from '../../common';
 import { MethodRequestBuilderBase } from '../../common/request-builder/request-builder-base';
 import { ODataCreateRequestConfig } from '../../common/request/odata-create-request-config';
@@ -13,6 +13,7 @@ import {
   DestinationNameAndJwt
 } from '../../../scp-cf/destination-service-types';
 import { oDataUri } from '../uri-conversion';
+import { HttpReponse } from '../../../http-client';
 /**
  * Create OData request to create an entity.
  *
@@ -86,7 +87,7 @@ export class CreateRequestBuilder<EntityT extends Entity>
       .then(request => request.execute())
       .then(response =>
         deserializeEntity(
-          response.data,
+          extractData(response),
           this._entityConstructor,
           response.headers
         )
@@ -95,4 +96,8 @@ export class CreateRequestBuilder<EntityT extends Entity>
         Promise.reject(errorWithCause('Create request failed!', error))
       );
   }
+}
+
+function extractData(response: HttpReponse): MapType<any> {
+  return response.data;
 }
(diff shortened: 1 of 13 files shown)


FrankEssenberger:
In this case the context is rather clear: create for odata v4. But in general one or two sentences would be nice.

packages/core/src/odata/v4/request-builder/create-request-builder.ts
```diff
@@ -86,7 +87,7 @@ export class CreateRequestBuilder<EntityT extends Entity>
       .then(request => request.execute())
       .then(response =>
         deserializeEntity(
-          response.data.d,
+          extractData(response),
```
FrankEssenberger, line 90: could you make a second PR, where you look for the many hard coded `.data.d` usage and replace them all.

Additional optimization: Is this something we can know from the system we are calling for example S/4 use `data.d`, C/4 something I do not know, concur XYZ. Perhaps it makes then sense to include this extraction in the builder or so. I do not know.

packages/core/src/odata/v4/request-builder/create-request-builder.ts
```diff
@@ -1,6 +1,6 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 
-import { errorWithCause } from '@sap-cloud-sdk/util';
+import { errorWithCause, MapType } from '@sap-cloud-sdk/util';
```
FrankEssenberger, line 3: I compared this once with the v2 create request builder. I did the same also for the get all and get by key and they are also very similar. Please create a follow up ticket that we do not forget to reduce this code duplication. Obvious way would be base classes but mixin could be a more TS way to do that.

packages/core/test/test-util/testPostRequestOutcome.ts
```diff
@@ -0,0 +1,7 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: the file name is not kebap case

test-packages/integration-tests/test/test-data/single-test-entity-response-v4.ts
```diff
@@ -0,0 +1,62 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: Also create a ticket that we remember to setup a real E2E test with a CAP service or one of the two v4 services.

packages/core/src/odata/v4/request-builder/create-request-builder.ts
jjtang1985: - had a discussion with Marika and decided to make abstract request builders and move the `.data.d` to a single file -> [ticket](https://github.wdf.sap.corp/MA/scp-rocket/issues/1257).
- for different systems, if there is a clear rule, we can use a system-to-extractData mapping. However, even for c/4, the known rule is only for v2 not v4. Also, we decided to also support collection in v2, so SFSF can also make use of the collection type now.

packages/core/src/odata/v4/request-builder/create-request-builder.ts
jjtang1985: see comment above.

test-packages/integration-tests/test/test-data/single-test-entity-response-v4.ts
jjtang1985: It's on the radar, but you are right, a ticket is good for tracking.
-> https://github.wdf.sap.corp/MA/scp-rocket/issues/1258

PR #186, opened 2020-06-11
feat:  Delete Odata v4

FrankEssenberger:
## Context

Like for Junjies PRs the etag handling is not included. I used the update and create PRs as a blue print. @jjtang1985 Could you comment why you added tests in the `core/test/request-builder/update-request-builder-v4.spect.ts` and in `integration-tests/test/request-builder-v4.spec.ts`? For me the tests seem to be equivalent since they both mock the response in the same way.

## Definition of Done

- [X] Tests created/adjusted for your changes.
~~Release notes updated.~~ 
I guess we add a big note once the v4 gets to the public export.
- [X] PR title adheres to [conventional commit guidelines](https://www.conventionalcommits.org) (please note that only `fix:` and `feat:` will end up in the release notes)
- [X] If applicable: Properly documented (JSDoc of public API)
- [X] If applicable: Check if `node run doc` still works.


jjtang1985:
@FrankEssenberger , to answer your question, the idea of the current test strategy is that we trust the v2 tests (unit/integration/type tests) are reliable enough so we can basically do the the same for v4. However, some common functions, e.g., destination/authentication/custom fields, should not be test for a second time.

Then your question is more or less like what's the difference between the `get-all-request-builder.spec.ts` (`unit test`) and `request-builder.spec.ts` (`integration-test`). 
- the unit test is testing all the functions of a `GetAllRequstBuilder`, e.g., `builder()`/`top()`/`execute()`. [1]
- the integration test is testing some complicated scenario that a unit test cannot cover, e.g., authentication/destination/csrf token. [2]
- for unit test, we create request builder from the code in the same package like below, which is not the same case as a user: [3]
```
import { GetAllRequestBuilder } from '../../src';
import { TestEntity } from '../test-util/test-services/v2/test-service';
new GetAllRequestBuilder(TestEntity);
```
- for integration test, we import the vdm from some other package, which is the same as how users do with the real vdms released in the internal npm: [4]
```
import {
  TestEntity
} from '@sap-cloud-sdk/test-services/test-service';
TestEntity.requestBuilder();
```
Yes, you are right. After removing the tests for common functions, the unit test and integration test look very similar, since the difference from [1] and [2] does not exist.
But I would argue, keeping them in different level still makes sense, since the difference from [3] and [4] is still there, especially when considering some typescript specific issues like circular dependencies.

FrankEssenberger:
I considered the changes. Feel free to re-review.

PR #187, opened 2020-06-12
feat: V4 generator

marikaner:
## Context

Adjust the generator to cover basic functionality of v4.
This includes:
* generating correct v4 VDM with correct imports (`/v4` suffix), `OneToManyLink` instead of `Link`
* parsing EDMX files
  * with adjusted navigation properties
  * with collection types
  * and not failing when there are base types or enum types
 
* Currently missing and out of scope for this PR:
  * Correct behavior for navigation properties when there is a base type
  * Function Imports (currently they are handled the same way as v2)
  * Better handling of derived types, currently all properties are just duplicated

TODO (in scope):
- [ ] Tests

**Notes for the Reviewer**: Please take note of the "new" file structure, it is quite verbose, some things could be merged, some things could be structured differently. Please share your ideas, I am not 100% happy with the current state.

## Definition of Done

- [ ] Release notes updated.


packages/generator/src/parser/parsed-service-metadata.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import {
+  EdmxMetadataBaseExtended,
+  SwaggerMetadata
+} from './parser-types-common';
+import { EdmxMetadata as EdmxMetadataV2 } from './parser-types-v2';
+import { EdmxMetadata as EdmxMetadataV4 } from './parser-types-v4';
+
+export interface ParsedServiceMetadata<
+  T extends 'v2' | 'v4' | unknown = unknown
```
jjtang1985, line 11: could you please explain a bit regarding the use case for unknown?

packages/generator/src/parser/edmx-parser-common.ts
```diff
@@ -0,0 +1,83 @@
...
+} from './parser-types-common';
+
+const logger = createLogger({
+  package: 'generator',
+  messageContext: 'edmx-parser'
+});
+
+export function parseBaseMetadata(
+  root,
+  oDataVersion: 'v2' | 'v4',
```
FrankEssenberger, line 21: Minor: I have checked and the literal type `'v2' | 'v4'` appears almost a dozen times in the code. I would prefer to define it  like ODataVersionType and use it. Makes find and replace easier.

test-resources/service-specs/v4/API_TEST_SRV/API_TEST_SRV.edmx
```diff
@@ -0,0 +1,247 @@
+<?xml version="1.0" encoding="utf-8"?>
```
FrankEssenberger, line 1: General question: When I run `npm run generate:test-services`. I thought it will generate the services for v2 and v4 given in these files and generate them also coping the files to the `packages/core/test/test-util/test-services/v4/test-service/` and `packages/core/test/test-util/test-services/v2/test-service/`. I think this is what happening but I wonder why there is no change in the entities? Is the v4 edmx files designed in a way to keep the existing code?

packages/generator/src/parser/parsed-service-metadata.ts
marikaner: `unknown` really means unknown. We don't which it will be yet, only after parsing this is known.

test-resources/service-specs/v4/API_TEST_SRV/API_TEST_SRV.edmx
marikaner: Yes, but not completely. You won't have changes if I already re-generated, though.

packages/generator/src/imports.ts
```diff
@@ -41,11 +42,13 @@ export function externalImportDeclaration(
...
 export function coreImportDeclaration(
-  namedImports: string[]
+  namedImports: string[],
+  oDataVersion: 'v2' | 'v4'
 ): ImportDeclarationStructure {
+  const moduleSuffix = oDataVersion === 'v4' ? '/v4' : '';
   return {
     kind: StructureKind.ImportDeclaration,
-    moduleSpecifier: '@sap-cloud-sdk/core',
+    moduleSpecifier: `@sap-cloud-sdk/core${moduleSuffix}`,
```
FrankEssenberger, line 51: If I search for `@sap-cloud-sdk/core/v4` I do not find a thing. I would expect that the generated test eintities need to import the request builders etc. from the v4 path?

FrankEssenberger: Or is there some test for this?

marikaner: This is because this currently does not generate a v4 version for the type tests. I will add this to the todo list. You can find a v4 in the generated test-utils of core, though

packages/generator/src/imports.ts
```diff
@@ -59,15 +62,17 @@ export function corePropertyTypeImportNames(
 export function corePropertyFieldTypeImportNames(
   properties: VdmProperty[]
 ): string[] {
-  return unique(
-    properties.filter(prop => !prop.isComplex).map(prop => prop.fieldType)
-  );
+  return unique([
+    ...properties.filter(prop => !prop.isComplex).map(prop => prop.fieldType),
+    ...(properties.some(prop => prop.isMulti) ? ['CollectionField'] : [])
```
FrankEssenberger, line 67: I find the name `isMulti` a bit missleading. Sounds like multi link or so. If isMulti releates to collection Field yes/no then I would call it: `isCollectionField`

marikaner: I was actually considering calling this `isCollection` and using it for both navigation props and props.

packages/generator/src/parser/edmx-parser-v4.ts
```diff
@@ -0,0 +1,115 @@
...
+    ...entityType,
+    Property: entityType.Property.filter(
+      prop => !enumTypeNames.includes(stripNamespace(parseTypeName(prop.Type)))
+    )
+  }));
+}
+
+export function parseEdmxV4(
+  root
+): Omit<EdmxMetadata, keyof ReturnType<typeof parseBaseMetadata>> {
```
FrankEssenberger, line 93: could this replaced be with: ` Omit<EdmxMetadata, keyof EdmxMetadataBase> `?

packages/generator/src/parser/parser-types-v2.ts
```diff
@@ -0,0 +1,69 @@
...
+  entityTypes: EdmxEntityType[];
+  associationSets: EdmxAssociationSet[];
+  associations: EdmxAssociation[];
+  functionImports: EdmxFunctionImport[];
+  complexTypes: EdmxComplexTypeBase[];
+}
+
+export function isV2Metadata(
+  metadata: EdmxMetadataBase
+): metadata is EdmxMetadata {
```
FrankEssenberger, line 22: is a boolean returned? what is metadata is EdmxMetadata for a type?

packages/generator/src/parser/edmx-parser.ts
```diff
@@ -1,111 +1,44 @@
...
-    selfLink: parseLink(root)
+  const metaData = parseBaseMetadata(
+    root,
+    getODataVersion(parsedEdmx),
+    edmxPath
+  );
+
+  return {
+    ...metaData,
+    ...(metaData.oDataVersion === 'v2' ? parseEdmxV2(root) : parseEdmxV4(root))
```
FrankEssenberger, line 38: try to find all swtiches between v2 and v4 and use the isV2() method you introduced already everywhere so that it is easier to find all the places where a switch is.

packages/generator/src/parser/parser-types-v2.ts
marikaner: This is called a type guard in Typescript (https://www.typescriptlang.org/docs/handbook/advanced-types.html#typeof-type-guards)

packages/generator/src/parser/parser-types-v4.ts
```diff
@@ -0,0 +1,56 @@
...
+  EdmxEntitySetBase,
+  EdmxFunctionImportBase,
+  EdmxComplexTypeBase,
+  EdmxNamed
+} from './parser-types-common';
+
+export interface EdmxMetadata extends EdmxMetadataBase {
+  entitySets: EdmxEntitySet[];
+  entityTypes: EdmxEntityType[];
+  enumTypes: EdmxEnumType[];
```
FrankEssenberger, line 15: In the final `VdmServiceMetadata` created by the parse methods I do not see a representation of the enumTypes. Only EntitySets, functionImports and complexTypes where the Entity sets reference the entityTypes. Where do the enumTypes go?

marikaner: This is not implemented, I only added enum types, to make sure it does not fail (and initially to have a difference between v2 / v4

FrankEssenberger:
## Problem:
- Parse is version dependent: parseEdmxV4 parseEdmxV2 with different return types
The read and parsing of the data is the same up th the point when on the root the two different parseEdmxV2 are called in parseEdmxFile.
- Transformation is also version dependent transformServiceMetadata(). The transformation is mainly the same.

The result of parseService() is VdmServiceMetadata. This does not depend on the service.
So perhaps on can get rid of all the intermediate parser-types where all the objects like:
entitySets, functionImports, complexTypes. Potential flow

- You keep ony the real base containing the root entity holding all plain read data longer.
- Then in the transformServiceMetadata()  you build the entites, compexTypes etc. with the right methods. The results do not depend on the version e.g. VdmEntity.
- General the transformServiceMetadata is too large and a lot of stuff is going on. Most of it also the same for both versions. Consider using some pipe to add all version independent content like:
oDataVersion: metadata.edmx.oDataVersion, namespace, originalFileName, directoryName, npmPackageName,  speakingModuleName and then call slimm methods for the rest.

packages/generator/src/parser/parser-types-v4.ts
marikaner: They are not implemented yet. In the end I assume we will have a common representation, that just won't be used in the v2 case.

packages/generator/src/parser/parsed-service-metadata.ts
marikaner: I removed that conditional type, as this is only relevant for two modules and can also be achieved by type assertions.

packages/generator/src/parser/edmx-parser-common.ts
marikaner: done

packages/core/src/odata/common/selectable/collection-field.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export class CollectionField<EntityT extends EntityBase> extends Field<EntityT>
+  implements SelectableEdmTypeField {
+  readonly selectable: true;
+  constructor(
+    readonly _fieldName: string,
+    readonly _entityConstructor: Constructable<EntityT>,
+    // TODO: for simple type fields this can be improved to just use the edm type
+    // for complex types this is a little more complex due to the nature of complex types and should be revisted latest for v2.0
+    readonly _fieldType: SimpleTypeFields<EntityT> | ComplexTypeField<EntityT>
```
FrankEssenberger, line 18: General question, I know this is the way before. But should the collection type have two generic parameters: one for the entity and one for the type of the internal data type?

FrankEssenberger: Perhaps a thing we could als add a ticket for?

marikaner: I agree, we actually said, this would be done in a follow up, not sure what happened here. @jjtang1985 do we already have a follow up for this?

packages/generator/src/complex-type/file.ts
```diff
@@ -1,20 +1,21 @@
...
 export function complexTypeSourceFile(
-  complexType: VdmComplexType
+  complexType: VdmComplexType,
+  service: VdmServiceMetadata
 ): SourceFileStructure {
   return {
     kind: StructureKind.SourceFile,
     statements: [
-      ...importDeclarations(complexType),
+      ...importDeclarations(complexType, service),
```
FrankEssenberger, line 18: If you just use the odata version also just pass the odata version

packages/generator/src/entity/imports.ts
```diff
@@ -26,13 +27,17 @@ export function importDeclarations(
       [
         ...corePropertyTypeImportNames(entity.properties),
         ...corePropertyFieldTypeImportNames(entity.properties),
-        ...coreNavPropertyFieldTypeImportNames(entity.navigationProperties),
+        ...coreNavPropertyFieldTypeImportNames(
+          entity.navigationProperties,
+          service
```
FrankEssenberger, line 32: same here. If you just use the version in the end to decide which type of link to put, only pass the version.

packages/generator/src/entity/namespace.ts
```diff
@@ -44,28 +46,64 @@ function properties(entity: VdmEntity): VariableStatementStructure[] {
   return entity.properties.map(prop => property(prop, entity));
 }
 
-function property(
+function getPropertyFieldClassName(prop: VdmProperty): string {
+  return prop.isMulti ? 'CollectionField' : prop.fieldType;
+}
+
+function getPropertyFieldInitializer(
```
FrankEssenberger, line 53: Is there a property not related to a field? Is getFieldClassName not enough as a name? Also for the other methods.

packages/generator/src/complex-type/file.ts
marikaner: Yeah, I thought of that, but found the other way a little more consistent with the rest. Anyways, no big deal, I changed it.

packages/generator/src/entity/namespace.ts
marikaner: done

packages/generator/src/function-import/import.ts
```diff
@@ -26,21 +26,24 @@ export function importDeclarations(
...
-      'FunctionImportRequestBuilder',
-      'FunctionImportParameter'
-    ]),
+    coreImportDeclaration(
+      [
+        ...corePropertyTypeImportNames(functionImportParameters),
+        ...returnTypes.map(returnType =>
+          responseTransformerFunctionName(returnType)
+        ),
+        ...(returnTypes.some(
```
FrankEssenberger, line 35: Minor: could you perhaps these 10 lines in a little helper function like responseTransformerFunctionName so that the code looks a bit cleaner.

marikaner: Can you specify the lines you mean?

marikaner: I now just assumed 35 and ongoing

marikaner:
Tests are failing as long as I don't regenerate, which I will do only after the review, to make it easier.

packages/generator/src/function-import/import.ts
FrankEssenberger:
```
...(returnTypes.some(
          returnType =>
            returnType.returnTypeCategory ===
            VdmFunctionImportReturnTypeCategory.EDM_TYPE
        )
          ? ['edmToTs']
          : [])
```

packages/generator/src/generator-utils.ts
```diff
@@ -134,8 +134,15 @@ export function endWithDot(text: string): string {
   return !text || text.endsWith('.') || text.endsWith(':') ? text : `${text}.`;
 }
 
-export function linkClass(navProperty: VdmNavigationProperty): string {
-  return navProperty.isMultiLink ? 'Link' : 'OneToOneLink';
+export function linkClass(
+  navProperty: VdmNavigationProperty,
+  service: VdmServiceMetadata
+): string {
+  return navProperty.isCollection
```
FrankEssenberger, line 141: also here only pass the odata version

marikaner: Already done. Make sure to look at "Changes from all commits". This is a little confusing on GH.

packages/generator/src/generator.ts
```diff
@@ -226,7 +226,7 @@ export async function generateSourcesForService(
     sourceFile(
       serviceDir,
       `${entity.className}RequestBuilder`,
-      requestBuilderSourceFile(entity),
+      requestBuilderSourceFile(entity, service),
```
FrankEssenberger, line 229: if I see it right also for this and the `complexTypeSourceFile()` below only the version from the service is actually needed.

packages/generator/src/imports.ts
```diff
@@ -59,15 +62,17 @@ export function corePropertyTypeImportNames(
 export function corePropertyFieldTypeImportNames(
   properties: VdmProperty[]
 ): string[] {
-  return unique(
-    properties.filter(prop => !prop.isComplex).map(prop => prop.fieldType)
-  );
+  return unique([
+    ...properties.filter(prop => !prop.isComplex).map(prop => prop.fieldType),
+    ...(properties.some(prop => prop.isCollection) ? ['CollectionField'] : [])
```
FrankEssenberger, line 66: Minor: Isn`t there a more elegant way to add the 'CollectionField' to the list? Wrapping it inside an [] and then ... it feels kind of brute force.

packages/generator/src/imports.ts
```diff
@@ -59,15 +62,17 @@ export function corePropertyTypeImportNames(
...
+  ]);
 }
 
 export function coreNavPropertyFieldTypeImportNames(
-  navProperties: VdmNavigationProperty[]
+  navProperties: VdmNavigationProperty[],
+  service: VdmServiceMetadata
 ): string[] {
-  return unique(navProperties.map(navProp => linkClass(navProp)));
+  return unique(navProperties.map(navProp => linkClass(navProp, service)));
```
FrankEssenberger: See above if service or just version is needed.

packages/generator/src/parser/edmx-parser-v4.ts
```diff
@@ -0,0 +1,113 @@
...
+  return types.map(type =>
+    type.BaseType ? addBaseTypeToType(type, types, joinTypes) : type
+  );
+}
+
+function addBaseTypeToType<T extends EdmxDerivedType>(
+  type: T,
+  types: T[],
+  joinTypes: (type: T, baseType: T) => T
+) {
```
FrankEssenberger, line 33: Could you add the return type also here. Most likely just T.

packages/generator/src/imports.ts
marikaner: The only way I can think of would be:
```
const x = unique(properties.filter(prop => !prop.isComplex).map(prop => prop.fieldType));
return properties.some(prop => prop.isCollection) ? [...x, 'CollectionField'] : x;
```
I guess then it is a matter of taste. I think I like my approach above more ;)

packages/generator/src/parser/edmx-parser-v4.ts
```diff
@@ -0,0 +1,113 @@
...
+  const baseType = types.find(e => e.Name === stripNamespace(type.BaseType!));
+
+  if (!baseType) {
+    throw new Error(
+      `Type ${type.BaseType} not found, but defined as BaseType of Type ${type.Name}.`
+    );
+  }
+
+  if (baseType.BaseType) {
+    return joinTypes(type, addBaseTypeToType(baseType, types, joinTypes));
```
FrankEssenberger, line 43: recursion always breaks my head. Could you perhaps add a little JS doc which explains what this message does.

packages/generator/src/parser/edmx-parser-v4.ts
```diff
@@ -0,0 +1,113 @@
...
+  parseEntityTypes,
+  parseFunctionImports,
+  parseComplexTypes
+} from './edmx-parser-common';
+import { stripNamespace, parseTypeName } from './parser-util';
+import { EdmxMetadataBase } from './parser-types-common';
+
+function joinTypesWithBaseTypes<T extends EdmxDerivedType>(
+  types: T[],
+  joinTypes: (type: T, baseType: T) => T
```
FrankEssenberger, line 22: I know this is all private., but could you create a type and then mark the two implementations `joinComplexTypes` and `joinEntityTypes` to have this type.

packages/generator/src/parser/edmx-to-vdm-v2.ts
```diff
@@ -0,0 +1,140 @@
...
+export function joinAssociationMetadata(
+  associationSets: EdmxAssociationSet[],
+  associations: EdmxAssociation[]
+): JoinedAssociationMetadata[] {
+  return associationSets.map(assocSet => {
+    const matchingAssoc = associations.find(
+      a => a.Name === stripNamespace(assocSet.Association)
+    );
+
+    if (!matchingAssoc) {
```
FrankEssenberger, line 32: Why this can never or should happen? Can you explain what is the content of the associationSets and the associations.

packages/generator/src/parser/edmx-parser-v4.ts
marikaner: I would prefer to keep it simple here. The function clearly states that we need something that takes two things of the same type and returns something of the same type. If I add an additional Type, this will be obfuscated, because I would have to check the type definition to know this.
I agree, that it would be nicer to somehow know that the existing joinXXX methods are the ones used for that, but I don't think it is worth changing that.

packages/generator/src/parser/edmx-to-vdm-v2.ts
marikaner: This is the old v2 behavior. If the edmx file is corrupt is it possible that there is no corresponding association found.

packages/generator/src/parser/edmx-to-vdm-v4.ts
```diff
@@ -0,0 +1,70 @@
...
+  transformEntity,
+  navigationPropertyBase
+} from './edmx-to-vdm-common';
+
+function navigationProperties(
+  entityMetadata: JoinedEntityMetadata,
+  classNames: { [originalName: string]: string },
+  formatter: ServiceNameFormatter
+): VdmNavigationProperty[] {
+  const entityType = entityMetadata.entityType as EdmxEntityType;
```
FrankEssenberger, line 22: Ah here is the cast now. I like it much better than the conditional type.

packages/generator/src/parser/edmx-to-vdm-v4.ts
```diff
@@ -0,0 +1,70 @@
...
+    return {
+      ...navigationPropertyBase(
+        navProp.Name,
+        entityMetadata.entitySet.Name,
+        formatter
+      ),
+      from: entityMetadata.entityType.Name,
+      to: navBinding.Target,
+      toEntityClassName: classNames[navBinding.Target],
+      multiplicity: isCollectionType ? '1 - *' : '1 - 1',
```
FrankEssenberger, line 47: What about an enum? Is the string used somewhere and if so where? Also the `isMultiLink` not redundant information?

FrankEssenberger: or multiplicity is redundant if you have the boolean.

packages/generator/src/parser/parser-types-common.ts
```diff
@@ -1,109 +1,66 @@
...
-  Name: string;
+export interface EdmxComplexTypeBase extends EdmxNamed {
   Property: EdmxProperty[];
 }
 
-export interface EdmxEntitySet {
+export interface EdmxEntitySetBase extends EdmxNamed {
   EntityType: string;
-  Name: string;
   'sap:content-version': string;
```
FrankEssenberger, line 47: I find these kind of strings e.g. `sap:content-version` in multiple places in the code. Does it make sense to replace them via a key collection or strings?

packages/generator/src/parser/parser-types-common.ts
```diff
@@ -163,3 +120,14 @@ export interface SwaggerPathParameter extends SwaggerDescribed {
 export interface SwaggerDescribed {
   description?: string;
 }
+
+export interface JoinedEntityMetadata {
```
FrankEssenberger, line 124: Minor: I find this name not so good. `SetWithType` perhaps? The swagger definition is perhaps part of the entitySet?

packages/generator/src/parser/parser-util.ts
```diff
@@ -0,0 +1,24 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+
+export function stripNamespace(name: string): string {
+  const nameParts = name.split('.');
```
FrankEssenberger, line 4: Here an exception would be good so that we realize if something goes wrong and the input is not: `myNamesape.myName`

packages/generator/src/parser/parser-util.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function isCollection(typeName: string): boolean {
+  return collectionRegExp.test(typeName);
+}
+
+export function parseTypeName(typeName: string): string {
+  return isCollection(typeName) ? parseCollectionTypeName(typeName) : typeName;
+}
+
+export function parseCollectionTypeName(typeName: string): string {
```
FrankEssenberger, line 18: Why this is exported?

packages/generator/src/parser/parser-util.ts
marikaner: As long as `name` is a string this won't fail.

packages/generator/src/parser/parser-util.ts
```diff
@@ -0,0 +1,24 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+
+export function stripNamespace(name: string): string {
+  const nameParts = name.split('.');
+  return nameParts[nameParts.length - 1];
+}
+
+const collectionRegExp = /Collection\((?<collectionType>.*)\)/;
```
FrankEssenberger, line 8: In the API v4 edmx I find the following collection type: `Collection(API_TEST_SRV.A_TestComplexType)`. This would mean the regex would take: `API_TEST_SRV.A_TestComplexType` right? But the namespace in front should be part of the type? Is this tested?

packages/generator/src/typedoc.ts
```diff
@@ -73,7 +73,7 @@ export function getNavPropertyDescription(
   property: VdmNavigationProperty
 ): string {
   return `${
-    property.isMultiLink ? 'One-to-many' : 'One-to-one'
+    property.isCollection ? 'One-to-many' : 'One-to-one'
```
FrankEssenberger, line 76: I know it si only doc, but perhaps align it with the `1 -1` and `1-N` case mentioned above.

packages/generator/src/vdm-types.ts
```diff
@@ -58,7 +61,13 @@ export interface VdmNavigationProperty extends VdmPropertyDescriptor {
   from: string;
   to: string;
   toEntityClassName: string;
-  multiplicity: string;
+  /**
+   * @deprecated Since v1.21.3. Will not be replaced.
+   */
+  multiplicity?: string;
+  /**
+   * @deprecated Since v1.21.3. Use `isCollection` instead.
```
FrankEssenberger, line 69: In this interface I do not see a `isCollection` what is this referring to?

packages/generator/test/parser/edmx-to-vdm-v4.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+      service.edmx.complexTypes,
+      formatter,
+      new Set()
+    );
+
+    const entity = transformEntitiesV4(service, vdmComplexTypes, formatter)[0];
+    expect(entity.properties[0]).toMatchObject({
+      isCollection: true,
+      isComplex: true,
+      edmType: 'namespace.TestComplexType',
```
FrankEssenberger, line 57: Ah, I found the test. So namespace is part of the type?

packages/generator/src/vdm-types.ts
```diff
@@ -1,8 +1,10 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 
 import { PathLike } from 'fs';
+import { ODataVersion } from '@sap-cloud-sdk/util';
 
 export interface VdmServiceMetadata {
+  oDataVersion: ODataVersion;
```
FrankEssenberger, line 7: Isn't this addition and the `isCollection` a breaking change?  I assume nobody will use this object outside of our code but if you see some risk we should do the usual depreciation and add a new object.

packages/generator/src/parser/parser-util.ts
FrankEssenberger: true, but give a unexpected result. For a exported method I would like to have input validation. Even if it is that simple.

packages/generator/src/parser/edmx-parser-v4.ts
FrankEssenberger: Ok not critical and only private.

packages/generator/src/imports.ts
FrankEssenberger: aggreed.

packages/generator/src/complex-type/file.ts
FrankEssenberger: thanks. I try to keep interfaces as tight as possible. Then you can already guess ok it will have a v2/v4 switch.

packages/generator/src/parser/parser-util.ts
marikaner: I intended to use it directly, but then didn't. I can remove the export.

packages/generator/src/parser/parser-util.ts
marikaner: Good point, I didn' test this directly. I will add some unit tests.

packages/generator/src/vdm-types.ts
marikaner: It is in the `PropertyDescriptor`.

packages/generator/test/parser/edmx-to-vdm-v4.spec.ts
marikaner: Yeah, here it refers to the type as found in the edmx file. I added some explicit tests.

packages/generator/src/vdm-types.ts
marikaner: We could also make this optional and add default values, wherever this is passed. I don't love that and would rather prefer to add a compatibility note. What do you think?

packages/generator/src/parser/edmx-parser-v4.ts
marikaner: done

packages/generator/src/parser/edmx-parser-v4.ts
marikaner: done

packages/generator/src/parser/edmx-to-vdm-v4.ts
marikaner: cool 😊

packages/generator/src/parser/edmx-to-vdm-v4.ts
marikaner: Yeah, totally, that is why I deprecated it. It isn't even used anywhere. I still left it here, so that it is consistent and less of a breaking change.

packages/generator/src/parser/parser-types-common.ts
marikaner: What do you mean by key collection? I think those are read when parsing at some point, but maybe when we do your proposed refactoring this can also be considered? I would not want to add this to this PR ;)

packages/generator/src/parser/parser-types-common.ts
marikaner: This has been here forever, I just moved it from one file to another. Not sure if this was exported before, though. Anyways, I don't love the name either, but I still like it better than `SetWithType`. I couldn't come up with anything substantially better and at least this one wins with "Gewohnheitsrecht" ;)

packages/generator/src/parser/parser-util.ts
marikaner: Hmm, it isn't unexpected. It gives you everything after the last dot in the name. That means that if there is no dot, it just returns the input. I added some explicit tests where this becomes visible. Is there anything other way we can make this more clear?
Or do you mean you want to test it for inputs other than string?
(Btw, this has also been here for a long time, I just moved it to a separate file and used it consistently, where it was implemented redundantly.)

packages/generator/src/typedoc.ts
marikaner: I don't get this comment? Can you explain a little more what you mean?

marikaner:
Let's discuss some of the open questions tomorrow and then merge 😊

packages/generator/src/parser/edmx-to-vdm-v2.ts
FrankEssenberger: ok

packages/generator/src/parser/parser-types-common.ts
FrankEssenberger: JoinedData could be anything. It does not tell you what it is. But let's keep it since refactoring a name is easy also later. Which means never :-).

packages/generator/src/parser/parser-util.ts
FrankEssenberger: I think I mean more, that if a value is passed inside not of the form `NameSpace.Name` something is wrong in the EDMX file. In general a namespace should be there and no dot in the name or name space is allowed. The method will handle it gracefully and return just the full thing or in case of WordA.WordB.WordC the WordC. However, I would like to get a stop of the generation. Since it is not clear that WordC is the name or that the generator works if there is no namespace at all.

packages/generator/src/vdm-types.ts
FrankEssenberger: yes, sorry.

packages/generator/src/vdm-types.ts
FrankEssenberger: as discussed we go for a note since it is not really a interface for the outside world and the usage should be limited and we take the risk over making the properties optionals with defaults since this is not a good concept to do just for avoid a breaking change.

packages/generator/src/typedoc.ts
FrankEssenberger: The other multiplicity is deprecated so ignore the comment.

packages/generator/src/parser/parser-types-common.ts
FrankEssenberger: ok fair, I think about having a interface or so with these keys and use it then in this type to minimize the occurrence of any explicit string.

FrankEssenberger:
Summary: Only the `stripNamespace()` discussion is open in other words if this should throw an illegal argument type of exception (Java way of saying it) in case the input was not WordA.WordB where words do not contain a dot.

packages/generator/src/parser/parser-util.ts
marikaner: A namespace can contain multiple dots (e. g. in the reference services this is the case). I am not sure whether it is normally possible that a name can have no namespace. If this is normally "forbidden" it would make sense to add a warning or even an error, but I would probably even do this somewhere else...

packages/generator/src/parser/parser-types-common.ts
marikaner: We can probably lose them alltogether in the refactoring.

marikaner:
> Summary: Only the `stripNamespace()` discussion is open in other words if this should throw an illegal argument type of exception (Java way of saying it) in case the input was not WordA.WordB where words do not contain a dot.

Ok great, thank you! I will go ahead and merge this one. As I am already working on the next PR for the generator, let's continue this open discussion and adjust in the new PR if necessary.

Issue #135, opened 2020-05-13
WhiteSource security vulnerability finding regarding lodash 4.17.15


jjtang1985:
We received such feedback regarding a WhiteSource finding:

> Hi colleagues,
>
> Recently our project encountered a WhiteSource security vulnerability (WS-2020-0070) of NodeJS component "lodash-4.17.15".
>
> Can you please help fix it?
> Related Cloud SDK component: @sap-cloud-sdk/util
> Version: 1.20.0

jjtang1985:
There is an open issue [here](https://github.com/lodash/lodash/issues/4738).

HoffmannThomas:
Looking at the lodash track record (4738,4518, etc.) you should consider not using it at all...

marikaner:
We aren't, but we have a dependency that has a dependency that uses it.

naresh-pidugu:
Is there a fix for the issue reported here ?

avidj:
Hello @marikaner,
now we have the problem this dependency comes into our project (SAP Graph) transitively through the cloud sdk while the lodash maintainer(s) seem reluctant to fix the issue. Therefore, our mandatory whitesource scans fail due to that transitive dependency.

marikaner:
Thanks for letting us know. Yeah, the reluctance to fix this issue on lodash side is a major problem.  I was hoping for them to come around eventually. We know that this is occurring transitively through `winston > async > lodash`. `async` already got rid of `lodash`. `winston` already updated to the latest version of `async` on their master, but did not make a new release for this yet. There already is a ticket in the winston repo: https://github.com/winstonjs/winston/issues/1812.
As `async` is just a small convenience library it might be possible, that winston is compatible with a newer version, then I might provide a hacked hotfix. I will let you know.

marikaner:
We solved this manually and will probably make a new release next week.

PR #201, opened 2020-06-19
fix: Manually update version of vulnerable dependency

marikaner:
## Context

`lodash` has a vulnerability that the refuse to solve causing security checks of our customers to fail. `lodash` comes into our dependency tree transitively through `winston`:
`winston > async > lodash`. The `async` library has already released a new version (3 and onwards) where they got rid of `lodash`. `winston` is already referencing `async` v3 on their master, but they did not release a new version yet (there is an issue for that https://github.com/winstonjs/winston/issues/1812).  As we cannot tell how long this will take, maybe we should just update that version manually.

I modified the package-lock.json and added a newer version as a direct dependency. I am aware that this is not a clean way to do this, but it should close https://github.com/SAP/cloud-sdk/issues/135.

We should keep in mind, that this should be removed once winston releases a new version and make sure to not override this in the meantime.

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/util/package-lock.json
+++ b/packages/util/package-lock.json
@@ -1,5673 +1,5102 @@
 {
-	"name": "@sap-cloud-sdk/util",
-	"version": "1.21.2",
-	"lockfileVersion": 1,
-	"requires": true,
-	"dependencies": {
-		"@babel/code-frame": {
-			"version": "7.8.3",
-			"resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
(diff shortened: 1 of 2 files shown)


jjtang1985:
After merging, please update it in the issue: #135.

jjtang1985:
We should create a follow up to check whether Winston makes a new release then remove the direct dependency if possible.

marikaner:
> We should create a follow up to check whether Winston makes a new release then remove the direct dependency if possible.

Done.

marikaner:
> LGTM please remeber to create the follow up ticket with an instruction which dependency to bring back and that the package-lock.json needs to be regenerated. Also I think we must be careful, because a run of `npm i` will revert your manual changes and will lead to to a failure because the " async" is now missing? We have to keep this in mind.

Yes, I created a ticket. `npm i` does not break it, but removing the package-lock of course will break it.

PR #203, opened 2020-06-19
feat: Support generation of v4 function imports

marikaner:
## Context

The generator now should support v2 and v4 function import parsing (and generation).

## Definition of Done

- [ ] Release notes updated.


packages/generator/test/generator.spec.ts
```diff
@@ -1,46 +1,123 @@
...
+    });
+  });
+
+  describe('v4', () => {
+    beforeAll(async () => {
+      files = await getGeneratedFiles('v4');
+    });
+
+    it('generates expected number of files', () => {
+      expect(files.length).toBe(29);
```
jjtang1985, line 50: There is a similar test above about v2.
Could you please compare these numbers and explain a bit regarding the difference?
In this case, what's the additional file?

packages/generator/test/generator.spec.ts
```diff
@@ -1,46 +1,123 @@
...
+        file => file.getBaseName() === 'TestEntity.ts'
+      );
+
+      expect(testEntityFile).toBeDefined();
+      expect(testEntityFile!.getClasses().length).toBe(1);
+      expect(testEntityFile!.getInterfaces().length).toBe(2);
+      expect(testEntityFile!.getNamespaces().length).toBe(1);
+
+      const entityClass = testEntityFile!.getClass('TestEntity');
+      expect(entityClass!.getProperties().length).toBe(26);
```
jjtang1985, line 64: Also here.

packages/generator/test/generator.spec.ts
```diff
@@ -1,46 +1,123 @@
...
+
+      checkStaticProperties(entityClass!);
+
+      const entityNamespace = testEntityFile!.getNamespace('TestEntity');
+      expect(entityNamespace!.getVariableDeclarations().length).toBe(27);
```
jjtang1985, line 69: Also here.

packages/generator/src/parser/parser-types-v4.ts
```diff
@@ -54,3 +57,13 @@ export interface EdmxNavigationPropertyBinding {
   Path: string;
   Target: string;
 }
+
+export interface EdmxFunction extends EdmxNamed {
+  ReturnType: { Type: string };
+  Parameter: EdmxParameter[];
+  IsBound: boolean;
```
jjtang1985, line 64: Is there any tests/use cases about this?

test-packages/test-services/srv/v4/test-service/TestComplexType.ts
```diff
@@ -102,7 +102,7 @@ export interface TestComplexType {
 /**
  * @deprecated Since v1.6.0. Use [[TestComplexType.build]] instead.
  */
-export function createTestComplexType_1(json: any): TestComplexType {
+export function createTestComplexType(json: any): TestComplexType {
```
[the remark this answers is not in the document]
marikaner: The reason here is, that in the v2 example we have a test to check that names of function imports do not clash with names of the builders of complex types. Initially we just copied the v2 service. I now removed the case that tests the name formatting from v4 because I thought it would be the same. Maybe I am wrong thought, what do you think?

packages/generator/test/generator.spec.ts
marikaner: `TestComplexBaseType` - there are now base types in the edmx definitions for entities and complex types.

packages/generator/test/generator.spec.ts
marikaner: There are some variable declarations collection types

packages/generator/test/generator.spec.ts
marikaner: There are some collection type properties.

packages/generator/src/parser/parser-types-v4.ts
marikaner: Isbound? No it is not used anywhere yet.

packages/core/test/test-util/test-services/v4/test-service/TestComplexType.ts
```diff
@@ -102,7 +102,7 @@ export interface TestComplexType {
 /**
  * @deprecated Since v1.6.0. Use [[TestComplexType.build]] instead.
  */
-export function createTestComplexType_1(json: any): TestComplexType {
+export function createTestComplexType(json: any): TestComplexType {
```
FrankEssenberger, line 105: I guess this change is related to the changes introduced in the edmx file?

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
```diff
@@ -147,54 +127,6 @@ export function testFunctionImportComplexReturnTypeCollection(parameters: TestFu
   return new FunctionImportRequestBuilder('get', '/sap/opu/odata/sap/API_TEST_SRV', 'TestFunctionImportComplexReturnTypeCollection', (data) => transformReturnValueForComplexTypeList(data, TestComplexType.build), params);
 }
 
-/**
```
FrankEssenberger, line 150: Why were these test cases removed? Is this the same in v2 and v4? Perhaps it also makes sense to have a edmx common and then a v2 v4 edmx for testing specific things?

packages/generator/src/parser/edmx-parser-common.ts
```diff
@@ -42,7 +42,7 @@ export function parseEntityTypes(root): EdmxEntityTypeBase[] {
   });
 }
 
-export function parseFunctionImports(root): EdmxFunctionImportBase[] {
+export function parseFunctionImportsBase(root): EdmxFunctionImportBase[] {
```
FrankEssenberger, line 45: General refactor suggestion: Perhaps we do also here a odata/[v2/v4/common] then we can keep the file names clean. Now the function is called base and the file common.

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
marikaner: Yeah, maybe you're right - it is not 100% the same and for regressions it is good to have it.

packages/generator/src/parser/edmx-to-vdm-common.ts
```diff
@@ -418,56 +419,49 @@ function parseReturnType(
...
-        swaggerDefinition,
-        functionImport.originalName
-      )
+      originalName: p.Name,
+      parameterName: formatter.originalToParameterName(
+        edmxFunctionImport.Name,
+        p.Name
+      ),
+      edmType: parseType(p.Type),
+      jsType: edmToTsType(p.Type),
```
FrankEssenberger, line 448: the attribute is called jsType but the method TsType? Which type is it?

packages/generator/src/parser/service-parser.ts
```diff
@@ -132,6 +135,9 @@ function transformServiceMetadata(
     edmxPath: serviceDefinitionPaths.edmxPath
   };
 
+  const transformFunctionImports = isV2Metadata(metadata.edmx)
```
FrankEssenberger, line 138: I am really looking forward to refactor this long and messy function.

packages/generator/src/parser/edmx-to-vdm-common.ts
marikaner: I also stumbled upon this. As it is a type I guess it is ts ¯\_(ツ)_/¯. I didn't want to change it, because it would be a breaking change again. I could add a tsType property and deprecate this one, not sure if it is worth it though. What do you think?

packages/generator/src/parser/service-parser.ts
marikaner: yes...

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
marikaner: Ah, also FunctionImports in v4 can only be GET

packages/core/test/test-util/test-services/v4/test-service/TestComplexType.ts
marikaner: Yes, there is no createTestComplexType function import, see below.

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
marikaner: First, I was actually thinking that one of the functions could make sense (createTestComplexType because its name conflicts with the testcomplextype builder), but I just checked and this is done in the common part, so I don't see the necessity to have an additional test here.
Secondly, Functions in v4 can only be GET therefore I removed the other test cases.

packages/generator/src/parser/edmx-parser-common.ts
marikaner: I restructured the files, that's actually a good idea, I didn't even think of that ¯\_(ツ)_/¯. I didn't rename the functions though - those that are called `xxxBase` also return some kind of base type for the most part.

packages/generator/src/parser/edmx-to-vdm-common.ts
marikaner: Also, maybe this just doesn't belong in this PR?

PR #204, opened 2020-06-22
fix: imports in test entity (v4)

jjtang1985:
## Context

All TestEntities have compilation errors, which are not detected.
- [x] created a `v4.js` in the root as a config pointing to `dist/v4`
- [x] `npm install` from the root should also compile the TestEntities for verifying. 

## Definition of Done

- [ ] Release notes updated.


packages/generator/src/imports.ts
```diff
@@ -44,7 +44,7 @@ export function coreImportDeclaration(
   namedImports: string[],
   oDataVersion: ODataVersion
 ): ImportDeclarationStructure {
-  const moduleSuffix = oDataVersion === 'v4' ? '/v4' : '';
+  const moduleSuffix = oDataVersion === 'v4' ? '/src/odata/v4' : '';
```
marikaner, line 47: This is not how this was intended. Why is it not possible to get it from v4?

PR #212, opened 2020-06-25
doc: Minor improvements

marikaner:
## Context

Yesterday I merged the VDM documentation, but when merging I messed something up. Just minor improvements otherwise.


docs/js/features/odata/use-generated-client.md
```diff
@@ -70,25 +64,29 @@ MyEntity.requestBuilder()
...
+Setting an `authorization` or `apikey` header (regardless of lowercase or uppercase spelling) will skip any automatic authorization header building that the SAP Cloud SDK would normally do.
+:::
+
 ### Setting custom query parameters
 
-```suggestion
-The SAP Cloud SDK automatically sets some necessary request headers on every request. You can specify additional custom headers using the `withCustomHeaders` function:
+The SAP Cloud SDK adds necessary query parameters for a request based on your configuration. You can add custom parameters by using the `withCustomQueryParameters` function. Custom query parameters take precedence over those created by the SAP Cloud SDK.
+
+In the example below an additional query parameter `language=en` will be added to the request:
```
CharlesDuboisSAP, line 78: I'm not sure about this one
```suggestion
In the example below an additional query parameter `"language": "en"` will be added to the request:
```

docs/js/features/odata/use-generated-client.md
```diff
@@ -405,9 +403,11 @@ The below example leverages OData v4 exclusive features to build a more complex
...
 
 ```ts
-//Fetch all business partners where:
-//- the last name is at least twice as long as the first name
-//- AND the combined string of first and last name does not contain 'bob'
+/*
+  Fetch all business partners where:
+    - the last name is at least twice as long as the first name
+    - AND the combined string of first and last name does not contain 'bob'
+*/
```
CharlesDuboisSAP, line 410: How did you do that, the compiler gave me an error if I had a `/*comment*/`

docs/js/features/odata/use-generated-client.md
marikaner: Maybe I have to rephrase a bit.

docs/js/features/odata/use-generated-client.md
marikaner: Good catch, I didn't even check that.

PR #217, opened 2020-06-26
feat: Add more filter functions

marikaner:
## Context

Implemented all filter functions that are currently possible

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/core/src/odata/common/filter/filter-function-base.ts
+++ b/packages/core/src/odata/common/filter/filter-function-base.ts
@@ -1,0 +1,97 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+
+import { EdmTypeShared } from '../edm-types';
+import { EntityBase, ODataVersionOf } from '../entity';
+import { Field, FieldType } from '../selectable';
+import { Filter } from './filter';
+
+/**
+ * Data structure to represent OData V2 filter functions.
+ * Use the factory function [[filterFunction]] to create instances of `FilterFunction`.
+ */
+export abstract class FilterFunction<
+  EntityT extends EntityBase,
+  ReturnT extends FieldType
+> {
+  /**
+   * EdmType of the return type of the filter function.
+   */
+  public edmType: EdmTypeShared<ODataVersionOf<EntityT>>;
+
+  /**
+   * Creates an instance of FilterFunction.
+   * @param functionName - Name of the function
+   * @param parameters - Representation of the parameters passed to the filter function
+   */
+  constructor(
+    public functionName: string,
+    public parameters: FilterFunctionParameterType<EntityT>[]
+  ) {}
+
+  /**
+   * @deprecated Since v1.21.0. There will be no replacement. Let us know if you were using this functionality.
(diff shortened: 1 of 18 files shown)


packages/core/src/odata/common/filter/filter-function-base.ts
```diff
@@ -0,0 +1,97 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+
+import { EdmTypeShared } from '../edm-types';
+import { EntityBase, ODataVersionOf } from '../entity';
+import { Field, FieldType } from '../selectable';
+import { Filter } from './filter';
+
+/**
+ * Data structure to represent OData V2 filter functions.
```
FrankEssenberger, line 9: Is this true? If so this class and the related ones should not be in the common folder but in the V2 specific one.

packages/core/src/odata/common/filter/filter-function.ts
```diff
@@ -1,97 +1,62 @@
...
-   * Serializes the filter function into a string
-   * @param parentFieldNames - Names of parents in case the function is part of a filter on a navigation property
-   * @returns The filter function as string
-   */
-  toString(parentFieldNames: string[] = []): string {
-    const params = this.parameters
-      .map(param => this.transformParameter(param, parentFieldNames))
-      .join(', ');
-    return `${this.functionName}(${params})`;
+export function filterFunction<EntityT extends EntityBase>(
```
FrankEssenberger, line 35: I would call this method a bit different to indicate, that it builds, construcuts filterfunction but is not itself a filter function.

test-packages/integration-tests/test/v2/filter-functions.spec.ts
```diff
@@ -0,0 +1,23 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+import {
```
FrankEssenberger, line 2: genral comment: In the `filter-factory.ts` of the test util I see many filter object beeing created for testing. But this has not been adjusted for all the new filter function beeing defined here. Is this obvious?

FrankEssenberger: As discussed: Only do the specific cases.
Add type tests to `filter.spec.ts`.

packages/core/src/odata/common/filter/filter-function-base.ts
marikaner: No it's not. It is a left over from before.

packages/core/src/odata/common/filter/filter-function.ts
marikaner: Phew, this function has been there for a while now. I think as it is supposed to also be used by our users the name is quite good. I would generally like to keep it, but independent of that this is out of scope for this PR is guess ;)

test-packages/integration-tests/test/v2/filter-functions.spec.ts
marikaner: Added some more tests as discussed.

PR #228, opened 2020-06-30
Documentation: JS added expand, and deep create / update

CharlesDuboisSAP:


docs/js/features/odata/use-generated-client.md
```diff
@@ -123,14 +123,18 @@ The example above creates a request to get all BusinessPartner entites.
...
 #### Select
 
-When reading entities, the API offers `select( ... )` on the builders. Through it, the query parameter `$select` is set. It restricts the response to the given selection of properties in the request.
+When reading entities, the API offers `select( ... )` on the builders. Through it, the query parameters `$select` and `$expand` are set. It restricts the response to the given selection of properties in the request.
 
 <!-- OData v4
 When reading entities, the API offers `select( ... )` on the builders. Through it, the query parameters `$select` and `$expand` are set. It takes in properties of the entity being queried. Primitive properties are added to `$select` while complex and navigational properties are added to `$expand`. This handling is done automatically by the SDK.
+
+maybe add this?
+To include navigation properties, OData offers the expand operation. The VDM allows you to use navigation properties in the select function just like normal properties.
```
marikaner, line 132: Yes 😊 but only for v4 as you mentioned. Good to have it in a comment! 👍

docs/js/features/odata/use-generated-client.md
```diff
@@ -146,7 +150,7 @@ BusinessPartner.requestBuilder()
 The above translates to the following query parameters:
 
 ```sql
-$select=FirstName,LastName,to_BusinessPartnerAddress
+$select=FirstName,LastName,to_BusinessPartnerAddress/*&$expand=to_BusinessPartnerAddress
```
marikaner, line 153: 👍 good catch

docs/js/features/odata/use-generated-client.md
```diff
@@ -287,6 +290,33 @@ BusinessPartner.requestBuilder().create(businessPartner);
...
+      .build()
+  ])
+  .build();
+
+// execute the create request
+BusinessPartner.requestBuilder()
+  .create(businessPartner)
+  .execute(myDestination);
+```
+
```
marikaner, line 319: Optional: I don't remember this myself, but maybe you can find out: When reading this section I think, so what? That is exactly the behavior I would expect ;)
Is there something special that we do? Is the url different than with the normal create? Is there some additional header that we set?
Maybe you can add a comment for future reference.

docs/js/features/odata/use-generated-client.md
```diff
@@ -318,6 +348,10 @@ BusinessPartner.requestBuilder()
   .ignoreVersionIdentifier();
 ```
 
+#### Deep Update
+
+It is also possible to update an entity together with related entities in a single request, same as [deep create](#deep-create).
+
```
marikaner, line 354: That is not true ;) Deep update still needs to be implemented and will only be available for OData v4. Maybe, again, just add a comment for future reference when somebody documents v4?

docs/js/features/odata/use-generated-client.md
CharlesDuboisSAP: I don't understand, what did you expect? why do you think you need to do something special?
I just wanted to show the reader some code, but I'll add this in a comment anyway

marikaner: I mean - what is the difference between create and deep create?

CharlesDuboisSAP: Create Request Builder^:
to create a new entity

Deep Create:
to create an entity together with related entities in a single request

Maybe I'm missing some feature but it seems really simple to use and not worth explaining

marikaner: That's exactly what I mean. It is so obvious, why is it even documented as a feature? I remember that this was not working from the beginning and according to Junjie there are differences in v2 / v4. But it is possible that there is nothing special happening and we just promote the fact that this can also be done. Anyways, I am totally fine with the comment ;)

PR #238, opened 2020-07-06
Fix: Winston Issue Update Dependencies

FrankEssenberger:
## Context

we were using the 3.3.2 of winston containing a generic Logger class. In version 3.3.3 they changed this to a Logger interface. This lead to errors in the cloud-sdk-logger.d.ts declaring an object with a non existing type. 

I updated all package-lock.json files and the typescript version. I also needed to fix the rambda version to 5.8.X since the super new 5.9.0 had a type error. In a test I needed to adjust the imports a bit after the version updates.

**Question**: How do we say that a version of the SDK should not be used. The version 1.23.0 should be backlisted because it contains this faulty type definition.

## Definition of Done

- [x] Release notes updated.


jjtang1985:
To be honest, I'm still confused. 
- [`3.33` is broken] This [issue(https://github.com/winstonjs/winston/issues/1822) says, `3.3.2` is stable, but `3.3.3` is broken.
- [`3.33` is working] The [release note](https://github.com/winstonjs/winston/releases/tag/v3.3.3) said the issue is fixed by reverting.

But if we can test our `1.23.0` and the hot fix `1.23.1`, I don't care what they did.
To mark `1.23.0` as in the blacklist, you can [deprecate this version](https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions).

FrankEssenberger:
as repository owner @marikaner has to run:
```
npm deprecate @sap-cloud-sdk/util@1.23.0 "This version only works with winston version 3.3.2 and therefore is deprecated."
```

to depricate version 1.23.0. We decide if we make a patch or a minor release today.

PR #241, opened 2020-07-06
feat: Support v4 edm types in generator

marikaner:
## Context

Add field generation for v4 edm types.

## Definition of Done

- [x] Release notes updated.


packages/generator/src/generator-utils.ts
```diff
@@ -42,7 +42,7 @@ export function isNullableParameter(parameter: any) {
   return !!parameter['Nullable'] && parameter['Nullable'] !== 'false';
 }
 
-type EdmTypeMapping = { [key in EdmType]: string };
```
FrankEssenberger, line 45: For the moment this is ok, but we should keep in mind, that in the unlikely case that we have in v2/v4... the same Edm type represented by different data. But I guess it is fine and will never happen.

PR #247, opened 2020-07-07
Fix: Conversion number TS to edm formats

FrankEssenberger:
## Context

we found a mistake in the code that for the conversion from the TS representation to edmx also the parseNumber method was used. This is not correct since the values 'INF','-INF' and 'NaN' are casted to Number.POSITIVE_INFINITY etc. So one needs a different conversion method to bring the number values back to the 'INF','-INF' and 'Nan' value. This fix provides this fix.

## Definition of Done

- [x] Release notes updated.

### DIFF
--- a/packages/core/src/odata/common/payload-value-converter.ts
+++ b/packages/core/src/odata/common/payload-value-converter.ts
@@ -54,26 +54,23 @@ export function toNumberFromEdm(value: string | number): number {
 /**
  * @hidden
  */
-export function toNumberFromTs(value: Number): number|string {
-  if (typeof value === 'number') {
-    return value;
-  }
-
+export function toNumberFromTs(value: number): number | string {
   if (value === Number.POSITIVE_INFINITY) {
     return 'INF';
   }
   if (value === Number.NEGATIVE_INFINITY) {
     return '-INF';
   }
-  if (value === Number.NaN) {
-    return 'NaN'
+  if (Number.isNaN(value)) {
+    return 'NaN';
   }
 
-  const num = Number(value);
+  if (typeof value === 'number') {
+    return value;
+  }
 
   throw new Error(`TS->edm: Cannot create number from input "${value}"`);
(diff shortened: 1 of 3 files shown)


packages/core/test/odata/payload-value-converter.spec.ts
```diff
@@ -388,3 +433,19 @@ describe('edm to moment and back', () => {
...
+}
+
+function checkInfinityCasesToTs(edmType: EdmType) {
+  expect(edmToTs('INF', edmType)).toBe(Number.POSITIVE_INFINITY);
+  expect(edmToTs('-INF', edmType)).toBe(Number.NEGATIVE_INFINITY);
+  expect(edmToTs('NaN', edmType)).toBe(Number.NaN);
+}
+
+function checkInfinityCasesRoundTrip(edmType: EdmType) {
+  ['NaN'].forEach(s => expect(tsToEdm(edmToTs(s, edmType), edmType)).toBe(s));
```
marikaner, line 450: Why a list?

packages/core/src/odata/common/payload-value-converter.ts
```diff
@@ -27,7 +27,7 @@ const fromBigNumber = (value: BigNumber): string =>
 /**
  * @hidden
  */
-export function parseNumber(value: string | number): number {
+export function toNumberFromEdm(value: string | number): number {
```
marikaner, line 30: I am not a fan of the names here, because somehow it does not resonate in my head. Alternatives:
`edmToNumber` + `numberToEdm`, `fromEdmToNumber` + `fromNumberToEdm`

packages/core/test/odata/payload-value-converter.spec.ts
FrankEssenberger: ah had a failing test for NaN and forgot to put the other back. Thanks

PR #252, opened 2020-07-08
chore: add OData V4 e2e tests

jjtang1985:
## Context
Add a new e2e test stage for verifying OData V4 implementation. 
The reason is simple, the existing integration test strategy where we mocked the requests/response, is not enough.

### which OData v4 service are we using?
We build a service ourselves by using cap with an in-memory db, because the public service is not stable enough.

### what's new in the GitHub action?
- e2e test stage
  - build the app (generate db file for sqlite)
  - start the app (wrapped in `pm2` to make it as daemon mode)
  - run the jest tests

### how do we test?
- start the server locally, open the metadata page, download it and put it into the `test-resources` [1]
- generate the vdms by using the same command as you used for test entities. [2]
- write your test in the `e2e-test` just like the integration tests.

The [1] and [2] are not automated, which is the same as the quarterly release and the test entities. It should be stable.

### any other highlights?
- the whole e2e tests takes < 100s in the GitHub Actions, which can be improved further when using cache.
- the poc projects also tried to used cypress GitHub action, which can be used for frontend e2e tests later, also only needed < 100s.
- the `@sap/cds` lib is fetched from the npmjs.org

### future planning
- if the e2e tests takes longer or becomes flaky , we can consider disable it for internal developers, but make it as prerequisites of the release pipeline.

## Definition of Done

- [ ] Release notes updated.


scripts/generate-test-services.ts
```diff
@@ -109,3 +109,10 @@ generateTestServicesPackage(packageOutputDir, 'v2');
 generateTestServicesPackage(packageOutputDir, 'v4');
 generateTestServicesWithLocalCoreModules(coreUnitTestOutputDir, 'v2');
 generateTestServicesWithLocalCoreModules(coreUnitTestOutputDir, 'v4');
+
+generate({
```
FrankEssenberger, line 113: Cloud you perhaps wrap this generate also in a function with a nice name like `generateTestServiesE2Etests` or something like this.

package.json
```diff
@@ -13,9 +13,13 @@
   "repository": "github:SAP/cloud-sdk",
   "scripts": {
     "test": "npm run test:unit && npm run test:integration && npm run test:type",
-    "test:unit": "npx lerna run test --ignore \"@sap-cloud-sdk/integration-tests\" --ignore \"@sap-cloud-sdk/type-tests\"",
+    "test:unit": "npx lerna run test --ignore \"@sap-cloud-sdk/integration-tests\" --ignore \"@sap-cloud-sdk/type-tests\" --ignore \"@sap-cloud-sdk/e2e-tests\"",
     "test:integration": "npx lerna run test --scope \"@sap-cloud-sdk/integration-tests\"",
     "test:type": "npx lerna run test --scope \"@sap-cloud-sdk/type-tests\"",
+    "test:e2e": "npx lerna run test --scope \"@sap-cloud-sdk/e2e-tests\"",
+    "e2e-test:build": "cd test-packages/e2e-tests && npm run deploy:to-sqlite",
+    "e2e-test:start-server": "cd test-packages/e2e-tests && npm start",
```
FrankEssenberger, line 21: the command `e2e-test:start-server` is not used in the the build pipeline. I guess it is for local testing. Perhaps it makes sense to include a `test-local:start` and `test-local:watch` here and link it to the two scripts in the e2e-tests/package.json. Also I would like to split the generate block in more  things: generate-vdm-test-entties-v2,
generate-vdm-test-entties-v4, generate-e2e-service and generate-all (naming is super bad, just to get the four different things clear.

scripts/generate-test-services.ts
FrankEssenberger: Or perhaps it makes sense to split the generation in multiple tasks see comment above.

lerna.json
```diff
@@ -6,7 +6,9 @@
     "packages/test-util",
     "packages/util",
     "test-packages/integration-tests",
+    "test-packages/e2e-tests",
```
FrankEssenberger, line 9: Just to be 100% sure. I see that the two packages are marked as `private` is this sufficient that they are not getting published?

jjtang1985: I think yes, see the [doc](https://docs.npmjs.com/files/package.json) about the `package.json`.

package.json
jjtang1985: I would remove the watch command, since for such a small app, you don't need it and it asks for additional dependency: '@sap/cds-dk'.

scripts/generate-test-services.ts
jjtang1985: I'm a bit afraid of adding more scripts to the main `package.json`. But I see the use case, since currently, we are touching the v4 and e2e parts. Let's add 3 additional scripts then. I hope we can remove them at the end of the quarter.

package.json
FrankEssenberger: ok

test-packages/e2e-tests/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "@sap-cloud-sdk/core": "^1.24.0",
+    "@sap-cloud-sdk/test-services-e2e": "^1.24.0",
+    "@sap/cds": ">=3.33.1",
+    "express": "^4.17.1",
+    "sqlite3": "^4.2.0"
+  },
+  "scripts": {
+    "start:daemon": "pm2 start start.js",
+    "start": "cds run",
+    "watch": "cds watch",
```
FrankEssenberger, line 19: I thought you wanted to remove the watch command?

package.json
```diff
@@ -13,10 +13,17 @@
...
     "test:type": "npx lerna run test --scope \"@sap-cloud-sdk/type-tests\"",
-    "generate:test-services": "npx ts-node scripts/generate-test-services.ts && npm run lint:fix && lerna link --force-local",
+    "test:e2e": "npx lerna run test --scope \"@sap-cloud-sdk/e2e-tests\"",
+    "e2e-test:build": "cd test-packages/e2e-tests && npm run deploy:to-sqlite",
+    "e2e-test:start-server": "cd test-packages/e2e-tests && npm start",
+    "e2e-test:start-server:deamon": "cd test-packages/e2e-tests && npm run start:daemon",
+    "generate:test-services:v2": "npx ts-node scripts/generate-test-services.ts v2 && npm run lint:fix && lerna link --force-local",
+    "generate:test-services:v4": "npx ts-node scripts/generate-test-services.ts v4 && npm run lint:fix && lerna link --force-local",
+    "generate:test-services:e2e": "npx ts-node scripts/generate-test-services.ts e2e && npm run lint:fix && lerna link --force-local",
+    "generate:test-services:all": "npx ts-node scripts/generate-test-services.ts all && npm run lint:fix && lerna link --force-local",
```
marikaner, line 26: This could be just `generate:test-services`

test-packages/e2e-tests/start.js
```diff
@@ -0,0 +1,8 @@
+var process = require('child_process');
+process.exec('cds run',function (err,stdout,stderr) {
```
marikaner, line 2: It looks like there is no linter running on this ;) (We don't need to do this right away)

PR #259, opened 2020-07-13
Add a note about overriding dependencies

MatKuhr:
This adds a small note on how to override SDK dependencies.


docs/java/guides/dependencies.md
```diff
@@ -89,3 +89,32 @@ Google the error message.
...
+You can achieve this by listing it in the dependency management _before the SDK BOM_.
+
+For example to override the version of SLF4J:
+```xml
+<dependencyManagement>
+    <dependencies>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>your-sfl4j-version</version>
```
newtork, line 104:
```suggestion
            <version>your-slf4j-version</version>
```

or even a property...

```suggestion
            <version>${slf4j.version}</version>
```

MatKuhr: I went with the `your-version` style because we use it in other places as well :)

PR #263, opened 2020-07-15
Fix: deserialiser for navigation properties

jjtang1985:
## Context

In v4, the value of the one-to-many navigation property (an array) is not wrapped in an object with a key `results` anymore, which needs v4 specific implementation.

See example below:
``` 
    "Friends": [
        {
            "UserName": "russellwhyte",
        }
    ]
```

Please note: this pr also introduced the following the edmx and vdm files of the travel service, because I added an e2e test for that. The test is disabled (will be skipped in the pipeline), but is useful when reproducing issues or showing real demo for some reasons. Also I think the cap service doesn't support e.g., collection type, additional service might be helpful.
Let know whether you like it.

## Definition of Done

- [ ] Release notes updated.


FrankEssenberger:
What is the reason for not putting the test in the pipeline?

jjtang1985:
This is a public service that is not stable, based on the feedback from Alex and Artem.

PR #265, opened 2020-07-16
[Java] [OData VDM] Correct & update code example for OData 4.0 API

MatKuhr:
The previous section was only correct for OData 2.0. Update according to the current state of the API.


docs/java/features/odata/use-vdm.md
```diff
@@ -110,13 +110,32 @@ An [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) is a v
...
+{ label: 'OData V2', value: 'v2', },
+{ label: 'OData V4', value: 'v4', }]}>
+<TabItem value="v4">
+
+```java
+partner = service.getBusinessPartnerByKey("id")
+                 .execute(destination);
+response = service.updateBusinessPartner(partner)
+                 .execute(destination);
+partner = response.getModifiedEntity();
```
newtork, line 123: I don't have a suggestion really, but if I saw this a PR, then I would ask for introducing a new variable or at least a comment like `// update partner reference`.

Maybe it would be sufficient  already if you'd put another newline/space between `response = ` and `partner =`.


What do you think?

MatKuhr: Great suggestion, I also added a comment to the v2 variant :)

PR #275, opened 2020-07-20
fix: lambda filter on the expanded one-to-many link

jjtang1985:
## Context
This PR will fix the compilation error, when creating request like:

```
People.requestBuilder()
    .getAll()
    .expand(
      People.FRIENDS.filter(
        People.FRIENDS.filter(
          all(People.LAST_NAME.equals('test'))
        )
      )
    )
```

## Definition of Done

- [ ] Release notes updated.


packages/core/src/odata/common/entity-deserializer.ts
```diff
@@ -176,7 +176,8 @@ export function entityDeserializer(
         ([_, field]) =>
           (field instanceof EdmTypeField ||
             field instanceof ComplexTypeField) &&
-          typeof json[field._fieldName] !== 'undefined'
+          typeof json[field._fieldName] !== 'undefined' &&
```
FrankEssenberger, line 179: perhaps you can create a little helper function: isUndefinedOrNull. Perhaps this exists already? Perhaps there is a build in funciton for this check?

packages/core/src/odata/common/entity-deserializer.ts
```diff
@@ -195,11 +196,15 @@ export function entityDeserializer(
...
       .map(property => ({
-        ...(typeof json[property.originalName] !== 'undefined' && {
-          [property.name]: isComplexTypeNameSpace(property.type)
-            ? deserializeComplexType(json[property.originalName], property.type)
-            : edmToTs(json[property.originalName], property.type)
-        })
+        ...(typeof json[property.originalName] !== 'undefined' &&
+          json[property.originalName] !== null && {
+            [property.name]: isComplexTypeNameSpace(property.type)
+              ? deserializeComplexType(
```
FrankEssenberger, line 202: I find this really hard to read. Relating the spread (...) with the tenary operator in the third boolean check is in my opinion to much. I would prefer a seperate function returning the object with a proper return and the condition separated.

packages/core/src/odata/common/selectable/one-to-many-link.ts
```diff
@@ -30,9 +30,18 @@ export class OneToManyLink<
...
+      | Filterable<LinkedEntityT>
+      | OneToManyLink<LinkedEntityT, any>
+    )[]
+  ): this {
     const link = this.clone();
-    link._filters = new FilterLink(this, expressions);
+
+    link._filters = new FilterLink(
+      this,
+      expressions.map(e => (e instanceof OneToManyLink ? e._filters : e))
```
FrankEssenberger, line 43: just to be sure. Will this work sort of recursively if you would add a third layer:
```
  TestEntity.TO_MULTI_LINK_1.filter(
            TestEntityMultiLink_1.TO_MULTI_LINK_2.filter(
                 TestEntityMultiLink_2.TO_MULTI_LINK_3.filter(
...
```
with an if you could get two levels but the third not. Perhaps it is already working. Just want to be sure.

jjtang1985: Please keep in mind, the same function  names `filter()` are used in difference classes, which makes it confusing. There is a ticket:
https://github.wdf.sap.corp/MA/scp-rocket/issues/1287

```
People.requestBuilder()
      .getAll()
      .expand(
        People.FRIENDS.filter( [1]
          People.FRIENDS.filter( [2]
            any(People.USER_NAME.equals('russellwhyte'))
          )
        )
      )
```
will generate url like
```
$expand=Friends($filter=Friends/any(a0:a0/UserName eq 'russellwhyte'))
```
so, the [1] filter is actually a "setter" for setting the filter of the expanded navigation property.
the [2] filter is static method for building the lambda expression.

Therefore, there are only two levels.
Using the same function for different purpose is not good. Since this is the agreed api design, we consider as the initial version. The follow up ticket should solve it.

jjtang1985:
@marikaner , here are two examples:
- [example](https://services.odata.org/TripPinRESTierService/(S(miemz3v3blexv3n23th5mzju))/People('russellwhyte')) that returns `null` for normal property and complex property
- [example](https://services.odata.org/V4/TripPinServiceRW/People('russellwhyte')) that hides some properties, which might be null or undefined.

current logic of the core, when `undefined`/`null` is detected from the response:
- [normal property] returns the value as it is in the response
- [complex type] error

I would like to keep it consistent and makes difference between `undefined` and `null`, so I'll apply the same rule for complex type as the normal property.

However, currently, only the disabled e2e tests (travel service) can verify the null handling for both normal property and for complex type. Unit test is not possible because we have to touch the vdm. See example here:
```
      const testEntity = new TestEntityV4();
      testEntity.stringProperty = null; (compilation error, unless null is allowed)
      testEntity.complexTypeProperty = null; (compilation error, unless null is allowed)
```

If we agree my latest change, then we either find a way to make the unit test or modify the vdm.

packages/core/src/odata/common/selectable/one-to-many-link.ts
FrankEssenberger: ok

packages/core/src/odata/common/entity-deserializer.ts
FrankEssenberger: have you seen this @jjtang1985

packages/core/src/odata/common/entity-deserializer.ts
FrankEssenberger: have you seen this @jjtang1985

jjtang1985: We treat undefined and null differently. So the helper function is not useful in this case.

packages/core/src/odata/common/entity-deserializer.ts
jjtang1985: Actually, initially I had the same feeling. Once I read it carefully, I felt the readability is not bad. So I don't have strong opinion about it. Fortunately, I don't touch this code block. So I'll give this opportunity to the next person;)

PR #303, opened 2020-07-24
chore: Use Yarn

marikaner:
## Context

Lerna hoisting / not hoisting? Dependabot broken. I couldn't resist and hope that this will make our lives better.

## Definition of Done

- [ ] Release notes updated.


scripts/check-licenses.js
```diff
@@ -0,0 +1,53 @@
...
+);
+
+function isAllowedLicense(licenses) {
+  return Array.isArray(licenses)
+    ? isAllowedLicense(licenses.join(','))
+    : allowedLicenses.some(allowedLicense => licenses.includes(allowedLicense));
+}
+
+function isSapDependency(dependency) {
+  return dependency.startsWith('@sap');
```
jjtang1985, line 52: I would be careful about this logic.
e.g., this function returns true for this dependency: `https://www.npmjs.com/package/@saparanj/maxbot1`.
```suggestion
  return dependency.startsWith('@sap/');
```
of course, for all the known scopes like `@sap-cloud-sdk`, we can add them.

package.json
```diff
@@ -43,7 +43,7 @@
     "lint": "yarn eslint --ext .ts .",
     "lint:fix": "yarn lint -- --fix",
     "lint:fix-cached": "yarn lint:fix -- --cache",
-    "check:license": "yarn license-checker -direct --excludePackages \"@sap/edm-converters@1.0.21;@sap/xsenv@2.2.0;weak-map@1.0.5\" --onlyAllow \"MIT;Apache;ISC;BSD;WTFPL;Artistic;CC-BY;CC0;Unlicense;Public Domain;GPL-3.0\"",
+    "check:license": "node scripts/check-licenses.js",
```
jjtang1985, line 47: btw, you said, this was a fix, so calling the same command with yarn was not possible?

.github/dependabot.yml
```diff
@@ -1,6 +1,6 @@
 version: 2
 updates:
-- package-ecosystem: npm
+- package-ecosystem: yarn
```
jjtang1985, line 3: I saw "Dependabot broken.", but master build was successful. Could you please share a bit more about the major trigger of this PR?

scripts/check-licenses.js
marikaner: Yeah you're right. I thought about it and thought it might be overkill, but you're right ;)

package.json
marikaner: It is possible, but because there are workspaces now and license-checker does not support this correctly, local packages are recognized, but because they are private license-checker marks them as "unlicensed". I considered multiple solutions, like other packages or `yarn licenses list` (which might be better in the end, but because it is in yml format I decided against it).

marikaner: Btw I just saw, that yarn has a `--json` flag. I'll check whether I can work with that.

.github/dependabot.yml
marikaner: Of course. Dependabot was reporting that there were vulnerabilities in some packages. But it could not generate automatic PRs for this. I found that the reason most probably was, that there were package-locks in those packages, that were broken, e. g. the package-lock of integration did not contain the local packages the integration-test package relies on. So I ended up fixing this manually with an ugly workaround (Remove the local dependencies first, run npm install, add dependencies again). Obviously these packages cannot be handled by dependabot.
Another problem is, that dependabot somehow succeeded in fixing some dependencies, but not all - the update check crashes. I believe that this is has the same root cause.

package.json
marikaner: I checked and I think it is not worth it.

FrankEssenberger:
Why there is such a big number of files changes. That the package-lock.json and so one change is clear. But I see a lot of other stuff going on:
- .map files removed from one service
- A new `legacy` serivce`? 
- Thr `srv` dir in many paths for the test services has gone away. Why is this?
- There is a new `check-licence.js` is this new?

Could you please give some context about the

marikaner:
Of course, I'll add notes inline below:

> Why there is such a big number of files changes. That the package-lock.json and so one change is clear. But I see a lot of other stuff going on:
> 
> * .map files removed from one service

Are you sure? If yes, which service are you referring to? If no, see below.

> * A new `legacy` serivce`?

There shouldn't be any new legacy service, only the old one. Only difference is, it should have been moved, see below.

> * Thr `srv` dir in many paths for the test services has gone away. Why is this?

We used some feature from lerna that is non-standard - [publishConfig.directory](https://github.com/lerna/lerna/blob/d04ce8e10503003e498c44d0db5ff455054d7d71/commands/link/README.md#publishconfigdirectory). It allows you to define what is exposed when linking lerna packages. As we are not linking using lerna anymore, we cannot use this feature. We introduced the `srv` directory before we had the separation of `v2 / v4` to be able to delete the service directory without touching the legacy service. Since the services are now encapsulated in their own directories, the `srv` directory is obsolete and we do not have to rely on this lerna feature.

> * There is a new `check-licence.js` is this new?

Yes it is, but the behavior remains the same. Check this [discussion](https://github.com/SAP/cloud-sdk/pull/303#discussion_r460567156).

scripts/check-licenses.js
marikaner: Done

.github/dependabot.yml
marikaner: I now also added an ADR, to reflect this.

marikaner:
@FrankEssenberger I added the ADR, please review.

knowledge-base/architectural-decisions-records/use-yarn-workspaces.md
```diff
@@ -0,0 +1,20 @@
...
+Further, because `lerna bootstrap --hoist` changes the root `package-lock.json` we are not using hoisting to ensure we respect the `build once` princible. Therefore, we currently don't benefit from hoisting.
+
+## Decision
+
+Use yarn workspaces with hoisting.
+
+## Consequences
+
+Yarn supposedly has better performance than npm. Yarn's concept of workspaces and lerna packages are designed to work together. `yarn install` "bootstraps" the packages. Combined with the `--frozen-lockfile` flag, the `yarn.lock` will not be overwritten. These concepts are quite straight forward and should reduce confusion. As there is only one lock file created by yarn directly, the expectation is, that external tools (e. g. dependabot) will be able to parse it.
+The team has to adjust to the new commands and different behavior of yarn in some places. Some commands are more verbose (e. g. `npm ci` vs `yarn --frozen-lockfile`), others are more strict (e. g. `npm install` by default does not fail for wrong engines, i. e. node versions, yarn needs a flag to ignore wrong engines, `yarn --ignore-engines`).
```
FrankEssenberger, line 20: is there a `install` missing after yarn for --forzen-lockfile and --ignore-engies?

CONTRIBUTING.md
```diff
@@ -2,13 +2,13 @@
 Thank you for taking your time to contribute to the SAP Cloud SDK!
 
 ## Project Structure
-This project containes multiple packages, that are managed using [lerna](https://github.com/lerna/lerna).  Productive packages are located in the [`packages`](./packages) directory, test packages are located in the [`test-packages`](./test-packages) directory.
-All dependencies of the packages are hoisted into the root project. Some of the packages are interdependent, therefore `npm install` won't work from within those packages. Run `npm install` in the root directory instead.
+This project containes multiple packages, that are managed using [lerna](https://github.com/lerna/lerna) and [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/).  Productive packages are located in the [`packages`](./packages) directory, test packages are located in the [`test-packages`](./test-packages) directory.
+All dependencies of the packages are hoisted into the root project. Some of the packages are interdependent, therefore `yarn install` won't work from within those packages. Run `yarn install` in the root directory instead.
```
FrankEssenberger, line 6: Is this true? I guess only dependencies used by more then one package are hoisted?

knowledge-base/architectural-decisions-records/use-yarn-workspaces.md
marikaner: No, this is the shortcut.

CONTRIBUTING.md
marikaner: No you're right - not all dependencies are moved up. I fixed it.

knowledge-base/architectural-decisions-records/use-yarn-workspaces.md
FrankEssenberger: So  if you do not give an argument yarn does the `install`. Good to know.

PR #314, opened 2020-07-29
fix: Fix de-/serialization of collection properties in complex types

marikaner:
## Context

Complex types did not support collections. I fixed that and added some tests for this. I also exposed the serializeComplexType function for consistency.
I also extended the `PropertyMetadata` to handle collections and added collection properties to our test edmx file.
In addition I added shortcuts to our most prominent packages (core, generator and util) for easier usage (`yarn workspace @sap-cloud-sdk/generator test` vs `yarn generator test`).

@FrankEssenberger is this a fix or new functionality in your opinion (based on this, should our next version increase be `patch` or `minor`)?

## Definition of Done

- [x] Release notes updated.


packages/core/src/odata/v2/request-builder/function-import-request-builder.ts
```diff
@@ -13,7 +19,9 @@ export class FunctionImportRequestBuilder<
   // reuse
   ParametersT,
   ReturnT
-> extends ActionFunctionImportRequestBuilderBase<ParametersT, ReturnT> {
+> extends MethodRequestBuilderBase<
```
FrankEssenberger, line 22: I think here you should not overwrite the changes for the introduction of actions?

FrankEssenberger: See also removed files below

packages/core/src/odata/common/entity-deserializer.ts
```diff
@@ -202,7 +205,12 @@ export function entityDeserializer(
     return complexType._propertyMetadata
       .map(property => ({
         ...(typeof json[property.originalName] !== 'undefined' && {
-          [property.name]: isComplexTypeNameSpace(property.type)
+          [property.name]: property.isCollection
```
FrankEssenberger, line 208: I find this kind of chaining of tenary operators in combination with the spread very hard to read. I would prefer a function(json,originalName){
-undfined check->return,
-isCollection->return deserializeCollectionType(),
-isComplect->return deserializeComplexType().
-isEdmToTs->return edmToTs
Throw exception because no case is found which should not be right?

packages/core/src/odata/common/entity-serializer.ts
```diff
@@ -136,7 +136,9 @@ export function entitySerializer(tsToEdm) {
     return complexType._propertyMetadata
       .map(property => ({
         ...(typeof fieldValue[property.name] !== 'undefined' && {
-          [property.originalName]: isComplexTypeNameSpace(property.type)
+          [property.originalName]: property.isCollection
```
FrankEssenberger, line 139: same comment as above.

packages/core/src/odata/common/selectable/collection-field.ts
```diff
@@ -26,9 +27,9 @@ export class CollectionField<
 
   constructor(
     fieldName: string,
-    entityConstructor: Constructable<EntityT>,
+    fieldOf: ConstructorOrField<EntityT>,
```
FrankEssenberger, line 30: The JSdoc for this states:
```
/**
 * Union type to represent the parent of a complex type field. This can either be an entity constructor or another complex type field.
 */
export type ConstructorOrField<EntityT extends EntityBase, ComplexT> =
```

So it is strange that this is used here in the collection field context. Can you explain?

packages/core/src/odata/v2/request-builder/function-import-request-builder.ts
marikaner: I have no idea, what happened here, I hope I restored order, but please check again, carefully.

packages/core/src/odata/common/entity-deserializer.ts
marikaner: I extracted this to a separate function

packages/core/src/odata/common/selectable/collection-field.ts
marikaner: In our Field Hierarchy we have (I abstracted the naming) the BaseField, the EdmTypeFields, the ComplexTypeFields, the EdmTypeComplexTypeFields and the CollectionField. The BaseField is abstract and out of the equasion. The EdmTypeFields are fields of an entity, that have an edm type. EdmTypeComplexTypeFields can only occur as fields of a complex type, so they have the `fieldOf`. The complex type field can occur as a field of a property or another complextype, so it has the `fieldOf`. The same applies to CollectionField. 😊

packages/core/src/odata/common/entity-deserializer.ts
```diff
@@ -192,6 +196,21 @@ export function entityDeserializer(
       );
   }
 
+  function deserializeComplexTypeProperty(
```
FrankEssenberger, line 199: ah, readability +1

FrankEssenberger: Question: Applies also to the serialize analogon below. Should we for the last step return the edmToTs or should we also do a check if it is edm and throw an exception otherwise so that we have a better error handling?

packages/core/src/odata/common/selectable/collection-field.ts
FrankEssenberger: ok fair, but then we should adjust the JSDoc description.

packages/core/src/odata/common/entity-deserializer.ts
marikaner: Hmm, what would a check like this look like? Pseudocode: `propertyMetadata.type is edmType`? For this our generation must be broken, I guess.
`edmToTs` is quite forgiving here. If there is no serializer for this type, we just pass the value as is, if there is and the format is not correct, the corresponding converters fail. I don't have any strong opinion here ¯\_(ツ)_/¯

packages/core/src/odata/common/selectable/collection-field.ts
marikaner: Good point, I moved it to a separate file and adjusted the docs.

packages/core/src/odata/common/entity-deserializer.ts
FrankEssenberger: ok if the edm is forgiving we can use it as a fallback.

PR #330, opened 2020-07-30
Fix: Use correct response accessors

marikaner:
## Context

Use centralized accessors for OData responses, that are different for v2/v4.

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/core/src/odata/v2/request-builder/response-data-accessor.ts
+++ b/packages/core/src/odata/v2/request-builder/response-data-accessor.ts
@@ -1,6 +1,6 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 
-export function getCollectionResult(data): any[] | any {
+export function getCollectionResult(data): any[]s {
   return data.d.results;
 }
 
@@ -10,7 +10,7 @@ export function isCollectionResult(data): boolean {
 
 export function getSingleResult(data): Record<string, any> {
   return isSingleResultAsCollection(data)
-    ? getCollectionResult(data)
+    ? data.d.results
     : data?.d || {};
 }
 
--- a/packages/core/src/odata/v2/request-builder/response-data-accessor.ts
+++ b/packages/core/src/odata/v2/request-builder/response-data-accessor.ts
@@ -4,11 +4,15 @@ export function getCollectionResult(data): any[] {
   return data.d.results;
 }
 
+export function isCollectionResult(data): boolean {
+  return Array.isArray(data.d.results);
+}
+
 export function getSingleResult(data): Record<string, any> {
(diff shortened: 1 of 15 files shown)


packages/core/src/odata/v2/request-builder/response-data-accessor.ts
```diff
@@ -0,0 +1,20 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+
+export function getCollectionResult(data): any[] | any {
```
FrankEssenberger, line 3: Minor: I find it strange that the `getCollectionResult` can return `any` and not only collection. I see the reason in the workaround. However I would do the follwoing. make the getCollection return always a collection since this is also exported.

Use for the getSingleResult() the following:
```
export function getSingleResult(data): Record<string, any> {
  return isSingleResultAsCollection(data)
    ? data.d.results
    : data?.d || {};
}
```

marikaner: Yes, I also did that in the beginning and changed it back and fro.

packages/core/src/odata/v2/request-builder/response-data-accessor.ts
```diff
@@ -0,0 +1,20 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: General: Perhaps it makes sense to anticipate the fact that C4C,SFSF, S/4 all not obey the OData specs to 100% and include some checks on the types of the results i.e. include a warn/error statement in the  getSingle and  getCollection in case the data type is not object or array. This would make potential issue opene by consumer easier to classify.

packages/core/src/odata/v4/request-builder/create-request-builder.ts
```diff
@@ -87,7 +87,7 @@ export class CreateRequestBuilder<EntityT extends Entity>
       .then(request => request.execute())
       .then(response =>
         deserializeEntity(
-          extractData(response),
+          getSingleResult(response.data),
```
FrankEssenberger, line 90: Is it always `.data` or should we move also this extraction part to the getter methods?

marikaner: It is not always data. In batch for example we first have to parse the reponse.data, to get this.

packages/core/src/odata/v2/request-builder/response-data-accessor.ts
marikaner: True. Would you also add that for the existing workaround? (My opinion: yes).

PR #336, opened 2020-07-31
Feature: Add unbound action to generator

FrankEssenberger:
## Context

Sorry that so many files are touched but a lot of them are just new generated services files introduced by the action. 
The two main code changes are:
- In the core I adjusted the signature of the request builder and request config to meet the requirements required by the action. Note that action do not have URL parameters. Everything is passed in the body.
- In the generator I created the necessary files to generate the actions. I created `action-function-import` as the combined home for actions and funcitons since they are so similar.

I am not super happy with the naming and reuse but I had no better idea. Feel free to contact me for improvements.

## Definition of Done

- [X] Tests created/adjusted for your changes.
- [X] Release notes updated.
- [X] PR title adheres to [conventional commit guidelines](https://www.conventionalcommits.org) (please note that only `fix:` and `feat:` will end up in the release notes)
- [X] If applicable: Properly documented (JSDoc of public API)
- [X] If applicable: Check if `node run doc` still works.


packages/generator/src/action-function-import/parameters-interface.ts
```diff
@@ -1,25 +1,25 @@
...
+    properties: actionFunctionImport.parameters.map(parameter => ({
       name: parameter.parameterName,
       type: parameter.jsType,
       hasQuestionToken: parameter.nullable,
       docs: [addLeadingNewline(parameter.description)]
     })),
     docs: [
       addLeadingNewline(
-        `Type of the parameters to be passed to [[${functionImport.functionName}]].`
+        `Type of the parameters to be passed to [[${actionFunctionImport.name}]].`
```
marikaner, line 22: Btw - one more reason to call it parameters instead of payload.

packages/generator/src/action-function-import/request-builder-arguments.ts
```diff
@@ -0,0 +1,22 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+import {
+  VdmActionImport,
+  VdmFunctionImport,
+  VdmServiceMetadata
+} from '../vdm-types';
+import { responseTransformerFunctionName } from './response-transformer-function';
+
+export function getRequestBuilderArgumentsBase(
+  parameterName: 'params' | 'payload',
```
marikaner, line 10: I would even call this params for both functions and actions.

packages/generator/src/action-function-import/function.ts
```diff
@@ -51,37 +51,18 @@ function getFunctionImportStatements(
       }, 'const params = {\n') + '\n}'
     : '{}';
 
-  const parameters =
-    service.oDataVersion === 'v2'
-      ? getFunctionImportRequestBuilderArgumentsV2(functionImport, service)
-      : getFunctionImportRequestBuilderArgumentsV4(functionImport, service);
+  let parameters = getRequestBuilderArgumentsBase(
```
marikaner, line 54: I can't fully wrap my head around this - `getRequestBuilderArgumentsBase` returns `string[]` and the last  item in the returned list is the first argument passed into the function? So the last item would be `'params'` in this case here?
Probably I'm just reading it wrong, but if not, I would argue that this should not be necessary.

packages/generator/src/edmx-to-vdm/common/action-function-parameters.ts
```diff
@@ -0,0 +1,65 @@
...
+): VdmParameter[] {
+  return getParameter(
+    edmxActionImport,
+    edmxParameters,
+    swaggerDefinition,
+    formatter
+  );
+}
+
+export function getFunctionImportParameters(
```
marikaner, line 29: As those two functions are basically the same, why don't we just export `getParameters`?

packages/generator/src/edmx-to-vdm/common/action-function-parameters.ts
```diff
@@ -0,0 +1,65 @@
...
+): VdmParameter[] {
+  return getParameter(
+    edmxFunctionImport,
+    edmxParameters,
+    swaggerDefinition,
+    formatter
+  );
+}
+
+function getParameter<T extends EdmxNamed>(
```
marikaner, line 43:
```suggestion
function getParameters<T extends EdmxNamed>(
```

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -0,0 +1,139 @@
...
+  complexTypes: Omit<VdmComplexType, 'factoryName'>[]
+): VdmActionImportReturnType {
+  return parseReturnTypes(
+    returnType,
+    entities,
+    complexTypes
+  ) as VdmActionImportReturnType;
+}
+
+function parseReturnTypes(
```
marikaner, line 39: Same as for the parameters, why not just exporting this function instead of the two above. Does anyone down the road even care about this being returned `as VdmActionImportReturnType` (or VdmActionImportReturnType)?

packages/generator/src/edmx-to-vdm/common/action-import.ts
```diff
@@ -0,0 +1,47 @@
...
+export function transformActionImportBase(
+  edmxActionImport: EdmxActionImport,
+  edmxParameters: EdmxParameter[],
+  swaggerDefinition: SwaggerPath | undefined,
+  formatter: ServiceNameFormatter
+): VdmActionImportBase {
+  const actionName = formatter.originalToActionImportName(
+    edmxActionImport.Name
+  );
+  const actiononImport = {
```
marikaner, line 26:
```suggestion
  const actionImport = {
```

packages/generator/src/edmx-to-vdm/common/action-import.ts
```diff
@@ -0,0 +1,47 @@
...
+  messageContext: 'function-import'
+});
+
+export function transformActionImportBase(
+  edmxActionImport: EdmxActionImport,
+  edmxParameters: EdmxParameter[],
+  swaggerDefinition: SwaggerPath | undefined,
+  formatter: ServiceNameFormatter
+): VdmActionImportBase {
+  const actionName = formatter.originalToActionImportName(
```
marikaner, line 23:
```suggestion
  const name = formatter.originalToActionImportName(
```

Then on line 28 you can simplify a bit as well.

packages/generator/src/edmx-to-vdm/common/function-import.ts
```diff
@@ -2,64 +2,40 @@
...
-export function transformFunctionImportBase<T extends EdmxNamed>(
-  edmxFunctionImport: T,
+export function transformFunctionImportBase(
+  edmxFunctionImport: EdmxFunctionImportV2 | EdmxFunctionImportV4,
   edmxParameters: EdmxParameter[],
   swaggerDefinition: SwaggerPath | undefined,
   formatter: ServiceNameFormatter
-): Omit<VdmFunctionImport, 'httpMethod' | 'returnTypeEdmx' | 'returnType'> {
+): VdmFunctionImportBase {
   const functionName = formatter.originalToFunctionImportName(
```
marikaner, line 24:
```suggestion
  const name = formatter.originalToFunctionImportName(
```
Same as for action.

packages/generator/src/edmx-to-vdm/v2/function-import.ts
```diff
@@ -29,7 +30,11 @@ export function generateFunctionImportsV2(
         formatter
       ),
       httpMethod,
-      returnType: parseReturnTypes(f.ReturnType, entities, complexTypes)
+      returnType: parseFunctionImportReturnTypes(
```
marikaner, line 33: See my comment in `action-function-return-types.ts`. If the type is important here just add the `as` here.

packages/generator/src/edmx-to-vdm/v4/action-import.ts
```diff
@@ -0,0 +1,65 @@
...
+  parseActionImport,
+  parseActions
+} from '../../edmx-parser/v4';
+import { ServiceMetadata } from '../../edmx-parser/edmx-file-reader';
+import { VdmActionImport, VdmComplexType, VdmEntity } from '../../vdm-types';
+import { parseActionImportReturnTypes } from '../common/action-function-return-types';
+import { stripNamespace } from '../edmx-to-vdm-util';
+import { transformActionImportBase } from '../common/action-import';
+
+export function findActionForActionImport(
```
marikaner, line 16: It seems that you could reuse the generic functionality here? I guess there should be something in common?

packages/generator/src/edmx-to-vdm/v4/function-import.ts
```diff
@@ -3,14 +3,30 @@ import { ServiceNameFormatter } from '../../service-name-formatter';
...
   parseFunctions
 } from '../../edmx-parser/v4';
 import { ServiceMetadata } from '../../edmx-parser/edmx-file-reader';
-import { stripNamespace } from '../edmx-to-vdm-util';
 import { VdmComplexType, VdmEntity, VdmFunctionImport } from '../../vdm-types';
-import { parseReturnType } from '../../edmx-parser/legacy/common';
+import { parseFunctionImportReturnTypes } from '../common/action-function-return-types';
+import { stripNamespace } from '../edmx-to-vdm-util';
+
+export function findFunctionForFunctionImport(
```
marikaner, line 16: Same as for action.

packages/generator/src/action-function-import/parameters-interface.ts
FrankEssenberger: yes

packages/generator/src/edmx-to-vdm/v4/action-import.ts
FrankEssenberger: I add a shared function.

packages/core/src/odata/v4/request/action-import-payload.ts
```diff
@@ -1,6 +1,6 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 
 export {
-  FunctionImportParameter as ActionImportParameter,
-  FunctionImportParameters as ActionImportParameters
+  FunctionImportParameter as ActionImportPayloadElement,
```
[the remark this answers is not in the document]
FrankEssenberger: I put the name back to Parameter and only in the config made the cast to the payload.

packages/generator/src/edmx-parser/v4/edmx-parser.ts
```diff
@@ -69,9 +72,21 @@ export function parseFunctionImports(root): EdmxFunctionImport[] {
...
+
 export function parseFunctions(root): EdmxFunction[] {
   return forceArray(root.Function).map(f => ({
     ...f,
-    Parameter: forceArray(f.Parameter)
+    Parameter: forceArray(f.Parameter),
+    IsBound: false
+  }));
+}
+export function parseActions(root): EdmxAction[] {
```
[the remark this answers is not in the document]
FrankEssenberger: as for the other cases I create a private generic method and exported the two concrete ones delegating the work.

test-resources/service-specs/v4/API_TEST_SRV/API_TEST_SRV.edmx
```diff
@@ -146,6 +146,31 @@
         <Member Name="Member2" Value="2"/>
       </EnumType>
 
+
+      <Action Name="TestActionImportNoParameterNoReturnType">
+      </Action>
+
+<!--
```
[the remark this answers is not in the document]
FrankEssenberger: ah yes. Since the code is roughly the same for action and funciton I thought two test cases: an action with nothing and an action with everything is enough.

packages/generator/src/edmx-to-vdm/v2/function-import.ts
FrankEssenberger: see above

PR #347, opened 2020-08-03
Pull up the docs section, adjust levels

artemkovalyov:
## Context

Pull up the docs as they are the first thing developers are searching for and adjusting levels to be consistent with the hierarchy.


README.md
```diff
@@ -20,47 +22,44 @@ An SDK to reduce your development effort when building side-by-side extension ap
 - [Links](#links)
 - [License](#license)
 
-## Packages
+### Packages
```
marikaner, line 25: Hmm, I would prefer this and the following sections to be one level up (## instead of ###), they are not subsections of the TOC, are they?
Why did you move them down?

PR #378, opened 2020-08-12
feat: support enum

jjtang1985:
## Context

- [x] commit all the generated files

## Definition of Done

- [ ] Release notes updated.


packages/core/src/odata/common/selectable/enum-field.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Represents a selectable property with a enum value.
+ *
+ * @typeparam EntityT - Type of the entity the field belongs to
+ */
+export class EnumField<EntityT extends EntityBase>
+  extends EnumFieldBase<EntityT>
+  implements SelectableEdmTypeField {
+  readonly selectable: true;
+
+  edmEnumType: string;
```
FrankEssenberger, line 34: In my IDE these two are marked as unused. Are they used somewhere and if so for what?

FrankEssenberger:
As discussed we stick to version 4.0.1 of OData v4 which does not need the namespace with the enum.

FrankEssenberger:
Just let me know when this is ready for review.

FrankEssenberger:
Just let me know when this is ready for review.

jjtang1985:
@FrankEssenberger , the generator part is added. For test entities, only the core part is committed. I'll add all the generated files when necessary or before merging.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 ## New Functionality
 
+- Support Enum type.
```
FrankEssenberger, line 21: perhaps add a bit more context that it is about enum types in OData v4 in the generated client. Type safety is there and so on.

packages/core/src/odata/common/selectable/collection-field.ts
```diff
@@ -18,7 +18,8 @@ import { ConstructorOrField } from './constructor-or-field';
  */
 export class CollectionField<
   EntityT extends EntityBase,
-  FieldT extends EdmTypeShared<'any'> | Record<string, any> = any
+  // todo there is not super class for enum like TestEnumType
```
FrankEssenberger, line 21: I do not really get this comment. Is it about that there is not proper type we can put for the enums, because each one is dynamically generated?

packages/core/src/odata/common/selectable/enum-field.ts
```diff
@@ -0,0 +1,73 @@
...
+import { EdmTypeField, SelectableEdmTypeField } from './edm-type-field';
+
+/**
+ * Represents a property with a enum value.
+ *
+ * @typeparam EntityT - Type of the entity the field belongs to
+ */
+class EnumFieldBase<EntityT extends EntityBase> extends EdmTypeField<
+  EntityT,
+  string
```
FrankEssenberger, line 16: Add a  comment here why string is used as a temporary solution to the type problem.

FrankEssenberger:
@jjtang1985 Could you merge master. I just reached the generator part in the review and I think I changed a few things there which will lead to conflicts for you. Sorry for that :-(

jjtang1985:
@FrankEssenberger , I merged the master and resolved the merge conficts.

packages/generator/src/edmx-to-vdm/common/entity.ts
```diff
@@ -90,9 +93,20 @@ function properties(
       entity.entitySet.Name,
       p.Name
     );
-    const isComplex = isComplexType(p.Type);
     const isCollection = isCollectionType(p.Type);
-    const typeMapping = getTypeMappingEntityProperties(p.Type, complexTypes);
+    const isComplex = isCollection
+      ? isComplexType(parseCollectionTypeName(p.Type), complexTypes)
```
FrankEssenberger, line 98: Minor: It is just a personal taste but I try to avoid mutiple calls like `parseCollectionTypeName` So I woul introdoced a `const parsed = isCollection ? arseCollectionTypeName(p.Type) : pType. And then call the `isEnumType(parsed)` and `isComplexType(parsed). Then you also have one tenary operator.

packages/generator/src/edmx-to-vdm/common/complex-type.ts
```diff
@@ -65,11 +63,23 @@ export function transformComplexTypesBase(
           c.Name,
           p.Name
         );
-        const isComplex = isComplexType(p.Type);
         const isCollection = isCollectionType(p.Type);
+        const isComplexOrEnum = isComplexTypeOrEnumType(
```
FrankEssenberger, line 67: goes also in the direction of my comment above. I find this very complicated, which is not avoidable since there are so many cases. But I would do the following. 
- Put all the isSomething methods in the edmx-to-vdm-util class.
- Include the collection check inside the isSomething method. So isEnum() returns true on a collection of enums and a enum input. Since it is used this way anyway it will reduce complexity in the coden
- I know it is more efficient the way you done it. For example if `isComplexOrEnum=true` and you later find isEnum=true it can not be a complex. However, I would find a less efficient but more symmetric flow nice:
```
isCollection = isCollection(string)
isEnum= isEnum(string,enums) //uses the enumOrComplex internally
isComplex = isComplex(string,complexTypess) //uses the enumOrComplex internally
```

packages/generator/src/edmx-to-vdm/common/complex-type.ts
```diff
@@ -88,16 +98,21 @@ export function transformComplexTypesBase(
...
-      }).filter(filterUnknownPropertyTypes)
+      })
     };
   });
 }
 
 export function getTypeMappingComplexProperties(
   typeName: string,
-  formattedTypes: MapType<any>
+  enumTypes: VdmEnumType[],
```
FrankEssenberger, line 111: Minor & genral understanding: It seems like for complex type there is a rule between the typeName and the actual type. See `complexTypeFieldType(formattedTypes[withoutPrefix]`. For enums it also seems like this:  `e.originalName === getPostfix(name)` if you look in the `findEnumType`. For me both approaches are fine. Using a list of types and find the right one or extract from the typeName. However, we should not mix.

packages/generator/src/edmx-to-vdm/edmx-to-vdm-util.ts
```diff
@@ -33,6 +33,10 @@ export function complexTypeName(type: string): string | undefined {
   return last(type.split('.'));
 }
 
+export function enumTypeName(type: string): string | undefined {
```
FrankEssenberger, line 36: This is marked unused for me.

packages/generator/src/edmx-to-vdm/edmx-to-vdm-util.ts
```diff
@@ -111,7 +116,14 @@ export function typesForCollection(
       fieldType: 'CollectionField'
     };
   }
-  if (isComplexType(typeInsideCollection)) {
+  if (isComplexTypeOrEnumType(typeInsideCollection)) {
```
FrankEssenberger, line 119: Also here I would make it symmetric between enum and complex type:
```
if(isEnumType(string,enums)){
return enumStuff
}
if(isComplex(string,complexTypes)){
return ComplexStuff
}

thorw execption cause nothing is matched.
```

packages/generator/src/enum-type/enum.ts
```diff
@@ -0,0 +1,81 @@
...
+    isExported: true,
+    members: enumType.members.map(m => ({
+      name: m,
+      value: m
+    }))
+  };
+}
+
+export function fieldTypeClass(
+  complexType: VdmComplexType
```
FrankEssenberger, line 32: rename argument `complexType` argument. And also if the VdmComplexType can represent an VdmEnumType I would export it once more to make the typing nicer.

packages/generator/src/enum-type/enum.ts
```diff
@@ -0,0 +1,81 @@
...
+): ClassDeclarationStructure {
+  return {
+    kind: StructureKind.Class,
+    name: `${complexType.fieldType}<EntityT extends Entity>`,
+    extends: `ComplexTypeField<EntityT, ${complexType.typeName}>`,
```
FrankEssenberger, line 37: I guess this is also a copy paste error and it should use the enum-field?

packages/generator/src/enum-type/enum.ts
```diff
@@ -0,0 +1,81 @@
...
+            name: 'fieldName',
+            type: 'string'
+          },
+          {
+            name: 'fieldOf',
+            type: 'ConstructorOrField<EntityT>'
+          }
+        ],
+        docs: [
+          `\nCreates an instance of ${complexType.fieldType}.\n\n@param fieldName - Actual name of the field as used in the OData request.\n@param fieldOf - Either the parent entity constructor of the parent complex type this field belongs to.`
```
FrankEssenberger, line 54: Sentence also needs adjustment.

packages/generator/src/enum-type/enum.ts
```diff
@@ -0,0 +1,81 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: General: Search and replace for all `complex`. Should be removed, renamed or adjusted.

packages/generator/src/edmx-to-vdm/edmx-to-vdm-util.ts
jjtang1985: The caller (and the caller of the caller) of this function is creating the `complexTypes`, that's why you have to determine `isComplex` by using `isComplexTypeOrEnumType` and `isEnum`

packages/generator/src/generator-utils.ts
```diff
@@ -68,7 +70,8 @@ const edmToTsTypeMapping: EdmTypeMapping = {
   // OData v4 specific
   'Edm.Date': 'Moment',
   'Edm.Duration': 'Duration',
-  'Edm.TimeOfDay': 'Time'
+  'Edm.TimeOfDay': 'Time',
+  'Edm.Enum': undefined
```
FrankEssenberger, line 74: is this correct? Isn't it string on a JS level?

packages/generator/src/enum-type/enum.ts
jjtang1985: good catch

packages/generator/src/enum-type/enum.ts
jjtang1985: same

packages/generator/src/enum-type/enum.ts
jjtang1985: same

packages/generator/src/enum-type/enum.ts
jjtang1985: same

packages/generator/src/generator-utils.ts
```diff
@@ -185,23 +189,37 @@ export function getGenericParameters(
   entityClassName: string,
   prop: VdmProperty
 ): string {
-  const param = prop.isCollection
-    ? prop.isComplex
-      ? [`${prop.jsType}`]
-      : [`'${prop.edmType}'`]
-    : [];
+  let param;
```
FrankEssenberger, line 192: Minor: You can initialize with [] here and get rid of the `else` black later on.

packages/generator/src/generator-utils.ts
```diff
@@ -185,23 +189,37 @@ export function getGenericParameters(
...
 export function createPropertyFieldInitializer(
   property: VdmProperty,
   entityClassName: string
 ) {
-  const edmOrComplexType = property.isComplex
-    ? property.jsType
-    : `'${property.edmType}'`;
-  const collectionTypeOrEdmType =
-    property.isComplex && !property.isCollection ? undefined : edmOrComplexType;
+  const edmOrComplexTypeOrEnumType =
```
FrankEssenberger, line 211: This block is a bit messy I would say. I would refactor it in the following way:
- create a helper function for creating the ``new ${property.fieldType}(${[``
- This helper function needs the `isCollection` and `type:string` as input.
- handle the special case of (isCollection=true and enum) 
- Then you call this helper for the concrete three cases:
```
if(isEnum){
return helper(isCollection,property.jsType)
}
if(isComplex){
return helper(isCollection,property.jsType)
}
if(isEdm){
helper(isCollection,property. edmType)
}

jjtang1985: I personal feel this is hard to understand as well, but it's not easy to refactor like what you said.
Because we want to decide `collectionTypeOrEdmType` by checking 3 boolean variables (isEnum, isComplex and isCollection), which has 2^3 = 8 cases. For 3 cases out of 8, we have "special treatment" and the rest of them (5 cases), we return something like a "default value".
In this way, the if-else condition block is a bit longer than your example.
I refactored a bit, so you really see 3 "special cases" and the "default behaviour".

packages/generator/src/edmx-to-vdm/common/complex-type.ts
jjtang1985: This function is only used only once, where we iterate all the properties of a complex type.
Without enum, the current master checks the following condition like a `switch` keyword:
- isEdm
- isCollection
- isComplex
- error if not match (enum is filtered before)
Now enum comes, I removed the filter and added a new condition so the `switch` it becomes:
- isEdm
- isCollection
- isComplex
- isEnum
- error if not match

PR #383, opened 2020-08-14
Fix: Add warning if auth headers from two sources

FrankEssenberger:
## Context

We had an issue reported [here](https://answers.sap.com/questions/13112827/sap-cloud-sdk-executehttprequest-request-failed-wi.html). It is not yet 100% confirmed, but I am pretty sure that the reason was that the user passed a request in the execute HTTP request with auth headers and the destination was a normal onPrem one also leading to auth headers. In this case the custom one wins. Which is fine, but in most cases unintentional I would say. This PR adds a warning if the a user has auth headers from two sources (destination & request config).

## Definition of Done

- [x] Release notes updated.
~~- [] If applicable: Properly documented (JSDoc of public API)~~


CHANGELOG.md
```diff
@@ -22,7 +22,7 @@
 
 ## Improvements
 
--
+- Add warning in case the destination, and the custom request config contain both `authorization` headers.
```
mr-flannery, line 25:
```suggestion
- Log a warning in case the destination and the custom request config contain both `authorization` headers. 
```

packages/core/src/header-builder/authorization-header.ts
```diff
@@ -49,11 +50,31 @@ export async function addAuthorizationHeader<
...
   destination: Destination,
   customHeaders?: MapType<any>
 ): Promise<MapType<string>> {
   const customAuthHeaders = getHeader('authorization', customHeaders);
+
+  if (Object.keys(customAuthHeaders).length && hasAuthHeaders(destination)) {
+    logger.warn(`You provided authorization headers in the request config. 
+    However, your destination ${destination.name} also provides authorization headers.
+    To have authorization information from both sources is often unintended.
+    The custom headers from the request config will be used.`);
```
mr-flannery, line 76: Using a string template here might result in unintentional formating of the output, I think the line breaks and indentation will be reflected one-to-one. Consider manually concatenating strings instead.

FrankEssenberger: ah good point. Thanks

PR #385, opened 2020-08-14
Fix: Wrong type in argument of Collections 

FrankEssenberger:
## Context

This PR fixes two things which are not related but are hard to fix seperately.
1. In this [commit](https://github.com/SAP/cloud-sdk/commit/fbcae2834c65bc8992bf18dbe791287aab648327) @marikaner fixed the links for the TSDoc of the generated clients. However, the clients were not regenerated.  That is why so many files are changes
2. I introduced an error in the generator in [this commit](https://github.com/SAP/cloud-sdk/commit/27be8568a7f918395a79c7156df0e2d496b52e0b). The error was the following. For collections containing and Edm type the isComplex should be false so that at the very end in the `createPropertyFieldInitializer` the edm type is put and not the js as it was now:
```
CollectionField<TestEntity, 'Edm.String'> = new CollectionField('CollectionProperty', TestEntity, 'Edm.String')//right
CollectionField<TestEntity, 'string'> = new CollectionField('CollectionProperty', TestEntity, 'string')//wrong
```

Since the bug lead to in import like `import {string} from './string'`  we should create a patch release as soon as possible. 

## Definition of Done

- [x] Release notes updated.


FrankEssenberger:
@jjtang1985 PR #378 will most likely also fix this issue. So will merge his PR first. Resolve the conflicts on this one and then see if the test is still green. Then we can merge it for the extra test.

PR #394, opened 2020-08-18
Chore: Fix moment warning

FrankEssenberger:
## Context
 
I added a warning for non named imports and moment showed up. Since github shows the warnings in every review for PRs I was annoyed and wanted to fix the warning. 

I think the moment guys did not export the namespace/module properly see here: https://github.com/moment/moment/issues/5679

Nevertheless I did not want to wait for their fix so I imported the namespace more explicitly or imported everything via *. Not so beautiful like the real ES6 imports but the warning is goen.


FrankEssenberger:
@mr-flannery thanks for doing a review for the SDK. Currently @marikaner @jjtang1985 are on holiday so you are kind of my obi wan.

- It would fail in the lint phase in the pipeline or locally, if somebody changes it.
- I will add some document explaining the reasons
- I thought I needed both. Was wrong changed it to the one which is working for all methods.

PR #398, opened 2020-08-19
Feature: Export OData v4 to root for API generation

FrankEssenberger:
## Context

In this PR I moved the export of the v4 entities to the root. Due to ambiguities like `Entity` we decided to append a `V2` or `V4` if needed and reexport the `EntityV2` as `Entity` for compatibility. Unfortunately the generator needed quite some adjustment, because always the right `V2` or `V4` versions needs to be imported and used. Please have a look at the generator changes because these are the only real ones. You can find them easily by searching for the

`caps()` methods. I used this to transform the OData version v2/v4 to V2/V4 and to have a reference where I adjusted the generator.

Besides that a lot of tests needed adjustment so this huge number of file has been touched. I check the generated API docs and only the `EntityV2` and `EntityV4` appear. Not the reexported entry. Also the legacy service still compiles which means that I have not forgotten to reimport a used object. At least this is the hope.

**Since this PR is so huge, I would like to merge it before somebody does anything else to avoid conflicts.**

## Definition of Done

- [x] Release notes updated.


packages/core/src/odata/common/selectable/complex-type-field.ts
```diff
@@ -117,7 +117,7 @@ export function getEntityConstructor<EntityT extends EntityBase, ComplexT>(
  * Convenience method to get the [[EdmTypeShared]] from the overloaded constructor.
  * @param complexTypeNameOrEdmType - Either the name of the complex type or the EdmType
  * @param edmTypeOrUndefined - Either the EdmType or undefined.
- * @returns EdmType
+ * @returns EdmTypeV4
```
jjtang1985, line 120:
```suggestion
 * @returns EdmType
```

packages/core/src/odata/common/entity.ts
```diff
@@ -234,7 +234,7 @@ export abstract class EntityBase {
    * Returns all changed properties compared to the last known remote state.
    * The returned properties does not include custom fields. Use [[getUpdatedCustomFields]], if updated custom fields are needed.
    *
-   * @returns Entity with all properties that changed
+   * @returns EntityV4 with all properties that changed
```
jjtang1985, line 237:
```suggestion
   * @returns Entity with all properties that changed
```

packages/core/src/odata/common/entity.ts
```diff
@@ -256,7 +256,7 @@ export abstract class EntityBase {
   /**
    * Returns a map of all defined fields in entity to their current values.
    *
-   * @returns Entity with all defined entity fields
+   * @returns EntityV4 with all defined entity fields
```
jjtang1985, line 259:
```suggestion
   * @returns Entity with all defined entity fields
```

packages/core/src/odata/common/entity-builder.ts
```diff
@@ -49,7 +49,7 @@ export class EntityBuilder<EntityT extends EntityBase, JsonT> {
    * If you have obtained the JSON as a request payload use the [[deserializeEntity]] methods.
    * Note that fields not mappable to a field in the target entity are silently ignored.
    * @param json - Representation of the entity in JSON format.
-   * @returns Entity constructed from JSON representation.
+   * @returns EntityV4 constructed from JSON representation.
```
jjtang1985, line 52:
```suggestion
   * @returns Entity constructed from JSON representation.
```

packages/core/src/odata/v2/entity-deserializer.ts
```diff
@@ -4,19 +4,19 @@ import {
...
-  extractODataETag,
+  edmToTsV2,
+  extractODataEtagV2,
   extractDataFromOneToManyLink
 );
 
-export const deserializeEntity = deserializer.deserializeEntity;
-export const deserializeComplexType = deserializer.deserializeComplexType;
+export const deserializeEntityV2 = deserializer.deserializeEntity;
+export const deserializeComplexTypeV2 = deserializer.deserializeComplexType;
```
jjtang1985, line 22: breaking changes

packages/core/src/odata/v2/entity-serializer.ts
```diff
@@ -4,15 +4,15 @@ import {
...
  */
-const serializer: EntitySerializer = entitySerializer(tsToEdm);
+const serializer: EntitySerializer = entitySerializer(tsToEdmV2);
 
-export const serializeEntity = serializer.serializeEntity;
-export const serializeComplexType = serializer.serializeComplexType;
-export const serializeEntityNonCustomFields =
+export const serializeEntityV2 = serializer.serializeEntity;
+export const serializeComplexTypeV2 = serializer.serializeComplexType;
+export const serializeEntityNonCustomFieldsV2 =
```
jjtang1985, line 17: breaking changes

packages/core/src/odata/v2/entity-v2.ts
```diff
@@ -1,18 +1,20 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 
 import { EntityBase, Constructable } from '../common';
-import { CustomField } from './selectable/custom-field';
+import { CustomFieldV2 } from './selectable/custom-field-v2';
```
jjtang1985, line 4:
```suggestion
import { CustomField } from './selectable/custom-field-v2';
```

jjtang1985 (changes requested):
Since there are 600+ files, I chose some vdm files/core files/test files for reviewing as first step. Basically the main question is how to keep minimal breaking changes for the users. I think we should use class/function names without `v2` suffix if it's used only for v2. e.g., We have `export { EntityV2 as Entity }` in `entity-v2.ts`, we can use `import { Entity } from './entity-v2';` instead of `import { EntityV2 } from './entity-v2';`

packages/core/src/odata/v2/entity-deserializer.ts
FrankEssenberger: after discussing also with marika, we add the export `deserializeEntity` to ensure compatibility since only the ` entityDeserializer` was @experimental.

test-packages/test-services/v2/test-service/TestEntity.ts
```diff
@@ -7,12 +7,12 @@ import { TestEntityRequestBuilder } from './TestEntityRequestBuilder';
 import { Moment } from 'moment';
 import { BigNumber } from 'bignumber.js';
 import { TestComplexType, TestComplexTypeField } from './TestComplexType';
-import { AllFields, AnyField, BigNumberField, BooleanField, CustomField, DateField, Entity, EntityBuilderType, Field, Link, NumberField, OneToOneLink, StringField, Time, TimeField } from '@sap-cloud-sdk/core';
+import { AllFields, AnyField, BigNumberField, BooleanField, CustomFieldV2, DateField, EntityBuilderType, EntityV2, Field, Link, NumberField, OneToOneLink, StringField, Time, TimeField } from '@sap-cloud-sdk/core';
```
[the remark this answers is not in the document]
FrankEssenberger: as discussed we use V2,V4 for consistency.

packages/core/src/odata/common/entity.ts
FrankEssenberger: V4 removed

packages/core/src/odata/common/entity-builder.ts
FrankEssenberger: +1

packages/core/src/odata/v2/entity-serializer.ts
FrankEssenberger: adjusted

packages/core/src/odata/v2/entity-deserializer.ts
FrankEssenberger: adjusted

packages/core/src/odata/v2/entity-v2.ts
FrankEssenberger: I am also not sure if this is correct.

knowledge-base/architectural-decisions-records/v4-api.md
```diff
@@ -251,7 +251,7 @@ because it replaces the existing data with the payload.
   - delete entities (`@removed`)
   - delete links (`reason = deleted`)
   - add links (with id/key)
-  - update entities (with id/key + Partial<Entity>)
+  - update entities (with id/key + Partial<EntityV2>)
```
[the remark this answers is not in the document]
FrankEssenberger: I would revert all changes, because for this MD files I think it is not relevant if in some code snipped the EntityV4 or Entity is given. I think we understand what is meant.

packages/core/src/odata/common/edm-types.ts
```diff
@@ -31,8 +31,8 @@ export type EdmTypeV4 =
 // The generic parameter is currently unused. We still have to revise whether we can use it in a later version of typescript.
 export type EdmTypeShared<VersionT extends ODataVersion | 'any'> =
```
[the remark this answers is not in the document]
FrankEssenberger: I would keep it, so that we remember to fix this sometime in the future. (Later means never...)

packages/core/src/odata/common/filter/collection-filter-function.ts
```diff
@@ -1,6 +1,7 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 import { EntityBase } from '../entity';
-import { EdmTypeShared, FieldType } from '../../v2';
+import { FieldType } from '../selectable';
```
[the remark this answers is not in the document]
FrankEssenberger: I think the compiler complained after I changed the index.ts of the v2 and v4 not to export things.

packages/core/src/odata/common/filter/orderable-filter-function.ts
```diff
@@ -1,8 +1,7 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 
 import { EntityBase, ODataVersionOf } from '../entity';
-import { EdmTypeShared } from '..';
-import { FieldType } from '../../v2';
+import { EdmTypeShared, FieldType } from '..';
```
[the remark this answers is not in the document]
FrankEssenberger: hm, I wonder how this happened. Adjusted.

packages/core/src/odata/common/uri-conversion/get-keys.ts
```diff
@@ -3,7 +3,7 @@
 import { MapType } from '@sap-cloud-sdk/util';
 import { EntityBase, Constructable } from '../entity';
 import { toPropertyFormat } from '../../../util';
-import { Field } from '../../v2';
+import { Field } from '../selectable';
```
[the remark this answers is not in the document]
FrankEssenberger: added a ticket

packages/core/src/odata/common/uri-conversion/uri-value-converter.ts
```diff
@@ -40,7 +40,7 @@ export function convertToUriForEdmString(value: any): string {
 
 /**
  * Interface defining the methods of the URI converter.
- * The concrete implementations are created in odata/v2/uri-conversion/uri-value-converter.ts and odata/v4/uri-conversion/uri-value-converter.ts
+ * The concrete implementations are created in odata/v2/uri-conversion/uri-value-converter-v2.ts and odata/v4/uri-conversion/uri-value-converter-v2.ts
```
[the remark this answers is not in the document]
FrankEssenberger: When I refactored the names of the classes the IDE asked  `should I change also the filename?` I have no trong opinion here. In cases where we have a v2 parent folder I think we can get rid of this.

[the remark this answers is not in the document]
FrankEssenberger: @jjtang1985 what do you think?

packages/core/src/odata/v2/entity-v2.ts
jjtang1985: I think yes, but please correct me. In the `custom-field-v2.ts` file, there is one export: `export { CustomFieldV2 as CustomField };`.
But I guess the decision is to use `V2` as suffix for all classes/interfaces like `ABCV2`. So this comment is not relevant any more.

packages/core/src/odata/common/uri-conversion/uri-value-converter.ts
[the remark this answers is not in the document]
jjtang1985: For me, we decided to use the class/interface name with `v2` suffix, because of the limit of the documentation. If the doc can live with the same file name, I would go without specifying the suffix, since the folder name has this information.

packages/core/src/odata/v2/entity-v2.ts
FrankEssenberger: yes this was the idea. It will import the same thing just via two names

packages/core/src/odata/common/uri-conversion/uri-value-converter.ts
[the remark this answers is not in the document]
FrankEssenberger: I check with the documentation. If this works I will get rid fo the v2 v4 in folders.

jjtang1985 (approved):
Once the the file names are adjusted so the `v2`/`v4` info is removed, I'm fine for the current status.
Please check the docs manually to make sure we see new v4 docs and there is no unexpected changes for v2.
This is a temporary solution since the "chosen" Plan C (multiple packages) introduces breaking changes but we don't have plan to make a version 2 SDK for this quarter.

packages/core/src/odata/v2/entity-v2.ts
FrankEssenberger: I remove the suffix if we are in a `v2` `v4` folder.

packages/core/src/odata/common/uri-conversion/uri-value-converter.ts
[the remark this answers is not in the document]
FrankEssenberger: suffix removed if possible.

PR #404, opened 2020-08-20
Feature: Nested Expand

FrankEssenberger:
## Context

**Merge #398 first**

I did test driven design and it seems like nested expands are already working? I added a test to the `get-all-request-builder-v4.spec.ts` as well as the `TripPin/request-builder.spec.ts` and it works. Perhaps I missunderstood the requirement @artemkovalyov? So this is just adding test coverage.

I also adjusted the test entity to have three level of navigation to be sure that the recursive approach works and is not somehow just hard coding the first level case. But all fine.

The reason I made this PR against the branch `generateApiDocWithODataV4` is that I would hope to get this merged as the first thing when you two are back, since it is creating conflicts for sure with everything else. Once this is merged I will change it to master.

## Definition of Done

- [x] Release notes updated.


test-packages/test-services-e2e/TripPin/microsoft-o-data-service-sample-trippin-in-memory-models-service/Airlines.d.ts
```diff
@@ -0,0 +1,91 @@
+import { AirlinesRequestBuilder } from './AirlinesRequestBuilder';
```
jjtang1985, line 1: I'm a bit confused.
All the changes in this folder are about adding `.d.ts` files and `.d.ts.map` files, which do exist in the current master branch.

test-packages/test-services-e2e/v4/admin-service/TestEntity.d.ts
```diff
@@ -0,0 +1,213 @@
+import { TestEntityRequestBuilder } from './TestEntityRequestBuilder';
```
jjtang1985, line 1: Changes in this folder have the similar situation as the ones in `TripPin`

test-packages/test-services/v2/multiple-schemas-service/MultiSchemaTestEntity.d.ts
```diff
@@ -0,0 +1,79 @@
+import { MultiSchemaTestEntityRequestBuilder } from './MultiSchemaTestEntityRequestBuilder';
```
jjtang1985, line 1: This folder as well.

test-packages/test-services/v2/test-service/TestEntity.d.ts
```diff
@@ -0,0 +1,334 @@
+import { TestEntityRequestBuilder } from './TestEntityRequestBuilder';
```
jjtang1985, line 1: This folder as well

jjtang1985 (changes requested):
The expand was implemented recursively, so it's possible all the existing functions are working properly.
The 3rd level expand looks good for quality assurance.
However, in `test-packages`, I see lots of "new" files, which are not new at all.
Could you please check?

FrankEssenberger:
I think this could be related starting from the different branches. Once the big PR is merged I changed to Master and check your comments.

FrankEssenberger:
Ok after merging master it looks ok from the changes. Have a look @Junjie adn @marikaner

jjtang1985 (approved):
Ok, so all the strange `d.ts` file changes are gone.

FrankEssenberger:
I am also not sure what happened there.

PR #419, opened 2020-08-26
Fix:  Generator error for services without Entities

FrankEssenberger:
## Context

In the VDM we found an error for the `API_PRODUCT_AVAILY_INFO_BASIC` service. This service does not have any entities. Only function imports and some types. In this case the `Batch.ts` files were erroneous. I just skipped the generation of the batch file in this case. 

Should we add an extra test for this? This would mean a edmx without an entity?

## Definition of Done

- [x] Release notes updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 
 ## Fixed Issues
 
--
+- [Generator] Skip generation of Batch.ts for services without entities.
 
 
 # 1.27.0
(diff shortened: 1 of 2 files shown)


jjtang1985 (approved):
LGTM, if this solves the issue.
For me, creating a new test entity is not necessary.
It would be great if we can add some unit tests, when possible.
If you want to merge it without tests, please manually test it.

FrankEssenberger:
I tested it with the broken service. So I will merge it. Once we added the "break on generating syntactically wrong code" this corner case will also be tested in the E2E tests.

PR #421, opened 2020-08-26
docs: Align redirection

marikaner:
## Context

The existing generator redirect was broken. All of the redirect pages displayed a debug message saying "Feature Title". I think telling the user that this is about "Redirecting" is more straight forward. Also, because we had some very complicated links, I aligned links with file names.
I tested this locally and all links still work.

## Definition of Done

- [ ] Release notes updated.


jjtang1985:
I have some stupid questions. 
- How could you reproduce to see the "Feature Title"?
- Why do we have to use the redirect md files?
For example, `use-typed-odata-v4-client-for-javascript-and-typescript` will redirect to `odata-v4-client`. Couldn't we just link to `odata-v4-client` instead of `use-typed-odata-v4-client-for-javascript-and-typescript`?

marikaner:
> I have some stupid questions.
> 
> * How could you reproduce to see the "Feature Title"?
> * Why do we have to use the redirect md files?
>   For example, `use-typed-odata-v4-client-for-javascript-and-typescript` will redirect to `odata-v4-client`. Couldn't we just link to `odata-v4-client` instead of `use-typed-odata-v4-client-for-javascript-and-typescript`?

You can try for example: https://sap.github.io/cloud-sdk/docs/js/features/odata/use-typed-odata-client-for-js, but you have to look quickly because it jumps to the redirected page after half a second.

I think the redirects are there for historic reasons. If the links were sent to anyone before, they would be dead otherwise. I tried to find out whether there is a shorter notation for this, maybe in a central place, but I didn't find anything.

jjtang1985:
> but you have to look quickly because it jumps to the redirected page after half a second.

Yes, it's redirected to "Page Not Found" 😞

jjtang1985 (approved):

marikaner:
> > but you have to look quickly because it jumps to the redirected page after half a second.
> 
> Yes, it's redirected to "Page Not Found" 😞

Oh, I didn't even realize. I hope after my changes it works, if not, you will get another PR ;)

PR #423, opened 2020-08-27
docs: Try to fix redirect specific broken link

marikaner:
## Context

Not sure why this one redirect does not work, because it works locally. This is to try, whether the notation is different here.

## Definition of Done

- [ ] Release notes updated.


jjtang1985 (approved):
go go go.
Btw, I see the same redirect multiple times, only this is broken?
Does the local one works with the new version?

marikaner:
> go go go.
> Btw, I see the same redirect multiple times, only this is broken?
> Does the local one works with the new version?

Do you mean you get redirected to a redirect and then again to a redirect? if so, where?

PR #424, opened 2020-08-27
docs: More verbose, but redirects should work now

marikaner:
## Context

As in the title - I made the redirect more verbose, I don't see a reason why it wouldn't work correctly now...

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -17,7 +17,7 @@ module.exports = {
       apiKey: '84d46c71e9f2445436400effad7c4e1b',
       indexName: 'sap_cloud-sdk',
       // appId: 'app-id', // Optional, if you run the DocSearch crawler on your own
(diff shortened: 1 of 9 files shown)


docusaurus.config.js
```diff
@@ -198,7 +198,7 @@ module.exports = {
         theme: {
           customCss: require.resolve('./src/css/custom.css')
         },
-        sitempa: {
+        sitemap: {
```
jjtang1985, line 201: this config never works?

jjtang1985 (approved):
looks good

docusaurus.config.js
marikaner: It can't ;) but it is only supposed to improve performance, locally I can't see a difference. The search was fast before as well.

PR #425, opened 2020-08-27
Add test for write with CRUD service

FrankEssenberger:
## Context

In this PR I added a test fro write/update and delte operation in interaction with a CAP generated service. The functionality works more or less out of the box.

However, I found that the `etag` is not returned from the server. The entity is persisted but for productive usage the CAP colleagues should add this right?

## Definition of Done

~~- [ ] Release notes updated.~~
~~- [ ] PR title adheres to [conventional commit guidelines](https://www.conventionalcommits.org) (please note that only `fix:` and `feat:` will end up in the release notes)~~


test-packages/e2e-tests/db/schema.cds
```diff
@@ -10,7 +10,5 @@ entity TestEntity {
   DoubleProperty: Double;
   DecimalProperty: Decimal(9,2);
   DateProperty: Date;
-  TimeOfDayProperty: Time;
```
jjtang1985, line 13: This property is about the `Edm.TimeOfDay`, which is a valid v4 data type.

See [here](https://github.wdf.sap.corp/MA/cap-odata-v4/blob/master/edm_type_to_cap_type.md) for cap data type to edm type mapping.
See [here](http://docs.oasis-open.org/odata/odata/v4.0/csd01/part3-csdl/odata-v4.0-csd01-part3-csdl.html#_Toc355092835) for OData V4 primitive data types.

test-packages/e2e-tests/db/schema.cds
```diff
@@ -10,7 +10,5 @@ entity TestEntity {
...
-  DataTimeOffsetDataTimeProperty: DateTime;
```
jjtang1985, line 14: This property is about the `Edm.DateTimeOffset`, which is a valid v4 data type.

packages/core/test/test-util/test-services/v4/test-service/TestComplexType.ts
```diff
@@ -62,16 +62,6 @@ export interface TestComplexType {
    * @nullable
    */
   floatProperty?: number;
-  /**
-   * Time Property.
-   * @nullable
-   */
-  timeProperty?: Time;
```
jjtang1985, line 69: I know there is a task to remove invalid data type in v4, but maybe we can make a separate pr.

test-packages/e2e-tests/test/request-builder.spec.ts
```diff
@@ -1,11 +1,48 @@
...
+    .getByKey(key.keyInt, key.keyString)
+    .execute(destination);
+}
+
+async function deleteEntity(key: EntityKey): Promise<void> {
+  try {
+    const fetched = await queryEntity(key);
+    return TestEntity.requestBuilder().delete(fetched).execute(destination);
+  } catch (e) {
+    // Entity was not present nothing to delete
```
jjtang1985, line 25: maybe check the error message and only ignore when it's about "nothing to delete"

jjtang1985 (changes requested):
Looks good.
Please consider the following:
- To make a cleaner PR, I prefer moving the vdm changes to a separate PR.
- The 2 edm types you removed are valid data type, see comments for details.
- For etag, please make a follow up ticket or create an issue in the the cap repo.

packages/core/test/test-util/test-services/v4/test-service/TestComplexType.ts
FrankEssenberger: I started on a separate PR.

test-packages/e2e-tests/db/schema.cds
FrankEssenberger: I will do a separate PR to remove the Edm.Time from v4.

jjtang1985 (approved):

PR #427, opened 2020-08-28
chore: ensure test services are up-to-data

jjtang1985:
## Context

The pipeline now generates the test services (v2 and v4 but without the ones used in e2e tests) to check whether they are up-to-date.

## Definition of Done

- [ ] Release notes updated.


package.json
```diff
@@ -48,6 +48,7 @@
     "check:dependencies": "yarn workspaces run check:dependencies",
     "check:license": "node scripts/check-licenses.js",
     "check:vulnerability": "yarn audit-ci -h",
+    "check:test-service": "yarn generate:test-services:v2 && yarn generate:test-services:v4 && yarn ts-node scripts/check-test-service.ts",
```
FrankEssenberger, line 51: Minor: Just a matter of taste. You could add the generation of the test services also to the typescript file.

scripts/check-test-service.ts
```diff
@@ -0,0 +1,9 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
```
FrankEssenberger, line 1: Minor: Perhaps you can adjust the name of the script file to something like: `check-test-service-for-changes.ts` then it is clear from the name what the script does.

FrankEssenberger (approved):
LGTM only a very little comment

marikaner (approved):
I like it! We might want to consider to actually update the test services on this branch based on this instead of having additional manual steps.

PR #429, opened 2020-08-28
Add Licence File

FrankEssenberger:
## Context

we were a bit unsure which licence to put for the documentation. The OSPO clarfied that for the documentation the same licence can be used as for the code which of OS is the apache 2.0 licence. 

In principle we have this licence already on master. But since the documentation branch is so different I added the licence once more also to this branch. This can not hurt and we are on the save side.


florian-richter (approved):
LGTM

PR #431, opened 2020-08-28
Fix: Serialize TimeOfDay

FrankEssenberger:
## Context

We had a bug in the serializer for single digit values in the `Time` object. The padding with zeros was missing.
So `{hours:1,minutes:2,seconds:3}` was serialized to `1:2:3` and not `01:02:03` as it should be.

## Definition of Done

- [x] Release notes updated.


packages/core/src/odata/v4/payload-value-converter.ts
```diff
@@ -107,9 +107,22 @@ function durationToEdmDuration(value: moment.Duration): string {
...
+function padSeconds(seconds: number): string {
+  const secondsString = seconds.toString(10);
+  if (!secondsString.includes('.')) {
+    return secondsString.padStart(2, '0');
+  }
+  const secondsStringParts: string[] = secondsString.split('.');
+  return [secondsStringParts[0].padStart(2, '0'), secondsStringParts[1]].join(
+    '.'
+  );
+}
```
marikaner, line 125: Actually you could use this function for all parts of the timeofday but even if you don't like that take a look below:
```suggestion
function padTimeComponent(timeComponent: number): string {
  const [wholeNumber, fractionalNumber] = timeComponent.toString().split('.');
  return [wholeNumber.padStart(2, '0'), fractionalNumber].join('.');
}
```

packages/core/src/odata/v4/payload-value-converter.ts
```diff
@@ -107,9 +107,22 @@ function durationToEdmDuration(value: moment.Duration): string {
 }
 
 function timeToEdmTimeOfDay(value: Time): string {
-  return `${value.hours}:${value.minutes}:${value.seconds}`;
+  const hoursPadded = value.hours.toString(10).padStart(2, '0');
+  const minutesPadded = value.minutes.toString(10).padStart(2, '0');
+  const secondsPadded = padSeconds(value.seconds);
+  return `${hoursPadded}:${minutesPadded}:${secondsPadded}`;
```
marikaner, line 113: I think you can repeat yourself a little less, also see suggestion below.
```suggestion
  return [value.hours, value.minutes, value.seconds].map(timeComponent => padTimeComponent(timeComponent)).join(':');
```

marikaner (changes requested):
Minor changes, let me know what you think!

FrankEssenberger:
liked your suggestions

marikaner (approved):

FrankEssenberger:
Very good remarks marika.

PR #437, opened 2020-08-31
fix: generate filter based on the user input instead of flattened version

jjtang1985:
## Context
A user reported that one of the s/4 service cannot handle 6 filter expression with `and` operator in one level like `A and B ... and F`. However, if you split the expression into 2 pieces like `(A and B) and (C and D and E and F)`, it works. Therefore, it makes sense to build such filter from the SDK. Unfortunately, SDK removed all the brackets as optimisation, making it not possible. This PR aims to disable the optimisation, so users have the freedom to build their own structure for filters.

Here is the original stackoverflow question: https://stackoverflow.com/questions/63615746/sap-cloud-sdk-for-javascript-multiple-and-filters-are-directly-appended-leads-t

## Definition of Done

- [ ] Release notes updated.


FrankEssenberger:
Please add a PR description and perhaps the link to the issue raised by the user.

CHANGELOG.md
```diff
@@ -43,6 +43,8 @@ Blog: TBD<br>
 
 - [Generator] Skip generation of Batch.ts for services without entities.
 - [RequestBuilder] Fix serialization of "Edm.Time" fields in OData V4. 
+- Fix the generation of the `$filter` parameter in URL so that parentheses are not removed even logically they are not needed.
```
FrankEssenberger, line 46: I would call it Change not Fix since we did nothing wrong before.

FrankEssenberger (commented):
LGTM. Just the minor things like PR description and Changelog.md suggestion.

FrankEssenberger (approved):

PR #438, opened 2020-08-31
Script to get reponse time information

FrankEssenberger:
## Context

A little script to get the response time for issues on cloud-sdk and cloud-sdk-cli. Question is where to put it? Should we put it in a closed repo or here? What do you think?


jjtang1985:
I guess these script can be stored in the same repo as the one used for the feedback monitor?

FrankEssenberger:
> I guess these script can be stored in the same repo as the one used for the feedback monitor?

This is a good idea. I close this PR and move it to the other repo.

PR #442, opened 2020-09-01
docs: Odata v4 overview

marikaner:
## Context

Most of this was already done. I extracted the general parts and referenced it in both versions. After a short discussion with Junjie - we should probably rethink the structure from time soon.

I initially wanted to embed this, which worked like a charm, but it does not update the TOC on the right -.- so this isn't really an option.


docs/js/features/odata/execute-odata-request.mdx
```diff
@@ -0,0 +1,129 @@
+---
+title: Executing a request using a generated OData client
+sidebar_label: Executing a request
+description: Use the SAP Cloud SDK for JavaScript to build and run OData requests in a typesafe way.
```
jjtang1985, line 4:
```suggestion
description: Use the SAP Cloud SDK for JavaScript to build and run OData requests in a type-safe way.
```

jjtang1985 (dismissed):
Looks good. I left some minor comments.

marikaner:
Thank you for the suggestions! 😊

jjtang1985 (approved):

PR #444, opened 2020-09-02
fix: generated index file should not export batch when no batch can be found

jjtang1985:
## Context

We stopped generating the `Batch.ts` when there is no entity. In this case, the index file should not export the batch.
I manually tested the generated folder for compilation checks, which should be done automatically soon.

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/generator/src/generator-utils.ts
+++ b/packages/generator/src/generator-utils.ts
@@ -3,7 +3,11 @@
 import { EdmTypeShared } from '@sap-cloud-sdk/core';
 import { createLogger, ODataVersion } from '@sap-cloud-sdk/util';
 import { pipe } from 'rambda';
-import { VdmNavigationProperty, VdmProperty } from './vdm-types';
+import {
+  VdmNavigationProperty,
+  VdmProperty,
+  VdmServiceMetadata
+} from './vdm-types';
 
 const logger = createLogger({
   package: 'generator',
@@ -308,3 +312,7 @@ const makeNpmCompliant = pipe(
 
 const npmMaxLength = 214;
 const npmRegex = /^(?:@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/;
+
+export function hasBatchRequest(service: VdmServiceMetadata): boolean {
+  return service.entities && service.entities.length > 0;
+}
--- a/packages/generator/src/generator-utils.ts
+++ b/packages/generator/src/generator-utils.ts
@@ -313,6 +313,6 @@ const makeNpmCompliant = pipe(
 const npmMaxLength = 214;
 const npmRegex = /^(?:@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/;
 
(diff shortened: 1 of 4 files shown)


packages/generator/src/service/index-file.ts
```diff
@@ -6,23 +6,26 @@ import {
...
-      ...service.complexTypes.map(complexType =>
-        exportStatement(complexType.typeName)
-      ),
-      ...(service.functionImports && service.functionImports.length
-        ? [exportStatement('function-imports')]
-        : []),
-      exportStatement('BatchRequest')
-    ]
+    statements: hasBatchRequest(service)
+      ? basicStatements.concat(exportStatement('BatchRequest'))
```
FrankEssenberger, line 27: Minor: I prefer the spread  [...basicStatements,exportStatement('BatchRequest')]  instead of concat.

FrankEssenberger (approved):

packages/generator/src/generator-utils.ts
```diff
@@ -308,3 +312,7 @@ const makeNpmCompliant = pipe(
 
 const npmMaxLength = 214;
 const npmRegex = /^(?:@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/;
+
+export function hasBatchRequest(service: VdmServiceMetadata): boolean {
+  return service.entities && service.entities.length > 0;
```
marikaner, line 317:
```suggestion
  return !!service.entities?.length;
```

packages/generator/src/generator-utils.ts
```diff
@@ -308,3 +312,7 @@ const makeNpmCompliant = pipe(
 
 const npmMaxLength = 214;
 const npmRegex = /^(?:@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/;
+
+export function hasBatchRequest(service: VdmServiceMetadata): boolean {
```
marikaner, line 316: The function name is a little misleading here. How about `hasEntities`?

marikaner (changes requested):
Minor stuff

packages/generator/src/service/index-file.ts
```diff
@@ -6,23 +6,26 @@ import {
...
+    ...service.entities.map(entity => exportStatement(entity.className)),
+    ...service.entities.map(entity =>
+      exportStatement(`${entity.className}RequestBuilder`)
+    ),
+    ...service.complexTypes.map(complexType =>
+      exportStatement(complexType.typeName)
+    ),
+    ...(service.functionImports && service.functionImports.length
+      ? [exportStatement('function-imports')]
+      : [])
```
marikaner, line 22: You could consider adding the batch export similar to this one.
```
...(hasBatchRequest(service)
      ? [exportStatement('BatchRequest')]
      : [])
```

packages/generator/src/service/index-file.ts
marikaner: +1

marikaner (approved):
LGTM, a test would be nice...

jjtang1985:
Discussed

PR #448, opened 2020-09-03
Chore: Added E2E tests for deep create

FrankEssenberger:
## Context

I added some test for create and deep create. It is difficult since many servie do not support the full OData scope or only version 4.0 and not 4.01. So here is what I have done:

- For the tripping service I change to the `TripPinServiceRW` service because I realized in Postman that only for this one I can persist something
- I also removed the token in the trippin service URL to the test and out of the generated entities, because when you persist data and I had some corrupted state I wanted to switch to a fresh new token i.e. clean database.
- Only creation of entity without navigation (no deep create) was possible for the trippin service. Error on attempt was `System.InvalidOperationException", "stacktrace": "   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)\r\n ....`
- I extended the CDS model to contain a 1:N relation and was able to make a deep create. Update/Create was not possible (only v4.01)
- I reduced the key of the CDS TestEntity to just be one column. I did this to make the join between the two related entites easier and shorter. I think we have a lot of tests with the two property keys in the TestEntites.
- I kept all the non working tests, because they could be valuable in the later development

## Definition of Done

~~- [ ] Release notes updated.~~


test-packages/e2e-tests/test/request-builder.spec.ts
```diff
@@ -98,6 +103,76 @@ describe('Request builder test', () => {
     );
   });
 
+  it('should create an entity with related entities (deep create)', async () => {
```
jjtang1985, line 106: So, this is THE activated test for deep create.

jjtang1985 (changes requested):
I ignored most of the  generated files and only checked the `People.ts`.
The tests look good.
The only open topic is about where should we put the token.

test-packages/e2e-tests/test/request-builder.spec.ts
FrankEssenberger: yes

test-packages/e2e-tests/test/TripPin/request-builder.spec.ts
```diff
@@ -2,12 +2,32 @@
 
 import { any } from '@sap-cloud-sdk/core';
 import { resetDataSource } from '@sap-cloud-sdk/test-services-e2e/TripPin/microsoft-o-data-service-sample-trippin-in-memory-models-service/action-imports';
+import { PersonGender } from '@sap-cloud-sdk/test-services-e2e/TripPin/microsoft-o-data-service-sample-trippin-in-memory-models-service/PersonGender';
 import { People } from '../../../test-services-e2e/TripPin/microsoft-o-data-service-sample-trippin-in-memory-models-service';
 
-const url = 'https://services.odata.org/';
+const url = 'https://services.odata.org/V4/(S(yym4o5sbztcrqnvx5vjqr1z2))';
```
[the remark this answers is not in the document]
FrankEssenberger: Ok, I keep it in the mapping, but if we realize that we have to update the token from time to time for some reason I would prefer having only a change in a single file and not so many files touched.

test-packages/e2e-tests/test/request-builder.spec.ts
FrankEssenberger: more was not supported by the services, but at least something.

test-packages/e2e-tests/test/TripPin/request-builder.spec.ts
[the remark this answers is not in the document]
jjtang1985: I agree.

jjtang1985 (approved):

PR #453, opened 2020-09-04
feat: generator will fail when compilation errors are found in ts files

jjtang1985:
## Context

- use `tsc` to transpile .ts files instead of using `ts-morph` so we can show compilation errors
- remove the `outDir` config in the `tsconfig.json` so that when transpiring .ts files, all .js files are in the same folder instead of the `dist` folder.
- use typescript version `4.0.2` to align with the ts-morph version that we are using. Because if we keep using `3.9.6`, all the .js files will be modified to the old states. In fact, when we updated the `ts-morph` (using typescript `4.0.2`) in this [PR](https://github.com/SAP/cloud-sdk/pull/417/files), we upgraded all the .js files. I think we should use `4.0.2` to keep the .js file constant instead of moving back. Let me know if you have concerns.
- negative tests: if you change the `service-mapping.json` and append `\n` to the end of `servicePath`, the generator will fail and show you the compilation error.

## Definition of Done

- [ ] Release notes updated.


FrankEssenberger (changes requested):
From a function point of view it looks good to me. I have a few questions:
- @marikaner you mentioned libs not compatible with TS 4.X.Y. So is the update a problem?
- In the  tsConfig() we include the `./dist` as output dir for generated VDMs. Should we perhaps also change this? Also what was the advantage in removing the dist? Just better overview of the files?

jjtang1985:
Discussed with Marika, we'll update the typescript version to 4.0.2 via this PR first: https://github.com/SAP/cloud-sdk/pull/406

jjtang1985:
I removed the config `./dist` as output dir in this pr, because for all vdm packages (as well as the test entities), we are using `ts-morph` for generating `.js` files. All the generated `.js` files are in the same folder as the `.ts` file. 

To keep the consistency, I have to touch the config, so that `tsc` will not put the `.js` files in the `./dist` folder.

jjtang1985:
@FrankEssenberger , as mentioned in the standup, this is the pr that we should merge before the `e2e test generator`, so that you don't need the `compile` command.

FrankEssenberger (approved):
My questions regarding `dist` folder and ts have been answered

PR #459, opened 2020-09-07
fix: use full url for linking to api doc

jjtang1985:
## Context

I made a mistake to commit to the `document` branch directly, which tried to use the `useBaseUrl `.
Switched to the full url.

Here is an related [issue](https://github.com/facebook/docusaurus/issues/3303), saying there can be false positive and how to disable it via config.
However, the config is about to change `navbar/footer` items, which is not applicable in our case.

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/docs/js/features/odata/odata-v4-client.mdx
+++ b/docs/js/features/odata/odata-v4-client.mdx
@@ -24,7 +24,7 @@ For more details on **how to execute requests** using a (pre-)generated OData cl
 
 #### Filter on One-To-Many Navigation Properties
 
-OData V4 introduces [lambda operators](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_LambdaOperators) e.g., `any`/`all`, so that the root property of the one-to-many navigation properties can be filtered. Below is an example that demonstrates how to use the lambda operator [any](pathname:///api/1.28.1/modules/sap_cloud_sdk_core#any).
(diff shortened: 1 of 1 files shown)


FrankEssenberger (approved):
Fine for me, but what will happen locally? The link will point to the internet and not the local static copy of the API right?

docs/js/features/odata/odata-v4-client.mdx
```diff
@@ -26,7 +24,7 @@ For more details on **how to execute requests** using a (pre-)generated OData cl
 
 #### Filter on One-To-Many Navigation Properties
 
-OData V4 introduces the [lambda operators](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_LambdaOperators) e.g., `any`/`all`, so that the root property of the one-to-many navigation properties can be filtered. Below is an example that demonstrates how to use the lambda operator [any]({useBaseUrl(/api/1.28.1/modules/sap_cloud_sdk_core#any)}).
+OData V4 introduces the [lambda operators](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_LambdaOperators) e.g., `any`/`all`, so that the root property of the one-to-many navigation properties can be filtered. Below is an example that demonstrates how to use the lambda operator [any](https://sap.github.io/cloud-sdk/api/1.28.1/modules/sap_cloud_sdk_core#any).
```
marikaner, line 27:
```suggestion
OData V4 introduces [lambda operators](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_LambdaOperators) e.g., `any`/`all`, so that the root property of the one-to-many navigation properties can be filtered. Below is an example that demonstrates how to use the lambda operator [any](pathname:///api/1.28.1/modules/sap_cloud_sdk_core#any).
```

marikaner (approved):

PR #480, opened 2020-09-14
[Fix] Renaming to .mdx broke the Documentation Build

FrankEssenberger:
## Context

I renamed the .md files consistently to .mdx but this broke the build. This PR tries to fix it.

### DIFF
--- a/docs/dzen/getting-started.mdx
+++ b/docs/dzen/getting-started.mdx
@@ -100,7 +100,7 @@ git checkout -b java-client-generator-how-to
 
 1. All the documentation resides in the `docs` folder.
 2. Let's start with creating a file containing our _How To_. In my case it will be `java/how-to/generate-client.md`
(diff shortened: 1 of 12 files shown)


FrankEssenberger:
Checked it locally. `npm run build` worked after deleting the `docusaurus` folder. Please go ahead and merge so that others are not blocked by my mistake

artemkovalyov (approved):
As I know by default Docusaurus considers MD and MDX as well. At least react components work fine. That's why I was keeping it simple before we need any advanced functionality unlocked by explicit MDX resolution. I'll check the docs regarding this again.

FrankEssenberger:
Per default they have a `.mdx` parser. At least this is how I understood it and also some of the things we use already are not plain .md.

artemkovalyov:
@FrankEssenberger , you're right, and probably this default parsing does the tick. MDX should also work. Do you have anything not yet covered currently without having MDX resolution?

FrankEssenberger:
I hope not. I wrote a little script moving all `.md` file in the doc folder.

PR #487, opened 2020-09-15
[Chore] Add E2E test for asChildOf

FrankEssenberger:
## Context

I picked the documentation issue for the `asChildOf` functionality. When I was going over the code samples I realized, that we do not have an E2E test for this functionality. So I added a test against the CAP service. Works just fine.


marikaner (approved):
lgtm

PR #496, opened 2020-09-18
chore: Deprecate MapType

marikaner:
## Context

TypeScript provides the type `Record` for this.

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/analytics/src/analytics-data.ts
+++ b/packages/analytics/src/analytics-data.ts
@@ -3,7 +3,7 @@
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
-import { findProjectRoot, MapType } from '@sap-cloud-sdk/util';
+import { findProjectRoot } from '@sap-cloud-sdk/util';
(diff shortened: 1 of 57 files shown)


FrankEssenberger (commented):
Question: Are MapType<string> and Record<string,string> the same type in the typescript structural typing sense? Would:
```
const myMap:MapType = ...
getAutHeaders(destination,myMap) 
```
compile? If not this would be a breaking change I guess.

marikaner:
> Question: Are MapType and Record<string,string> the same type in the typescript structural typing sense? Would:
> 
> ```
> const myMap:MapType = ...
> getAutHeaders(destination,myMap) 
> ```
> 
> compile? If not this would be a breaking change I guess.

This is a very good question. I think yes, but now that you are asking, I'm not sure. I will check it.

FrankEssenberger (approved):

jjtang1985 (approved):

PR #498, opened 2020-09-18
feat: Allow to mute loggers

marikaner:
## Context

I am quite annoyed by the many log outputs we get in the tests. This is the first step towards making them disappear.

## Definition of Done

- [x] Release notes updated.

### DIFF
--- a/packages/util/src/logger/cloud-sdk-logger.ts
+++ b/packages/util/src/logger/cloud-sdk-logger.ts
@@ -16,11 +16,29 @@ const container = new Container();
 const exceptionTransport = new transports.Console();
 const customLogLevels = {};
 const DEFAULT_LOGGER__MESSAGE_CONTEXT = '__DEFAULT_LOGGER__MESSAGE_CONTEXT';
+let silent = false;
 
 const moduleLogger = createLogger({
   package: 'util',
   messageContext: 'cloud-sdk-logger'
 });
+
+function toggleMuteLoggers(silence: boolean) {
+  silent = silence;
+  container.loggers.forEach(logger => toggleSilenceTransports(logger, silence));
+}
+
+function toggleSilenceTransports(logger: Logger, silence: boolean) {
+  logger.transports.forEach(transport => (transport.silent = silence));
+}
+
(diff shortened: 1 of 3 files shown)


packages/util/src/logger/cloud-sdk-logger.ts
```diff
@@ -16,12 +16,36 @@ const container = new Container();
...
 const customLogLevels = {};
 const DEFAULT_LOGGER__MESSAGE_CONTEXT = '__DEFAULT_LOGGER__MESSAGE_CONTEXT';
+let silent = false;
 
 const moduleLogger = createLogger({
   package: 'util',
   messageContext: 'cloud-sdk-logger'
 });
 
+function toggleMuteLoggers(silence: boolean) {
```
FrankEssenberger, line 26: Minor: Method is not exported so not so important, but I am a bit confused `silent` and `silence` both a boolean muting so why not use the same name.

marikaner: Because silent is a local variable 😕 I just didn't have a better idea.

FrankEssenberger: ah shadow, consider using a namespace for the private scope above.

FrankEssenberger (approved):
LGTM

marikaner:
I checked how I would do it with a namespace and it is not as straight forward as I would think. I guess variables cannot be changed from the outside (except using a method). I will refrain from changing to namespaces right now, but if it is important to you, lets talk about another PR.

```
namespace CloudSDKLogger {
  export const format = process.env.VCAP_SERVICES ? kibana : local;
  export const loggerReference = 'sap-cloud-sdk-logger';
  export const exceptionLoggerId = 'sap-cloud-sdk-exception-logger';

  export const container = new Container();
  export const exceptionTransport = new transports.Console();
  export const customLogLevels = {};
  export const DEFAULT_LOGGER__MESSAGE_CONTEXT =
    '__DEFAULT_LOGGER__MESSAGE_CONTEXT';
  export let silent = false;
}
```

PR #499, opened 2020-09-18
chore: Unify test config and mute loggers

marikaner:
## Context

This will only work once #498 is merged. We had the same config duplicated for all tests. I put this into a central place and also muted the logging output so it isn't cluttering the output. Hopefully now we can see more when there are errors in the pipeline.


FrankEssenberger (approved):
LGTM

PR #500, opened 2020-09-18
[Chore] Remove BL security warning

FrankEssenberger:
## Context

we have a security warning due to `bl@1.2.2` which is fixed in version `1.2.3` of `bl`. This comes in via the `dtslint` type test package, so it does not affect any productive situation. But it good to not have this yellow banner every time I open the SDK.

Yarn is very convinient:
- `yarn why bl` tells you why the dependency is installed
- with an entry in the resolution entry `libHavingItAsDependency/libToReplaceVersion: newVersion`  you can overrule the version used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
   "dependencies": {},
   "resolutions": {
     "@types/node": "^12.0.0",
-    "dot-prop": "^5.1.1"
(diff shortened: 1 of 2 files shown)


marikaner (approved):
finally ;)

Issue #508, opened 2020-09-23
The manual OData call to C4C works fine, while same call via SDK doesn't


artemkovalyov:
**Describe the bug**

Querying IndividualCustomers from C4C
To do this we tried out two variations:
    1. Querying the ODATA API directly using a lambda function in Kyma (see file get-indiv-customers.js). Using the $expand syntax it is possible to list all the current customer’s addresses, i.e. this works as expected (see file response-indiv-customers-native.json)
    2. Using the CloudSDK to query C4C. In this case the syntax is slightly different (see file get-indiv-customers-sdk.js). In this case we always get back an empty array of addresses, even if the IndividualCustomer has addresses in C4C (see file response-indiv-customers-sdk.json). We observe this behavior for all of the IndividualCustomer child entities we have tested.

Thanks for your help!

**To Reproduce**
Please note that to use the Cloud SDK Nazar has set up a node.js module which is installed as a dependency in the lambda function. The library he set up is available at https://github.com/PTA-Kyma/pta-kyma-c4-odata.git (to install: @nazarkulyk/pta-kyma-c4-odata)

**Expected behavior**
The call returns data with Expand

**Used Versions:**
We'll 

**Code Examples**
If applicable, add code snippets as examples to help explain your problem. Please remove sensitive information as it will be shown in the internet.

FrankEssenberger:
I can not access the repo.

FrankEssenberger:
or share the ` get-indiv-customers.js` and `get-indiv-customers-sdk.js`. In general the expand should work.

FrankEssenberger:
I just made checked and we have a [test](https://github.com/SAP/cloud-sdk/blob/45af0f068c77976daae1eb7ab4297dfe47a3d063/test-packages/e2e-tests/test/TripPin/request-builder.spec.ts#L31-L40) for that against the  `https://services.odata.org/TripPinRESTierService` 
service. It is working as expected.

artemkovalyov:
Yeah, I was also unable to access the repo @FrankEssenberger. I'm working on it and on repro scenario. Thanks for looking into this.

FrankEssenberger:
[...]

PR #510, opened 2020-09-23
[Fix] Response parsing of 1:N relations in C4C

FrankEssenberger:
## Context

In general for OData v2 an expanded nav property as the form `{results:[]} i.e. the navigation entities are wrapped in a results object. For C4C this seems to be not the case see here #508 
This PR makes the data extraction more flexible.

## Definition of Done

- [x] Release notes updated.

### DIFF
--- a/packages/core/src/odata/v2/extract-data-from-one-to-many-link.ts
+++ b/packages/core/src/odata/v2/extract-data-from-one-to-many-link.ts
@@ -2,9 +2,18 @@
 
 /**
  * Data extractor for one to many links for v2 entity used in [[entityDeserializer]]
+ * It first tries if data.result is an array and returns it.
+ * Then it tires if data itself is an array and return it.
+ * If both are not the case it returns undefined
  * @param data - One to many link response data
  * @returns The content of the one to many link
  */
 export function extractDataFromOneToManyLink(data): any[] | undefined {
-  return data.results;
+  if (data.results && Array.isArray(data.results)) {
+    return data.results;
+  }
+  if (Array.isArray(data)) {
+    return data;
+  }
+  return undefined;
 }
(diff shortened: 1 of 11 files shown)


packages/core/src/odata/v2/extract-data-from-one-to-many-link.ts
```diff
@@ -2,9 +2,18 @@
...
  * Data extractor for one to many links for v2 entity used in [[entityDeserializer]]
+ * It first tries if data.result is an array and returns it.
+ * Then it tires if data itself is an array and return it.
+ * If both are not the case it returns undefined
  * @param data - One to many link response data
  * @returns The content of the one to many link
  */
 export function extractDataFromOneToManyLink(data): any[] | undefined {
-  return data.results;
+  if (data.results && Array.isArray(data.results)) {
```
marikaner, line 12: I don't understand why we have this function in the first place...
Maybe you can consider deprecating it, but I guess this is nothing for the quick fix.

Anyways, even for a quick fix, please use the existing response data accessor: https://github.com/SAP/cloud-sdk/blob/master/packages/core/src/odata/v2/request-builder/response-data-accessor.ts

I must have missed this function when applying it everywhere.

marikaner (changes requested):

packages/core/src/odata/v2/extract-data-from-one-to-many-link.ts
FrankEssenberger: The data accessor is a v2 thing. We thought there is a clear cut between v2 and v4. But it seems like it is not so strict so I mad now a single method handling both cases. Let me know what you think.

artemkovalyov:
Looks like a reasonably quick fix for quirks of service backend implementation details. Design-wise I won't currently have a say, but as that's not a public API we can always refactor as OData v4 is still in progress.

marikaner (approved):

PR #514, opened 2020-09-24
Response accessor

marikaner:
## Context

This is what I meant with "use the response accessor". I think we should definitely pass the response accessor to the deserializer not some specific functions, but this should happen on a separate ticket. This would only be the first step to have the "data accessing" functionality in one place. I created a separate PR for your branch, @FrankEssenberger, to make it easier for you to see the changes (that I reverted to before my comments and then used the accessor).

### DIFF
--- a/packages/core/src/odata/v4/request-builder/response-data-accessor.ts
+++ b/packages/core/src/odata/v4/request-builder/response-data-accessor.ts
@@ -45,7 +45,7 @@ function validateCollectionResult(data): void {
  * @param data - Response of the one to many link
  * @returns any[] - Collection extracted from the response
  */
-export function getLinkedCollectionResult(data): any[] | undefined {
+export function getLinkedCollectionResult(data): any[] {
   return Array.isArray(data) ? data : [];
 }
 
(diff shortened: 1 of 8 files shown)


packages/core/src/odata/v4/request-builder/response-data-accessor.ts
```diff
@@ -39,6 +39,16 @@ function validateCollectionResult(data): void {
   }
 }
 
+/**
+ * Extract the collection data from the one to many link response.
+ * If the data does not contain a collection an empty array is returned.
+ * @param data - Response of the one to many link
+ * @returns any[] - Collection extracted from the response
+ */
+export function getLinkedCollectionResult(data): any[] | undefined {
```
FrankEssenberger, line 48:
```suggestion
export function getLinkedCollectionResult(data): any[] {
```

FrankEssenberger (approved):
Ah, ok now I got it. You wanted to extend the data-accessor (now also for v4) and pass this to the deserializer. Makes sense.

PR #517, opened 2020-09-25
chore: Add last function to util

marikaner:
## Context

We are slowly getting rid of rambda. `last` can be quite useful, so I added it to our util package.

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/core/src/odata/v2/odata-batch-request-builder.ts
+++ b/packages/core/src/odata/v2/odata-batch-request-builder.ts
@@ -1,6 +1,6 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 
-import { errorWithCause, last } from '@sap-cloud-sdk/util';
+import { errorWithCause, first, last } from '@sap-cloud-sdk/util';
 import { v4 as uuid } from 'uuid';
 import {
   Destination,
@@ -327,7 +327,7 @@ function toConstructableFromRetrieveResponse(
   entityToConstructorMap: Record<string, Constructable<EntityV2>>
 ): Constructable<EntityV2> {
   const entityJson = isCollectionResult(responseBody)
-    ? getCollectionResult(responseBody)[0]
+    ? first(getCollectionResult(responseBody))
     : getSingleResult(responseBody);
 
   return entityToConstructorMap[
--- a/packages/core/src/odata/v2/odata-batch-request-builder.ts
+++ b/packages/core/src/odata/v2/odata-batch-request-builder.ts
@@ -1,7 +1,6 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 
-import { errorWithCause } from '@sap-cloud-sdk/util';
-import { head, last } from 'rambda';
+import { errorWithCause, last } from '@sap-cloud-sdk/util';
 import { v4 as uuid } from 'uuid';
 import {
   Destination,
@@ -252,7 +251,7 @@ function partitionBatchResponse(response: string, lineBreak: string): string[] {
   }
 
   // E.g., --batch_1234
-  const partSeparator = head(response.split(lineBreak, 1));
+  const partSeparator = response.split(lineBreak, 1)[0];
   if (!partSeparator) {
     throw Error(`Failed to get part separator of the response: ${response}`);
   }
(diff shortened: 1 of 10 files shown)


packages/core/src/odata/v2/odata-batch-request-builder.ts
```diff
@@ -252,7 +251,7 @@ function partitionBatchResponse(response: string, lineBreak: string): string[] {
   }
 
   // E.g., --batch_1234
-  const partSeparator = head(response.split(lineBreak, 1));
+  const partSeparator = response.split(lineBreak, 1)[0];
```
FrankEssenberger, line 254: I am confused but the limit and [0]. This means the results array contains only on element and this is the one we take, right? I wonder when can the if (!partSeparator) case happen. If the response is defined the split will always return an element.

packages/generator/src/edmx-parser/legacy/common/edmx-to-vdm.ts
```diff
@@ -1,8 +1,7 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
 
-import { last } from 'rambda';
 import { toTitleFormat, toTypeNameFormat } from '@sap-cloud-sdk/core';
-import { createLogger } from '@sap-cloud-sdk/util';
+import { createLogger, last } from '@sap-cloud-sdk/util';
```
FrankEssenberger, line 4: was this method unused?

packages/generator/src/edmx-to-vdm/common/entity.ts
```diff
@@ -1,6 +1,5 @@
 /* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
-import { createLogger } from '@sap-cloud-sdk/util';
-import { last } from 'rambda';
+import { createLogger, last } from '@sap-cloud-sdk/util';
```
FrankEssenberger, line 2: also here.

packages/core/src/scp-cf/environment-accessor.ts
```diff
@@ -41,7 +40,7 @@ export function getDestinationBasicCredentials(): BasicCredentials {
  * @returns The 'destination' credentials object or null if it does not exist.
  */
 export function getDestinationServiceCredentials(): any {
-  return head(getDestinationServiceCredentialsList());
+  return getDestinationServiceCredentialsList()[0];
```
FrankEssenberger, line 43: I would prefer to have a `first` or `head` funciton in the util to get the element. I know it is trivial, but if we use it for last I want to symmetric on first.

FrankEssenberger (changes requested):
QUestion

packages/core/src/odata/v2/odata-batch-request-builder.ts
marikaner: Yeah don't look at this. I removed that in my other PR ;) Here I really just replaced last and head.

packages/core/src/scp-cf/environment-accessor.ts
marikaner: Hmm ok ;)

packages/generator/src/edmx-parser/legacy/common/edmx-to-vdm.ts
marikaner: No I added it ;)

packages/generator/src/edmx-to-vdm/common/entity.ts
marikaner: Same here ;)

packages/generator/src/edmx-parser/legacy/common/edmx-to-vdm.ts
FrankEssenberger: But I do not see a usage in this file.

packages/generator/src/edmx-to-vdm/common/entity.ts
FrankEssenberger: But I do not see a usage in this file.

packages/core/src/odata/v2/odata-batch-request-builder.ts
```diff
@@ -268,7 +267,7 @@ function partitionChangeSetResponse(
   responseOfSingleRetrieveRequest: string,
   lineBreak: string
 ): string[] {
-  const firstLine = head(responseOfSingleRetrieveRequest.split(lineBreak, 1));
+  const firstLine = responseOfSingleRetrieveRequest.split(lineBreak, 1)[0];
```
FrankEssenberger, line 270: use also first here.

FrankEssenberger (approved):
Perhaps replace the other [0] also with first. But you mentioned that this is changed in your PR anyhow.

PR #519, opened 2020-09-25
chore: Expose entity de-/serializer

marikaner:
## Context

Expose the entity de-/serializer modules to be used in a type safe manner.

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/core/src/odata/v2/entity-deserializer.ts
+++ b/packages/core/src/odata/v2/entity-deserializer.ts
@@ -12,14 +12,15 @@ import { getLinkedCollectionResult } from './request-builder/response-data-acces
  * Entity deserializer instance for v2 entities.
  * See [[EntityDeserializerType]] for the provided methods.
  */
-const deserializer: EntityDeserializer = entityDeserializer(
+export const entityDeserializerV2: EntityDeserializer = entityDeserializer(
   edmToTsV2,
   extractODataEtagV2,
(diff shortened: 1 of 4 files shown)


FrankEssenberger (changes requested):
In the main PR it looked like you only use one method the `deserializeEntity` is so consider ot exposing the whole serializer.

marikaner:
> In the main PR it looked like you only use one method the `deserializeEntity` is so consider ot exposing the whole serializer.

I initially passed around only the deserializeEntity method. But there are two reasons I changed my mind while working on it:
1. It is not easy to type that function, meaning that I would probably then just end up using `any` wherever this is used. We could also add types for the specific deserializer functions but I question that this is good style.
2. At the moment we are only using the entity deserialization, but is possible that we need this for complex types at some point (improbable).
3. I wanted it to be consistent with the response accessor that I also pass to some functions. Here I use two functions, but not all of them. It would probably be weird to pass those two functions instead of the module itself.

I don't have a very strong opinion here, but I am slightly leaning towards passing the module. If you don't agree I will be happy to change it.

This discussion however is more about the specific usage of what this PR exposes, not on the content of the PR iteself.

FrankEssenberger (approved):
Argument of symmetry to accessor with multiple methods and typing and complex types future prooves counts for me stronger than the minimal interface rule.

PR #520, opened 2020-09-25
chore: Improve auth header warning

marikaner:
## Context

The warning message had some weird formatting. Also, I got this message, when I passed a destination directly. Those don't have a name so I got a message telling me that the "destination undefined" had auth headers.

In general I think we need to rethink how we go about headers. At the moment this warning also occurs when there are no custom headers set. I guess this is because the generic http client uses them. As this is only a warning, maybe this is low priority.
 
## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/core/src/odata/common/response-data-accessor.ts
+++ b/packages/core/src/odata/common/response-data-accessor.ts
@@ -1,7 +1,0 @@
-/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
-
-export interface ResponseDataAccessor {
-  getCollectionResult: (data) => any[];
-  isCollectionResult: (data) => boolean;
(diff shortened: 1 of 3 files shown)


packages/core/src/odata/common/response-data-accessor.ts
```diff
@@ -0,0 +1,7 @@
+/* Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. */
+
+export interface ResponseDataAccessor {
```
FrankEssenberger, line 3: I guess this slipped in?

FrankEssenberger (approved):

packages/core/src/odata/common/response-data-accessor.ts
marikaner: yes -.-

PR #523, opened 2020-09-25
fix: Remove dangerous prepare method

marikaner:
## Context

The `prepare` function was introduced a long time ago. The name does not say what this function does, it has a side effect and in order to execute a request it had to be called before. I deprecated it and replaced its usage.

## Definition of Done

- [ ] Release notes updated.


FrankEssenberger (commented):
I put a question regarding this in the initial PR. So I am happy with the changes, but perhaps you can comment if this is a change in behaviour i.e. breaking change? In general I think it is much safer this way.

marikaner:
It is not a breaking change. I deprecated the publicly visible function, but replaced all its internal usage with an update of the payload itself.

FrankEssenberger (approved):

PR #525, opened 2020-09-25
chore: Remove all copyright messages

marikaner:
## Context

They are not necessary anymore because we have one central file for this now.

## Definition of Done

- [ ] Release notes updated.


.eslintrc.js
```diff
@@ -1,295 +1,253 @@
 module.exports = {
-    "env": {},
```
jjtang1985, line 2: I assume this change is only about removing the copyright check

jjtang1985 (approved):
LGTM. Please fix the lint error.

.eslintrc.js
marikaner: Yes, and on save it was formatted...

PR #531, opened 2020-09-28
[Fix] Transpile services in chunks and await result

FrankEssenberger:
## Context

I had problems with running the generate with transpile locally. The system got overloaded by around 200 processes in parallel. I made an adjustment to do this in chunks of 16 services and await for these to get finished.

I think this could be wrong in the E2E tests since we did not await the result properly.

### DIFF
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -6,7 +6,6 @@ import { createLogger } from '@sap-cloud-sdk/util';
 import { emptyDirSync } from 'fs-extra';
 import {
   Directory,
-  EmitResult,
   IndentationText,
   ModuleResolutionKind,
   Project,
@@ -55,31 +54,45 @@ const logger = createLogger({
   messageContext: 'generator'
 });
 
-export async function generate(
-  options: GeneratorOptions
-): Promise<void | EmitResult[]> {
+const threadsForCompiling = 16;
+
+export async function generate(options: GeneratorOptions): Promise<void> {
   const project = await generateProject(options);
   if (!project) {
     throw Error('The project is undefined.');
   }
 
   await project.save();
-
   if (options.generateJs) {
-    project
+    let directories = project
       .getDirectories()
-      .filter(dir => !!dir.getSourceFile('tsconfig.json'))
-      .forEach(dir => {
-        logger.info(`Transpiling files in the directory: ${dir.getPath()}.`);
-
-        execa('tsc', { cwd: dir.getPath() }).catch(err => {
-          logger.error(`Error: Failed to generate js files: ${err}`);
-          process.exit(1);
-        });
-      });
+      .filter(dir => !!dir.getSourceFile('tsconfig.json'));
+    while(directories.length > 0){
+      const chunk = directories.splice(0,threadsForCompiling)
+      await transpileDirectories(chunk)
+    }
   }
 
   return;
+}
+
+export async function transpileDirectories(directories:Directory[]):Promise<void[]>{
+  return Promise.all(directories.map(directory=>transpileDirectory(directory)))
+}
+
+export async function transpileDirectory(dir: Directory): Promise<void> {
+  try {
+    logger.info(
+      `Transpiling files in the directory: ${dir.getPath()} started.`
+    );
+    await execa('tsc', { cwd: dir.getPath() })
+      logger.info(
+        `Transpiling files in direcotory:  ${dir.getPath()} finished.`
+    );
+  } catch (err) {
+    logger.error(`Error: Failed to generate js files: ${err}`);
+    process.exit(1);
+  }
 }
 
 export async function generateProject(
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -2,7 +2,7 @@
 
 import { PathLike, readFileSync } from 'fs';
 import { resolve, basename } from 'path';
-import { createLogger } from '@sap-cloud-sdk/util';
+import { createLogger, splitInChunks } from '@sap-cloud-sdk/util';
 import { emptyDirSync } from 'fs-extra';
 import {
   Directory,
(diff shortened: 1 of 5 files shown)


packages/generator/src/generator.ts
```diff
@@ -55,33 +54,47 @@ const logger = createLogger({
...
-        logger.info(`Transpiling files in the directory: ${dir.getPath()}.`);
-
-        execa('tsc', { cwd: dir.getPath() }).catch(err => {
-          logger.error(`Error: Failed to generate js files: ${err}`);
-          process.exit(1);
-        });
-      });
+      .filter(dir => !!dir.getSourceFile('tsconfig.json'));
+    while(directories.length > 0){
+      const chunk = directories.splice(0,threadsForCompiling)
```
marikaner, line 71: I am not a big fan of changing the input here(`directories`) with splice. Especially with my comment above to put into a separate function it will have a side effect. Instead consider something like this:
```js
const chunk = (d) => {
  let chunks = [];
  for(let i = 0; i < d.length/threadsForCompiling; i++) {
    chunks = [...chunks, d.slice(i * threadsForCompiling, (i + 1) * threadsForCompiling)];
  }
  return chunks;
}
```

marikaner (changes requested):
Fix linting, that makes it easier to read ;)
Especially consider my comment about splice.

packages/generator/src/generator.ts
```diff
@@ -55,33 +54,47 @@ const logger = createLogger({
   messageContext: 'generator'
 });
 
-export async function generate(
-  options: GeneratorOptions
-): Promise<void | EmitResult[]> {
+const threadsForCompiling = 16;
```
jjtang1985, line 57: Moving it to the cli args?

packages/generator/src/generator.ts
```diff
@@ -55,33 +54,47 @@ const logger = createLogger({
...
+export async function transpileDirectories(directories:Directory[]):Promise<void[]>{
+  return Promise.all(directories.map(directory=>transpileDirectory(directory)))
+}
+
+export async function transpileDirectory(dir: Directory): Promise<void> {
+  try {
+    logger.info(
+      `Transpiling files in the directory: ${dir.getPath()} started.`
+    );
+    await execa('tsc', { cwd: dir.getPath() })
```
jjtang1985, line 88: Please correct me.
I guess you had the feeling of the difference between these two versions in terms of the performance (CPU usage).
For the code changes, I found this is the line (with `await` keyword) that changes your life and the rest of them is doing the same, including the `16` thing.

FrankEssenberger: sorry. I do not really get what you mean.

jjtang1985: Discussed, `await` will be removed.

packages/generator/src/generator.ts
FrankEssenberger: I had the same initial hesitation, but thought then: this is so easy and feels so 2012 Physics code doing things in parallel using MPI or openMP. But you are right and we should have pure functions. No side effects. So I make at least a shallow copy with slice.

packages/generator/src/generator.ts
FrankEssenberger: I also head the idea, but was to lazy because this would also mean an adjustment in the cli. But I will add it and make a follow up ticket.

packages/generator/src/generator.ts
marikaner: I also agree that it is a little easier to read. Maybe just do it in a function and then:
`function (input) { const directoriesYouWorkWith = [...input] }` and then keep it as is?

FrankEssenberger: I created a splitInChunks in the array util. Then this can be used somewhere else perhaps. I think it is nicer now.

jjtang1985 (approved):

packages/generator/src/generator.ts
```diff
@@ -55,31 +57,46 @@ const logger = createLogger({
...
-
-        execa('tsc', { cwd: dir.getPath() }).catch(err => {
-          logger.error(`Error: Failed to generate js files: ${err}`);
-          process.exit(1);
-        });
-      });
+      .filter(dir => !!dir.getSourceFile('tsconfig.json'));
+    for (const chunk of splitInChunks(
+      directories,
+      options.processesJsGeneration || defaultValueProcessesJsGeneration
```
marikaner, line 73: This should not be necessary. Either it was overwritten or it is using the default, right?
```suggestion
      options.processesJsGeneration
```

marikaner: That would also mean you can get rid of the exported `defaultValueProcessesJsGeneration` variable.

packages/generator/src/generator.ts
```diff
@@ -55,31 +57,46 @@ const logger = createLogger({
...
-      .forEach(dir => {
-        logger.info(`Transpiling files in the directory: ${dir.getPath()}.`);
-
-        execa('tsc', { cwd: dir.getPath() }).catch(err => {
-          logger.error(`Error: Failed to generate js files: ${err}`);
-          process.exit(1);
-        });
-      });
+      .filter(dir => !!dir.getSourceFile('tsconfig.json'));
+    for (const chunk of splitInChunks(
```
marikaner, line 71: To make this more javascripty you could consider doing that in a reduce instead of a for loop.

packages/generator/src/generator.ts
FrankEssenberger: If somebody uses the `generate` method directly without the CLI the default value is not set form the command line parser.

packages/generator/src/generator.ts
FrankEssenberger: changed it but I do not know if this is easier to grasp than the for loop.

packages/generator/src/generator.ts
marikaner: True that should not be the case but as long as it is, go for it ;)

marikaner (approved):

PR #537, opened 2020-09-30
[Chore] Reuse the reusable code parts

FrankEssenberger:
## Context

We had already example and documentation for the OData v2. Try to reuse this also for v4 where possible. 
I would like to have better blocks of documentation but the problem with the TOC and the headers blocks this.

So this is the best I could do.


docs/js/features/odata/common/query-parameter/index.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+export function Top(){
+  return <TopContent/>
+}
+
+export function Skip(){
+  return <SkipContent/>
+}
+
+export function Select(){
```
jjtang1985, line 16: Is this a v2 item? the import is missing

FrankEssenberger: oh yes. v2 has a select item and v4 a expand-select item which are rather different.

docs/js/features/odata/odata-v4-client.mdx
```diff
@@ -13,112 +13,92 @@ keywords:
...
+
+  #### Filter on One-To-Many Navigation Properties
+  <FilterOneToMany/>
+
+  #### More Filter Expressions
+  More advanced filter expressions can be found [here](#available-filter-expressions).
+
+  ### Expand
+
+  #### Expand and Select
```
jjtang1985, line 50: I think `$select` should have the same level as `$expand`/`$filter`

docs/js/features/odata/odata-v4-client.mdx
```diff
@@ -13,112 +13,92 @@ keywords:
...
+  ### Deep Create
+  <DeepCreate/>
+
+  ### Create as Child of
+  <CreateAsChildOf/>
+
+## Update Request Builder
+<Update/>
+
+## Delete Request Builder
```
jjtang1985, line 80: Etag and csrf token are missing.

jjtang1985 (approved):
I like the way how we "reuse" the doc with the `jsx` files.
Please consider some minor suggestions.

jjtang1985 (approved):

PR #538, opened 2020-10-01
Remove duplicate entry in sidebar

newtork:
## Context

Logging is listed twice for Java/Guides


artemkovalyov (approved):
Nicely spotted!

PR #548, opened 2020-10-02
Add example MTA file to logging guide

MatKuhr:
This PR enhances the logging guide to also cover MTA build files.


cschubertcs (commented):
Looks like it could work (I never tried it, so no definite clue from my side). Just two questions:
- Did you verify it/know where this is coming from?
- Where is the `java.tomcat` coming from?

MatKuhr:
> * Did you verify it/know where this is coming from?

I did not verify it myself but a user of the SDK was using it like this. In itself it's identical to the `manifest.yml` usage, only the surroundings are different. That's why I thought it's fine we just include as an example as well without further testing.

> * Where is the `java.tomcat` coming from?

I copied an example from [the documentation](https://help.sap.com/viewer/4505d0bdaf4948449b7f7379d24d0f0d/2.0.03/en-US/33548a721e6548688605049792d55295.html) that is also linked now.

cschubertcs (approved):
Ah, okay, somewhat found that in the documentation. Thanks for the link!
LGTM then.

PR #551, opened 2020-10-05
Removed insensitive terms from docs

MatKuhr:
## Context

This PR introduces inclusive language as [announced here](https://blogs.wdf.sap.corp/sapnews_en/2020/09/replacing-insensitive-terms-to-make-the-software-industry-more-inclusive/).


artemkovalyov (approved):
Thanks for taking care of this @MatKuhr , I think we should stay compliant here.

PR #557, opened 2020-10-07
chore: Refactor Destination Accessor

FrankEssenberger:
## Context

Clean up the long logic of the `getDestinationFromDestinationService` method. The main thing I did was creating a class holding the state i.e. options, selction strategy, isolation and so on. Doing so one could have slim methods and there was no need to pass many arguments around or have the super long methods arguments.

Points for discussion:
I used `Destination | undefined` but the older methods used `Destination|null` so I needed to adjust there. Should I also use `null` instead of `undefiend`?

It seems like I broke a lot of tests. So perhaps wait with the review when it is green.


FrankEssenberger:
So test are working now. I adjusted the logic so that if the provider/subscriber is not needed it is not called. This lead to problems in the nocks as discussed already for our new test strategy we will improve there.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+    // The information if the destination originates from a instance of account service is not relevant for the caching.
+    destinationCache.cacheRetrievedDestinations(
+      this.decodedUserJwt || this.decodedProviderToken,
+      { instance: [], subaccount: [destination] },
+      this.isolationStrategy
+    );
+    return destination;
+  }
+
+  private async tryService(): Promise<Destination | undefined> {
```
mr-flannery, line 482: maybe `tryFetchingFromService`?

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+  private get selectionStrategy(): DestinationSelectionStrategy {
+    return this.options.selectionStrategy
+      ? this.options.selectionStrategy
+      : subscriberFirst;
+  }
+
+  private get isolationStrategy(): IsolationStrategy {
+    return this.options.isolationStrategy
+      ? this.options.isolationStrategy
+      : IsolationStrategy.Tenant;
```
mr-flannery, line 363: I think `return this.options.isolationStrategy ?? IsolationStrategy.Tenant` is equivalent and more concise (same applies to `selectionStrategy`)

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+    const da = new DestinationAccessor(
+      name,
+      options,
+      decodedUserJwt,
+      providerToken
+    );
+
+    return (
+      da.trySubscriberCache() ??
+      (await da.tryProviderCache()) ??
```
mr-flannery, line 312: shouldn't this also be subject to the `selectionStrategy`?

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+        return fetchDestination(
+          destinationService.credentials.uri,
+          this.providerToken,
+          this.name,
+          this.options
+        );
+      }
+      if (!this.options.userJwt) {
+        throw Error(
+          `No user token (JWT) has been provided! This is strictly necessary for principal propagation. Value of the JWT: ${this.options.userJwt}.`
```
mr-flannery, line 394: Not sure if adding unsanitized JWTs to error messages is the most secure thing to do. at least the signature could be redacted/falsified.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+      decodedUserJwt,
+      providerToken
+    );
+
+    return (
+      da.trySubscriberCache() ??
+      (await da.tryProviderCache()) ??
+      (await da.processDestinationResult(await da.tryForOAuth())) ??
+      (await da.processDestinationResult(await da.tryForCert())) ??
+      (await da.processDestinationResult(await da.tryService())) ??
```
mr-flannery, line 315: two observations:
1. I think you're calling `tryService` in all three functions. given the cache, that shouldn't be an issue, but at least in my mental model I would've first called `tryService` explicitly and then passed it through these functions. But that's rather minor, and you could probably also make a case for why it makes sense to do it the way you've done it.
2. What bugs me more is that if you call `processDestinationResult` in every case, then using some kind of pipeline-thingy would make more sense here. Maybe not exactly that fp pipeline thingy, but write in a way that you don't have to explicitly call that function everytime. like:
```
const result = x ?? y ?? z;
processResult(result)
return result;
```

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+    // We are in selection strategy subscriberFirst here. So if we get something from subscriber ignore provider.
+    if (await this.destinationFoundFromSubscriber()) {
+      return false;
+    }
+    return true;
+  }
+
+  private async tryProviderCache(): Promise<Destination | undefined> {
+    if (this.useCache) {
+      const providerDestinationCache = destinationCache.retrieveDestinationFromCache(
```
mr-flannery, line 613: I'd rename this to something that indicates that the thing is a destination, e.g. `destinationFromProviderCache` or just `destination` (since the provider-cache-context is already given by the function name).

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+        this.isolationStrategy
+      );
+      if (subscriberDestinationCache) {
+        logger.info(
+          'Successfully retrieved destination from destination service cache for subscriber destinations.'
+        );
+        return subscriberDestinationCache;
+      }
+    }
+  }
```
mr-flannery, line 576: missing newline

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+      this.updateCacheService(
+        this.decodedProviderToken,
+        this.providerDestinationCache
+      );
+    }
+
+    return this.providerDestinationCache;
+  }
+
+  private async getSubscriberDestination() {
```
mr-flannery, line 536: missing return type

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+  }
+
+  private subscriberDestinationCache: DestinationsByType | undefined;
+  private providerDestinationCache: DestinationsByType | undefined;
+
+  private constructor(
+    private name: string,
+    private options: DestinationOptions,
+    private decodedUserJwt: DecodedJWT | undefined,
+    private providerToken: string
```
mr-flannery, line 344: strongly consider making these `readonly`

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+
+  private static async getProviderToken(
+    options: DestinationOptions
+  ): Promise<string> {
+    const { userJwt, ...optionsWithoutJwt } = options;
+    return serviceToken('destination', optionsWithoutJwt);
+  }
+
+  private subscriberDestinationCache: DestinationsByType | undefined;
+  private providerDestinationCache: DestinationsByType | undefined;
```
mr-flannery, line 338: I have to say that these two variables are confusing me immensely.
Both `trySubscriberCache` as well as `tryProviderCache` are reading from the actual `destinationCache` that is defined outside of this class, and since the tests are passing I'm guessing that cross-request caching still works (unless the tests are broken, of course). 
If that is the case, then what are these variables doing? They're called `somethingCache`, but they're not actually caches (as also indicated by the type). From how they're used it looks like you're trying to prevent that the same HTTP requests are executed multiple times in the scope of a single conceptual "hey gimme destination X" operation. But how would that happen in the first place?

mr-flannery (changes requested):
I like the refactoring in general! 👍 
I think there are some things that can be improved, and I find the usage of `cache` confusing (and either I didn't understand it correctly or the code is more complex than it needs to be).

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+      this.name
+    );
+
+    if (!selectedFromService) {
+      logger.info('Could not retrieve destination from destination service.');
+    }
+    return selectedFromService ?? undefined;
+  }
+
+  private async destinationFoundFromSubscriber(): Promise<boolean> {
```
marikaner, line 498: This function is only used to check whether there are subscriber destinations for the subscriber first strategy, right? Is it even necessary to execute the selection strategy here? (If yes, I guess a comment would be helpful).
Consider something like this: `canFindSubscriberDestination`

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+    if (this.useCache) {
+      destinationCache.cacheRetrievedDestinations(
+        token,
+        destinations,
+        this.isolationStrategy
+      );
+    }
+  }
+
+  private updateCacheSingle(
```
marikaner, line 467: Why does this return a destination?

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+  ): Promise<Destination | undefined> {
+    if (!destination) {
+      return;
+    }
+
+    logger.info('Successfully retrieved destination from destination service.');
+
+    return this.updateCacheSingle(
+      await this.addProxyConfiguration(destination)
+    );
```
marikaner, line 451: Although those are just two lines of code, this is kind of hidden magic...

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+
+    return (
+      da.trySubscriberCache() ??
+      (await da.tryProviderCache()) ??
+      (await da.processDestinationResult(await da.tryForOAuth())) ??
+      (await da.processDestinationResult(await da.tryForCert())) ??
+      (await da.processDestinationResult(await da.tryService())) ??
+      undefined
+    );
```
marikaner, line 317: `tryForOAuth` and `tryForCert` both call `tryService`. Also, `processDestinationResult` is always called.
```suggestion
    if (this.options.useCache) {
      return da.trySubscriberCache() || (await da.tryProviderCache());
      }
      let destination = await da.tryService();
      destination = await da.tryForOAuth(destination) || await da.tryForCert(destination) || destination;
      return da.processDestinationResult(destination);
    );
```

marikaner (commented):
I think the destination accessor is still very complex, so I just looked at some places that struck me so that we can find a sweet spot step by step. I think this can probably be split into multiple files (this files is currently > 600 loc). I found one function that is unused, maybe there is more.
I didn't try to understand too deeply yet, but I am not sure what an instance of DestinationAccessor represents. I'll think about it more next week ;)

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+    options: DestinationOptions & { iss?: string }
+  ): Promise<Destination | undefined> {
+    const decodedUserJwt = await DestinationAccessor.getDecodedUserJwt(options);
+    const providerToken = await DestinationAccessor.getProviderToken(options);
+    const da = new DestinationAccessor(
+      name,
+      options,
+      decodedUserJwt,
+      providerToken
+    );
```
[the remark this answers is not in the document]
FrankEssenberger: I wanted to but async/await is not allowed in a constructor as well. So I would have promises as class properties and I would need to await then always when I access them, which cluttered the code so I decided to move it outside.

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: I also was not happy with the name. These variables are used so that the services are called a minimal amount of times for one fetch round trip. If the cache destination is switched on they are useless. But  if the cache is switched of by the config it prevents the code to be highly inefficient compared to the old implementation, where all was in one method so that also without a cache only one call was done.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -132,9 +132,9 @@ export async function getDestination(
 ): Promise<Destination | null> {
   const destination = await (tryDestinationFromEnv(name) ||
     tryDestinationForServiceBinding(name) ||
-    getDestinationFromDestinationService(name, options));
+    DestinationAccessor.getDestinationFromDestinationService(name, options));
 
-  return destination;
+  return destination ?? null;
```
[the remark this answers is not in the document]
FrankEssenberger: I will do the following. I add a v2 ticket to get rid of all null usages as return type because we use `undefined` much more often.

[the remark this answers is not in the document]
FrankEssenberger: https://github.wdf.sap.corp/MA/scp-rocket/issues/1477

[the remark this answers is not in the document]
FrankEssenberger: but to avoid breaking changes and these check I will undefined for the moment.

packages/core/test/scp-cf/destination-accessor-proxy-configuration.spec.ts
```diff
@@ -135,5 +128,5 @@ describe('proxy configuration', () => {
     });
     expect(actual).toEqual(expected);
     httpMocks.forEach(mock => expect(mock.isDone()).toBe(true));
-  });
+  }, 9999999);
```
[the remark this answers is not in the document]
FrankEssenberger: ah, I did this for debugging because jest killed the process. I forgot to remove it.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
-    !destinationByType.instance.length && !destinationByType.instance.length
+    !destinationByType.instance.length && !destinationByType.subaccount.length
   );
 }
+
+class DestinationAccessor {
+  public static async getDestinationFromDestinationService(
+    name: string,
+    options: DestinationOptions & { iss?: string }
+  ): Promise<Destination | undefined> {
```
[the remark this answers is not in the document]
FrankEssenberger: I tried will use `null` here but keep the `undefined` everywhere else because it is more convenient because you do not need explicit return statements.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
-function subscriberDestinationIsSelected(
-  selectionStrategy: DestinationSelectionStrategy
-): boolean {
-  return (
-    selectionStrategy === subscriberFirst ||
-    selectionStrategy === alwaysSubscriber
-  );
-}
-
 function emptyDestinationsByType(
```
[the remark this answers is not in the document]
FrankEssenberger: yes I like it with verbs.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+  }
+
+  private static async getDecodedUserJwt(
+    options: DestinationOptions & { iss?: string }
+  ): Promise<DecodedJWT | undefined> {
+    return options.userJwt
+      ? verifyJwt(options.userJwt, options)
+      : options.iss
+      ? { iss: options.iss }
+      : undefined;
```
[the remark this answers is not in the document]
FrankEssenberger: good idea.

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: yes

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: yes

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: ah yes

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: I would love: private and readonly because I do not want to expose them and also make them not changeable and be forced to initialize them. But I guess readonly is the more modern and more used keyword.

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: yes but since the logic to determine if you should consider the caches is rather complex I put it in the methods.

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: I copy pasted that but I think it does not make sense. If `!this.options.userJwt` there is nothing to print... I just remove it.

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: I adjust it. Marika commented the same thing :-).

packages/core/src/scp-cf/destination-accessor.ts
mr-flannery: Okay I get the reasoning, but I still don't understand this: In which scenario in the old code would the same endpoint have been called twice?

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: I tried to make it more clear

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: ehm, for chaining, but this is not good.

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: I put the method calls in the root method

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: yes you need to execute the selection strategy so that you know if you would find a dest on the subscriber result.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+  ) {
+    this.providerDestinationCache = undefined;
+    this.subscriberDestinationCache = undefined;
+  }
+
+  private get useCache(): boolean {
+    return this.options.useCache || false;
+  }
+
+  private get selectionStrategy(): DestinationSelectionStrategy {
```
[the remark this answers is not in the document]
FrankEssenberger: matter of taste I would say, but it is also a bit more efficient to do that once in the constructor so I get rid of the getters.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
   return (
-    !destinationByType.instance.length && !destinationByType.instance.length
+    !destinationByType.instance.length && !destinationByType.subaccount.length
   );
 }
+
+class DestinationAccessor {
+  public static async getDestinationFromDestinationService(
+    name: string,
+    options: DestinationOptions & { iss?: string }
```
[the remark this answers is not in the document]
FrankEssenberger: I put it there with a FIXME to provide a proper method

FrankEssenberger:
Ok I updated and I am happy with the current state. Thanks for the comments. Please have a rereview.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+      const providerDestinationCache = destinationCache.retrieveDestinationFromCache(
+        this.decodedProviderToken,
+        this.name,
+        this.isolationStrategy
+      );
+      if (!providerDestinationCache) {
+        return;
+      }
+
+      if (await this.isProviderNeeded()) {
```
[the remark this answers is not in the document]
FrankEssenberger: changed flow

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -482,19 +285,346 @@ const emptyDestinationByType: DestinationsByType = {
...
+      this.providerDestinationCache = await getInstanceAndSubaccountDestinations(
+        destinationServiceCreds.uri,
+        this.providerToken,
+        this.options
+      );
+      this.updateCacheService(
+        this.decodedProviderToken,
+        this.providerDestinationCache
+      );
+    }
```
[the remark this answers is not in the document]
FrankEssenberger: tried to remove the side effect. The global cache is still there but this is a smaller side effect.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
+      try {
+        const destination = getDestinationFromEnvByName(name);
+        if (destination) {
+          logger.info(
+            'Successfully retrieved destination from environment variable.'
+          );
+          return destination;
+        }
+      } catch (error) {
+        logger.info(error.message);
```
marikaner, line 275: I think we should do more than just log the error here and add some context. `Could not read destinations env variable` or something in that direction.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
+      }
     }
+
+    logger.info('Could not retrieve destination from environment variable.');
```
marikaner, line 279: This is quite misleading. When we get here we didn't even try. Consider something like this:
```suggestion
    logger.info('No environment variable set.');
```

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
+      const destination = destinationForServiceBinding(name);
+      logger.info('Successfully retrieved destination from service binding.');
+      return destination;
+    } catch (error) {
+      logger.info(error.message);
+      logger.info('Could not retrieve destination from service binding.');
       logger.info(
-        'Successfully retrieved destination from destination service cache for provider destinations.'
+        'If you are not using SAP Extension Factory, this information probably does not concern you.'
+      );
```
marikaner, line 247: I would rather recommend adding one error message. It seems unnecessary to separate this and could be misinterpreted as you see the error first.
```ts
 logger.info(`Could not retrieve destination from service binding. If you are not using SAP Extension Factory, this information probably does not concern you. Original error: ${error.message}`);
```

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
+      return undefined;
```
marikaner, line 248: This is not necessary
```suggestion
```

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
+        this.destionationServiceCredentials.uri,
+        accessToken,
+        this.options
+      )
+    ]);
+
+    return {
+      instance: destinations[0],
+      subaccount: destinations[1]
+    };
```
marikaner, line 346: very minor
```suggestion
    const [instance, subaccount] = await Promise.all([
      fetchInstanceDestinations(
        this.destionationServiceCredentials.uri,
        accessToken,
        this.options
      ),
      fetchSubaccountDestinations(
        this.destionationServiceCredentials.uri,
        accessToken,
        this.options
      )
    ]);

    return {
      instance,
      subaccount
    };
```

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
-      ? await getAllProviderDestinations(providerToken, options)
-      : emptyDestinationByType;
-
-  if (options.useCache) {
-    if (options.userJwt) {
-      destinationCache.cacheRetrievedDestinations(
-        decodedUserJwt!,
-        subscriberDestinations,
-        isolation
+  private get destionationServiceCredentials() {
```
marikaner, line 349:
```suggestion
  private get destinationServiceCredentials() {
```

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
-const emptyDestinationByType: DestinationsByType = {
-  instance: [],
-  subaccount: []
-};
+    if (this.useCache) {
+      return (
+        this.getProviderDestinationCache() ??
+        this.getProviderDestinationService()
+      );
+    }
```
marikaner, line 566: I think this is slightly better (same for subscriber):
```suggestion
    let destination;
    if (this.useCache) {
      destination = this.getProviderDestinationCache();
    }
    return destination || this.getProviderDestinationService();
```

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
-): boolean {
-  return (
-    !destinationByType.instance.length && !destinationByType.instance.length
-  );
+  private async searchSubscriberAccountForDestination(): Promise<
+    DestinationSearchResult | undefined
+  > {
+    if (!this.isSubscriberNeeded(this.decodedUserJwt)) {
+      return;
+    }
```
marikaner, line 576: I think semantically it would fit better to move this up to the calling function.

marikaner (changes requested):
I have to be honest, I really don't see the point of the destination accessor class at the moment.
All the static methods could be just functions as they have been before and they wouldn't have to be forwarded to the destination accessor (`getDestination` calls `DestinationAccessor.getDestination`).
The only use at the moment is the destination service.
We have separate modules handling env destinations and service binding destinations (with currently inconsistent names: `env-destination-accessor` and `vcap-service-destination`). I think we should also introduce one for the destination service. The DestinationAccessor class is only really used for this purpose. With a rename in a separate module it will make much more sense. With this you could also reduce the size of the destination-accessor module which is still almost 600 LOC and has a lot of different concerns.
I currently find that the subscriber / provider retrieval process is quite hidden and I had difficulties following the flow, but I will review once more tomorrow ;)

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -86,12 +85,32 @@ export interface DestinationAccessorOptions {
    * The user token of the current request.
    */
   userJwt?: string;
+
+  /**
+   * @hidden
+   */
+  iss?: string;
+  // FIXME This is used to put a subscriber domain in without having a JWT like for packground processes.
+  // We will create a seperate method for this on the destination accessor wit proper JS doc. This will be deprecated.
```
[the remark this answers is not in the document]
FrankEssenberger: I would put the deprecated once we added a proper method exposing this.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
+class DestinationAccessor {
+  public static async getDestination(
+    name: string,
+    options: DestinationOptions = {}
+  ): Promise<Destination | null> {
+    return (
+      DestinationAccessor.searchEnvVariablesForDestination(name) ??
+      DestinationAccessor.searchServiceBindingForDestination(name) ??
+      DestinationAccessor.getDestinationFromDestinationService(name, options)
+    );
```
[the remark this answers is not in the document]
FrankEssenberger: now I am confused. I wanted to go to the next only for null/undefined. So I thought `??` is only considering `null/undefined` and `||` falsy values like `undefined,null,0,''`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
-  let providerDestinationCache;
-
-  if (
-    options.useCache &&
-    options.userJwt &&
-    subscriberDestinationIsSelected(selectionStrategy)
-  ) {
-    const subscriberDestinationCache = destinationCache.retrieveDestinationFromCache(
-      decodedUserJwt!,
+  return DestinationAccessor.getDestinationFromDestinationService(
```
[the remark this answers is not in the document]
FrankEssenberger: For me it makes sense to have the iss options next to the JWT since they are both the sources for the subaccount url.

packages/core/src/scp-cf/destination-accessor.ts
```diff
@@ -153,348 +168,420 @@ export async function getDestinationFromDestinationService(
...
+    if (getDestinationsEnvVariable()) {
+      logger.warn(
+        "Environment variable 'destinations' is set. Destinations will be read from this variable. " +
+          'This is discouraged for a productive application! ' +
+          'Unset the variable to read destinations from the destination service on SAP Cloud Platform.'
       );
-      return providerDestinationCache;
+
+      try {
+        const destination = getDestinationFromEnvByName(name);
```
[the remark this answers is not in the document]
FrankEssenberger: ok

packages/core/src/scp-cf/destination-accessor.ts
FrankEssenberger: yes

FrankEssenberger:
I splt the service from the destination-accessor part. Let me know if we are moving in the right direction.

packages/core/src/scp-cf/destination/destination-service.ts
```diff
@@ -55,24 +65,48 @@ export function fetchSubaccountDestinations(
   );
 }
 
-enum DestinationType {
+export enum DestinationType {
```
marikaner, line 68: This is now public api. It should be documented.

packages/core/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -0,0 +1,424 @@
...
+      options
+    );
+    const da = new DestinationFromService(
+      name,
+      options,
+      decodedUserJwt,
+      providerToken
+    );
+    const destinationResult =
+      (await da.searchSubscriberAccountForDestination()) ??
```
marikaner, line 92: ??

packages/core/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -0,0 +1,424 @@
...
+  name: string,
+  options: DestinationOptions
+): Promise<Destination | null> {
+  return DestinationFromService.getDestinationFromDestinationService(
+    name,
+    options
+  );
+}
+
+class DestinationFromService {
```
marikaner, line 74: consider if you can find a better name, enterprisy manager, retriever, accesor etc.

packages/core/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -0,0 +1,424 @@
...
+      options.isolationStrategy ?? IsolationStrategy.Tenant;
+    this.options.isolationStrategy = this.isolationStrategy;
+
+    this.selectionStrategy =
+      options.selectionStrategy ??
+      DestinationSelectionStrategies.subscriberFirst;
+    this.options.selectionStrategy = this.selectionStrategy;
+
+    this.useCache = options.useCache ?? false;
+    this.options.useCache = this.useCache;
```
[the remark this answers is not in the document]
marikaner: Consider using just this.options.

packages/core/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -0,0 +1,424 @@
...
+    const providerToken = await DestinationFromService.getProviderToken(
+      options
+    );
+    const da = new DestinationFromService(
+      name,
+      options,
+      decodedUserJwt,
+      providerToken
+    );
+    const destinationResult =
```
marikaner, line 91: Consider adding a comment

marikaner: Or if you have a good idea extract multiple functions and make this more explicit.

marikaner (approved):
LGTM , thanks!

packages/core/src/scp-cf/destination-accessor.ts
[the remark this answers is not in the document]
FrankEssenberger: clarified. Will be changed

packages/core/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @param name - The name of the destination to be retrieved.
+ * @param options - Configuration for how to retrieve destinations from the destination service.
+ * @returns A promise returning the requested destination on success.
+ */
+export async function getDestination(
+  name: string,
+  options: DestinationOptions = {}
+): Promise<Destination | null> {
+  return (
+    searchEnvVariablesForDestination(name) ??
```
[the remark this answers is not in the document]
FrankEssenberger: removed

packages/core/src/scp-cf/destination/destination-from-service.ts
FrankEssenberger: a added a one more method to make it more clear

PR #565, opened 2020-10-09
chore: Simplify e2e test scripts

marikaner:
## Context

We have a lot of scripts for the e2e tests in the root. I simplified them a little. Now, wenn you locally run `yarn test` the tests will pass and start and stop the e2e-test server.

I also considered removing the `test:integration` and `test:e2e` scripts because the difference between `yarn test:integration` and `yarn integration-tests test` is very small, but I left it for symmetry.

## Definition of Done

- [ ] Release notes updated.


FrankEssenberger (approved):
Great. This means that now also the server comes up and stops automatically.

marikaner:
> Great. This means that now also the server comes up and stops automatically.

yes.

FrankEssenberger (changes requested):
Great. This means that now also the server comes up and stops automatically. This means the github workflow needs to be adjusted?

PR #580, opened 2020-10-13
feat: Initial rest client

jjtang1985:
## Context

- migrate the current version to os repo
- ~~add poc code/tests~~

## Definition of Done

- [ ] Release notes updated.


packages/core/src/http-client/http-client.ts
```diff
@@ -83,6 +83,25 @@ export function execute(executeFn: ExecuteHttpRequestFn) {
   };
 }
 
+/**
+ *
+ * @experimental This is an experimental function, which might be removed later.
+ * @param destination - A destination or a destination name and a JWT.
+ * @param requestConfig - Any object representing an HTTP request.
+ */
+
```
marikaner, line 92:
```suggestion
```
I think this will not be recognized if there is an empty line beween

packages/generator-rest/test/poc/promotion.service.ts
```diff
@@ -0,0 +1,29 @@
+import { AxiosRequestConfig } from 'axios';
+import { buildAxiosRequestConfig } from '@sap-cloud-sdk/core';
+import { CalculationApi, Percent } from './generated/calculation';
+import { transaction } from './promotion-data';
+
+export async function getPromotion(): Promise<Percent> {
+  const token = 'ask the team';
```
marikaner, line 7: I don't really get why this is in the promotion-service. But I guess this is just a test for now right...

yarn.lock
```diff
@@ -1711,9 +1716,9 @@
     "@types/responselike" "*"
 
 "@types/fs-extra@^9.0.1":
-  version "9.0.1"
-  resolved "https://registry.npmjs.org/@types/fs-extra/-/fs-extra-9.0.1.tgz#91c8fc4c51f6d5dbe44c2ca9ab09310bd00c7918"
-  integrity sha512-B42Sxuaz09MhC3DDeW5kubRcQ5by4iuVQ0cRRWM2lggLzAa/KVom0Aft/208NgMvNQQZ86s5rVcqDdn/SH0/mg==
+  version "9.0.2"
```
marikaner, line 1719: Just for my understanding - did you delete the yarn.lock? I thought changes in the lock file should only be new additions, but not the old versions, but maybe I am mistaken 🤔

packages/core/src/http-client/http-client.ts
marikaner: In light of the fact that this is only used in tests for now, maybe rather add this there?

marikaner (approved):
As discussed, please remove the poc code from this pr and merge the initial version ;)

yarn.lock
jjtang1985: The pipeline failed due to `check:test-service`, which detected the changes in the `yarn.lock`.
I tried to install again, which did not help:
I also tried to run the script locally, no errors were shown.
Therefore, I deleted the lock file completely and re-generated it again.

PR #591, opened 2020-10-16
[chore] Remove rambda

FrankEssenberger:
## Context

Since it was to big (over 2MB) we removed rambda. Please check if I made no mistake in the reimplementation of the few methods we used.

## Definition of Done

- [x] Release notes updated.


packages/util/test/object.spec.ts
```diff
@@ -60,4 +60,31 @@ describe('renameKeys', () => {
 
     expect(renameKeys({ a: 'A', b: 'B' }, input)).toEqual(expected);
   });
+
+  it('flattens a object', () => {
```
jjtang1985, line 64: May be you can add a test for e.g., `flatten(1)`, which should return `[]`?.

packages/util/test/pipe.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { identity, pipe } from '../src';
+
+describe('pipe', () => {
+  const addAlice = (str: string): string => str + 'Alice';
+  const addBob = (str: string): string => str + 'Bob';
+  const addCharlie = (str?: string): string =>
+    str ? str + 'Charlie' : 'noCharlie';
+
+  it('should execute sync pipe with argument', () => {
```
jjtang1985, line 9: I think the original `pipe` supports the type check of the parameters.
Could you please add a test for example:
```
pipe(
        () => {return 1},
        (s: string)=> {return ''}
      )();
```

packages/util/test/object.spec.ts
FrankEssenberger: I adjusted the signature to expect an array as input and also added a test for the simple case.

packages/util/test/pipe.spec.ts
FrankEssenberger: I was breaking my head how to do that properly and checked finally in the `index.d.ts` of the rambda and found explicit typing like: 
```
export function pipe<V0, T1, T2, T3, T4, T5, T6, T7, T8>(
  fn0: (x: V0) => T1,
  fn1: (x: T1) => T2,
  fn2: (x: T2) => T3,
  fn3: (x: T3) => T4,
  fn4: (x: T4) => T5,
  fn5: (x: T5) => T6,
  fn6: (x: T6) => T7,
  fn: (x: T7) => T8): (x: V0) => T8;
```

and this for all funcitons from one to 10 arguments. So this is how they achieve type safety (up to ten arguments). I can put this but then we have multiple overloaded implementations.

jjtang1985: ok, you can also argue, the home made pipe function doesn't support the type safe in a chain.
In this way, maybe git rid of pipe? because the vanilla code is safer?
Or we check all the usage in the sdk, if they only handle functions like `(string) => string`, we can also limit the signature of the home made pipe.

marikaner (commented):
Just one minor comment from me: We already discussed, that we might want to consider to just get rid of `pipe`. It is rarely helpful in the cases where we use it.
I am not a big fan of `assoc` either (especially because I never know what it does from the name). I would even consider deprecating `assocSome` - if I am not mistaken this is not used anywhere.

marikaner:
`mergeSome` is also unused.

packages/util/test/pipe.spec.ts
FrankEssenberger: I removed the pipe.

FrankEssenberger:
I deprecated `assocSome` and `mergeSome`.

packages/util/src/pipe.ts
```diff
@@ -1,5 +1,22 @@
+/**
+ * Same as pipe but for asynchronous functions:
+ * asynvPipe(finc1,func2,...funcN)(start)  executes all functions in sequence awaiting the result and piping the response through in other words await funcN(... await func2(await func1(start))...)
+ * @deprecated We will remove this in version 2.0 of the SDK.
+ *
+ * @param fns - a list of asynchronous functions to be executed in sequence
+ * @param start - start value for the first function in the list
+ * @returns the reponse of the last function in the list.
+ */
 export const asyncPipe = (...fns) => (start: any): Promise<any> =>
```
jjtang1985, line 10: This can also be removed?

FrankEssenberger: we do not use it, but it was there before and customers could use it

jjtang1985: maybe we can also deprecate this for 2.0

jjtang1985 (approved):

PR #600, opened 2020-10-19
feat: Rest client request builder

jjtang1985:
## Context

As discussed, we wanted to create a wrapper class to handle e.g., the destination/connectivity service and provide type safe features. 

- [x] wrapper class with potential to be extended with multiple functions (execute/customHeader/customParameter) for every request (CRUD).
- [x] wrapper class consumes the vanilla open api client
- [x] the connectivity service is handled under the hood by calling `execute`
- [x] generator is created
- [ ] grooming

## Definition of Done

- [ ] Release notes updated.


packages/generator-rest/test/poc/generated/PriceCalculation/request-builder.ts
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved.
+ *
+ * This is a generated file powered by the SAP Cloud SDK for JavaScript.
+ */
+import { Destination, DestinationNameAndJwt, RestRequestBuilder, RestRequestConfig } from '@sap-cloud-sdk/core';
+import { AxiosRequestConfig } from 'axios';
+import { PriceCalculationApi, PriceCalculate } from './api';
+
+export class PriceCalculationApiRequestBuilder {
```
jjtang1985, line 10: This is the generated request builder.

packages/generator-rest/test/poc/promotion.service.ts
```diff
@@ -0,0 +1,57 @@
+import {
+  PriceCalculateResponse, CalculationApi
+} from './generated/PriceCalculation';
+import { transaction } from './promotion-data';
+import { PriceCalculationApiRequestBuilder } from './generated/PriceCalculation/request-builder';
+
+const token = 'ask the team';
+
+export async function getPromotionV3(): Promise<PriceCalculateResponse> {
```
jjtang1985, line 9: This is how the APIs look like.

jjtang1985:
We have to be careful about switching the version of the open api generator.
I used `1.0.18-4.3.1` and `1.0.18-4.2.1` and the generated code are not compatible. (the former generates models with capital camel names `MyName`, the later generates lower camel cases `myName`)

packages/generator-rest/package.json
```diff
@@ -30,14 +30,18 @@
     "test:local": "yarn jest",
     "test:debug": "node --inspect-brk ../../node_modules/.bin/jest --runInBand",
     "test:watch-debug": "node --inspect-brk ../../node_modules/.bin/jest --watch --runInBand",
+    "generate:poc": "ts-node src/generator-cli.ts -i test/resources/PriceCalculation  -o test/poc/generated-bk",
     "check:dependencies": "depcheck . --ignore-bin-package=true"
   },
   "dependencies": {
-    "@openapitools/openapi-generator-cli": "cli-4.2.1",
```
marikaner, line 36: Can we use version 2 here?

marikaner (commented):
I haven't reviewed yet, just want to make you aware, that I renamed the directory in #611.

FrankEssenberger (commented):
I do not see a test for the `.execute()` wrapped method.

jjtang1985:
> I do not see a test for the .execute() wrapped method.

here is the [line](
https://github.com/SAP/cloud-sdk/pull/600/files/5c01c0bf807460af088c5c6b0711074b151302ec#diff-c888c8ff9f4c2463364e8081627385097f1a97ea51a1843114bf04a36b840458R17)

packages/generator-rest/src/generator.ts
```diff
@@ -13,71 +13,128 @@ import {
   removeSync,
   writeJsonSync
 } from 'fs-extra';
+import { Project } from 'ts-morph';
```
marikaner, line 16: Why do we need ts-morph here?

packages/generator-rest/src/generator.ts
```diff
@@ -13,71 +13,128 @@ import {
...
-      }
-    })
+async function generateOneApi(
+  file: string,
+  options: GeneratorOptions,
+  pathToTemplates: string,
+  pathToMustacheValues: string
+) {
+  const folderForService = getFolderForService(options.outputDir, file);
+  mkdirSync(folderForService);
```
marikaner, line 56: Let's refer to directories rather then folders. Folder can be ambiguous (which it is not here, but anyways).

packages/generator-rest/src/generator.ts
```diff
@@ -13,71 +13,128 @@ import {
...
+    '-i',
+    pathToAdjustedOpenApiDefFile,
+    '-g',
+    'typescript-axios',
+    '-o',
+    folderForService,
+    '-t',
+    pathToTemplates,
+    '--api-package',
+    'api',
```
FrankEssenberger, line 116: I wonder why there is no`mode` and `api` folder in the generated code. I find it a bit easier to investigate if they are that is why I set them. But I also remember that under certain conditions (perhaps with the templates or so) they were ignored.

packages/generator-rest/src/request-builder/api-request-builder-class.ts
```diff
@@ -0,0 +1,73 @@
...
+  ClassDeclarationStructure,
+  MethodDeclarationStructure,
+  OptionalKind,
+  ParameterDeclarationStructure,
+  StructureKind
+} from 'ts-morph';
+import { toPascalCase, toPropertyFormat } from '@sap-cloud-sdk/core';
+import { OpenApiPath, OpenApiServiceMetadata } from '../open-api-types';
+
+export function apiRequestBuilderClass(
```
FrankEssenberger, line 11: add some doc, that this methods creates the wrapper class right?

packages/generator-rest/test/poc/promotion.service.ts
FrankEssenberger: I found it now. Why did you put this in a seperate file and not in the spec test file? Just a question/

packages/generator-rest/test/poc/promotion.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  return PriceCalculationApiRequestBuilder.calculateViaRestWithTenant(
+    'oppsapihub',
+    transaction
+  )
+    .withCustomHeaders({
+      apiKey: token
```
FrankEssenberger, line 13: use `process.env.sandboxtoken` or something else here. Because if you have a constant I will put the token there and then I push it and we have to clean up the repo...

packages/core/src/rest/rest-request-builder.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  withCustomHeaders(headers: Record<string, string>): this {
+    this.requestConfig.addCustomHeaders(headers);
+    return this;
+  }
+
+  buildRequestConfig(
+    destination: Destination | DestinationNameAndJwt
+  ): Promise<AxiosRequestConfig> {
+    return buildAxiosRequestConfig(destination, {
```
FrankEssenberger, line 17: I think the CSRF header are not fetched automatically. You could hand them in via the custom headers of course but it would be nice to include the header automatically as in the `headers()` method of the OData request.

packages/generator-rest/src/generator.ts
```diff
@@ -13,71 +13,128 @@ import {
...
+    if (response !== undefined) {
+      logger.info(`Generated the client ${response.stdout}`);
+    }
+  } catch (err) {
+    logger.error('In exception block');
+    logger.error(err);
+  }
 }
 
 function getServiceNameCamelCase(openApiFileName: string) {
```
marikaner, line 140: I know this has been here before, but maybe this can be put in a separate module (like a name formatter). Also `getServiceNameCamelCase` calls `toPascalCase`.

packages/generator-rest/test/poc/promotion.service.ts
marikaner: I like the api in general. I was also wondering why this is here and not in the spec file.

packages/generator-rest/test/poc/promotion-data.ts
```diff
@@ -0,0 +1,80 @@
+/* !
+ * Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved.
+ */
+
+import {
```
marikaner, line 5: Is this data used for something?

packages/generator-rest/test/poc/generated/PriceCalculation/index.ts
```diff
@@ -0,0 +1,16 @@
+// tslint:disable
+/**
+* Calculation
```
marikaner, line 3: I think the file structure would be a bit clearer like this:
* `outputDir/` (like in the other generator)
  * `price-calculation/`
    * `generated/`
      * `api.ts` (the files generated by the openapi generator)
    * `request-builder.ts`

With this we would also avoid naming conflicts.

packages/generator-rest/test/rest-generator.spec.ts
```diff
@@ -5,7 +5,7 @@ import * as fs from 'fs-extra';
 import { generateRest } from '../src/generator';
 
 describe('rest generator test', () => {
-  const inputDir = path.resolve(__dirname, 'resources');
+  const inputDir = path.resolve(__dirname, 'resources', 'test-apis');
```
FrankEssenberger, line 8: I understand that the integration test needs the token, but you could add a test on the generated classes and methods.

FrankEssenberger: So that the generated class has the expected files and the class has the method, but this is also implicitliy tested already

packages/generator-rest/test/poc/promotion.service.ts
FrankEssenberger: also what about mocking the response of the called server?

packages/generator-rest/test/poc/generated/PriceCalculation/request-builder.ts
marikaner: For discussion: @FrankEssenberger @jjtang1985 Do you think the name should be `-RequestBuilder` or rather `-Api`? Although structurally it is very similar to our other request builders, here you don't really build construct a request (with a fluent api) as in the other one.

packages/generator-rest/src/request-builder/api-request-builder-class.ts
```diff
@@ -0,0 +1,73 @@
...
+  return {
+    kind: StructureKind.Class,
+    name: `${serviceMetadata.apiName}ApiRequestBuilder`,
+    methods: method(serviceMetadata.paths),
+    isExported: true
+  };
+}
+
+function method(paths: OpenApiPath[]): MethodDeclarationStructure[] {
+  return paths.flatMap(p => getMethods(p));
```
FrankEssenberger, line 23: perhaps you can find adjust the spec file that the sample service also contains more than one method  (`calculateViaRestWithTenant`)

packages/core/src/rest/rest-request-config.ts
```diff
@@ -0,0 +1,9 @@
+export class RestRequestConfig {
```
marikaner, line 1: This concept of the odata request builders was there because we potentially construct a request (with top, skip, orderby, filter, etc.).
Here it is not really needed and seems like overkill to me.
I see two options:
1. I think it would be acceptable if we base the requestbuilder on the methodrequestbuilderbase (that would need some changes to be independent of odata). So probably add a layer below methodrequestbuilderbase (and maybe rename it to odatarequestbuilder).
2. Just store this in the request builder

packages/generator-rest/test/poc/generated/PriceCalculation/request-builder.ts
```diff
@@ -0,0 +1,25 @@
...
+    return new CalculateViaRestWithTenantRequestBuilder(new RestRequestConfig(), tenantName, priceCalculate);
+  }
+}
+
+export class CalculateViaRestWithTenantRequestBuilder extends RestRequestBuilder {
+  constructor(public requestConfig: RestRequestConfig, public tenantName: string, public priceCalculate: PriceCalculate) {
+    super(requestConfig);
+  }
+
+  async execute(destination: Destination | DestinationNameAndJwt) {
```
FrankEssenberger, line 21: Could you extract the response type `PriceCalculateResponse` from the `PriceCalculationApiFp` and the method. I do not know if there is an elegant way to map the type with the method.

packages/generator-rest/src/request-builder/operation-request-builder-class.ts
```diff
@@ -0,0 +1,110 @@
...
+function method(
+  metadata: OpenApiServiceMetadata,
+  openApiPath: OpenApiPath,
+  operation: OpenApiOperation
+): MethodDeclarationStructure[] {
+  return [
+    {
+      kind: StructureKind.Method,
+      name: 'execute',
+      isAsync: true,
```
FrankEssenberger, line 63: here the return type must be included. But I guess it is not see easy...

packages/generator-rest/test/poc/generated/PriceCalculation/request-builder.ts
```diff
@@ -0,0 +1,25 @@
...
+    const requestConfig: AxiosRequestConfig = await this.buildRequestConfig(destination);
+    return new PriceCalculationApi({ basePath: requestConfig.baseURL }).calculateViaRestWithTenant(this.tenantName, this.priceCalculate, requestConfig);
+  }
```
marikaner, line 24: I think the underlying requestbuilder should be even more generic.
It should take the arguments of the request, e. g. `...args: any[]` in the constructor. If I understand correctly the response is always an axios response, right?

packages/generator-rest/test/poc/generated/PriceCalculation/git_push.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/sh
```
marikaner, line 1: Is there a way to not generate this and the other unused files?

packages/generator-rest/test/get-promotion.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import {
+  setTestDestination,
+  unmockAllTestDestinations
+} from '@sap-cloud-sdk/test-util';
+import { getPromotionV3 } from './poc/promotion.service';
+import { PercentActionEnum } from './poc/generated/PriceCalculation';
+
+describe('Get promotion', () => {
+  xit('should get value and action', async () => {
```
marikaner, line 11: Disabled?

marikaner (commented):
@FrankEssenberger @jjtang1985 Should we consider calling the package openapi-generator?

Other than that I have some questions regarding the structure and generality of the request builder. I think we should definitely use the latest version of the openapi generator before building our own stuff on top.

packages/generator-rest/test/poc/promotion.service.ts
jjtang1985: I use the version from Charles in the book store project, where the service is used by some cap components?
Then I use this file for keeping 3 different versions of the API
- v1, vanilla openapi client
- v2, with a single wrapper class with `execute()` method only
- v3, in addition to `execute`, we are free to extend the request builder with more functions

I can move it to the spec file, but I would keep the version 1 with a test, because the open api generator introduced some breaking changes. With the vanilla client test, it's easier to navigate to the root cause.

packages/generator-rest/test/poc/generated/PriceCalculation/request-builder.ts
jjtang1985: You are right, the reasons why I use `RequestBuilder` are:
- the naming conflict between this and the one generated by openapi
  - if we decide to hide the generated open api client, then we can only export this
  - if we want to export both, we have to make a second name
- the similar structure.

packages/generator-rest/package.json
jjtang1985: I'll update it.

packages/generator-rest/src/generator.ts
jjtang1985: ts-morph? you mean the why using the lib?
Because we need to generate `generated/PriceCalculation/request-builder.ts`

packages/generator-rest/src/request-builder/api-request-builder-class.ts
jjtang1985: yes, all the files in this folder are ts-morph related files for generating our wrapper classes.

jjtang1985: Added more comments.

packages/generator-rest/src/request-builder/api-request-builder-class.ts
jjtang1985: I'll add more spec files for testing the generator.
I would also keep this service as e2e test.

packages/generator-rest/test/get-promotion.spec.ts
jjtang1985: Because without the token, it will fail. We should decide to:
- keep the e2e test and save the token by using secret mechanism
- mock the e2e test

I would like to keep the e2e test.

packages/generator-rest/test/poc/generated/PriceCalculation/git_push.sh
jjtang1985: Yes, you have to add the file names into the `.openapi-generator-ignore` file.
The problem is you have to know all the unused files, but when updating versions, I see more files are generated.

packages/generator-rest/test/poc/generated/PriceCalculation/request-builder.ts
jjtang1985: Yes, in this example, it's `Promise<AxiosResponse<PriceCalculateResponse>>`

packages/generator-rest/test/poc/promotion-data.ts
jjtang1985: This is a json object (80 lines) used as the request body of the e2e test.

packages/generator-rest/src/generator.ts
jjtang1985: Will move to core

packages/generator-rest/src/request-builder/operation-request-builder-class.ts
jjtang1985: Without the explicit return type, it is `Promise<AxiosResponse<PriceCalculateResponse>>` now.
The reason why it's not so easy is because you have to know which response code you refer to.
In this case,  the it looks like:
```
200: {
  content:{
    application-json:{
      schema: {
        $ref: "#/components/schemas/PriceCalculateResponse"
      }
    }
  }
}
400:{}
default:{} 
```
You can argue, just use the one with the 2XX pattern.
We can also try to check the number of 2XX, 
- if there is one occurrence, we use it as return type
- else, we leave it empty

packages/generator-rest/src/request-builder/api-request-builder-class.ts
FrankEssenberger: sure

packages/generator-rest/test/poc/generated/PriceCalculation/index.ts
FrankEssenberger: good point. One folder from the generated stuff and outside of this our wrapper

packages/generator-rest/src/request-builder/operation-request-builder-class.ts
jjtang1985: tested with the VS Code, the `execute()` function returns the type `Promise<AxiosResponse<PriceCalculateResponse>>`.
I will add the return type with the strategy mentioned above so the potential compilation errors are detected earlier when breaking changes are introduced.

packages/generator-rest/src/generator.ts
jjtang1985: Discussed with Frank, if we keep the ts-morph as the way for the odata generator, I would keep the same way for the rest, because it's easy for maintenance.

packages/generator-rest/test/poc/generated/PriceCalculation/request-builder.ts
jjtang1985: Discussed with Frank, I still think the vanilla version of the open api client can work as some work around in some cases. Therefore, choosing a new name might make sense. The name of the class still needs to be decided though.

packages/generator-rest/src/generator.ts
jjtang1985: https://github.wdf.sap.corp/MA/scp-rocket/issues/1524

packages/generator-rest/test/poc/promotion.service.ts
jjtang1985: e2e tests with credential/env are adjusted

packages/core/src/rest/rest-request-builder.ts
jjtang1985: https://github.wdf.sap.corp/MA/scp-rocket/issues/1525

packages/generator-rest/test/poc/generated/PriceCalculation/index.ts
jjtang1985: I slightly changed the structure you proposed. I'll use `open-api` instead of `generated`, because:
- it's clear, these are generated by using the open-api tool
- the outputDir might also be named as `generated` in real life.

packages/generator-rest/src/generator.ts
jjtang1985: This is really open-api/json related implementation. I guess this should not be reused outside the package.

packages/rest-generator/src/parse-open-api-json.ts
```diff
@@ -0,0 +1,88 @@
+import { readJsonSync } from 'fs-extra';
+import { last } from '@sap-cloud-sdk/util';
+import {
+  OpenApiOperation,
+  OpenApiPath,
+  OpenApiServiceMetadata,
+  SupportedOperation
+} from './open-api-types';
+
+export function toOpenApiModel(
```
FrankEssenberger, line 10: no doc. Also it returns a `ApiServiceMetadata` and is called `toOpenApiModel`. I would consider renaming either the method or the interface.

FrankEssenberger (approved):
LGTM. Only one minor comment left for me.

packages/rest-generator/test/test-services/PriceCalculation/open-api/base.ts
```diff
@@ -0,0 +1,84 @@
+/**
```
FrankEssenberger, line 1: Question: What about the folders/package for model and api for better readability since we set this options in the generator/

marikaner (approved):
LGTM 😊

packages/rest-generator/test/test-services/PriceCalculation/open-api/base.ts
jjtang1985: I tried these two options with vanilla open api command without templates, still didn't work.
Follow up tickets:
https://github.wdf.sap.corp/MA/scp-rocket/issues/1524

PR #611, opened 2020-10-21
chore: Loosen versions

marikaner:
## Context

Use `^` for moment and also the open api generator (+ some test deps).

## Definition of Done

- [ ] Release notes updated.


jjtang1985 (approved):

packages/rest-generator/package.json
```diff
@@ -33,7 +33,7 @@
     "check:dependencies": "depcheck . --ignore-bin-package=true"
   },
   "dependencies": {
-    "@openapitools/openapi-generator-cli": "cli-4.2.1",
+    "@openapitools/openapi-generator-cli": "^1.0.18-4.3.1",
```
jjtang1985, line 36: I'm afraid of making dynamic version for this. But we should have tests for protection.

marikaner: Why are you afraid? Did we have some issues?

PR #614, opened 2020-10-22
chore: add debug/warning msg when using custom header 

jjtang1985:
## Context

Custom header are supported when executing http request. Since it will be used instead of the ones fetched from the destination service, more warning messages should be helpful.

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/packages/core/src/http-client/http-client.ts
+++ b/packages/core/src/http-client/http-client.ts
@@ -1,6 +1,6 @@
 import * as http from 'http';
 import * as https from 'https';
-import { errorWithCause } from '@sap-cloud-sdk/util';
+import { createLogger, errorWithCause } from '@sap-cloud-sdk/util';
 import axios from 'axios';
 import { buildHeadersForDestination } from '../header-builder/header-builder-for-destination';
 import {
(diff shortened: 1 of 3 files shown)


packages/core/src/http-client/http-client.ts
```diff
@@ -30,6 +35,13 @@ export async function buildHttpRequest(
   destination: Destination | DestinationNameAndJwt,
   customHeaders?: Record<string, any>
 ): Promise<DestinationHttpRequestConfig> {
+  if (customHeaders) {
+    logger.warn(
+      `The custom headers are provided with the keys: ${Object.keys(
```
FrankEssenberger, line 40: Perhaps also add a sentence what this means: `These keys will overwrite the headers created by the SDK`.

packages/core/src/http-client/http-client.ts
```diff
@@ -30,6 +35,13 @@ export async function buildHttpRequest(
...
+        customHeaders
+      )}.`
```
jjtang1985, line 42:
```suggestion
      )}. These keys will overwrite the headers created by the SDK.`
```

FrankEssenberger (approved):
For the moment this is fine for me, but I added a backlog item that we should unify the header handling and merging between custom and calculated headers in one central place. Then we can also give a nicer message. "There have been provded" and "replaced XYZ"

PR #618, opened 2020-10-25
fix: the original error is shown when retrieving csrf token

jjtang1985:
## Context

In e2e test vdm, batch failed with OP destination, the error was:
`The error response is undefined.\n    at /home/vcap/app/node_modules/@sap-cloud-sdk/core/src/header-builder/csrf-token-header.ts`

## Definition of Done

- [x] Release notes updated.

### DIFF
--- a/packages/core/src/header-builder/csrf-token-header.ts
+++ b/packages/core/src/header-builder/csrf-token-header.ts
@@ -49,7 +49,9 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
     .then(response => response.headers)
     .catch(error => {
       if (!error.response) {
-        throw new Error(`The error response is undefined. The original error is: ${error}.`);
+        throw new Error(
+          `The error response is undefined. The original error is: ${error}.`
+        );
       }
       return error.response.headers;
     });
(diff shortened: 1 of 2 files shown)


packages/core/src/header-builder/csrf-token-header.ts
```diff
@@ -49,7 +49,9 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
     .then(response => response.headers)
     .catch(error => {
       if (!error.response) {
-        throw new Error('The error response is undefined.');
+        throw new Error(
```
FrankEssenberger, line 52: throw errorWithCause('The message'.error)

FrankEssenberger (approved):
LGTM align with marika if we should adjust the `errorWithCause` method to also include the `message` property of the exception goin in if it is present.

jjtang1985:
Discussed with Marika, the 'stack' is not standard, but supported by all the browsers and node versions, which is reliable.

FrankEssenberger:
What about her thoughts on the `message` property?

jjtang1985:
> What about her thoughts on the message property?
Usually, the useful information is included in the stack so message is relatively useless. So we decided to use the existing `errorWithCause` unless we see some issues from the stack.

PR #624, opened 2020-10-26
chore: Generator templates

marikaner:
## Context

This is the preliminary result of my findings with regards to replacing ts-morph with string templates. I tried this against the local test vdm and the s4 vdm. In this POC I am only generating entities (which are the most complex file that we generate I believe). The code could probably be improved quite a bit, the purpose here it to show what it is about and how it compares to the current approach.

In my opinion the template approach is better readable, testable and more convenient. The (non scientific) measurements prove that it is much faster too (10x). It would be interesting to know whether the results are similar for you. To run both versions execute:
```shell
# template approach
$ ts-node generate-templates.ts t > templates.log

# ts-morph approach
$ ts-node generate-templates.ts m > ts-morph.log
``` 

I would like to consider this for v2 (and potentially start #before).

## Definition of Done

- [ ] Release notes updated.


FrankEssenberger:
I tried it out and `template` took 2.8sec on my i9 six core machine. The `ts-morph` took 20sec so a factor of 7.

FrankEssenberger:
Let me know when we should review. My personal opinion is that I prefer stronger typing via TS-morph so that standard object are created safe. Howver, I now that in many places we build things on a string basis. With this apporach this becomes consistent and I think easier to test.

marikaner:
> Let me know when we should review. My personal opinion is that I prefer stronger typing via TS-morph so that standard object are created safe. Howver, I now that in many places we build things on a string basis. With this apporach this becomes consistent and I think easier to test.

It is ready for review. This will remain POC state unless we make a decision to proceed ;)

marikaner:
@FrankEssenberger Do we really achieve stronger typing with ts-morph? As you said in many places we use strings. But in addition to that it is still up to us to ensure the data we pass to ts-morph is correct. If the strings we use to represent our types are wrong, there is no additional type safety (e. g. I am still responsible to type `{ type: 'string' }` and I can still make the mistake of writing `{ type: 'bing' }`. So I think it seems to give us more type safety, while in fact it really doesn't. The only additional "safety" that ts-morph gives us is that every bracket really is closed, i. e. it has internal mechanisms to ensure syntactical correctness of typescript code.

marikaner:
Closing as I split this PR into #687 and #688.

PR #626, opened 2020-10-26
fix: client credential cache

jjtang1985:
## Context

Currently, the cache is never hit even with 2 get all requests within a second.

## Definition of Done

- [ ] Release notes updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@
 ## Fixed Issues
 
 - [Generator] Fix the function import (OData V2 + V4) and action import, where the return type is a primitive edm type like `Edm.String`.
-- [Cache] Fix the client credentials cache, which is never hit due to wrong expire time.
+- [Cache] Fix expiration times in the client credentials cache.
 - [CSRF] Fix the error message of fetching the csrf token, it should show original error.
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@
 ## Fixed Issues
 
 - [Generator] Fix the function import (OData V2 + V4) and action import, where the return type is a primitive edm type like `Edm.String`.
-- [Cache] Fix the client credential cache, which is never hit due to wrong expire time.
+- [Cache] Fix the client credentials cache, which is never hit due to wrong expire time.
 - [CSRF] Fix the error message of fetching the csrf token, it should show original error.
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,8 @@
 
 ## Fixed Issues
 
-- [Generator] Fix the function import (OData V2 + V4) and action import, where the return type is an primitive edm type like `Edm.String`.
+- [Generator] Fix the function import (OData V2 + V4) and action import, where the return type is a primitive edm type like `Edm.String`.
+- [Cache] Fix the client credential cache, which is never hit due to wrong expire time.
 - [CSRF] Fix the error message of fetching the csrf token, it should show original error.
 
 
--- a/packages/core/src/scp-cf/client-credentials-token-cache.ts
+++ b/packages/core/src/scp-cf/client-credentials-token-cache.ts
@@ -1,3 +1,4 @@
+import moment from 'moment';
 import { Cache } from './cache';
 import { headerForClientCredentials } from './xsuaa-service';
 import {
@@ -21,7 +22,11 @@ const ClientCredentialsTokenCache = (
     credentials: ClientCredentials,
     token: ClientCredentialsResponse
   ): void => {
-    cache.set(getGrantTokenCacheKey(url, credentials), token, token.expires_in);
+    cache.set(
+      getGrantTokenCacheKey(url, credentials),
+      token,
+      moment().add(token.expires_in, 'second').unix() * 1000
+    );
   },
   clear: (): void => {
     cache.clear();
--- a/packages/core/src/scp-cf/client-credentials-token-cache.ts
+++ b/packages/core/src/scp-cf/client-credentials-token-cache.ts
@@ -27,7 +27,7 @@ const ClientCredentialsTokenCache = (
       token,
       token.expires_in
         ? moment().add(token.expires_in, 'second').unix() * 1000
-        : 0
+        : undefined
     );
   },
   clear: (): void => {
--- a/packages/core/src/scp-cf/client-credentials-token-cache.ts
+++ b/packages/core/src/scp-cf/client-credentials-token-cache.ts
@@ -27,7 +27,7 @@ const ClientCredentialsTokenCache = (
       token,
       token.expires_in
         ? moment().add(token.expires_in, 'second').unix() * 1000
-        : undefined
+        : 0
     );
   },
   clear: (): void => {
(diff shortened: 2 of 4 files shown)


packages/core/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -21,7 +22,11 @@ const ClientCredentialsTokenCache = (
     credentials: ClientCredentials,
     token: ClientCredentialsResponse
   ): void => {
-    cache.set(getGrantTokenCacheKey(url, credentials), token, token.expires_in);
+    cache.set(
+      getGrantTokenCacheKey(url, credentials),
+      token,
+      moment().add(token.expires_in, 'second').unix() * 1000
```
jjtang1985, line 28: A typical `expires_in` value is `43199`, which means 12 hours later it will expire.

packages/core/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -21,7 +22,11 @@ const ClientCredentialsTokenCache = (
     credentials: ClientCredentials,
     token: ClientCredentialsResponse
   ): void => {
-    cache.set(getGrantTokenCacheKey(url, credentials), token, token.expires_in);
+    cache.set(
+      getGrantTokenCacheKey(url, credentials),
+      token,
+      token.expires_in && moment().add(token.expires_in, 'second').unix() * 1000
```
jjtang1985, line 28: > The number of seconds until the access token expires.

see doc: https://docs.cloudfoundry.org/api/uaa/version/74.27.0/index.html#authorization-code-grant-2

CHANGELOG.md
```diff
@@ -27,7 +27,8 @@
 
 ## Fixed Issues
 
-- [Generator] Fix the function import (OData V2 + V4) and action import, where the return type is an primitive edm type like `Edm.String`.
+- [Generator] Fix the function import (OData V2 + V4) and action import, where the return type is a primitive edm type like `Edm.String`.
+- [Cache] Fix the client credential cache, which is never hit due to wrong expire time.
```
FrankEssenberger, line 31: I think it is `client credentials` with an "s"

packages/core/src/scp-cf/client-credentials-token-cache.ts
FrankEssenberger: What does the `token.expires_in && moment().add(token.expires_in, 'second').unix() * 1000` do? you want to return the value now+expires time if the `token.expires_in` is present right? I see that the test is working but I find this hard to read because `&&` is returning a boolean?

FrankEssenberger (dismissed):
I would like to see a different way to set the number in the client credentials token beucause the `&&` triggers boolean for me.

packages/core/src/scp-cf/client-credentials-token-cache.ts
jjtang1985: `token.expires_in && moment().add(token.expires_in, 'second').unix() * 1000` equals `token.expires_in ?moment().add(token.expires_in, 'second').unix() * 1000 : token.expires_in`
Since the `expires_in` is typed `expires_in: number;`, I'll make it simple.

CHANGELOG.md
```diff
@@ -27,7 +27,8 @@
 
 ## Fixed Issues
 
-- [Generator] Fix the function import (OData V2 + V4) and action import, where the return type is an primitive edm type like `Edm.String`.
+- [Generator] Fix the function import (OData V2 + V4) and action import, where the return type is a primitive edm type like `Edm.String`.
+- [Cache] Fix the client credentials cache, which is never hit due to wrong expire time.
```
marikaner, line 31:
```suggestion
- [Cache] Fix expiration times in the client credentials cache.
```

packages/core/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -21,7 +22,13 @@ const ClientCredentialsTokenCache = (
     credentials: ClientCredentials,
     token: ClientCredentialsResponse
   ): void => {
-    cache.set(getGrantTokenCacheKey(url, credentials), token, token.expires_in);
+    cache.set(
+      getGrantTokenCacheKey(url, credentials),
+      token,
+      token.expires_in
```
marikaner, line 28: What does undefined mean? Is it valid forever or once? If once we don't need to cache it, I guess?

packages/core/test/scp-cf/client-credentials-token-cache.spec.ts
```diff
@@ -4,18 +4,18 @@ import { clientCredentialsTokenCache } from '../../src/scp-cf';
 describe('ClientCredentialsTokenCache', () => {
   it('should return token when valid, return undefined otherwise', () => {
     const clock = install();
-    const now = Date.now();
     const validToken = {
       access_token: '1234567890',
       token_type: 'UserToken',
-      expires_in: now + 100000,
+      // x second later
```
marikaner, line 10: I do not think that this comment belongs here, because it is documentation of the jwt. If you think this information should be added somewhere, maybe the knowledge-base is a better place

packages/core/test/scp-cf/client-credentials-token-cache.spec.ts
```diff
@@ -31,8 +31,8 @@ describe('ClientCredentialsTokenCache', () => {
       credentials,
       expiredToken
     );
-
-    clock.tick(75000);
+    // millisecond
+    clock.tick(75000 * 1000);
```
marikaner, line 35: Same as above. I am rather wondering where the magic number 75000 comes from. 

Maybe create variables?

marikaner:
@mr-flannery Do you remember why we handled this as a point in time instead of a duration? Maybe some systems handle it wrong?

mr-flannery:
I think Ahmed wrote most of the caching related code, and I can't remember questioning that. I'm assuming that by "handled _this_ as a point in time" you mean "why do call `cache.set` with a point in time (moment instance) instead of a duration". What problem do you see with that approach?

marikaner:
> @marikaner I think Ahmed wrote most of the caching related code, and I can't remember questioning that. I'm assuming that by "handled _this_ as a point in time" you mean "why do call `cache.set` with a point in time (moment instance) instead of a duration". What problem do you see with that approach?

I meant handling `expires_in` as a point in time. I guess it was just a mistake then ;)

packages/core/src/scp-cf/client-credentials-token-cache.ts
jjtang1985: since the cf doc doesn't tell whether this is mandatory, this might not be there.
as discussed, we don't use cache when it's undefined.

packages/core/test/scp-cf/client-credentials-token-cache.spec.ts
jjtang1985: moved to the js doc of the interface

packages/core/test/scp-cf/client-credentials-token-cache.spec.ts
jjtang1985: use variables with meaningful names

FrankEssenberger (approved):
LGTM

packages/core/src/scp-cf/client-credentials-token-cache.ts
marikaner: Maybe we really just don't need to set it?

PR #645, opened 2020-10-29
chore: Investigate moment replacement

marikaner:
## Context

This is the outcome of my investigation towards replacing moment. I think we should go the native way and either wait with v2 for it to be there or allow for both ways until Temporal becomes standard, while clearly marking that the moment support will be deprecated with a minor version update.

I also considered other alternatives out there, it seems that these are the most promising ones (there are some more):
4.21MB - moment
2.98MD - luxon
533KB - dayjs

Luxon is quite big and with this minor gain, I would rather keep the old API, I don't think this would be worth it. I thought that DayJS is worth a closer look, because it is as small as it is. It has abstractions for Dates and Durations, but I didn't find anything for time and the time zone support does not seem to be enough. It is based on a plugin mechanism where this could probably be added (and maybe there already is something).
In light of the native developments towards `Temporal`, I think this is the "ultimate" solution and changing the API in between is not very user friendly.


FrankEssenberger (changes requested):
I would like to add a decision to the `date-time-handling.md`. IMO it makes sense to wait with version 2.0 and get rid of moment. The lib has costed so much pain in the past so it is fair. You mentioned some movement in the status of the `Temporal`. Could you add the timeline of a better discussion.

marikaner:
> I would like to add a decision to the `date-time-handling.md`. IMO it makes sense to wait with version 2.0 and get rid of moment. The lib has costed so much pain in the past so it is fair. You mentioned some movement in the status of the `Temporal`. Could you add the timeline of a better discussion.

I don’t there is a timeline. I couldn’t find one at least. I‘ll set up a discussion for the decision.

jjtang1985:
- I think we all agree `Temporal` should be the destination we want to go in the future.
- Before that I personally would avoid switching from `moment`-> `lib abc` -> `Temporal`, before that's like `sdk 1.0` -> `sdk 2.0` -> `sdk 3.0` that is not user friendly, unless you have to wait for long time for `Temporal`, which you never know.
- In order to keep us save as plan B, because `moment` is in maintenance mode and we have to switch to some other solution, we can either go with native way or make some plugin patter that a user can provide the time lib via e.g., generator, which is a nice architecture but costs a bit from development perspective.

```
while (true){
    if( canWeWaitForTemporalAndStillUseMoment() ){
        if( isTemporalReady() ){
            UseTemporal();
        }
    } else {
        if( letUserDecideWhichLib() ){
            implementPluginPattern();
        } else {
            implementNativeWay();
        }
    }
}
```

marikaner:
> * I think we all agree `Temporal` should be the destination we want to go in the future.
> * Before that I personally would avoid switching from `moment`-> `lib abc` -> `Temporal`, before that's like `sdk 1.0` -> `sdk 2.0` -> `sdk 3.0` that is not user friendly, unless you have to wait for long time for `Temporal`, which you never know.
> * In order to keep us save as plan B, because `moment` is in maintenance mode and we have to switch to some other solution, we can either go with native way or make some plugin patter that a user can provide the time lib via e.g., generator, which is a nice architecture but costs a bit from development perspective.
> 
> ```
> while (true){
>     if( canWeWaitForTemporalAndStillUseMoment() ){
>         if( isTemporalReady() ){
>             UseTemporal();
>         }
>     } else {
>         if( letUserDecideWhichLib() ){
>             implementPluginPattern();
>         } else {
>             implementNativeWay();
>         }
>     }
> }
> ```

I was considering something in that direction as well. Something like custom serializers and deserializers (but then also payload serializers and deserializers, because they are based on each other). This might be overkill, though. Let's make this part of the discussion.

marikaner:
I added the notes from our discussion. Please review again.

knowledge-base/architectural-decision-records/date-time-handling.md
```diff
@@ -0,0 +1,58 @@
+# Date and Time handling in the SDK
+
+## Status
+
+decided
+
+## Context
+
+Currently we are using moment.js for handling date and time related types in the SDK. Unfortunately moment is quite big in size and the project itself recommends to not use it for new projects: https://momentjs.com/docs/#/-project-status/recommendations/.
```
florian-richter, line 9:
```suggestion
Currently we are using moment.js for handling date and time related types in the SDK. 
Unfortunately moment is quite big in size and the project itself recommends to not use it for new projects: https://momentjs.com/docs/#/-project-status/recommendations/.
```

knowledge-base/architectural-decision-records/date-time-handling.md
```diff
@@ -0,0 +1,58 @@
...
+* `Edm.Time` (v2) -> `Temporal.Time`
+* `Edm.Date` (v4) -> `Temporal.Date`
+* `Edm.Duration` (v4) -> `Temporal.Duration`
+* `Edm.TimeOfDay` (v4) -> `Temporal.Time`
+
+This is a Stage 2 proposal at the time being and will most likely reach stage 3 in 2020 still, there is no official timeline as to when stage 4 will be reached and when this is planned to become a standard.
+
+## Decision
+
+Allow to configure different serializers / deserializers for dates, but maybe even for all field types. We will provide adapters for moment, string and Temporal serialization and swich the default implementation from moment to either string or Temporal depending on the state of the proposal at the time of switching.
```
florian-richter, line 24:
```suggestion
Allow to configure different serializers / deserializers for dates, but maybe even for all field types. 
We will provide adapters for moment, string and Temporal serialization and switch the default implementation from moment to either string or Temporal depending on the state of the proposal at the time of switching.
```

knowledge-base/architectural-decision-records/date-time-handling.md
```diff
@@ -0,0 +1,58 @@
...
+  "Edm.DateTime": temporalMiddleware.dateTime
+}).execute(destination)
+
+bupa.date // Temporal
+```
+
+
+## Consequences
+
+This will allow users to easily switch implementation to what they see best fit for their use case and it will even allow them to write their own transformers. We have to consider, that at the moment `moment` does not handle times for example, while temporal does, therefore transformation needs to be handled separately for those cases. Also, url and payload serialization differ, while they are quite similar, therefore this approach needs to always consider two transformers for url and payload serialization.
```
florian-richter, line 44:
```suggestion
This will allow users to easily switch implementation to what they see best fit for their use case and it will even allow them to write their own transformers. 
We have to consider, that at the moment `moment` does not handle times for example, while temporal does, therefore transformation needs to be handled separately for those cases. 
Also, url and payload serialization differ, while they are quite similar, therefore this approach needs to always consider two transformers for url and payload serialization.
```

knowledge-base/architectural-decision-records/date-time-handling.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Alternative solutions
+
+### Migrate to a different library first
+Declined because this would mean that we introduce breaking changes twice.
+
+### Accept stage 3 proposal risk
+Declined as we think it is too risky and potentially too early for users to adopt it through the SDK. Stage 3 proposals have been downgraded to stage 2 in the past.
```
florian-richter, line 52:
```suggestion
Declined as we think it is too risky and potentially too early for users to adopt it through the SDK. 
Stage 3 proposals have been downgraded to stage 2 in the past.
```

knowledge-base/architectural-decision-records/date-time-handling.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Wait for Temporal to reach stage 4
+Declined, as we are planning to release a version 2 of the SDK in the first half of 2021 and it is unlikely this would fit our timeplan unless we keep the old implementation until a version 3.
```
florian-richter, line 55:
```suggestion
Declined, as we are planning to release a version 2 of the SDK in the first half of 2021 and it is unlikely this would fit our timeline unless we keep the old implementation until a version 3.
```

florian-richter (changes requested):
I would recommend to keep one sentence per line. Also some typo fixes.

marikaner:
@FrankEssenberger @jjtang1985 @florian-richter Feel free to review again.

jjtang1985 (approved):
LGTM, I think the decision is clear but the implementation details are not clarified and you just provided some example, right?

marikaner:
> LGTM, I think the decision is clear but the implementation details are not clarified and you just provided some example, right?

Correct, quote: "Exemplary code snippet (API not yet decided)"

PR #651, opened 2020-10-30
feature: Var Args and Array

FrankEssenberger:
## Context

Make the main public method lenient for usage with either arrays or variable args:
```
and(a,b) //was working already
and([a,b]) //is now also ok
``` 

I did not adjust the `batch` and `changeSet` methods because the current TS morph implementation does not allow for method overloading. This will change after we switch to the string template approach.

## Definition of Done

- [x] Release notes updated.


CHANGELOG.md
```diff
@@ -18,7 +18,7 @@
 
 ## New Functionality
 
--
+- Most public methods using variable arguments allow now also arrays as input.
```
jjtang1985, line 21:
```suggestion
- Most public methods using variable arguments now also allow arrays as input.
```

marikaner (commented):
I think we should include the changeset and batch still, but other than that it looks better than I expected.

packages/util/src/array.ts
```diff
@@ -49,3 +49,29 @@ export function splitInChunks<T>(arr: T[], chunkSize: number): T[][] {
...
+ * This wrapper methods makes it easy build an array from the input.
+ * @param firstOrArray - Either an array or the first element of the var args.
+ * @param rest - Second to last element if var args were used, undefined otherwise.
+ * @returns Array from the input or undefined if no var args was given.
+ */
+export function variableArgumentToArray<T>(
+  firstOrArray: undefined | T | T[],
+  rest: T[]
+): T[] {
+  if (typeof firstOrArray === 'undefined') {
```
jjtang1985, line 70: by reading the implementation, it means, 
- we ignore the `rest` if the `firstOrArray` is an array, which might be ok
- we do not accept `undefined` used as an element of varargs/array, which need to be documented at least

FrankEssenberger:
Added array input also to  changeset and batch

marikaner (approved):
LGTM. Maybe we should add a type test for one of those?

packages/util/src/array.ts
```diff
@@ -49,3 +49,29 @@ export function splitInChunks<T>(arr: T[], chunkSize: number): T[][] {
...
+
+/**
+ * We want to provide methods which accept a variable single number of elements and arrays.
+ * The overloaded signature to achieve this is:
+ * function doSomething(array:T[])
+ * function doSomething(...varArgs:T[])
+ * functiondoSomething(first:undefined|T|T[],...rest:T[]){
+ *   //implementation
+ * }
+ * This wrapper methods makes it easy build an array from the input.
```
marikaner, line 61:
```suggestion
 * function doSomething(first: undefined | T | T [], ...rest: T[]) {
 *   // implementation
 * }
 * This wrapper methods makes it easy to build an array from the input.
```

packages/util/src/array.ts
```diff
@@ -49,3 +49,29 @@ export function splitInChunks<T>(arr: T[], chunkSize: number): T[][] {
...
+  rest: T[]
+): T[] {
+  if (Array.isArray(firstOrArray)) {
+    return [...firstOrArray, ...rest];
+  }
+  return firstOrArray ? [firstOrArray, ...rest] : [...rest];
+}
+
+type Rest<T> = T extends T[] ? undefined : T[];
+type First<T> = undefined | T | T[];
```
marikaner, line 77: Are these types used anywhere?

packages/util/src/array.ts
```diff
@@ -49,3 +49,29 @@ export function splitInChunks<T>(arr: T[], chunkSize: number): T[][] {
...
+ * function doSomething(...varArgs:T[])
+ * functiondoSomething(first:undefined|T|T[],...rest:T[]){
+ *   //implementation
+ * }
+ * This wrapper methods makes it easy build an array from the input.
+ * @param firstOrArray - Either an array, the first element of the var args or undefined if no argument was given.
+ * @param rest - Second to last element if var args were used, empty array if the frist argument is an array.
+ * @returns Array from the input or empty array if no input was given.
+ */
+export function variableArgumentToArray<T>(
```
marikaner, line 66: Shouldn't it be `variadic` not `variable`?

packages/util/src/array.ts
FrankEssenberger: no I tried something there but my type kung fu was not strong enough and it would also not more understandable.

packages/util/test/array.spec.ts
```diff
@@ -1,7 +1,35 @@
...
+  unique,
+  last,
+  first,
+  splitInChunks,
+  variableArgumentToArray
+} from '../src';
 
 describe('array', () => {
   describe('flat', () => {
+    it('wraps variable arguments', () => {
```
marikaner, line 12: Why is this in flat?

packages/util/src/array.ts
FrankEssenberger: ?

FrankEssenberger: https://en.wikipedia.org/wiki/Variadic_function#:~:text=In%20mathematics%20and%20in%20computer,dating%20back%20to%201936%2D1937.

FrankEssenberger: was not aware of this word. Hm I am not sure is this common knowledge?

marikaner (commented):
By browsing through I found some minor things that could also be looked at.

packages/util/src/array.ts
marikaner: I only know it under this word (varargs) and rest parameters.

FrankEssenberger: but is is rather speaking so I renamed.

packages/util/test/array.spec.ts
FrankEssenberger: by mistake. I moved it to its own section.

PR #662, opened 2020-11-03
chore: Generate less code (REST)

marikaner:
## Context

In the review of the previous generator pr, we discussed a more generic approach that should be implemented on a separated PR. This is it. I replaced the generated additional classed by generics and transformed the api request builder class to an object.

## Definition of Done

- [ ] Release notes updated.


packages/core/src/rest/rest-request-builder.ts
```diff
@@ -1,9 +1,24 @@
-import { AxiosRequestConfig } from 'axios';
 import { Destination, DestinationNameAndJwt } from '../scp-cf';
 import { buildAxiosRequestConfig } from '../http-client';
 
-export abstract class RestRequestBuilder {
-  public customHeaders: Record<string, string> = {};
+type FunctionReturnType<ApiT, FnT extends keyof ApiT> = ApiT[FnT] extends (
```
FrankEssenberger, line 4: Very impressive what is possible with the conditional type. Please add some comment what is done here:
basic open ApiT is the class e.g. `PriceCalculationApi` which has multiple functions e.g. `calculateViaRestWithTenant` represented by the two types `ApiT` and `FnT`. This ensures that on new RestRequestBuilder<PriceCalculationApi>(    PriceCalculationApi,
    'calculateViaRestWithTenant') you need to pass a key as second argument.

FrankEssenberger: Request: Could you find or add a second method returning a union type e.g.
AxiosPromise<number|string>. I think the type magic should work but I would like to see a test.

packages/core/src/rest/api-request-builder.ts
```diff
@@ -0,0 +1,14 @@
+import { RestRequestBuilder } from './rest-request-builder';
+
+export type ParametersType<
```
FrankEssenberger, line 3: Where are these types used. I do not see a reference

packages/core/src/rest/rest-request-builder.ts
marikaner: Yes, I will add some documentation around this. For the second part I will see what I can do. I guess I can only achieve this by providing a custom service spec file, which I would rather want to do on a separate PR (which can be merged before this one).

packages/core/src/rest/api-request-builder.ts
marikaner: Yeah, at least the second one might not be necessary, I'll check.

packages/core/src/rest/rest-request-builder.ts
FrankEssenberger: Yes I think a custom spec file with all facets we want to check makes sense. A seperate PR for this makes also sense.

marikaner: Hey @FrankEssenberger 

I found a couple of problems in the openapi generator (not our additions, but the java thing below). Let's discuss them in the refinement. I documented everything, but would like to refrain from adding tests on this PR, unless you have anything more realistic related to my changes in mind ;) I added a note about you union type idea to the already existing ticket on testing this.

packages/core/src/rest/api-request-builder.ts
marikaner: I removed the second one and moved the structures a bit to separate better.

packages/rest-generator/src/request-builder/file.ts
```diff
@@ -8,56 +7,24 @@ import { toPascalCase } from '@sap-cloud-sdk/core';
...
 import { coreImportDeclaration } from '../utils';
-import { apiRequestBuilderClass } from './api-request-builder-class';
-import { operationRequestBuilderClass } from './operation-request-builder-class';
+import { operationsVariable } from './operations';
 
-/**
- * Used by the generator for generating the request builder source file.
- * @param serviceMetadata The service metadata model converted from the open api file.
- * @returns source file structure of the request builder file
- */
```
FrankEssenberger, line 18: why you removed the docs?

FrankEssenberger:
Question: All the test are commented out and also the code generation is not present anymore. So it is hard to tell for me what the generator actually does without the result beeing checked in.

FrankEssenberger (approved):
LGTM as I understood the tests and sample service will come back and have been removed due to IP issues.

PR #677, opened 2020-11-05
chore: New subdirectories

FrankEssenberger:
## Context

As a preparation for v2 of the SDK we already create a folder structure under core matching.
I tried to do the following:
- imports within on of the new folders use relative paths
- imports from outside of the folder use the root path of the folder
- I needed to remove the export of the `legacy-request-configs` because I also needed an export from the `odata-common`. 

I think we should merge this ASAP even if I did a few things not correctly because the conflict potential is so high.

## Definition of Done


packages/core/src/connectivity/scp-cf/destination/destination-selection-strategies.ts
```diff
@@ -1,9 +1,9 @@
 import { createLogger } from '@sap-cloud-sdk/util';
-import {
+import type {
   AllDestinations,
   DestinationsByType
 } from './destination-accessor-types';
-import { Destination } from './destination-service-types';
+import type { Destination } from './destination-service-types';
```
jjtang1985, line 6: 👍 
Do you see some warnings from the IDE or some command?

packages/core/src/connectivity/scp-cf/index.ts
```diff
@@ -10,3 +10,10 @@ export * from './resilience-options';
 export * from './token-accessor';
 export * from './xsuaa-service';
 export * from './xsuaa-service-types';
+export * from './tenant';
+export {
+  RegisteredJWTClaimsUser,
+  Scope,
+  mapping as mappingUserFields,
+  userFromJwt
+} from './user';
```
jjtang1985, line 19: Please update the release notes if they are missing.

packages/core/src/connectivity/scp-cf/destination/destination-selection-strategies.ts
FrankEssenberger: yes my IDE makes a red underline under the type because it seems not to understand it. But it compiles fine.

jjtang1985: Then another question, why you know you can optimise it by only importing the type, if your IDE cannot understand and suggest the improvement.

packages/core/src/odata-common/request-builder/batch/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './batch-change-set';
```
jjtang1985, line 1: Now we export more files, because it's necessary after slitting to multiple packages in v2.0.
I guess we can detect the potential name conflicts (e.g. same class names in common/v2/v4 folders) via tests.

packages/core/src/odata-common/request/odata-request.ts
```diff
@@ -1,22 +1,22 @@
...
   getHeaders,
   getHeaderValue,
+  buildCsrfHeaders,
+  buildHeadersForDestination,
   mergeHeaders
-} from '../../../header-builder';
+} from '../../header-builder';
 // TODO: The buildCsrfHeaders import cannot be combined with the rest of the other headers due to circular dependencies
-import { buildCsrfHeaders } from '../../../header-builder/csrf-token-header';
 // TODO: The buildHeadersForDestination import cannot be combined with the rest of the other headers due to circular dependencies
```
jjtang1985, line 19: The comments can be deleted?

packages/core/src/odata-v2/index.ts
```diff
@@ -9,6 +9,6 @@ export * from './payload-value-converter';
 export * from './request-builder';
 export * from './selectable/custom-field';
 export * from './uri-conversion';
-export * from './legacy-request-configs';
+// export * from './legacy-request-configs';
```
jjtang1985, line 12: I don't remember the reason why these are legacy files.
If we decide to export the same classes from the common instead of the legacy folder, we might want to double check the compatibility, because we don't have confident test coverage for legacy code.
I would maybe at least include some updates in the compatibility notes.

jjtang1985: I guess you also want to delete this line?

jjtang1985 (changes requested):
Looks good, but for renaming and import, it's hard to check all the cases, we have to trust the tests.
I would update the release notes for the `tenant` related changes and the potential compatibility notes.

packages/core/src/connectivity/scp-cf/destination/destination-selection-strategies.ts
FrankEssenberger: I did not buy the latest webstorm version so perhaps this is the reason. I agree that you could have a lint rule that checks if you use a import just as a `type` and adjust then the import.

packages/core/src/connectivity/scp-cf/index.ts
FrankEssenberger: yes

packages/core/src/odata-common/request/odata-request.ts
FrankEssenberger: yes

packages/core/src/odata-v2/index.ts
FrankEssenberger: I exported them with a different name and added a compatibility  note.

FrankEssenberger:
@jjtang1985 @marikaner I have the problem on this branch that the jest test runner is not exiting anymore. Can you try to reproduce the issue?

jjtang1985:
> I have the problem on this branch that the jest test runner is not exiting anymore. Can you try to reproduce the issue?
I can execute the integration tests.
Have you tried to use terminal outside of the IDE?
Also, is it a package specific issue?
For me, I did not run all tests, which tests are you talking about?

FrankEssenberger:
> > I have the problem on this branch that the jest test runner is not exiting anymore. Can you try to reproduce the issue?
> 
> I can execute the integration tests.
> Have you tried to use terminal outside of the IDE?
> Also, is it a package specific issue?
> For me, I did not run all tests, which tests are you talking about?

`yarn test:unit` for example. I also see the macOS node12 test running for 15min in the pipeline

packages/core/src/odata-common/request-builder/batch/index.ts
FrankEssenberger: for me the compiler already complains if there are conflicting exports.

FrankEssenberger:
So I tested and the yarn lock revert solved the issues. Do there is some thing sleeping for us...

jjtang1985 (approved):

jjtang1985:
So yarn lock might be relevant to the jest issue, but at least we know the force exit flag is helpful.

packages/core/src/connectivity/scp-cf/connectivity-service-types.ts
```diff
@@ -1,4 +1,4 @@
-import { Protocol } from '../protocol';
```
marikaner, line 1: I guess we should remove the `scp-cf` directory?

packages/core/src/header-builder/authorization-header.ts
```diff
@@ -5,9 +5,8 @@ import {
   DestinationAuthToken,
   getOAuth2ClientCredentialsToken,
   sanitizeDestination
-} from '../scp-cf';
```
marikaner, line 8: Don't the files in `header-builder` belong into one of the other packages?

marikaner (changes requested):
Is it possible to add a lintrule that does not allow submodule imports between those directories? To make sure we got them all?

FrankEssenberger:
> Is it possible to add a lintrule that does not allow submodule imports between those directories? To make sure we got them all?

such a rule 

> Is it possible to add a lintrule that does not allow submodule imports between those directories? To make sure we got them all?

We have an issue for the lint improvement in the backlog already. I added this case.

packages/core/src/header-builder/authorization-header.ts
FrankEssenberger: yes I move them around as I think they belong to.

marikaner (commented):
Please cross check with the [ADR](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/architectural-decisions-records/sdk-2.0.md). The src directory should only contain the listed directories (+rest, I guess) and an index.ts. Currently it still contains more. Consider whether util and the other files on root level can be moved into one of the packages.

marikaner (approved):
LGTM

PR #681, opened 2020-11-06
chore: Prepare move of the SDK

marikaner:
Adjust scripts, templates and urls for the move.


packages/generator/test/test-util/readme-util.ts
```diff
@@ -1,7 +1,7 @@
 export function getExpectedHelpfulLinks(): string {
   return `### Helpful Links
 
-- [SAP Cloud SDK](https://github.com/SAP/cloud-sdk)
+- [SAP Cloud SDK](https://github.com/SAP/cloud-sdk-js)
```
jjtang1985, line 4: All the vdm readme.md files have this line with different versions. you can also see it on the [npm](https://www.npmjs.com/package/@sap/cloud-sdk-vdm-sales-order-service)
If we do not release a new vdm version, the link will always be wrong on the npm.
Since this link will be used for documentation, you'll not see a 404, but redirect is also not possible.

jjtang1985 (approved):
Looks good for moving.
We have to consider the link of the readme.md

packages/generator/test/test-util/readme-util.ts
marikaner: True, the readme will show an info about that, but we should make another release soon probably.

PR #684, opened 2020-11-06
chore: Refactor header util

marikaner:
Move header util to util. I found that a previous PR made a breaking change, that is why I added the old header-util back. We should check whether there is more


FrankEssenberger:
> Move header util to util. I found that a previous PR made a breaking change, that is why I added the old header-util back. We should check whether there is more

I guess I moved the header-util to the util package. I overlooked in the many refactorings I made in the PR that only moves inside the core package are non breaking. I have a look at the changes of the PR and try to find additional things.

BTW besides the breaking change issue it makes sense IMO that the header util goes to the util packages because it contains methods used by OData, connectivity and http-client I guess.

packages/core/src/connectivity/scp-cf/csrf-token-header.ts
```diff
@@ -24,22 +24,24 @@ const logger = createLogger({
...
 function makeCsrfRequest<T extends HttpRequestConfig>(
   destination: Destination | DestinationNameAndJwt,
   requestConfig: Partial<T>
 ): Promise<Record<string, any>> {
-  const fetchHeader = !getHeaderValue(
-    'x-csrf-token',
-    requestConfig.headers
+  const fetchHeader = !pickValueIgnoreCase(
+    requestConfig.headers,
+    'x-csrf-token'
```
FrankEssenberger, line 44: should we perhaps have a const class with all the keys, or perhaps there is already a lib for that?

packages/core/src/odata-common/request/odata-request.ts
```diff
@@ -211,16 +210,16 @@ export class ODataRequest<RequestConfigT extends ODataRequestConfig> {
...
+    if (pickValueIgnoreCase(this.customHeaders(), 'if-match')) {
+      return pickIgnoreCase(this.customHeaders(), 'if-match');
     }
     const eTag = isWithETag(this.config)
       ? this.config.versionIdentifierIgnored
         ? '*'
         : this.config.eTag
       : undefined;
-    return filterNullishValues({ 'if-match': eTag });
+    return pickNonNullishIgnoreCase({ 'if-match': eTag });
```
FrankEssenberger, line 222: see comment above

packages/util/src/object.ts
```diff
@@ -108,24 +110,103 @@ export const assoc = <T>(key: string, value: any, obj: T) => ({
...
-  while (stack.length > 0) {
-    const current = stack.pop();
-    if (!Array.isArray(current)) {
-      flatResult.push(current);
-    } else {
-      stack.push(...current);
-    }
-  }
+/**
+ * Pick a partial object from the given object indepdendent of casing - if available.
```
FrankEssenberger, line 126: mention for all the pick methods that a reference on the found object is returned. I think pick is descriptive there but does not hurt to be extra explicit in the docs.

packages/util/src/object.ts
```diff
@@ -108,24 +110,103 @@ export const assoc = <T>(key: string, value: any, obj: T) => ({
...
+ * @returns - An object containing all keys from the `left` object, where headers present in the `right` object are replaced. Note that the casing used by `right` will be used.
+ */
+export function mergeLeftIgnoreCase<
+  LeftT extends Record<string, any>,
+  RightT extends Record<string, any>
+>(
+  left: LeftT = {} as LeftT,
+  right: RightT = {} as RightT
+): Record<string, any> {
+  return Object.entries(left)
```
FrankEssenberger, line 186: This does the right thing but I find it hard to understand. I would have expect: 
- Init the result with the right (because right wins)
- Loop over keys of left and only add the left key value if the pick does not find something
- Perhaps you can even create an inverse to the `pick` like unpick(keys) so that with a single call you get the stuff left in left and can add it.
- Also to call the single key method so often is highly inefficient because in each of them you loop over all keys.

packages/util/src/object.ts
```diff
@@ -108,24 +110,103 @@ export const assoc = <T>(key: string, value: any, obj: T) => ({
...
+  return Object.entries(obj)
+    .filter(([key, value]) => !isNullish(key) && !isNullish(value))
+    .reduce((filtered, [key, value]) => ({ ...filtered, [key]: value }), {});
+}
+
+/**
+ * Create an object by merging the `right` object into the `left` object ignoring casing, but keeping the right casing. Only keys present in the left object will be present in the merged object.
+ * @param left - Object to merge into. They keys of this object will be present in the returned object.
+ * @param right - Object to merge. Only keys in `left` will be considered for merging.
+ * @returns - An object containing all keys from the `left` object, where headers present in the `right` object are replaced. Note that the casing used by `right` will be used.
```
FrankEssenberger, line 177: mention for all merge methods that the resulting object contains shallow copies of the merged objects.

FrankEssenberger (changes requested):
Only minor suggestions e.g. some const for the header strings and a sugguestion for improving the merge implementation.

packages/core/src/connectivity/scp-cf/csrf-token-header.ts
marikaner: Not that I am aware of. Might have a nice touch to it (maybe not necessarily a class ;) ), but from the top of my head I don't see a big benefit. If you still think this should be done maybe create a ticket and let's discuss in the refinement?

packages/util/src/object.ts
marikaner: Done.

packages/util/src/object.ts
marikaner: Done.

packages/core/src/odata-common/request/odata-request.ts
marikaner: Let's discuss in the comment above.

packages/util/src/object.ts
marikaner: The steps you describe would not lead to the correct result, because consider the following case:
left: `{ a: 1, b: 2 }`, right: `{ A: 3, C: 4 }` => expected result `{ A: 3, b: 2, C: 4 }`.

I currently don't see a more efficient way than to:
1. Iterate over keys of left.
2. If right has the key (ignoring case), add it.
3. If not, add the one from left.

So, I could put everything into the reduce directly. When writing it I thought it might be a little simpler to split it into a map reduce.

marikaner:
> Only minor suggestions e.g. some const for the header strings and a sugguestion for improving the merge implementation.

Please check again, I adjusted the documentation, but have some doubts on the suggestions.

packages/util/src/object.ts
FrankEssenberger: init result: `{ A: 3, C: 4 }`
result pickInverse(keysRight,left) is `{b:2}`
merge the two leads the expected result.

You call the 
```
pickValueIgnoreCase(right, key)  //calls pickIgnoreCase
        ? pickIgnoreCase(right, key)
```

so you actually the method twice. But I think it is pre mature optimization and we can adjust the implementation if at some point in time we hear from customers.

FrankEssenberger (approved):

packages/util/src/object.ts
marikaner: Sorry, the example I gave was wrong. 🤦‍♀️ In that case you are right it would be easier the way you are proposing.
The correct example is:
left: `{ a: 1, b: 2 }`, right: `{ A: 3, C: 4 }` => expected result `{ A: 3, b: 2 }`

PR #687, opened 2020-11-09
chore: Use literal string templates for generation

marikaner:
## Context

This is the preliminary result of my findings with regards to replacing ts-morph with string templates. I tried this against the local test vdm and the s4 vdm. In this POC I am only generating entities (which are the most complex file that we generate I believe). The code could probably be improved quite a bit, the purpose here it to show what it is about and how it compares to the current approach.

In my opinion the template approach is better readable, testable and more convenient. The (non scientific) measurements prove that it is much faster too (10x). It would be interesting to know whether the results are similar for you. To run both versions execute:
```shell
# template approach
$ ts-node generate-templates.ts t > templates.log

# ts-morph approach
$ ts-node generate-templates.ts m > ts-morph.log
``` 

I would like to consider this for v2 (and potentially start #before).

## Definition of Done

- [ ] Release notes updated.


packages/generator/src/edmx-to-vdm/edmx-to-vdm-util.ts
```diff
@@ -88,11 +88,11 @@ export function isEnumType(name: string, enumTypes: VdmEnumType[]): boolean {
...
-    logger.warn(
-      `"CollectionKind" attribute found in the "${property.Name}" property. Currently, handling collection of properties is not supported by the generator.`
-    );
-  }
+  // if (property.hasOwnProperty('CollectionKind')) {
+  //   logger.warn(
+  //     `"CollectionKind" attribute found in the "${property.Name}" property. Currently, handling collection of properties is not supported by the generator.`
+  //   );
+  // }
 }
```
marikaner, line 96:
```suggestion
  if (property.hasOwnProperty('CollectionKind')) {
    logger.warn(
      `"CollectionKind" attribute found in the "${property.Name}" property. Currently, handling collection of properties is not supported by the generator.`
    );
  }
}
```

packages/generator/src/generator-utils.ts
```diff
@@ -123,9 +123,9 @@ export function getFallbackEdmTypeIfNeeded(
   if (edmType in edmToTsTypeMapping) {
     return edmType as EdmTypeShared<any>;
   }
-  logger.warn(
-    `The type ${edmType} is currently not supported by the sdk. Type "any" is used as fallback.`
-  );
+  // logger.warn(
+  //   `The type ${edmType} is currently not supported by the sdk. Type "any" is used as fallback.`
+  // );
```
marikaner, line 128:
```suggestion
  logger.warn(
    `The type ${edmType} is currently not supported by the sdk. Type "any" is used as fallback.`
  );
```

packages/generator/package.json
```diff
@@ -40,6 +40,7 @@
     "@sap/edm-converters": "^1.0.21",
     "@types/fs-extra": "^9.0.1",
     "@types/glob": "^7.1.2",
+    "common-tags": "^1.8.0",
```
marikaner, line 43:
```suggestion
```

packages/generator/package.json
```diff
@@ -50,6 +51,7 @@
     "yargs": "^16.0.3"
   },
   "devDependencies": {
+    "@types/common-tags": "^1.8.0",
```
marikaner, line 54:
```suggestion
```

yarn.lock
```diff
@@ -1779,6 +1779,11 @@
     "@types/node" "*"
     "@types/responselike" "*"
 
+"@types/common-tags@^1.8.0":
+  version "1.8.0"
+  resolved "https://registry.npmjs.org/@types/common-tags/-/common-tags-1.8.0.tgz#79d55e748d730b997be5b7fce4b74488d8b26a6b"
+  integrity sha512-htRqZr5qn8EzMelhX/Xmx142z218lLyGaeZ3YR8jlze4TATRU9huKKvuBmAJEW4LCC4pnY1N6JAm6p85fMHjhg==
+
```
marikaner, line 1786:
```suggestion
```

yarn.lock
```diff
@@ -3175,6 +3180,11 @@ commander@^5.1.0:
   resolved "https://registry.npmjs.org/commander/-/commander-5.1.0.tgz#46abbd1652f8e059bddaef99bbdcb2ad9cf179ae"
   integrity sha512-P0CysNDQ7rtVw4QIQtm+MRxV66vKFSvlsQvGYXZWR3qFU0jlMKHZZZgw8e+8DSah4UDKMqnknRDQz+xuQXQ/Zg==
 
+common-tags@^1.8.0:
+  version "1.8.0"
+  resolved "https://registry.npmjs.org/common-tags/-/common-tags-1.8.0.tgz#8e3153e542d4a39e9b10554434afaaf98956a937"
+  integrity sha512-6P6g0uetGpW/sdyUy/iQQCbFF0kWVMSIVSyYz7Zgjcgh8mgw8PQzDNZeyZ5DQ2gM7LBoZPHmnjz8rUthkBG5tw==
+
```
marikaner, line 3187:
```suggestion
```

jjtang1985:
Hi @marikaner , could we close this PR?

marikaner:
Yes, I think if we do this at some point it will have to be done anew ;)

PR #690, opened 2020-11-09
fix: Fix batch read response type

marikaner:
In my batch v4 refactoring a while back, I think I made a wrong assumption on a type. If the type remains as is, users would have to always cast the result from `.as(BusinessPartner) as BusinessPartner` otherwise it can be either an error or a business partner. Instead as should only be called, when we know it is not an error. I described this behavior in https://github.com/SAP/cloud-sdk/pull/11.

### DIFF
--- a/packages/core/src/odata-common/batch-response.ts
+++ b/packages/core/src/odata-common/batch-response.ts
@@ -17,7 +17,7 @@ export interface ReadResponse {
   httpCode: number;
   body: Record<string, any>;
   type: Constructable<EntityBase>;
-  as: <T extends EntityBase>(constructor: Constructable<T>) => Error | T[];
+  as: <T extends EntityBase>(constructor: Constructable<T>) => T[];
   isSuccess: () => boolean;
 }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/odata-common/batch-response.ts
```diff
@@ -17,7 +17,7 @@ export interface ReadResponse {
   httpCode: number;
   body: Record<string, any>;
   type: Constructable<EntityBase>;
-  as: <T extends EntityBase>(constructor: Constructable<T>) => Error | T[];
+  as: <T extends EntityBase>(constructor: Constructable<T>) => T[];
```
jjtang1985, line 20: If I remember correctly, the `Error | T[]` type is there from the beginning for some reasons.
In this way, at least release notes and the compatibility should be touched.

jjtang1985 (changes requested):
The improvement makes sense, but we should consider the release notes.

marikaner:
> The improvement makes sense, but we should consider the release notes.

Done.

jjtang1985 (approved):

PR #695, opened 2020-11-10
chore: Reduce code duplication create request builder

FrankEssenberger:
First PR for duplicated code reduction.
Started with the create request builder.


packages/core/src/odata-v2/request-builder/create-request-builder.ts
```diff
@@ -34,71 +26,14 @@ export class CreateRequestBuilderV2<EntityT extends EntityV2>
...
-
-  get entity(): EntityT {
-    return this._entity;
-  }
-
-  /**
-   * @deprecated Since v1.29.0. This method should never be called, it has severe side effects.   * Builds the payload of the query.
-   * @returns the builder itself
-   */
-  prepare(): this {
```
jjtang1985, line 52: This method is removed, which is fine since it was deprecated.
Please add one line to the compatibility notes.

jjtang1985 (changes requested):
LGTM.
Only one comment for the changelog.

packages/core/src/odata-v4/request-builder/create-request-builder.ts
```diff
@@ -90,17 +65,6 @@ export class CreateRequestBuilderV4<EntityT extends EntityV4>
...
-        deserializeEntityV4(
-          getSingleResult(response.data),
-          this._entityConstructor,
-          response.headers
-        )
-      )
-      .catch(error =>
-        Promise.reject(errorWithCause('Create request failed!', error))
-      );
+    return super.execute(destination, options);
```
marikaner, line 68: Oh no, why is `this.prepare()` still called here? Can you check whether it can be removed 🤦‍♀️

packages/core/src/odata-v2/request-builder/create-request-builder.ts
marikaner: Hmm, I think we should just keep it as it was public API. We should only remove it in version 2.

marikaner (dismissed):

packages/core/src/odata-v2/request-builder/create-request-builder.ts
FrankEssenberger: yes, my mistake. I put it back in.

packages/core/src/odata-v4/request-builder/create-request-builder.ts
FrankEssenberger: I see what the tests say if I remove it.

FrankEssenberger:
I put the prepare method to the base entity. So we have no breaking change. The tests also work without the prepare so I think we can remove it.

packages/core/src/odata-v4/request-builder/create-request-builder.ts
```diff
@@ -89,18 +52,6 @@ export class CreateRequestBuilderV4<EntityT extends EntityV4>
...
```
jjtang1985, line 55: The `execute()` method in this file can be removed.

jjtang1985 (changes requested):
Only needs to remove the `execute` method in the v4 version.

packages/core/src/odata-v4/request-builder/create-request-builder.ts
FrankEssenberger: ah good catch. Removed.

jjtang1985 (approved):

Issue #696, opened 2020-11-10
Cannot generate SAP Business One OData Client Library with SAP Cloud SDK


TrinidadMG:
**Describe the bug**
I have an error when trying to generate SAP Business One OData Client Library with SAP Cloud SDK

**To Reproduce**
I’m following the tutorial Generate Custom OData Client Library with SAP Cloud SDK's Generator for JavaScript: https://developers.sap.com/tutorials/cloudsdk-js-generator.html

1. Take the EDMX file attached here (changed extension to txt to be able to attach it here)
[B1API.edmx.txt](https://github.com/SAP/cloud-sdk-js/files/5517635/B1API.edmx.txt)
2. Create a folder called service-spefication
3. Copy the attached EDMX file into that folder
4. Execute generate-odata-client -i service-specification/ -o odata-client/
5. Error shown in the screen capture
<img width="918" alt="B1CloudSDKError" src="https://user-images.githubusercontent.com/18447094/98680965-6ac51480-2362-11eb-854d-83f1e5f021d6.png">

- The SAP Business One Odata APIs are Odata v4 APIs.
A single EDMX file contains all the entities of the APIs, all in one.

**Used Versions:**
 - SAP Cloud SDK version you used as dependency
@sap-cloud-sdk/cli/0.1.12 win32-x64 node-v12.19.0

jjtang1985:
Hi @TrinidadMG , thank you for creating an issue here.
as mentioned in the previous email, I can reproduce the same error with the latest version of the generator and the edmx file provided. Therefore, from the initial investigation I guess it's a bug from our side.
Please wait for the update.

One question from my side, is this a blocker that affects your live system or maybe you are working on some PoC project?

gregorwolf:
Hi @jjtang1985,

[...]

PR #699, opened 2020-11-10
Fix: generator can handle enum type with one member

jjtang1985:
This PR will fix the issue reported [here](https://github.com/SAP/cloud-sdk-js/issues/696), where the generator fails when it detects a enum type with one element.


packages/generator/src/edmx-parser/v4/edmx-parser.ts
```diff
@@ -31,7 +31,11 @@ export function parseComplexTypes(root): EdmxComplexType[] {
 }
 
 export function parseEnumTypes(root): EdmxEnumType[] {
-  return forceArray(root.EnumType);
+  const types: EdmxEnumType[] = forceArray(root.EnumType);
+  return types.map(edmxEnumType => ({
+    Name: edmxEnumType.Name,
+    Member: forceArray(edmxEnumType.Member)
```
jjtang1985, line 37: When there is only one member, the type of the `Member` is an object instead of an array.

marikaner: That is essentially the fix, right? ;)

marikaner (approved):
LGTM

PR #700, opened 2020-11-11
chore: Use beta version of the open api generator

marikaner:
The current "stable" version of the OpenAPI generator is broken. The beta version seems to work fine, I was able to transpile the results.

### DIFF
--- a/packages/rest-generator/openapitools.json
+++ b/packages/rest-generator/openapitools.json
@@ -2,6 +2,6 @@
   "$schema": "node_modules/@openapitools/openapi-generator-cli/config.schema.json",
   "spaces": 2,
   "generator-cli": {
-    "version": "4.3.1"
+    "version": "5.0.0-beta2"
   }
 }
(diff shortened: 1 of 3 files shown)


packages/rest-generator/openapitools.json
```diff
@@ -2,6 +2,6 @@
   "$schema": "node_modules/@openapitools/openapi-generator-cli/config.schema.json",
   "spaces": 2,
   "generator-cli": {
-    "version": "4.3.1"
+    "version": "5.0.0-beta2"
```
jjtang1985, line 5: Question: the `^2.1.4` is used in the package.json, which is a stable version instead of beta. Did you manually modify this line or maybe 2.1.4 points to the beta version?

packages/rest-generator/package.json
```diff
@@ -30,11 +30,11 @@
     "test:local": "yarn jest",
     "test:debug": "node --inspect-brk ../../node_modules/.bin/jest --runInBand",
     "test:watch-debug": "node --inspect-brk ../../node_modules/.bin/jest --watch --runInBand",
-    "generate:test-services": "ts-node src/generator-cli.ts -i test/resources/PriceCalculation  -o test/test-services",
+    "generate:test-services": "ts-node src/generator-cli.ts -i test/resources/test-apis -o test/test-services",
```
jjtang1985, line 33: 👍

packages/rest-generator/openapitools.json
marikaner: 2.1.4 is the version of the CLI, but with the CLI you can configure the version of the generator (through this openapitools.json).

jjtang1985 (approved):

PR #701, opened 2020-11-11
chore: Rename directories and change service url

marikaner:
I renamed the directories for the odata test specification so that we can add rest as well. I also changed the URL of the test service to be `/test-service` instead of `/admin`.

This PR looks big, but it is mostly configuration + regeneration.


jjtang1985 (approved):
LGTM.

PR #706, opened 2020-11-12
chore: eslint import improvement

florian-richter:
This should check for some of the mentioned issues:
- Don't allow circular imports (`no-circular`)
- Enforce root level imports between our core submodules (src/<submodule>) (`import/no-internal-modules`)
- Enforce root level imports between our packages (`import/no-internal-modules`)

- Don't allow imports from index (didn't find a matching rule, but I haven't looked for too long yet)


packages/core/src/connectivity/scp-cf/protocol.ts
```diff
@@ -1,9 +1,10 @@
+/* eslint-disable-next-line */
 export enum Protocol {
   HTTP = 'http',
   HTTPS = 'https'
 }
 
-/* eslint-disable-next-line  @typescript-eslint/no-redeclare */
+/* eslint-disable-next-line */
```
marikaner, line 7: Is that better?

marikaner (approved):
LGTM except for the one comment.

marikaner (approved):
LGTM except for the one comment.

packages/core/src/connectivity/scp-cf/protocol.ts
florian-richter: A second rule now complains here because you shouldn't export 2 things under the same name. But this seemed to be intentional so I though it would be fine to disable eslint for the next row here. Not sure what the best approach is here...

marikaner: What is the other rule? Why wasn't it complaining before? And can this be configured to ignore namespaces?

PR #709, opened 2020-11-13
chore: Document API dicussions

marikaner:
This is the documentation of our API improvement ideas. Please extend and improve the document for the upcoming discussions.

### DIFF
--- a/knowledge-base/implementation-documentation/api-improvements.md
+++ b/knowledge-base/implementation-documentation/api-improvements.md
@@ -76,7 +76,8 @@ const buPa: BusinessPartner = await BusinessPartnerApi.getAll()
   .execute(destination);
 ```
 * Cleaner separation between responsibilities.
-* Importing three things is quite complex
+* Importing three things is quite complex. 
+This could be avoided by using the function style filtering as discussed below.
 * Could be implemented as a non-breaking change initially.
 * Optionally call BusinessPartner => BusinessPartnerEntity.
 * Realistic for 2.0, but only useful if we consider function style filtering (at any point).
--- a/knowledge-base/implementation-documentation/api-improvements.md
+++ b/knowledge-base/implementation-documentation/api-improvements.md
@@ -194,6 +194,7 @@ const buPa: BusinessPartner = await BusinessPartner.requestBuilder().getAll().ex
 ```
 
 ### Issues
+This currently does not allow to access the response (or request) object, which means not having information on the HTTP code, raw response and other information, that might be relevant.
 
 ### Alternatives
 
--- a/knowledge-base/implementation-documentation/api-improvements.md
+++ b/knowledge-base/implementation-documentation/api-improvements.md
@@ -2,8 +2,11 @@
 
 This document contains some ideas for API changes, that would be a breaking change and would result in a major version upgrade.
 
-
 ## Filtering root elements based on 1-1 relationships
+
+**Decision** (Version 2 candidate)
+* Leaning towards option 2
+* Feedback from users would be helpful
 
 ### Current API
 ```ts
@@ -20,7 +23,7 @@ The inner filter is quite confusing, because `filter` implies that we are filter
 
 ### Alternatives
 
-`filterBy`
+**1. `filterBy`**
 
 ```ts
 BusinessPartner.requestBuilder().getAll()
@@ -30,8 +33,12 @@ BusinessPartner.requestBuilder().getAll()
     )
   )
 ```
+Cons:
+* Might not really solve the issue:
+  * The difference in language `filter` versus `filterBy` is quite subtle and could still be confused.
+  * Potentially from the context this is not clear
 
-flatten filters
+**2. flatten filters**
 
 ```ts
 BusinessPartner.requestBuilder().getAll()
@@ -41,12 +48,16 @@ BusinessPartner.requestBuilder().getAll()
     any(BusinessPartner.TO_MULTI_LINK.CITY.equals('Berlin'))
   )
 ```
-Concerns:
+Cons:
 * Matter of taste - not clearly better.
 * Longer for "deep" paths.
 * Not straight forward to implement. (Further investigation needed.)
 
 ## Seperate Entity(Set) API and entity
+
+**Decision** (Version 2 decided)
+* Option 1 required
+* Option 2 optional, only if we do function style filtering
 
 ### Current API
 ```ts
@@ -54,11 +65,11 @@ const buPa: BusinessPartner = await BusinessPartner.requestBuilder().getAll().ex
 ```
 
 ### Issues
-The entity is a god-like element here. Separating could clean up.
+The entity is a god-like element here and might cause confusion. Separating could clean up.
 
 ### Alternatives
 
-Split entity and api
+**1. Split entity and api**
 ```ts
 const buPa: BusinessPartner = await BusinessPartnerApi.getAll()
   .filter(BusinessPartner.FIRST_NAME.equals('Peter'))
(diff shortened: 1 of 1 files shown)


knowledge-base/implementation-documentation/api-improvements.md
```diff
@@ -0,0 +1,209 @@
...
+
+```ts
+BusinessPartner.requestBuilder().getAll()
+  .filter(
+    BusinessPartner.TO_BUSINESS_PARTNER_ADDRESS.filterBy(
+      BusinessPartnerAddress.CITY.equals('Berlin')
+    )
+  )
+```
+
```
FrankEssenberger, line 33:
```suggestion
Concerns:
* The difference in language `filter` versus `filterBy` is too subtle for me. I would just overlook it. Also since it is not directly on the root entity like: `BusinessPartner.filterBy`
```

knowledge-base/implementation-documentation/api-improvements.md
```diff
@@ -0,0 +1,209 @@
...
+### Alternatives
+
+Split entity and api
+```ts
+const buPa: BusinessPartner = await BusinessPartnerApi.getAll()
+  .filter(BusinessPartner.FIRST_NAME.equals('Peter'))
+  .execute(destination);
+```
+* Could fix circular dependencies.
+* Could be implemented as a non-breaking change initially.
```
FrankEssenberger, line 68:
```suggestion
* Could be implemented as a non-breaking change initially by deprecating the old `BusinessPartner.requestBuilder()` methods and delegate to the new `BusinessPartnerApi`.
```

knowledge-base/implementation-documentation/api-improvements.md
```diff
@@ -0,0 +1,209 @@
...
+
+
+Split entity instance, api, and "helpers" (schema)
+```ts
+const buPa: BusinessPartner = await BusinessPartnerApi.getAll()
+  .filter(BusinessPartnerSchema.FIRST_NAME.equals('Peter'))
+  .execute(destination);
+```
+* Cleaner separation between responsibilities.
+* Importing three things is quite complex
```
FrankEssenberger, line 79:
```suggestion
* Importing three things is quite complex. 
This could be avoided by using the function style filtering as discussed below.
```

knowledge-base/implementation-documentation/api-improvements.md
```diff
@@ -0,0 +1,209 @@
...
+### Alternatives
+
+Using a function:
+```ts
+BusinessPartner.requestBuilder().getAll()
+  .filter(bupa => bupa.FIRST_NAME.equals('Peter'))
+```
+* We are only using static variables here. This is not clear anymore and might be more confusing than before.
+* Makes more sense when the schema is split from entity and api.
+* The examples below could be non-breaking changes done sequentially over time.
```
FrankEssenberger, line 105:
```suggestion
* The examples below could be non-breaking changes done sequentially over time.
We would keep the original way to filter, deprecate it and allow for the new function in addition.
```

knowledge-base/implementation-documentation/api-improvements.md
```diff
@@ -0,0 +1,209 @@
...
+    bupa.FIRST_NAME.equals('Peter')
+  ))
+```
+
+```ts
+BusinessPartner.requestBuilder().getAll()
+  .filter((bupa, { and, equals }) => and(
+    equals(bupa.FIRST_NAME, 'Peter')
+  ))
+```
```
FrankEssenberger, line 120:
```suggestion
```
As mentioned in the meeting I like the last approach because it would make all filter functions like `equals`, `startsWith`, `greater`,... symmetric.
```

File: knowledge-base/implementation-documentation/api-improvements.md [hunk 6]
```diff
@@ -0,0 +1,209 @@
...
+
```
FrankEssenberger, line 121:
```suggestion
As mentioned in the meeting I like the last approach because it would make all filter functions like equals, startsWith, greater,... symmetric.
```

knowledge-base/implementation-documentation/api-improvements.md
```diff
@@ -0,0 +1,209 @@
...
+```ts
+const [buPa, req, res]: [BusinessPartner, Request, Response] = await BusinessPartner.requestBuilder().getAll().execute(destination);
+```
+
+```ts
+const [buPa, req, res]: [BusinessPartner, Request, Response] = await BusinessPartner.requestBuilder().getAll().executeRaw(destination);
+const buPa: BusinessPartner = await BusinessPartner.requestBuilder().getAll().execute(destination);
+```
```
FrankEssenberger, line 209:
```suggestion
```
I (FE) find this options most appealing since it will allow for extra information but in the cases the users do not like it they do not have do use the deconstruction. We can potentially also deprecate the plain `execute` and push users in the direction of the `executeRaw` method. BTW the name `executeRaw` is of course not fitting to the return type `[buPa, req, res]`.  What do you think about a `showRequest` and `showResponse` method which will change the return type form a `BusinessPartner` to a [BusinessPartner,Res,Req]?
```

Comment (FrankEssenberger) [inline knowledge-base/implementation-documentation/api-improvements.md:209]:
```suggestion
```I (FE) find this options most appealing since it will allow for extra information but in the cases the users do not like it they do not have do use the deconstruction. We can potentially also deprecate the plain execute and push users in the direction of the executeRaw method. BTW the name executeRaw is of course not fitting to the return type [buPa, req, res]. What do you think about a showRequest and showResponse method which will change the return type form a BusinessPartner to a [BusinessPartner,Res,Req]?
```

knowledge-base/implementation-documentation/api-improvements.md
marikaner: Can you formulate that without using "me"? I disagree, that the difference in language is too subtle, but I agree, that it is not 100 % clear what is meant by that, because the context does not disclose that.

knowledge-base/implementation-documentation/api-improvements.md
marikaner: I think this can be seen as a pro for this approach, can you express it without "I" and commit it?

knowledge-base/implementation-documentation/api-improvements.md
```diff
@@ -0,0 +1,209 @@
...
+
+
+## Execute raw
+
+### Current API
+```ts
+const buPa: BusinessPartner = await BusinessPartner.requestBuilder().getAll().execute(destination);
+```
+
+### Issues
```
marikaner, line 196:
```suggestion
### Issues
This currently does not allow to access the response (or request) object, which means not having information on the HTTP code, raw response and other information, that might be relevant.
```

knowledge-base/implementation-documentation/api-improvements.md
florian-richter: We could also consider overloading `.execute()`.

```ts
function execute(destination: Destination): BusinessPartner
function execute(destination: Destination, includeReqRes: true): [BusinessPartner, Request, Response]
function execute(destination: Destination, includeReqRes?: true): BusinessPartner | [BusinessPartner, Request, Response] {
  // stuff
  if (includeReqRes) {
    return [BuPa, Req, Res];
  } else {
    return BusinessPartner;
  }
}
```

marikaner: @florian-richter I think we decided against the arrays in favor of objects. I like the second parameters instead of `executeRaw` although I would also allow it to be false:
```
function execute(destination: Destination, includeReqRes = false): BusinessPartner;
function execute(destination: Destination, includeReqRes: true): HttpResponse;
function execute(destination: Destination, includeReqRes?: boolean): BusinessPartner | HttpResponse
```

marikaner: (Also could you add it to the document directly?)

marikaner:
@FrankEssenberger @jjtang1985 @florian-richter, I would like to merge this at some point. There are some open points, please add them or resolve them and give this a thumbs up. You can still change this document later of course.

knowledge-base/implementation-documentation/api-improvements.md
FrankEssenberger, line 168:
```suggestion
The latter approach makes the filter function very symmetric.
```

knowledge-base/implementation-documentation/api-improvements.md
FrankEssenberger: We should be careful not to add to many booleans as arguments e.g. `doSomething(true,false,true)` is a bad signature. I would prefer a object is we have potentially multiple booleans:`execute(destination:Destination,options:OptionsType)` And in `OptionsType` contains the `includeReqRes`.

FrankEssenberger:
> @FrankEssenberger @jjtang1985 @florian-richter, I would like to merge this at some point. There are some open points, please add them or resolve them and give this a thumbs up. You can still change this document later of course.

I added two last comments. But it is ok for me already.

FrankEssenberger (approved):

knowledge-base/implementation-documentation/api-improvements.md
florian-richter: @marikaner I've added it. You can merge now if you want.

PR #710, opened 2020-11-13
chore: Reduce code duplication get request builder 

FrankEssenberger:
Adjustment for the get-all (only minor because already done partially) and get-by-key request builders.


jjtang1985 (approved):
LGTM

PR #711, opened 2020-11-13
chore: Provide REST service

marikaner:
This is a very simple example of an OData API. We should be able to use the given specification for all kinds of tests.
I also added a mocked REST service to the existing CAP service.


test-packages/e2e-tests/odata.js
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const cds = require('@sap/cds');
+const express = require('express');
+
+module.exports = {
+  createODataApp() {
```
FrankEssenberger, line 6: Is the name ODataApp good? I think we refer to this often as `CAP` but I am not sure what is a good name.

marikaner: I have no strong opinion here. I called it odata, because it is the express behind `/odata`.

test-resources/rest-service-specs/test-service.json
```diff
@@ -0,0 +1,209 @@
+{
```
FrankEssenberger, line 1: I do not see any generated entities from this definition that are then called in E2E to the backend you created.

test-packages/e2e-tests/rest.js
```diff
@@ -0,0 +1,71 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const OpenAPIBackend = require('openapi-backend').default;
```
FrankEssenberger, line 2: cool stuff. That you can create a backend from spec.

test-packages/e2e-tests/server.js
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const express = require('express');
+const { createODataApp } = require('./odata');
+const { createRestApp } = require('./rest');
+
+async function startServer() {
+  const app = express();
+  const odataApp = createODataApp();
```
FrankEssenberger, line 8: why are you putting both APIs (OData and rest) in the same application?

test-resources/rest-service-specs/test-service.json
FrankEssenberger: no it is there.

marikaner: This is because I wanted to do it on a separate PR. I was hoping for some feedback on the setup first. ;)

FrankEssenberger: ok.

test-packages/e2e-tests/rest.js
marikaner: Yeah, but it is less cool than it looks like on first sight. The data still needs to be provided manually

test-packages/e2e-tests/server.js
marikaner: Because the setup is much simpler with this than starting an extra server - maintenance and performance-wise. Other than that the apps do not interfere with one another.
Would you prefer to separate them?

test-packages/e2e-tests/rest.js
FrankEssenberger: ehhh

test-packages/e2e-tests/server.js
FrankEssenberger: ok

test-packages/e2e-tests/odata.js
FrankEssenberger: I would call it CapApp but as you like.

FrankEssenberger (approved):

jjtang1985 (approved):
LGTM, a minor comment.
Looking forward to the rest test.

test-packages/e2e-tests/odata.js
marikaner: Ok, done.

PR #714, opened 2020-11-16
feat: support multiple schema

jjtang1985:
Previously, the generator could handle single schema in edmx file well. When detecting 2 schemas, we merged them, which did not work as expected. The generator did not support more than 2 schemas.

This PR will merge all the schema and utilise the namespace information for referencing.
The backwards solution, i.e., ignoring the namespace prefix, is still kept to ensure non-breaking changes.

- [x] Release notes updated.
- [x] PR title adheres to [conventional commit guidelines](https://www.conventionalcommits.org).


packages/generator/src/edmx-parser/legacy/common/edmx-to-vdm.ts
```diff
@@ -110,7 +110,7 @@ export function transformEntity(
   classNames: Record<string, any>,
   complexTypes: VdmComplexType[],
   formatter: ServiceNameFormatter
-): Omit<VdmEntity, 'navigationProperties'> {
+): Omit<Omit<VdmEntity, 'navigationProperties'>, 'entityTypeNamespace'> {
```
FrankEssenberger, line 113: I think you can do something like: `Omit<VdmEntity, 'navigationProperties' | 'entityTypeNamespace'>`. Or it is `&` instead of `|`. For types the union is `|` but for object the `&` so I do not know which one to use here.

packages/generator/src/edmx-parser/common/edmx-types.ts
```diff
@@ -38,6 +42,8 @@ export interface EdmxEntitySetBase extends EdmxNamed {
   'sap:updatable': string;
 }
 
+export type EdmxEntitySetBaseNamespaced = EdmxEntitySetBase & EdmxNamespaced;
```
FrankEssenberger, line 45: General: We did not use the extension `named` for the object e.g. `EdmxEntitySetBase`. Most of them include the `named` interface and do now the `namespaced` as well. So I would just add then `named` also to the Base sets and not introduce the `Named` name extensions. We could add a note to the competability notes that most of the types include a new property now. Technically this is a breaking change but I  think we can take the risk.

FrankEssenberger: Since this is only internal representation of the service metadata.

packages/generator/src/edmx-parser/edmx-file-reader.ts
```diff
@@ -57,19 +148,73 @@ function getODataVersion(edmx): ODataVersion {
...
+    const ret = mergeSchema(schema);
+    return ret;
   }
-  return schema;
+  const ret2 = mergeSchema(forceArray(schema));
+  return ret2;
+}
+
+function mergeSchema(schema: EdmxMetadataSchema[]) {
+  // : EdmxMetadataSchemaV2Merged | EdmxMetadataSchemaV4Merged
```
FrankEssenberger, line 161: leftover. I guess the return type

packages/generator/src/edmx-parser/edmx-file-reader.ts
```diff
@@ -1,22 +1,112 @@
 import { PathLike, readFileSync } from 'fs';
 import path, { basename } from 'path';
 import { parse } from 'fast-xml-parser';
-import { createLogger, ODataVersion } from '@sap-cloud-sdk/util';
+import { assoc, createLogger, flat, ODataVersion } from '@sap-cloud-sdk/util';
 import { forceArray } from '../generator-utils';
 import { SwaggerMetadata } from '../swagger-parser/swagger-types';
+import {
```
FrankEssenberger, line 7: Some time ago I refactored the generator https://github.com/SAP/cloud-sdk-js/pull/257

One part was to get rid of the EDM types and just pass the root down to the VDM methods and do the parsing when it is needed. At this point you know in general if you are V2 or V4 and can make some assumptions on the types. I guess the new types are just super unions containing all possible things. But then the question is how usefull they are.

packages/generator/src/edmx-to-vdm/v4/entity.ts
```diff
@@ -44,7 +47,9 @@ export function generateEntitiesV4(
   enumTypes: VdmEnumType[],
   formatter: ServiceNameFormatter
 ): VdmEntity[] {
-  const entitySets = parseEntitySets(serviceMetadata.edmx.root);
+  const entitySets = parseEntitySets(
+    serviceMetadata.edmx.root as EdmxMetadataSchemaV4Merged
```
FrankEssenberger, line 51: such casts indicate in my opinion that the typing before is not really usefull.

packages/generator/src/edmx-parser/edmx-file-reader.ts
FrankEssenberger: The problem, I am not able to understand yet if the logiv from line 170 onwards needs to be done here or if you could do it also on the fly in the parse methods.

packages/generator/src/edmx-parser/edmx-file-reader.ts
```diff
@@ -57,19 +148,73 @@ function getODataVersion(edmx): ODataVersion {
...
-    );
+    const ret = mergeSchema(schema);
+    return ret;
   }
-  return schema;
+  const ret2 = mergeSchema(forceArray(schema));
+  return ret2;
+}
+
+function mergeSchema(schema: EdmxMetadataSchema[]) {
```
FrankEssenberger, line 160: I think this is more than reading. It is parsing and should be done in the edmx parser files.

packages/generator/src/edmx-parser/v4/edmx-parser.ts
```diff
@@ -38,24 +44,30 @@ export function parseEnumTypes(root): EdmxEnumType[] {
...
-export function parseEntitySets(root): EdmxEntitySet[] {
-  return forceArray(root.EntityContainer.EntitySet).map(entitySet => ({
+export function parseEntitySets(
+  root: EdmxMetadataSchemaV4Merged
+): EdmxEntitySet[] {
+  return flat(root.EntityContainer.map(ec => ec.EntitySet)).map(entitySet => ({
     ...entitySet,
-    NavigationPropertyBinding: forceArray(entitySet.NavigationPropertyBinding)
+    NavigationPropertyBinding: forceArray(
+      entitySet['NavigationPropertyBinding']
```
FrankEssenberger, line 60: why not entitySet.NavigationPropertyBinding I think you have typing here.

packages/generator/src/edmx-to-vdm/common/entity.ts
```diff
@@ -130,19 +131,26 @@ const propertyFieldType = (type: string): string | undefined =>
...
+  namespace: string[],
   swagger?: SwaggerMetadata
 ): JoinedEntityMetadata<EntitySetT, EntityTypeT>[] {
   return entitySets.map(entitySet => {
-    // We assume metadata files to have a maximum of two schemas currently
-    // So entitySet.EntityType.split('.').slice(-1)[0] that we will only find one matching entry (and thus never forget anything)
-    const entityType = entityTypes.find(
-      t => t.Name === entitySet.EntityType.split('.').slice(-1)[0]
+    let entityType = entityTypes.find(
+      t => `${t.Namespace}.${t.Name}` === entitySet.EntityType
```
FrankEssenberger, line 144: can you explain give an example under which conditions the `entityType` us undefined and the second `split` on the set finds something.

jjtang1985: Below is an example of multi-schema, where the entitySet and entityType are linked. After this PR, our generator should be able to handle all the valid edmx that provides proper namespace information (`Namespace1` in the example). 
However, this PR will also apply stricter rules for the edmx file, which means if the namespace information is not proper, we cannot match the `Namespace1.A_TestEntityType1`.
I don't want to introduce breaking changes, if some S/4 edmx files have such issues, providing valid edmx is not our responsibility though. 
I created a ticket for handling/discussing this. 
For the time being, keeping the backwards solution that use the old logic can save us from this situation.

```
<Schema Namespace="Namespace1" xmlns="http://docs.oasis-open.org/odata/ns/edm">
      <EntityType Name="A_TestEntityType1" sap:content-version="1">
...
<Schema Namespace="Namespace5" xmlns="http://docs.oasis-open.org/odata/ns/edm">
      <EntityContainer Name="Container">
        <EntitySet Name="A_TestEntity1" EntityType="Namespace1.A_TestEntityType1" />
...
```

packages/generator/src/edmx-parser/edmx-file-reader.ts
FrankEssenberger: I think I have to understand firt (lets discuss tomorrow) How stront the schema separates the types and so on. Is it like a virtual to service definitions? If so one could also pull the changes up in the generator loop for the services and generate multiple services for each namespace?

jjtang1985:
As discussed with @FrankEssenberger , here are the space for improvement:
- [ ] get rid of v2/v4 union type to avoid casting, type is only used with v2/v4 context
- [ ] ADR for the design of merge() and backwards solution with example.

CHANGELOG.md
```diff
@@ -16,10 +16,11 @@
 
 - Batch `ReadResponse.as` does not parse errors anymore, but throws an error if the response in fact was an `ErrorResponse`. To avoid this check `response.isSuccess()` before casting.
 - Rename deprecated ODataRequestConfigs e.g. `ODataCreateRequestConfig` -> `ODataCreateRequestConfigLegacy` due to naming conflicts.
+- [Generator] In order to support multi-schema generation, the edmx schema/namespace related types are adjusted.
```
marikaner, line 19:
```suggestion
- [Generator] In order to support multi-schema generation, the edmx schema/namespace related types have been extended.
```

marikaner (changes requested):
Phew many comments, but also many changes ;)
Let's have a discussion on some of the things once you take a look at my comments!

packages/generator/src/typedoc.ts
```diff
@@ -102,7 +103,10 @@ export function getEntityDescription(
   entity: VdmEntity,
   service: VdmServiceMetadata
 ): string {
-  let description = entityDescription(entity.entitySetName, service.namespace);
+  let description = entityDescription(
+    entity.entitySetName,
+    getServiceName(service)
```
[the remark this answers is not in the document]
jjtang1985: because speakingModuleName is also namespace related...

packages/generator/src/edmx-parser/edmx-file-reader.ts
```diff
@@ -1,20 +1,110 @@
...
+  EdmxEntitySetBaseNamespaced,
+  EdmxNamed
+} from './common';
 
 const logger = createLogger({
   package: 'generator',
   messageContext: 'edmx-edmx-parser'
 });
 
+export interface EdmxMetadataSchema {
```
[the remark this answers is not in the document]
jjtang1985: notes: deep merge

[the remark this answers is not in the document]
jjtang1985: I removed the types and deep merge works.

packages/generator/test/service-mapping.spec.ts
```diff
@@ -14,7 +14,7 @@ describe('service-mapping', () => {
         enumTypes: [],
         entities: [],
         functionImports: [],
-        namespace: 'fghjkl',
+        namespaces: ['fghjkl'],
```
marikaner, line 17: I know we had this here before, by why this name. If it's easy to fix consider replacing it with a better name ;)

packages/generator/src/edmx-parser/common/edmx-parser.ts
```diff
@@ -19,10 +21,18 @@ export function parseEntityTypesBase(root): EdmxEntityTypeBase<any>[] {
...
 export function parseEntitySetsBase(root): EdmxEntitySetBase[] {
-  return forceArray(root.EntityContainer.EntitySet);
+  return extractPropertiesFromEntityContainer(root, ec => ec.EntitySet);
+}
+
+export function extractPropertiesFromEntityContainer(
+  root,
+  entityContainerToPropFn
+): any[] {
+  return flat(root.EntityContainer.map(entityContainerToPropFn));
```
marikaner, line 37: * Why `extract` not `get`?
* Why properties not property? It is always getting one property, right? (I guess it could do more, but it is never used like this)
* Consider whether passing a string as a second parameter makes it simpler and concentrates the "complexity" in this one place. That way you don't need to know the implementation of this function from the outside.
* Don't we need another forceArray around the property? I think `flat` will fail if it does not get a twodimensional array as input
```suggestion
export function getPropertyFromEntityContainer(
  root,
  entityContainerProperty: string
): any[] {
  return flat(root.EntityContainer.map(ec => forceArray(ec[entityContainerProperty])));
```

packages/generator/src/edmx-parser/edmx-file-reader.ts
```diff
@@ -57,19 +57,65 @@ function getODataVersion(edmx): ODataVersion {
...
+    EntityContainer: flat(
+      schema.map(s =>
+        forceArray(s.EntityContainer).map(ec =>
+          addNamespaceToEntityContainer(ec, s.Namespace)
+        )
+      )
+    ),
+    EntityType: flat(
+      schema.map(s => forceArray(s.EntityType).map(addNamespace(s.Namespace)))
+    ),
```
marikaner, line 76: Looks like code duplication to me ;)

Independently of that, I think we should merge this with the functionality of edmx-parser and move it to parser.

marikaner (commented):
Partial rereview

packages/generator/src/edmx-parser/edmx-file-reader.ts
```diff
@@ -57,19 +57,65 @@ function getODataVersion(edmx): ODataVersion {
...
+    Namespace: schema.map(s => s.Namespace)
+  };
+}
+
+function addNamespaceToEntityContainer(entityContainer, namespace: string) {
+  return {
+    Name: entityContainer.Name,
+    EntitySet: forceArray(entityContainer.EntitySet).map(
+      addNamespace(namespace)
+    ),
```
marikaner, line 104: Same here also looks a bit like code duplication.

packages/generator/src/vdm-types.ts
```diff
@@ -89,6 +90,7 @@ export interface VdmComplexType {
   typeName: string;
   factoryName: string;
   fieldType: string;
+  namespace: string;
```
marikaner, line 93: I'm sure there is a reason, but why does the `VdmComplexType` have a namespace. In `EdmxComplexType` it's obvious

jjtang1985: I add namespace info to `VdmEntity` as well, because both of them are used for matching the return type of the action/function import. This seems to be a temporary solution, I guess we will get rid of it when introducing additional abstraction.

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -91,17 +91,31 @@ function findEntityType(
...
 
 function findComplexType(
   returnType: string,
   complexTypes: Omit<VdmComplexType, 'factoryName'>[]
 ): Omit<VdmComplexType, 'factoryName'> | undefined {
   returnType = parseTypeName(returnType);
-  const parsedReturnType = returnType.split('.').slice(-1)[0];
-  return complexTypes.find(c => c.originalName === parsedReturnType);
+  const complexType = complexTypes.find(
+    e => `${e.namespace}.${e.originalName}` === returnType
```
jjtang1985, line 111: Here is why namespace is used.

packages/generator/src/edmx-parser/common/edmx-parser.ts
```diff
@@ -6,23 +7,52 @@ import {
...
+        forceArray(s['EntityContainer'][entityContainerProperty]).map(
+          addNamespace(schema.Namespace)
+        )
+      )
+  );
+}
+
+export function addNamespace(namespace) {
+  return obj => assoc('Namespace', namespace, obj);
+}
```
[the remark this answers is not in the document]
marikaner: Medium: Also consider, whether it would be nicer to not return a function here, but really only adding the namespace and using the callback in `.map`. I find this easier to read and then the name of the function fits what the function is actually doing.

```suggestion
export function addNamespace<T>(obj: T, namespace: string): T & { Namespace: string } {
  return { ...obj, Namespace: namespace };
}
```

Usage:
```ts
...map(obj => addNamespace(obj, schema.Namespace);
```

packages/generator/src/edmx-parser/common/edmx-parser.ts
```diff
@@ -6,23 +7,52 @@ import {
...
+
+export function getMergedPropertyWithNamespace(root, property: string): any[] {
```
marikaner, line 52: Consider adding a comment here

marikaner (approved):
I hope I didn't miss anything, but it seems that the new solution you came up with is even more concise than anything we discussed and it looks very good to me! I left some minor comments on the `addNamespace` function.

PR #724, opened 2020-11-19
feat: rest generator oclif command

FrankEssenberger:
Added a oclif command for executing the rest generator. I set the bin name to `sap-cloud-sdk-1` so that there is not conflict with the existing CLI. When we merge the repos this will of course also go under the normal name.

One thing was critical. I got an error on execution: ` Class constructor Command cannot be invoked without 'new'`.
[Stack overflow](https://stackoverflow.com/questions/50203369/class-constructor-cannot-be-invoked-without-new-typescript-with-commonjs/50203532#50203532) told me that this is related to the target beeing `es5` and classes beeing transpiled to this old standard. I added in the tsconfig of the rest-generator a new target `es2017` and it workds. 

I think we should evaluate if we can increase the target of the whole project to something more recent than es5.


packages/rest-generator/src/generator.ts
```diff
@@ -128,14 +128,12 @@ async function generateFilesUsingOpenAPI(
...
-      logger.info(`Generated the client ${response.stdout}`);
-    }
-  } catch (err) {
-    logger.error('In exception block');
-    logger.error(err);
+  const response = await execa.sync('npx', generationArguments);
+  if (response.stderr !== '') {
+    throw new Error(response.stderr);
+  }
+  if (response !== undefined) {
```
jjtang1985, line 135: here is the null check, then response is nullable.
If response is null, then line 132 will fail

packages/rest-generator/package.json
```diff
@@ -15,7 +15,7 @@
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "bin": {
-    "generate-rest-client": "./dist/generator-cli.js"
+    "sap-cloud-sdk-1": "./bin/run"
```
jjtang1985, line 18: Maybe `sap-cloud-sdk-rest` is more descriptive? I'm even not sure whether we will merge these two in this quarter.

packages/rest-generator/src/commands/generate-rest-client.ts
```diff
@@ -0,0 +1,88 @@
...
+      description: 'Output directory for the generated rest client.',
+      parse: input => resolve(input),
+      required: true
+    }),
+    clearOutputDir: flags.boolean({
+      name: 'clearOutputDir',
+      char: 'c',
+      description: 'Output directory for the generated rest client.',
+      default: false,
+      required: true
```
jjtang1985, line 53: There is a default value, so it can be optional?

packages/rest-generator/src/commands/generate-rest-client.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (response.exitCode !== 0 || !response.stderr.includes('version')) {
+      logger.error(
+        `Java -version did not return a Java version. Check Java version response: ${response.stderr}`
+      );
+      return Promise.reject();
+    }
+    return Promise.resolve();
+  } catch (err) {
+    logger.error(
+      `Error in checking the jave version. Is Java installed? Thisn is mandatory for the rest client generation. ${err.message}`
```
jjtang1985, line 84:
```suggestion
      `Error in checking the java version. Is Java installed? This is mandatory for the rest client generation. ${err.message}`
```

packages/rest-generator/src/commands/generate-rest-client.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export interface GeneratorOptions {
+  inputDir: string;
+  outputDir: string;
+  clearOutputDir?: boolean;
+}
+
+export class GenerateRestClient extends Command {
+  static description =
+    'Generates a Rest client from a openApi service file definition. For SAP solutions, you can find these definitions at https://api.sap.com/.';
```
jjtang1985, line 24:
```suggestion
    'Generates a Rest client from an openApi service file definition. For SAP solutions, you can find these definitions at https://api.sap.com/.';
```

packages/rest-generator/src/commands/generate-rest-client.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  static examples = [
+    '$ generate-rest-client generate-rest-client -i directoryWithOpenApiFiles -o outputDirectory',
+    '$ generate-rest-client generate-rest-client --help'
+  ];
+
+  static version = 'ABC';
```
jjtang1985, line 31: left over?

jjtang1985 (changes requested):
LGTM, only some minor comments.
Please create a ticket for upgrading/discussing the `es2017` topic, I found it's useful.
I tried to update to some newer version for some features, but the tests failed in the node 12 environment, so I gave up.
But it still makes sense to do so.

packages/rest-generator/src/generator.ts
FrankEssenberger: yes, the undefined check shoud be first.

packages/rest-generator/package.json
FrankEssenberger: ok you are right let's use a proper name and merge it when we add the CLI.

packages/rest-generator/src/commands/generate-rest-client.ts
FrankEssenberger: oh yes.

packages/rest-generator/src/commands/generate-rest-client.ts
FrankEssenberger: oh, yes I wanted to read the value from the package.json

FrankEssenberger:
@jjtang1985 I considered your good comments. Please have a second look.

jjtang1985 (approved):

PR #728, opened 2020-11-20
Fix api docs pipeline to work between two repos

artemkovalyov:
I updated the snippet that syncs the repos. This should do the work now. Please, take a look.


marikaner (approved):
Seems sensible

PR #733, opened 2020-11-23
chore: Add e2e test for the rest generator

marikaner:
I added some TODOs, that we should discuss and create tickets from. Ideally this would be done right away.


test-packages/e2e-tests/test/rest.spec.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  TestEntity,
+  TestServiceApiRequestBuilder
+} from '@sap-cloud-sdk/test-services/rest/test-service';
+import { destination } from './test-util';
+
+// TODO: How do I handle paths in rest requests?
+// TODO: Transpilation needed + tsconfig needs dom typings
+// TODO: Response is never "parsed"
```
marikaner, line 9: 1 and 3 are open for discussion. If you have opinions leave a comment.

jjtang1985: 1. Do you mean how to you test the generated url in the request or custom path or custom path parameter?
2. Discussed in the planning
3. As a minimal scope, I would like to see at least one required property (line 22 `keyProperty`) and one optional property (line 23 `stringProperty`) are tested. Is there any technical difficulty to parse the response?

test-packages/test-services/rest/test-service/open-api.json
```diff
@@ -1 +1 @@
...
...chema":{"type":"integer","format":"int32"}}}},"400":{"description":"bad input parameter"}},"tags":["TestService"]}}},"components":{"schemas":{"TestEntity":{"type":"object","required":["keyProperty"],"properties":{"keyProperty":{"type":"string","format":"uuid","example":"d290f1ee-6c54-4b01-90e6-d701748f0851"},"stringProperty":{"type":"string","example":"Example string"},"dateProperty":{"type":"string","format":"date","example":"2016-08-29"},"dateTimeProperty":{"type":"string","format":"date-time","example":"2016-08-29T09:12:33.001Z"},"int32Property":{"type":"integer","format":"int32","example":1},"int64Property":{"type":"integer","format":"int64","example":1},"floatProperty":{"type":"number","format":"float","example":1},"doubleProperty":{"type":"number","format":"double","example":1},"linkedSimpleTestEntity":{"$ref":"#/components/schemas/SimpleTestEntity"},"linkedSimpleTestEntityCollection":{"type":"array","items":{"$ref":"#/components/schemas/SimpleTestEntity"}}},"description":"TestEntity schema"},"SimpleTestEntity":{"type":"object","required":["stringProperty"],"properties":{"stringProperty":{"type":"string","example":"Example string"}},"description":"SimpleTestEntity schema"}}}}
```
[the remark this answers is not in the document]
marikaner: Sure

test-packages/e2e-tests/test/rest.spec.ts
marikaner: 1. We currently don't have a `withCustomServicePath` function here. That is why I have to touch the destination to get it from the correct url. I guess we should allow to set that.
3. The requestBuilder currently returns an AxiosResponse. I think we should rather return the data directly on execute, similar to how we do it for OData. We could then have a second executeRaw function that also returns the response (as discussed for OData).

What do you think about that?

As to the minimal scope of the test - do you think there is a case missing? Testing that there are required properties and optional properties should probably be part of a type test. Because this is covered by the type (in this case `TestEntity`) right?

test-packages/test-services/rest/test-service/open-api.json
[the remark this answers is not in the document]
marikaner: Ah no, this is not the one, that I created, this is the result of the generation. You can just refer to the test-service.json in the test resources.

jjtang1985 (approved):
LGTM, feel free to create follow up tickets or handle in different PRs for the TODOs
- `customServicePath` makes sense, that can be handled later without breaking changes
- `executeRaw` might be handled with higher priority to avoid future breaking changes.

Issue #741, opened 2020-11-24
New version (1.32.0) of @sap-cloud-sdk/core does not export `executeHttpRequest` function anymore


johannes-vogel:
**Describe the bug**
Hi colleagues,

when installing the newest version (1.32.0) of the `@sap-cloud-sdk/core` package, the functions of the `http-client` module are not exported anymore.

We rely on the `executeHttpRequest` method which was exported until `1.31.0`.
I did not find any related entry in the changelog and also the [documentation](https://sap.github.io/cloud-sdk/api/1.32.0/modules/sap_cloud_sdk_core#executehttprequest) suggests that this function should still be available. 

When looking at https://github.com/SAP/cloud-sdk-js/blob/v1.32.0/packages/core/src/index.ts, I can also see that it is not exported anymore.

**To Reproduce**
1. Install `@sap-cloud-sdk/core` in any node.js project.
2. `require('@sap-cloud-sdk/core').executeHttpRequest`  is not defined

**Expected behavior**
I'd expect the function to still be available.

marikaner:
Hey @johannes-vogel,

this should not have happened, I will check this and release a new version as soon as possible. Thanks for reporting!

johannes-vogel:
Hi @marikaner,

thanks for your quick support 👍

marikaner:
Hey @johannes-vogel, we have released a new version 1.32.1 that should fix this problem. Sorry for the inconvenience. I'll go ahead and close this issue.

johannes-vogel:
@marikaner I can confirm, that it is working again!

PR #742, opened 2020-11-24
fix: Fix exports

marikaner:
High priority fix for the broken version 1.32.0. We stopped exporting the http-client anymore. (This btw calls for an e2e or type test, I guess.)

Related to #741


FrankEssenberger (approved):
LGTM

test-packages/type-tests/test/http-client.spec.ts
```diff
@@ -0,0 +1,4 @@
+import { executeHttpRequest } from '@sap-cloud-sdk/core';
+
+// $ExpectType Promise<HttpResponse>
+executeHttpRequest({ url: 'https://example.com' }, { method: 'get' });
```
jjtang1985, line 4: 👍

jjtang1985 (approved):

PR #752, opened 2020-11-27
chore: move spec files package analytics & util

FrankEssenberger:
In this PR the spec files for the packages `analytics and `util` are moved.


jjtang1985 (approved):
LGTM.
Question: we also agreed that every `abc.ts` file should have `abc.spec.ts` in the same folder when possible, didn't we?
If so, I failed to find a ticket to reflect it.
Updated: added to https://github.wdf.sap.corp/MA/scp-rocket/issues/1146

FrankEssenberger:
@Junjie I also added the lint rule to this ticket: https://github.wdf.sap.corp/MA/scp-rocket/issues/1510

PR #760, opened 2020-12-01
chore: Move name converter and improve imports

marikaner:
As in the title. This should be fairly simple to review.


FrankEssenberger (approved):
LGTM

packages/util/src/string-formatter.ts
```diff
@@ -0,0 +1,43 @@
...
+  return voca.camelCase(str);
+}
+
+/**
+ * Convert a string to a human readable format, e.g. it transforms `to_BusinessPartner` to `To Business Partner`.
+ *
+ * @param str - The string to be transformed.
+ * @returns The transformed string.
+ */
+export function toTitleFormat(str: string): string {
```
marikaner, line 29: This function has a slightly different name than the others. The reason is, that we might want to add a `toTitleCase` function that transforms `content-type` to `Content-Type` as well.

jjtang1985:
👍 we should do the same for all unit tests.

packages/util/src/string-formatter.ts
FrankEssenberger: fine for me. `Format` is a good word since it is not changing the case.

PR #761, opened 2020-12-01
chore: Move spec files for package generator

FrankEssenberger:
Move the spec files for the generator package. Here I needed to sort a few tests to new files or to existing ones.
I kept the `test/test-util` because I had no good idea where to move the code. 
Have you any opinions where to put the test-util stuff?


packages/generator/src/edmx-parser/edmx-file-reader.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  parseEntitySets,
+  parseEntityType,
+  parseEnumTypes,
+  parseFunctionImports,
+  parseFunctions
+} from './v4';
+import { parseComplexTypesBase } from './common';
+
+describe('edmx-file-reader', () => {
+  it('does not fail for multiple schema entries in the edmx file', () => {
```
jjtang1985, line 14: Maybe remove the one from `packages/generator/src/edmx-parser/v2/edmx-parser.spec.ts`?

packages/generator/src/edmx-to-vdm/v4/action-import.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+  EdmxEntitySet,
+  EdmxEntityType
+} from '../../edmx-parser/v4';
+import { ServiceMetadata } from '../../edmx-parser/edmx-file-reader';
+import { generateEntitiesV4 } from './entity';
+import { generateActionImportsV4 } from './action-import';
+import { generateComplexTypesV4 } from './complex-type';
+import { createEntityType, getComplexType, getFormatter } from './entity.spec';
+
+describe('action-import', () => {
```
jjtang1985, line 15: I assume this this is an old test that has been moved from the new `packages/generator/src/edmx-to-vdm/v4/entity.spec.ts`

packages/generator/src/service-generator.spec.ts
```diff
@@ -0,0 +1,489 @@
+import { createOptions } from '../test/test-util/create-generator-options';
+import { GlobalNameFormatter } from './global-name-formatter';
+import { ServiceMapping } from './service-mapping';
+import { VdmReturnTypeCategory, VdmProperty } from './vdm-types';
+import { parseAllServices, parseService } from './service-generator';
+
+describe('service-generator', () => {
```
jjtang1985, line 7: As expected, we need to merge some tests (from v2/v4 folder) if the source code does not have v2/v4 structure. Having 500 lines as test is still fine, I guess we have to find a solution when this becomes huger.

jjtang1985 (approved):
LGTM, just minor comments.
> Have you any opinions where to put the test-util stuff?

I would keep package specific files in that package like `packages/generator/test/test-util/create-generator-options.ts`. If there are some very general functions that are/can be reused, then consider moving them out.

packages/generator/src/edmx-parser/edmx-file-reader.spec.ts
FrankEssenberger: oh yes good point. I copied it but forgot to remove it.

packages/generator/src/service-generator.spec.ts
FrankEssenberger: I agree. I just moved them all togehter.One thing we could do is to put the parseAll method and parseServce to a seperate file and have the tests for this file than seperate from the one of the service-generator. I think if the test files get too large this is a good indicator that the object under test does to much and should be split.

PR #766, opened 2020-12-02
chore: Move spec files package core (first part)

FrankEssenberger:
In this PR I moved all spec files which have a clear mapping i.e. `something.spec.ts` and `something.ts` do exist.
This covered most of the files but not all so that is why the `test` folder` is not completly empty yet
.


packages/core/src/connectivity/scp-cf/connectivity-service.spec.ts
```diff
@@ -1,17 +1,17 @@
 import { createLogger } from '@sap-cloud-sdk/util';
-import { Destination } from '../../src';
-import { addProxyConfiguration } from '../../src/connectivity/scp-cf';
 import {
   mockConnectivityServiceBinding,
   mockedConnectivityServiceProxyConfig,
   mockServiceBindings
-} from '../test-util/environment-mocks';
+} from '../../../test/test-util/environment-mocks';
```
florian-richter, line 6:
```suggestion
} from '@sap-cloud-sdk/test-util';
```

florian-richter: Also applies to all the imports below

packages/core/src/connectivity/scp-cf/destination/destination-service-cache.spec.ts
```diff
@@ -1,24 +1,23 @@
...
-  decodeJwt
-} from '../../src/connectivity/scp-cf';
 import {
   mockSingleDestinationCall,
   mockSubaccountDestinationsCall,
   mockVerifyJwt
-} from '../test-util/destination-service-mocks';
-import { mockServiceBindings } from '../test-util/environment-mocks';
-import { mockServiceToken } from '../test-util/token-accessor-mocks';
+} from '../../../../test/test-util/destination-service-mocks';
```
florian-richter, line 6:
```suggestion
} from '@sap-cloud-sdk/test-util';
```

packages/core/src/connectivity/scp-cf/destination/destination-service.spec.ts
```diff
@@ -1,13 +1,13 @@
 import nock from 'nock';
 import * as jwt123 from 'jsonwebtoken';
+import { destinationServiceUri } from '../../../../test/test-util/environment-mocks';
+import { privateKey } from '../../../../test/test-util/keys';
```
florian-richter, line 4:
```suggestion
import { privateKey } from '@sap-cloud-sdk/test-util';
```

packages/core/src/connectivity/scp-cf/jwt.spec.ts
```diff
@@ -2,12 +2,10 @@ import { IncomingMessage } from 'http';
...
 import {
-  audiences,
-  retrieveJwt,
-  verificationKeyCache,
-  verifyJwt
-} from '../../src';
-import { publicKey, signedJwtForVerification } from '../test-util/keys';
+  publicKey,
+  signedJwtForVerification
+} from '../../../test/test-util/keys';
```
florian-richter, line 7:
```suggestion
} from '@sap-cloud-sdk/test-util';
```

packages/core/src/connectivity/scp-cf/token-accessor.spec.ts
```diff
@@ -1,27 +1,24 @@
...
-  userApprovedServiceToken
-} from '../../src/connectivity/scp-cf';
 import {
   mockDestinationServiceBinding,
   mockServiceBindings,
   providerXsuaaUrl,
   subscriberXsuaaUrl
-} from '../test-util/environment-mocks';
-import { signedJwt } from '../test-util/keys';
+} from '../../../test/test-util/environment-mocks';
```
florian-richter, line 7:
```suggestion
} from '@sap-cloud-sdk/test-util';
```

packages/core/src/connectivity/scp-cf/xsuaa-service.spec.ts
```diff
@@ -1,13 +1,13 @@
 import { fail } from 'assert';
 import nock from 'nock';
-import { TokenKey } from '../../src';
+import { providerXsuaaUrl } from '../../../test/test-util/environment-mocks';
```
florian-richter, line 3:
```suggestion
import { providerXsuaaUrl } from '@sap-cloud-sdk/test-util';
```

packages/core/src/odata-common/request-builder/batch/batch-response-deserializer.spec.ts
```diff
@@ -43,37 +43,39 @@ describe('batch response transformer', () => {
     const entityToConstructorMap = {
       entity: 'entity' as any
     };
-    const batchTransformer = new BatchResponseDeserializer(
-      entityToConstructorMap,
-      responseDataAccessorV2,
-      {} as EntityDeserializer
-    );
+    describe('v2', () => {
```
florian-richter, line 46: I'm not a big fan of nested `describe`s. This caused errors in the past (but it may have been with jasmine not jest...) Maybe it would be better to start a new block here to flatten it a bit rather than nesting deeper

florian-richter (changes requested):
`test-util` should be imported through `@sap-cloud-sdk/test-util` if possible (I think there is also a linter rule you could adjust to automatically check for this).

FrankEssenberger:
@florian-richter on an other PR I got linter errors if I tried this. I will try to merge master and  see if I get the lint error, which is ensuring that we do not miss a wrong import.

jjtang1985:
If I remember correctly, for unit test, we did not use import like `@sap-cloud-sdk/util`, which is required for integration tests as well as type tests.

jjtang1985 (approved):
LGTM

packages/core/src/connectivity/scp-cf/connectivity-service.spec.ts
marikaner: @florian-richter But this is a different test-util.

FrankEssenberger:
@Junjie I would only use the direct import if I want to use the actual code without the compile step in between. For the pipeline this does not matter but for local development it has sometimes more convinience but I think the test util is something rather stable.

packages/core/src/connectivity/scp-cf/connectivity-service.spec.ts
jjtang1985: To make it clear: this is `test-util` directory under the `core` package instead of the `test-util` package.

jjtang1985:
> I would only use the direct import if I want to use the actual code without the compile step in between. For the pipeline this does not matter but for local development it has sometimes more convinience but I think the test util is something rather stable.

For me, if the `test-util` is the only one package that is related to this topic, I would keep the unit test as simple as possible like now. For api specific functions, we anyway covered (really?) in the integration/type tests. That's why I approved. We can change the lint rules if we have good reasons.

jjtang1985:
As discussed with @FrankEssenberger , for this PR, only `test-util` package is used as devDependency, both are fine, since `test-util` is stable. But we don't want to always use `@sap-cloud-sdk/core` for unit test, because you have to compile again and again during development.

Also, we detected a potential lint error, where you can import the files from other packages by using relative path instead of package names. This should only be possible in the `.spec.ts` file but NOT in the source code. Please create a follow up ticket if we cannot solve it within 5 min.

PR #767, opened 2020-12-02
feat: Fix many but not all issues in generator

marikaner:
What you will find in this PR:
* Replaced ts-morph with string templates
* Identified many (probably not all) issues and potential problems and solved many of them
  * I will create one ticket for all TODOs, but I will leave them here until they are done
* Implemented unit tests
* Renamed some functions in util (they have not yet been released so it is not a breaking change)


packages/rest-generator/src/openapi-types.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/**
+ * @experimental This API is experimental and might change in newer versions. Use with caution.
+ * Representation of an operation.
+ */
+export interface OpenApiOperation extends OpenAPIV3.OperationObject {
+  operationName: string;
+  method: string;
+  pattern: string;
```
FrankEssenberger, line 20: I do not see where the `pattern` is used.

FrankEssenberger: In the `parseOperation` the pattern is parse explicitly.

packages/rest-generator/src/wrapper-files/api-file.ts
```diff
@@ -0,0 +1,95 @@
...
+ * @returns The generated code for the SDK API wrapper.
+ */
+export function apiFile(openApiDocument: OpenApiDocument): string {
+  const requestBodyTypes = getRequestBodyTypes(openApiDocument);
+  return codeBlock`
+import { RestRequestBuilder } from '@sap-cloud-sdk/core';
+import { DefaultApi } from './openapi/api';
+${
+  requestBodyTypes
+    ? `import { ${requestBodyTypes} } from './openapi/model';`
```
FrankEssenberger, line 17: I guess you could use the `import type` here?  This never hurts if you just pass the type and could help avoid cyclic dependencies.

packages/rest-generator/src/parser/request-body.spec.ts
```diff
@@ -0,0 +1,81 @@
+import { createLogger } from '@sap-cloud-sdk/util';
+import { createRefs } from '../../test/test-util';
+import { parseRequestBody } from './request-body';
+
+describe('getRequestBody', () => {
```
FrankEssenberger, line 5: Question: Is it possible to define a schema also inline like for the parameters? I do not see a test for that if this is possible.

packages/rest-generator/src/parser/refs.ts
```diff
@@ -0,0 +1,36 @@
...
+ * This operates only on the current level.
+ * @param obj Object to resolve if necessary.
+ * @param refs References to resolve by.
+ * @returns A resolved object.
+ */
+export function resolveObject<T>(
+  obj: T | OpenAPIV3.ReferenceObject,
+  refs: $Refs
+): T {
+  return isReferenceObject(obj) ? refs.get(obj.$ref) : obj;
```
FrankEssenberger, line 35: does this `get` also handle layered resolution so if the ref consists of other refs consistings of refs... until you reach the actual definition. Also cyclic references could be an issue there.

packages/util/src/array.spec.ts
```diff
@@ -120,3 +122,43 @@ describe('array', () => {
     });
   });
 });
+
+describe('zip', () => {
```
FrankEssenberger, line 126: perhaps add a test where one array is really  a few elements longer than the other one to see that it is just appended. From the word `zip` with the docs it is clear, but the examples look as if the method is sorting. Perhaps you call the elements so that the result is not sorterd.

packages/util/src/array.spec.ts
```diff
@@ -120,3 +122,43 @@ describe('array', () => {
...
+  it('zips two arrays where left is longer than right', () => {
+    expect(zip([1, 3, 5], [2, 4])).toEqual([1, 2, 3, 4, 5]);
+  });
```
FrankEssenberger, line 129: perhaps also a type test that you can not mix different types with zip.

packages/util/src/code-block.ts
```diff
@@ -0,0 +1,25 @@
+import { zip } from './array';
+import { trim, trimRight } from './string';
+
+export const codeBlock = (strings, ...args) => {
```
FrankEssenberger, line 4: doc missing or `hidden` also I do not see where this is used.

packages/rest-generator/src/parser/operation.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    const anyParamNoSchema = { in: 'query', name: 'anyParamNoSchema' };
+    const anyParamWithSchema = {
+      in: 'query',
+      name: 'anyParamWithSchema',
+      schema: {}
+    };
+    const stringParam = {
+      in: 'query',
+      name: 'anyParamWithSchema',
+      schema: { type: 'string' }
```
FrankEssenberger, line 61: Is string and integer the only possible parameter? Boolean? Number?

packages/util/src/string.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { indent, trim, encodeBase64 } from './string';
+
+describe('encodeBase64', () => {
+  it('encodes a string', () => {
+    expect(encodeBase64('test')).toMatchInlineSnapshot('"dGVzdA=="');
```
FrankEssenberger, line 5: use this method in the `basicHeader` method.

packages/util/src/string.ts
```diff
@@ -1,3 +1,53 @@
...
+    .split('\n')
+    .map(subString => `${indentation}${subString}`)
+    .join('\n');
+
+/**
+ * Remove whitespace from the left side of a string.
+ * @param string String to trim.
+ * @returns String without whitespace on the left side.
+ */
+export const trimLeft = string => {
```
FrankEssenberger, line 27: General: Function definition via the arrows is a matter of taste. I like the funciton keyword. But any way you define it we should add stronger typing: Argument of type string and methods returns a string.

packages/util/src/code-block.ts
FrankEssenberger: Also here stronger typeing for arguments.

package.json
```diff
@@ -33,7 +33,7 @@
     "test:e2e": "yarn e2e-tests test",
     "generate:test-services:v2": "yarn generate v2 && yarn lint:fix && yarn --frozen-lockfile",
     "generate:test-services:v4": "yarn generate v4 && yarn lint:fix && yarn --frozen-lockfile",
-    "generate:test-services:rest": "yarn rest-generator ts-node src/generator-cli.ts -i ../../test-resources/rest-service-specs -o ../../test-packages/test-services/rest && yarn lint:fix && yarn --frozen-lockfile",
+    "generate:test-services:rest": "yarn rest-generator sap-cloud-sdk-rest generate-rest-client -i ../../test-resources/rest-service-specs -o ../../test-packages/test-services/rest --clearOutputDir && yarn --frozen-lockfile",
```
FrankEssenberger, line 36: cloud you include the generation of the rest client also in the script generate-rest-service.ts so that we have it consistent.

FrankEssenberger: Also I see that the service has been generated and it is also used in the `rest.spec.ts` but I think a compilation of the generated sevice is missing to ensure syntactic correct content.

packages/rest-generator/src/wrapper-files/api-file.ts
FrankEssenberger: I am also a bit afraid about stability here. Can we really rely on that there will be a file with the name in the models if the schema was named this way. What happens in case of naming conflict between a model schema type and a operation name. We will use a unique name and deviate from the initial name. If this is not an issue also explain why or if it is a valid concern create a follow up ticket.

FrankEssenberger (changes requested):
I like it a lot and had no real code changes just clarifications for me and minor missings things.

packages/rest-generator/src/wrapper-files/api-file.ts
marikaner: I think this is not an issue, but there are issues related to this: 
1. Why it is not an issue - we replace all tags in the original document before parsing. I found this weird but after some investigation I found that this is actually the only way to have one api for everything in one spec file. That tag is now 'default' so the name of the api is always `DefaultApi`, which is never exported on "package level"
2. Related issues - The stability around this whole wrapper is in fact questionnable. I have some todos left to check what the behavior of the openapi generator is in some cases. I already know of some incompatibilities that we already have. I will write them all down in a follow up - my concern is - have I considered all of them? ¯\\_(ツ)_/¯

packages/rest-generator/src/parser/request-body.spec.ts
marikaner: Yes it is, this would be a follow up then, because I did not yet check what the openapi generator does. I am not testing this for the parameters either (I think).

packages/rest-generator/src/parser/operation.spec.ts
marikaner: There are others, but I thought duplicating the type mapper here would be trivial, so I just checked that for two different cases I get the expected results. Do you think that is enough?

packages/rest-generator/src/parser/refs.ts
marikaner: It does not handle layered resolution (so also no cyclic refs). Maybe I should improve the docs a bit here.

marikaner: This test implicitly proves this: https://github.com/SAP/cloud-sdk-js/blob/ec895b4c3387309fb22e7919cdb3e4c3306225ba/packages/rest-generator/src/parser/operation.spec.ts#L89

The docs of this function aren't very helpful with this: https://apitools.dev/swagger-parser/docs/refs.html#getref-options

packages/util/src/array.spec.ts
marikaner: Will do

packages/util/src/code-block.ts
marikaner: I will add the docs and types. It is used in the string templates:
```
codeblock`bla`
```

packages/util/src/string.spec.ts
marikaner: Will do, just fyi, this was here before, but not tested.

packages/util/src/string.ts
marikaner: Will do

packages/rest-generator/src/wrapper-files/api-file.ts
marikaner: (BTW: you would have never seen this with ts-morph ;) )

marikaner: I cannot use import type because DefaultApi is also passed as the first parameter to the function.

packages/rest-generator/src/parser/refs.ts
marikaner: Improved docs.

packages/util/src/array.spec.ts
marikaner: Done.

packages/util/src/array.spec.ts
marikaner: Zip and most other array functions use a generic type. Testing this seems trivial to me and would rather check the correctness of generics in TypeScript than our code in my opinion. Or do you see other issues that I didn't?

packages/util/src/code-block.ts
marikaner: Done.

packages/util/src/string.spec.ts
marikaner: Done.

packages/util/src/string.ts
marikaner: Done.

packages/rest-generator/src/openapi-types.ts
marikaner: You are right, this is not used anywhere (yet). I think we might want to keep it anyways, because we will probably use it in the generated documentation later on (TSDoc for the generated functions).

package.json
marikaner: I don't understand the first comment.
To the second: Yes, we should definitely do that in the generator directly. Currently we don't even generate a tsconfig.json that is why this is not there yet. I will add it with high priority to the todo list.

marikaner:
@FrankEssenberger thank you for the thorough review! Ready for re-review.

packages/util/src/array.spec.ts
FrankEssenberger: no I also thought about the triviality and it is clear from the method signature/definition.

packages/rest-generator/src/parser/operation.spec.ts
FrankEssenberger: hm, ok we keep it this way but if we see an error related to an other type we add them :-).

package.json
FrankEssenberger: as discussed per phone to the first point.

FrankEssenberger (approved):
LGTM please go ahead and merge after you adjusted the generation script or added a TODO for this.

PR #777, opened 2020-12-04
chore: Rename rest to openapi

marikaner:
@FrankEssenberger sorry it took so long, it wasn't easy to change the oclif multi command to single (theoretically it's easy, it was just me not getting it.)


packages/openapi-generator/package.json
```diff
@@ -15,31 +15,21 @@
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "bin": {
-    "sap-cloud-sdk-rest": "./bin/run"
+    "generate-openapi-client": "./bin/run"
```
FrankEssenberger, line 18: This is the `sap-cloud-sdk` for the CLI so the command the generate will be called: `generate-openapi-client generate-openapi-client. ...` if this is the case perhaps you can give the comannd a shorter name or this bin.

FrankEssenberger (approved):
LGTM - after discussing the command name.

PR #784, opened 2020-12-07
fix: Use camel case for parameter name

jjtang1985:
Fix some special cases when the parameter names are like `"name": "$orderby"` and `"name": "attributes.ExampleCustomAttribute"`, the camel case is used to get rid of special characters.


test-packages/test-services/openapi/test-service/openapi/api/default-api.ts
```diff
@@ -232,11 +242,13 @@ export const DefaultApiFp = function(configuration?: Configuration) {
...
          * @param {string} [stringParameter] A parameter of type string
          * @param {number} [integerParameter] A parameter of type integer
+         * @param {string} [$dollarParameter] A parameter that contains $
+         * @param {string} [prefixSuffixParameter] A parameter that contains . in the middle
          * @param {*} [options] Override http request option.
          * @throws {RequiredError}
          */
-        async getAllEntities(stringParameter?: string, integerParameter?: number, options?: any): Promise<(axios?: AxiosInstance, basePath?: string) => AxiosPromise<Array<TestEntity>>> {
-            const localVarAxiosArgs = await DefaultApiAxiosParamCreator(configuration).getAllEntities(stringParameter, integerParameter, options);
+        async getAllEntities(stringParameter?: string, integerParameter?: number, $dollarParameter?: string, prefixSuffixParameter?: string, options?: any): Promise<(axios?: AxiosInstance, basePath?: string) => AxiosPromise<Array<TestEntity>>> {
```
jjtang1985, line 250: The `$` is still there in the default-api.

packages/util/src/string-formatter.spec.ts
```diff
@@ -48,6 +48,14 @@ describe('converts names', () => {
     it('for properties with numbers', () => {
       expect(camelCase('Field13Name')).toBe('field13Name');
     });
+
+    it('for properties with dot', () => {
```
FrankEssenberger, line 52: perhaps add propertiy names with 'dot' because for me it was not clear if it is about the name of the property or the value.

test-packages/test-services/openapi/test-service/openapi/api/default-api.ts
FrankEssenberger: is this the expected behavior that it is there?

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -38,20 +38,21 @@ export function parseOperation(
...
 ): OpenApiParameter[] {
   // TODO: What if this is a reference? What does OpenApi do?
-  // TODO: What about oneof and other operations?
+  // TODO: What about one of and other operations?
   return (
     operation.parameters
       ?.map(param => resolveObject(param, refs))
       .map(param => ({
         ...param,
+        name: camelCase(param.name),
```
FrankEssenberger, line 55: Is this the point where the `$` and `.` are removed?

FrankEssenberger (approved):
LGTM I just want to understand it better and added a few questions and small remarks.

packages/openapi-generator/src/parser/operation.ts
jjtang1985: Yes, the function `camelCase` removes the special characters and format it to camel case.

test-packages/test-services/openapi/test-service/openapi/api/default-api.ts
jjtang1985: @marikaner , maybe you can comment it here.
I personally would not touch the `default-api.ts` since it's the vanilla openapi client?
Then we either keep the same naming strategy (use $dollarParameter) or align with our own convention.
I prefer the latter one, but it's open to be discussed.

marikaner (approved):
LGTM, I left some minor comments

packages/util/src/string-formatter.spec.ts
```diff
@@ -48,6 +48,14 @@ describe('converts names', () => {
...
+      expect(camelCase('field.name')).toBe('fieldName');
+    });
+
+    it('for properties with dollar', () => {
+      expect(camelCase('$fieldName')).toBe('fieldName');
```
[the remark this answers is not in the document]
jjtang1985: I don't have strong opinion here, especially the vanilla open api client is using `$`.
Ok, then I'll align with the open api and keep using the `$` then.

test-packages/test-services/openapi/test-service/openapi/api/default-api.ts
marikaner: The $ is allowed in the beginning of a name of a parameter, therefore I don't see an issue. I would personally choose to keep the $, because it might be there for a reason and it does not hurt, but we would have to adjust the camelCase function for this (which is not that big a deal I guess). Not sure it is worth it.

packages/util/src/string-formatter.spec.ts
marikaner: This is a left over because I recently moved it from core to util. I forgot to align the names and make them more general, We can either do this as part of this PR or on a separate PR. Just let me know.

I guess something like this would be good.
```js
describe('string-formatter', () => {
  describe('camelCase', () => {
    it('transforms string containing a number', ...
    it('transforms string containing a dot', ...
    it('transforms string starting with $', ...
    etc.
  });
});
```

(I guess a separate PR would be better...)

test-resources/openapi-service-specs/test-service.json
```diff
@@ -51,6 +51,28 @@
...
+            "description": "A parameter that contains $",
+            "required": false,
+            "style": "form",
+            "explode": true,
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "doc.parameter",
```
marikaner, line 67: I guess that is a typo? ;)

jjtang1985:
To support the `$`, we cannot use the `camelCase()` any more. I'll leave it for now since the parameter names involved are the sdk internal names.

PR #785, opened 2020-12-07
fix: OAuth2SAMLBearerAssertion for technical user and subscriber destination

FrankEssenberger:
While investigating the `OAuth2UserTokenExchange` I was comparing the code blocks we have already from the `OAuth2SAMLBearerAssertion` case. Here the docs on the SAML2: https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/3cb7b81115c44cf594e0e3631291af94.html

And I think I found and error for the technical user case and subscriber/provider account destination. It is about this case:

image

If the destination was from the subscriber account one should also use a `clientCredentials` token using the subscriber subdomain not the one from the provider account. I adjusted this in the `addOAuthSamlAuth` method.


FrankEssenberger:
@mr-flannery it would be really great if you could perhaps also a quick look here since you are the expert on this whole token flow thingies.

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -214,19 +231,33 @@ class DestinationFromServiceRetriever {
         `Failed to fetch destination "${this.name}"! No binding to a destination service found.`
       );
     }
-
+    /* This covers the two technical user propagation cases https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/3cb7b81115c44cf594e0e3631291af94.html
+    If the destination comes from the provider account the  the client credentials token from the xsuaa.url is used (provider token).
```
mr-flannery, line 235:
```suggestion
    If the destination comes from the provider account the client credentials token from the xsuaa.url is used (provider token).
```

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -117,24 +120,37 @@ class DestinationFromServiceRetriever {
...
     return serviceToken('destination', optionsWithoutJwt);
   }
 
-  readonly decodedProviderToken: DecodedJWT;
+  private static async getSubscriberClientCredentialsToken(
+    options: DestinationOptions
+  ): Promise<string> {
+    if (!options.userJwt) {
+      throw new Error(
+        'User Jwt is needed to obtain a client credentials token for the subscriber account.'
```
marikaner, line 135:
```suggestion
        'User JWT is needed to obtain a client credentials token for the subscriber account.'
```

marikaner (approved):

mr-flannery (commented):
I haven't touched that topic yet, so unfortunately I'm not the expert here. But from what I understood your change looks correct to me. Before, when the `systemUser` was set, the code has always used the `providerToken`. Now, it checks for the origin and uses/fetches the token accordingly. Either way, I recommend getting a second opinion.

I'd double-check if the `destinationOrigin` is actually set the way you expect it here. I might mix something up here, but I think we had deprecated that...

FrankEssenberger:
> I'd double-check if the `destinationOrigin` is actually set the way you expect it here. I might mix something up here, but I think we had deprecated that...

No you are not mixing it up. It was deprecated but when I did the refactoring I found it useful to have a `destinationSearchResult` which had info like: fromCache and origin in it because it made the post-processing easier.

FrankEssenberger:
I will create also a follow up ticket to include E2E tests for all the OAuth cases since we are currently also implementing `OAuth2UserTokenExchange`.

PR #791, opened 2020-12-08
chore: Unwrap axios response

marikaner:
Introduce `executeRaw` that returns an AxiosResponse and return the plain response value in `execute`. Also adds tests.

### DIFF
--- a/packages/core/src/openapi/openapi-request-builder.spec.ts
+++ b/packages/core/src/openapi/openapi-request-builder.spec.ts
@@ -1,0 +1,52 @@
+import { AxiosRequestConfig, AxiosResponse } from 'axios';
+import { OpenApiRequestBuilder } from './openapi-request-builder';
+
+class TestApi {
+  constructor(public axiosConfig: AxiosRequestConfig) {}
+  fn: () => Promise<AxiosResponse<number>> = async () =>
+    ({
+      data: 1
+    } as AxiosResponse<number>);
+  nonAxiosResponse = async () => 1;
+}
+
+const destination = {
+  url: 'http://example.com'
+};
+
+describe('openapi-request-builder', () => {
+  it('executes request', async () => {
+    const requestBuilder = new OpenApiRequestBuilder(TestApi, 'fn');
+    const response = await requestBuilder.execute(destination);
+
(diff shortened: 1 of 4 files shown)


packages/core/src/openapi/openapi-request-builder.ts
```diff
@@ -45,11 +47,11 @@ export class OpenApiRequestBuilder<ApiT, FnT extends keyof ApiT> {
 
   /**
    * Execute request.
-   * @param destination  Destination to execute the request against.
+   * @param destination Destination to execute the request against.
```
FrankEssenberger, line 50: perhaps you can adjust the text form the `execute` and `executeRaw` a bit to make clear what the differences are. and when to use it.

I know it is in the return parameters but it is always nice to have a bit text in the beginning.

packages/core/src/openapi/openapi-request-builder.ts
```diff
@@ -45,11 +47,11 @@ export class OpenApiRequestBuilder<ApiT, FnT extends keyof ApiT> {
...
    * @param options Options to employ when fetching destinations.
-   * @returns A promise resolving to the requested return type.
+   * @returns A promise resolving to an AxiosResponse.
    */
-  async execute(
+  async executeRaw(
```
FrankEssenberger, line 54: Also add a ticket a quickly adjust the docs to reflect the new method on the openApi client.

FrankEssenberger: Wait the method is not exported. Is it supposed to be used just internally?

packages/core/src/openapi/openapi-request-builder.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { AxiosRequestConfig, AxiosResponse } from 'axios';
+import { OpenApiRequestBuilder } from './openapi-request-builder';
+
+class TestApi {
+  constructor(public axiosConfig: AxiosRequestConfig) {}
```
FrankEssenberger, line 5: add one test for executeRaw - unit test is enough.

FrankEssenberger (approved):
LGTM add the unit test and the documentation improvement.

PR #792, opened 2020-12-08
feature: Support OAuth2UserTokenExchange as authentication type for destinations

FrankEssenberger:
It is not a big change but it took me a while to understand what I have to do while reading the [docs](https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/39d42654093e4f8db20398a06f7eab2b.html).

I overloaded the `fetchDestination` method which is used to trigger the OAuth flows with a new signature containing the `X-user-token` if a user token is provided. 

I also adjusted the `authToken` type in the destination response from the service as given [here](https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/83a3f3b9cd314618aba651044ed5b9df.html#loio83a3f3b9cd314618aba651044ed5b9df__tokens). So we use the `http_header` value directly. This has the advantage that we do not need to build it.


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 
 - [Generator] Support parsing of multiple schemas in the edmx service specification.
 - [Util] Add string formatting functions.
+- [Core] Support for Destination with authentication type [OAuth2UserTokenExchange](https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/39d42654093e4f8db20398a06f7eab2b.html)
```
florian-richter, line 24:
```suggestion
- [Core] Support for `Destination` with authentication type [`OAuth2UserTokenExchange`](https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/39d42654093e4f8db20398a06f7eab2b.html)
```

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -220,21 +226,69 @@ class DestinationFromServiceRetriever {
     return credentials[0];
   }
 
-  private async addOAuthSamlAuth(
-    destination: Destination,
-    destinationOrigin: DestinationOrigin
+  private fetchDestinationByToken(
```
florian-richter, line 229: If this function returns a promise, it should be marked as async
```suggestion
  private async fetchDestinationByToken(
```

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -220,21 +226,69 @@ class DestinationFromServiceRetriever {
...
       throw Error(
-        `Failed to fetch destination "${this.name}"! No binding to a destination service found.`
+        'No user token (JWT) has been provided! This is strictly necessary for OAuth2UserTokenExchange.'
+      );
+    }
+    // This covers the three OAuth2UserTokenExchange cases https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/39d42654093e4f8db20398a06f7eab2b.html
+
+    // Case 1 Destination in provider and jwt issued for provider account
+    if (this.isProviderAndSubscriberSameTenant()) {
+      const authHeaderJwt = await userApprovedServiceToken(
```
florian-richter, line 254: This file is already pretty huge. I think this would be a good point to consider factoring out some functions. Maybe some methods could be turned into pure functions and moved out of this class?

packages/core/src/connectivity/scp-cf/destination/destination-service.ts
```diff
@@ -114,19 +116,48 @@ async function fetchDestinations(
...
+      options
+    );
+  }
+  return fetchDestinationByTokens(
+    destinationServiceUri,
+    token,
+    destinationName,
+    options
+  );
+}
```
florian-richter, line 156:
```suggestion
  return fetchDestinationByTokens(
    destinationServiceUri,
    typeof token === 'string' ? { authHeaderJwt: token } : token,
    destinationName,
    options
  );
}
```

packages/core/src/connectivity/scp-cf/destination/destination-service.ts
```diff
@@ -150,14 +181,18 @@ export async function fetchDestination(
       return destinationsFromCache[0];
     }
   }
+  let headers = wrapJwtInHeader(tokens.authHeaderJwt).headers;
+  if (tokens.exchangeHeaderJwt) {
+    headers = { ...headers, 'X-user-token': tokens.exchangeHeaderJwt };
+  }
```
florian-richter, line 187:
```suggestion
  const headers = tokens.exchangeHeaderJwt ? { ...headers, 'X-user-token': tokens.exchangeHeaderJwt } : wrapJwtInHeader(tokens.authHeaderJwt).headers;
```

packages/core/src/connectivity/scp-cf/environment-accessor.ts
```diff
@@ -110,6 +110,23 @@ export function getService(service: string): Service | undefined {
...
+ *
+ * @returns Destination service
+ * @throws Error in case no destination service is found in the VCAP variables
+ */
+export function getDestinationService() {
+  const destinationService = getService('destination');
+
+  if (!destinationService) {
+    throw Error(
+      `Failed to fetch destination "${this.name}"! No binding to a destination service found.`
```
florian-richter, line 124: What is `this.name` in this context? Is this something that is left from a copy-paste?

florian-richter (changes requested):

packages/core/src/connectivity/scp-cf/environment-accessor.ts
FrankEssenberger: copy paste error

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
FrankEssenberger: I agree. However, I tried this for a bit and somehow the methods are always related and it would be a bigger effort. So I would do this in a seperate PR.

FrankEssenberger:
Could someone please have a look again?

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -221,21 +227,69 @@ class DestinationFromServiceRetriever {
...
+    // Case 3 Destination in subscriber and jwt issued for subscriber account
+    if (destinationOrigin === 'subscriber') {
+      return this.fetchDestinationByToken({
+        authHeaderJwt: await DestinationFromServiceRetriever.getSubscriberClientCredentialsToken(
+          this.options
+        ),
+        exchangeHeaderJwt: this.options.userJwt
+      });
+    }
+    throw Error('');
```
marikaner, line 280: That message can be improved ;)

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -221,21 +227,69 @@ class DestinationFromServiceRetriever {
...
+    if (!this.options.userJwt) {
       throw Error(
-        `Failed to fetch destination "${this.name}"! No binding to a destination service found.`
+        'No user token (JWT) has been provided! This is strictly necessary for OAuth2UserTokenExchange.'
+      );
+    }
+    // This covers the three OAuth2UserTokenExchange cases https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/39d42654093e4f8db20398a06f7eab2b.html
+
+    // Case 1 Destination in provider and jwt issued for provider account
+    if (this.isProviderAndSubscriberSameTenant()) {
```
marikaner, line 254: Minor: I think this might be easier to read if there was a function that would get the `AuthAndExchangeTokens` before fetching the destination (you don't necessarily need an extra function for this).
Not that it is very hard to read, but there is boilerplate code in the way that could be avoided.

```ts
private async getAuthAndExchangeTokens(destinationOrigin: DestinationOrigin): Promise< AuthAndExchangeTokens> {
  // Option1, 2, 3
  // The magic happens here
}

private async addOAuth2UserTokenExchange(destinationOrigin: DestinationOrigin): Promise<Destination> {
  // ...
  return fetchDestination(getAuthAndExchangeTokens(destinationOrigin));
}
```

packages/core/src/connectivity/scp-cf/destination/destination-service.ts
```diff
@@ -114,19 +116,40 @@ async function fetchDestinations(
...
- * @param jwt - The access token
+ * @param token - The access token or AuthAndExchangeTokens if you want to include the X-user-token for OAuth2UserTokenExchange.
  * @param destinationName - The name of the desired destination
  * @param options - Options to use by retrieving destinations
  * @returns A Promise resolving to the destination
  */
 export async function fetchDestination(
   destinationServiceUri: string,
-  jwt: string,
+  token: string | AuthAndExchangeTokens,
```
marikaner, line 138: 1. Not entirely sure, but isn't that a breaking change?
2. Do we really need two functions here? Couldn't the token thingy be done on the first line?

marikaner (approved):
Only minor comments. I have not tried to follow the flow step by step for all cases, I am trusting our tests for this.

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
FrankEssenberger: good idea.

packages/core/src/connectivity/scp-cf/destination/destination-service.ts
FrankEssenberger: I think it is not breaking since the string input is still possible. I needed a way to allow to bring the `user exchange header` inside the call of the `callDestinationService`. However there is such a strong coupling to the cache that I did it this way. I think about something better for a while

PR #793, opened 2020-12-08
fix: Fix order of parameters

marikaner:
Fixes the order of parameters. The parameters in the default api are - required params, optional params, depending on order in the spec file. The order in our api is - required path params, required body, required query params, optional path params, optional query params.

After I had already implemented that, I had an insightful conversation with @jjtang1985: We should probably have a parameter object instead of multiple parameters, because if this gets bigger the API gets quite ugly. That way the order wouldn't matter, we would have to only keep it to call the default api.

@FrankEssenberger @florian-richter - I left the reordering for now, but I will be happy to still improve this on this PR.

### DIFF
--- a/packages/openapi-generator/src/parser/document.ts
+++ b/packages/openapi-generator/src/parser/document.ts
@@ -34,9 +34,7 @@ export function parseAllOperations(
       ...allOperations,
       ...methods()
         .filter(method => method in pathDefinition)
-        .map(method =>
-          parseOperation(pattern, pathDefinition, method, refs)
-        )
+        .map(method => parseOperation(pattern, pathDefinition, method, refs))
     ],
     []
   );
--- a/packages/openapi-generator/src/parser/document.ts
+++ b/packages/openapi-generator/src/parser/document.ts
@@ -35,7 +35,6 @@ export function parseAllOperations(
       ...methods()
         .filter(method => method in pathDefinition)
         .map(method =>
-          // The line above makes sure that the pathDefinition[method] is not undefined
           parseOperation(pattern, pathDefinition, method, refs)
         )
     ],
--- a/packages/openapi-generator/src/parser/document.ts
+++ b/packages/openapi-generator/src/parser/document.ts
@@ -36,7 +36,7 @@ export function parseAllOperations(
         .filter(method => method in pathDefinition)
         .map(method =>
           // The line above makes sure that the pathDefinition[method] is not undefined
-          parseOperation(pattern, method, pathDefinition[method]!, refs)
+          parseOperation(pattern, pathDefinition, method, refs)
         )
     ],
     []
--- a/packages/openapi-generator/src/parser/operation.ts
+++ b/packages/openapi-generator/src/parser/operation.ts
@@ -4,7 +4,7 @@ import {
   camelCase,
   partition,
   pascalCase,
-  removeDuplicatesLeft
+  filterDuplicatesRight
 } from '@sap-cloud-sdk/util';
 import { Method, OpenApiOperation, OpenApiParameter } from '../openapi-types';
 import { getType } from './type-mapping';
@@ -76,11 +76,12 @@ export function parseParameters(
 ): OpenApiParameter[] {
   // TODO: What if this is a reference? What does OpenApi do?
   // TODO: What about oneof and other operations?
-  return removeDuplicatesLeft(
+  return filterDuplicatesRight(
     operation.parameters?.map(param => resolveObject(param, refs)) || [],
     (left, right) => left.name === right.name && left.in === right.in
   ).map(param => ({
     ...param,
+    // TODO: Check whether types are correct here and whether we can use union types here.
     type: getType(resolveObject(param.schema, refs)?.type?.toString())
   }));
 }
--- a/packages/openapi-generator/src/parser/operation.ts
+++ b/packages/openapi-generator/src/parser/operation.ts
@@ -1,6 +1,11 @@
 import { $Refs } from '@apidevtools/swagger-parser';
 import { OpenAPIV3 } from 'openapi-types';
-import { camelCase, partition, pascalCase } from '@sap-cloud-sdk/util';
+import {
+  camelCase,
+  partition,
+  pascalCase,
+  removeDuplicatesLeft
+} from '@sap-cloud-sdk/util';
 import { Method, OpenApiOperation, OpenApiParameter } from '../openapi-types';
 import { getType } from './type-mapping';
 import { parseRequestBody } from './request-body';
@@ -71,30 +76,13 @@ export function parseParameters(
 ): OpenApiParameter[] {
   // TODO: What if this is a reference? What does OpenApi do?
   // TODO: What about oneof and other operations?
-  return filterDuplicates(
-    operation.parameters?.map(param => resolveObject(param, refs)) || []
+  return removeDuplicatesLeft(
+    operation.parameters?.map(param => resolveObject(param, refs)) || [],
+    (left, right) => left.name === right.name && left.in === right.in
   ).map(param => ({
     ...param,
     type: getType(resolveObject(param.schema, refs)?.type?.toString())
   }));
-}
-
-export function filterDuplicates(
-  parameters: OpenAPIV3.ParameterObject[]
-): OpenAPIV3.ParameterObject[] {
-  return parameters.reduce(
-    (params: OpenAPIV3.ParameterObject[], currentParam) => {
-      const duplicateIndex = params.findIndex(
-        param =>
-          param.name === currentParam.name && param.in === currentParam.in
-      );
-      if (duplicateIndex >= 0) {
-        params.splice(duplicateIndex, 1);
-      }
-      return [...params, currentParam];
-    },
-    []
-  );
 }
 
 export function parseOperationName(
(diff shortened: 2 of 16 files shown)


packages/openapi-generator/src/parser/document.ts
```diff
@@ -36,7 +36,7 @@ export function parseAllOperations(
         .filter(method => method in pathDefinition)
         .map(method =>
           // The line above makes sure that the pathDefinition[method] is not undefined
-          parseOperation(pattern, method, pathDefinition[method]!, refs)
+          parseOperation(pattern, pathDefinition, method, refs)
```
FrankEssenberger, line 39: The comment refers to this filter: `.filter(method => method in pathDefinition)` right? So perhaps put it directly to the line or above it.

packages/openapi-generator/src/parser/document.ts
```diff
@@ -36,7 +36,7 @@ export function parseAllOperations(
         .filter(method => method in pathDefinition)
         .map(method =>
           // The line above makes sure that the pathDefinition[method] is not undefined
```
marikaner, line 38:
```suggestion
```

packages/openapi-generator/src/parser/document.ts
marikaner: I removed the comment, it was outdated anyways.

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -47,13 +71,29 @@ export function parseParameters(
...
   // TODO: What if this is a reference? What does OpenApi do?
   // TODO: What about oneof and other operations?
-  return (
-    operation.parameters
-      ?.map(param => resolveObject(param, refs))
-      .map(param => ({
-        ...param,
-        type: getType(resolveObject(param.schema, refs)?.type?.toString())
-      })) || []
+  return filterDuplicates(
```
FrankEssenberger, line 74: Minor: In the openapi-types the `type` and `in` attribute for the parameter are `string` type. I would prefer some type literals lik `'path' | 'query' |...` so that one has stronger typing on the potential types and locations.

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -47,13 +71,29 @@ export function parseParameters(
...
+  ).map(param => ({
+    ...param,
+    type: getType(resolveObject(param.schema, refs)?.type?.toString())
+  }));
+}
+
+export function filterDuplicates(
+  parameters: OpenAPIV3.ParameterObject[]
+): OpenAPIV3.ParameterObject[] {
+  return parameters.reduce(
```
FrankEssenberger, line 85: why you do not use a `parameters.filter(param=>param.name === currentParam.name && param.in === currentParam.in) ?

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -47,13 +71,29 @@ export function parseParameters(
...
-      })) || []
+  return filterDuplicates(
+    operation.parameters?.map(param => resolveObject(param, refs)) || []
+  ).map(param => ({
+    ...param,
+    type: getType(resolveObject(param.schema, refs)?.type?.toString())
+  }));
+}
+
+export function filterDuplicates(
```
FrankEssenberger, line 82: I would prefer to have a removeDuplicates() method in the array util where you can hand in a custom comparator function `comperator(a:T,b:T):boolean`. Then we can reuse it better.

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -80,17 +77,52 @@ interface Parameter {
...
-        name: operation.requestBody.parameterName,
-        type: operation.requestBody.parameterType,
-        required: operation.requestBody.required
-      }
-    ];
-  }
+function getApiFunctionSignatureParams(
+  operation: OpenApiOperation
+): Parameter[] {
+  // const parameters = operation.operationName //getAllParameters(operation);
```
FrankEssenberger, line 83: comment

test-packages/test-services/openapi/test-service/open-api.json
```diff
@@ -168,6 +167,120 @@
           "default"
         ]
       }
+    },
+    "/test-cases/parameters/required-parameters/{requiredPathItemPathParam}": {
```
FrankEssenberger, line 171: nice idea to seperate the test cases via pathes

test-packages/test-services/openapi/test-service/api.ts
```diff
@@ -27,5 +27,25 @@ export const TestServiceApi = {
...
+    DefaultApi,
+    'testCaseGetRequiredParameters',
+    requiredPathItemQueryParam,
+    requiredPathItemPathParam,
+    requiredQueryParam,
+    optionalQueryParam,
+    optionalPathItemQueryParam,
+    simpleTestEntity
+  ),
+  testCasePostRequiredParameters: (requiredPathItemPathParam: string, simpleTestEntity: SimpleTestEntity, requiredPathItemQueryParam: string, requiredQueryParam: string, optionalPathItemQueryParam?: string, optionalQueryParam?: string) => new OpenApiRequestBuilder<DefaultApi, 'testCasePostRequiredParameters'>(
```
FrankEssenberger, line 41: I am a bit confused by the final singnature. I would have expected from the description an object with keys and some keys parameters are optional. I know that the name is not enough because only name and where it is used (path,query). If I want not to give the `optionalQueryParam` and not `optionalPathItemQueryParam` how would I do that?

FrankEssenberger: Something like: 
```
{requiredPath:{paraName1:{},paraName2:{}},requiredQuery:{paraName1:{},paraName2:{}},optionPath:{paraName1?:{},paraName2?:{}}}
```
or do I have a not the right understanding?

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -80,17 +77,52 @@ interface Parameter {
...
+  const [optionalPathParams, optionalQueryParams] = partition(
+    optional,
+    param => param.in === 'path'
+  );
+
+  return [
+    ...requiredPathParams,
+    ...getRequestBodyParams(operation),
+    ...requiredQueryParams,
+    ...optionalPathParams,
```
FrankEssenberger, line 103: see comment below on the flat flist of parameters.

test-packages/test-services/openapi/test-service/api.ts
FrankEssenberger: or are use supposed to fill the unwanted values with `undefinede`?

FrankEssenberger (changes requested):
Just one relevant question on the method signature in the API object.

packages/openapi-generator/src/parser/operation.ts
marikaner: Good idea. Done.

packages/openapi-generator/src/wrapper-files/api-file.ts
marikaner: Done.

test-packages/test-services/openapi/test-service/open-api.json
marikaner: Yeah, it's still quite complex...

test-packages/test-services/openapi/test-service/api.ts
marikaner: Currently the structure is: `fn(...requiredPathParams, ...requiredBody, ...requiredQueryParams, ...optionalPathParams, ...optionalBody, ...optionalQueryParams)`.

I was planning on doing the following, but I wanted to hear your opinion first:
```
fn(args: {
  requiredParam: string,
  optionalParam?: number
})
```
That way the order is completely irrelevant.

marikaner: Consider duplicates.

marikaner:
Hey @FrankEssenberger,

I implemented the variant with the object for parameters now. To be honest, I am not 100% happy with it yet, but I think it will do for this PR. The issue, that I have with the current implementation is, that there is no difference between path parameters, query parameters and the request body.

I could imagine something like this:
```ts
args: {
  pathParameters: {
    requiredPathParam: string,
    optionalPathParam? string
  },
  queryParameters: {
    requiredQueryParam: string,
    optionalQueryParam? string
  },
  body: BodyType
}
```

But then for simple cases the usage might not be optimal. Something in that direction:
```ts
fn({ pathParameters: { requiredPathParam: 'test' } })
```

FrankEssenberger:
> Hey @FrankEssenberger,
> 
> I implemented the variant with the object for parameters now. To be honest, I am not 100% happy with it yet, but I think it will do for this PR. The issue, that I have with the current implementation is, that there is no difference between path parameters, query parameters and the request body.
> 
> I could imagine something like this:
> 
> ```ts
> args: {
>   path: {
>     requiredPathParam: string,
>     optionalPathParam? string
>   },
>   query: {
>     requiredQueryParam: string,
>     optionalQueryParam? string
>   },
>   body: BodyType
> }
> ```


I share your concern. For little parameters it is overkill. I think it would be best to resolve name conflicts via some postfix and keep the signature clear. I also hope that conflicts are the exception because it is not a good API if you have two parameters with the same name as path and query. I mean how distinguish them? But for this PR I agree it is good enough.

FrankEssenberger (approved):
LGTM - duplicate parameter case will be tackled in a next PR.

PR #799, opened 2020-12-09
chore: adjust canary build versions

FrankEssenberger:
use a date timestamp before the commit id so that we have the info which code is in canary but the string sorting takes the the latest version first.

Try out: npm i @sap-cloud-sdk/util@1.32.2-20201209171510.47+6cadd29bb

In the package json I see:
```
  "_from": "@sap-cloud-sdk/util@1.32.2-20201209171510.47+6cadd29bb",
  "_id": "@sap-cloud-sdk/util@1.32.2-20201209171510.47",
...
```
So we have the build info in the from and the id containing the time. What bothers me is the `.47` I do not know where this comes from?

I have also published by hand already a version as you can see above and I would deprecate all the old canary version since they did not follow semantic versioning in a sense that the prerelease string (commit-id) was not ordered by release time. If you say this is OK I would use the script in this PR and let it run. Then I would remove the script.


scripts/deprecate-old-prerelease-versions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const packageName = '@sap-cloud-sdk/core';
+// const packageName ='@sap-cloud-sdk/util'
+// const packageName ='@sap-cloud-sdk/analytics'
+// const packageName ='@sap-cloud-sdk/test-util'
+// const packageName ='@sap-cloud-sdk/generator'
+
+const dryRun = true;
+
+async function deprecateVersions() {
```
jjtang1985, line 14: I guess this is a convenient tool for clean up?

jjtang1985 (approved):
I guess this is the candidate `1.32.2-20201209171510.47`?

In general, the version should be unique, because even you have 2 versions at the same time, the index should help. I wonder, the commit id is not that useful and we can git rid of it like what you did.

If the script is also part of this PR, then make sure, the stable version is protected, because they will not match the new pattern.

scripts/deprecate-old-prerelease-versions.ts
FrankEssenberger: yes I would like to use this to deprecate the old prerelease versions.

marikaner (approved):
LGTM

jjtang1985 (approved):

PR #801, opened 2020-12-10
fix: OAuth2ClientCredentials

jjtang1985:
`getTargetUri()` is used by `executeXsuaaPostRequest()`, which is used in 2 cases:
1. fetching service token (e.g., provider token)
2. fetching token for `OAuth2ClientCredentials` type.

In case 1, the base url is from the xsuaa service e.g., `https://xxx.ondemand.com`, where `/oauth/token` needs to be added.
In case 2, the whole url is from the destination service, where the user is responsible to provide the whole url e.g., `https://xxx.ondemand.com/oauth/token?grant_type=client_credentials`. So the no modification is needed.

This PR aims to solve the known issue with the minimal changes since both cases are not covered by e2e tests.
We should improve it once the e2e tests are there.


FrankEssenberger (approved):
LGTM

PR #808, opened 2020-12-11
[REST] feat: support yaml as input

jjtang1985:


scripts/generate-test-services.ts
```diff
@@ -137,8 +137,32 @@ if (arg === 'e2e' || arg === 'all') {
 
 if (arg === 'openapi' || arg === 'rest' || arg === 'all') {
   generateOpenApi({
-    inputDir: path.resolve('test-resources', 'openapi-service-specs'),
-    outputDir: path.resolve('test-packages', 'test-services', 'openapi'),
+    inputDir: path.resolve(
+      'test-resources',
+      'openapi-service-specs',
+      'test-service'
```
FrankEssenberger, line 143: perhaps for consistency call it `test-service-json`?

FrankEssenberger (approved):
LGTM - consider my comment about naming the two folders. Also perhaps if we change something in one spec file let's say JSON is there an automatic way to put this also in the yaml file?

scripts/generate-test-services.ts
jjtang1985: Fair point. I thought at least in the beta development, we rely on the json version. Just like the odata test entity, `test-entity`/`test-service` are the default names that we want to maintain.

jjtang1985:
> Also perhaps if we change something in one spec file let's say JSON is there an automatic way to put this also in the yaml file?

If we would like to maintain/sync both json and yaml, the automation is necessary. For example, we can call a script in the pipeline to check whether the json and the yaml files define the same specification. To synchronise, we can alway maintain one of these two (e.g., yaml), and create another script to update the json.

However, I'm not sure whether we should do this _now_. 
First of all, the test service will become huger as we found more complicated cases in the workflow service and might be even more when going GA. Keeping two variants with the same spec will duplicate the spec file itself and the generated files. 
Secondly, if you consider openapi version `2.0` and `3.0`, we then have 2*2=4 groups. Having more spec files and generated files is fine, but if half of them contains the redundant information, I would argue we can start doing this, when there are really some problems.

PR #814, opened 2020-12-14
chore: Support swagger and yml

marikaner:
Introduced a document converter that prepares the document for parsing returnin g a json openapi spec.


marikaner:
Opened a fixed version

Issue #816, opened 2020-12-14
getTargetUri in /packages/core/src/connectivity/scp-cf/xsuaa-service.ts automatically appends /oauth/token to the uri


blaart:
We are using a token server of Microsoft for a user token grant request. 
the url is like: https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token

see also: https://docs.microsoft.com/en-us/graph/auth-v2-user#3-get-a-token

We really want to use the sdk to resolve our destinations, but it does not work for this url as the getTargetUri function in /packages/core/src/connectivity/scp-cf/xsuaa-service.ts automatically appends /oauth/token to the uri, which results in https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token/oauth/token. Which results in a 404 Not Found.

I expect the given Token Service URL in the destination not te be changed.

blaart:
Never mind, I now see this pull request: https://github.com/SAP/cloud-sdk-js/pull/792

blaart:
I now see that this was fixed a few days ago in 1.33

artemkovalyov:
We're happy it worked for you @blaart. Let us know if anything else comes up.

blaart:
Hi Artem, Thanks for your reply, I based my conclusion on the changelog. While testing I still see that /oauth/token is added. 
I think this is the problem:
In /packages/core/src/connectivity/scp-cf/xsuaa-service.ts

In the post function the getTargetUri is still called: https://github.com/SAP/cloud-sdk-js/blob/70c0b652b63764b2a74f46bb1a6b2d568c6a37bf/packages/core/src/connectivity/scp-cf/xsuaa-service.ts#L231

Should it be fixed when you move this call to here: https://github.com/SAP/cloud-sdk-js/blob/70c0b652b63764b2a74f46bb1a6b2d568c6a37bf/packages/core/src/connectivity/scp-cf/xsuaa-service.ts#L221
?

jjtang1985:
Hi @blaart, the fix that you proposed makes sense, which should work as we planned a similar way. However, to do that, we would like to build our end to end test as our quality assurance, which was also planned in the backlog.

Now, since the previous workaround does not solve your issue, I would like ask whether your application is a PoC or live system so that we can discuss again in the refinement meeting.

Best regards,
Junjie

blaart:
[...]

PR #823, opened 2020-12-15
fix: Append `oauth/token` only for XSUAA urls

FrankEssenberger:
We had a bug addressed in #816. The problem was that we were appending `oauth/token` also if we got the URL from the destination service. I adjusted the three main methods of `xsuaa-service` so that they now accept all `string 
 |XSUaaServiceCredentials` as input. In case of string it is assumed as the URL of the service token endpoint and no adjustments are made. If it is `XSUaaServiceCredentials` the suffix is appended to the URL.


test-packages/integration-tests/test/v2/request-builder.spec.ts
```diff
@@ -531,7 +531,7 @@ describe('Request Builder', () => {
       URL: url,
       authTokens: [],
       Authentication: 'OAuth2ClientCredentials',
-      tokenServiceURL: 'https://token.example.com',
+      tokenServiceURL: 'https://token.example.com/some/token/endpoint',
```
jjtang1985, line 534: For the case that the token url of the destination is used, we only have the test in the integration tests. Could we maybe add one in the unit test?

jjtang1985 (approved):
LGTM. Only a request about a unit test if it is missing.

test-packages/integration-tests/test/v2/request-builder.spec.ts
FrankEssenberger: I added a test in the xsuaa spec file

jjtang1985 (approved):

PR #824, opened 2020-12-15
feat: [REST] support enum type

jjtang1985:
In the openapi specification, it's also possible to list all the potential values as an array with the key `enum`. In the example below, the vanilla openapi client will generate code with parameter `enumParameter?: 'value1' | 'value2',`. Aligned with the vanilla client, the SDK client should use the same type instead of `string` to fix the compilation errors.
```
          {
             "name": "enumParameter",
             "in": "query",
             "description": "A parameter of type string with enum",
             "required": false,
             "schema": {
               "type": "string",
               "enum": [
                 "value1",
                 "value2"
               ]
             }
           }
```

### DIFF
--- a/packages/openapi-generator/src/parser/parameters.spec.ts
+++ b/packages/openapi-generator/src/parser/parameters.spec.ts
@@ -27,12 +27,20 @@ describe('parseParameters', () => {
       name: 'numberParam',
       schema: { type: 'integer' }
     };
-    const enumParam = {
+    const enumStringParam = {
       in: 'query',
-      name: 'enumParam',
+      name: 'enumStringParam',
+      schema: {
+        type: 'string',
+        enum: ['value1', 'value2']
+      }
+    };
+    const enumNumberParam = {
+      in: 'query',
+      name: 'enumNumberParam',
       schema: {
         type: 'integer',
-        enum: ['value1', 'value2']
+        enum: [1, 2]
       }
     };
 
@@ -44,7 +52,8 @@ describe('parseParameters', () => {
             anyParamWithSchema,
             stringParam,
             numberParam,
-            enumParam
+            enumStringParam,
+            enumNumberParam
           ]
         },
         await createRefs()
@@ -54,7 +63,8 @@ describe('parseParameters', () => {
       { ...anyParamWithSchema, type: 'any' },
       { ...stringParam, type: 'string' },
       { ...numberParam, type: 'number' },
-      { ...enumParam, type: 'number', enum: ['value1', 'value2'] }
+      { ...enumStringParam, type: 'string' },
+      { ...enumNumberParam, type: 'number' }
     ]);
   });
 
--- a/packages/openapi-generator/src/parser/parameters.spec.ts
+++ b/packages/openapi-generator/src/parser/parameters.spec.ts
@@ -27,6 +27,14 @@ describe('parseParameters', () => {
       name: 'numberParam',
       schema: { type: 'integer' }
     };
+    const enumParam = {
+      in: 'query',
+      name: 'enumParam',
+      schema: {
+        type: 'integer',
+        enum: ['value1', 'value2']
+      }
+    };
 
     expect(
       parseParameters(
@@ -35,7 +43,8 @@ describe('parseParameters', () => {
             anyParamNoSchema,
             anyParamWithSchema,
             stringParam,
-            numberParam
+            numberParam,
+            enumParam
           ]
         },
         await createRefs()
@@ -44,7 +53,8 @@ describe('parseParameters', () => {
       { ...anyParamNoSchema, type: 'any' },
       { ...anyParamWithSchema, type: 'any' },
       { ...stringParam, type: 'string' },
-      { ...numberParam, type: 'number' }
+      { ...numberParam, type: 'number' },
+      { ...enumParam, type: 'number', enum: ['value1', 'value2'] }
     ]);
   });
 
--- a/packages/openapi-generator/src/parser/parameters.spec.ts
+++ b/packages/openapi-generator/src/parser/parameters.spec.ts
@@ -43,6 +43,14 @@ describe('parseParameters', () => {
         enum: [1, 2]
       }
     };
+    const enumBooleanParam = {
+      in: 'query',
+      name: 'enumBooleanParam',
+      schema: {
+        type: 'boolean',
+        enum: ['true', 'false']
+      }
+    };
 
     expect(
       parseParameters(
@@ -53,7 +61,8 @@ describe('parseParameters', () => {
             stringParam,
             numberParam,
             enumStringParam,
-            enumNumberParam
+            enumNumberParam,
+            enumBooleanParam
           ]
         },
         await createRefs()
@@ -64,7 +73,8 @@ describe('parseParameters', () => {
       { ...stringParam, type: 'string' },
       { ...numberParam, type: 'number' },
       { ...enumStringParam, type: "'value1' | 'value2'" },
-      { ...enumNumberParam, type: '1 | 2' }
+      { ...enumNumberParam, type: '1 | 2' },
+      { ...enumBooleanParam, type: 'boolean' }
     ]);
   });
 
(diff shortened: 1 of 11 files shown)


marikaner:
Please actually provide a description ;) There might also be users who are interested in what is going on.

marikaner:
Oh and please also adhere to the conventional commit guidelines in the title (should not start with [REST]).

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -106,3 +109,9 @@ function getRequestBodyParams(operation: OpenApiOperation): Parameter[] {
       ]
     : [];
 }
+
+function getType(parameter: Parameter) {
```
marikaner, line 113: This function has no return type.

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -106,3 +109,9 @@ function getRequestBodyParams(operation: OpenApiOperation): Parameter[] {
...
+  return parameter.enum
+    ? parameter.enum.map(e => `'${e}'`).join(' | ')
```
marikaner, line 115: I guess this is only correct for strings, but not for integers and booleans (and maybe others). Consider all possible types.

jjtang1985:
@marikaner , will you fine with the new PR name? I would like to put the `Rest` keyword in the `[]` like a tag to differentiate between `rest` and `odata`.

packages/openapi-generator/src/parser/parameters.ts
```diff
@@ -26,12 +26,25 @@ export function parseParameters(
...
+): OpenApiParameter {
+  const enumValue = resolveObject(param.schema, refs)?.enum;
+  const ret = {
     ...param,
     name: camelCase(param.name),
     // TODO: Check whether types are correct here and whether we can use union types here.
-    type: getType(resolveObject(param.schema, refs)?.type?.toString())
-  }));
+    type: getType(resolveObject(param.schema, refs)?.type?.toString()),
+    enum: resolveObject(param.schema, refs)?.enum
```
marikaner, line 42: Is there a good reason to differentiate these here? I couldn't find a case where anywhere further down in the code, where both are relevant at the same time. I think we should parse the enum type here already and transform it to a union type. This would also simplify the code for the wrapper.

packages/openapi-generator/src/parser/parameters.ts
```diff
@@ -26,12 +26,25 @@ export function parseParameters(
   parameters = filterDuplicateParams(parameters);
   parameters = reorderParameters(parameters);
   parameters = renameEquallyNamedParams(parameters);
-  return parameters.map(param => ({
+  return parameters.map(param => toOpenApiParameter(param, refs));
+}
+
+function toOpenApiParameter(
```
marikaner, line 32: I think it is a good rule of thumb to name functions using a verb. I also think it is not relevant to mention OpenApi in the function name
```suggestion
function parseParameter(
```

marikaner (changes requested):
I have 2 main concerns here:
1. I think the enums don't work for integers or maybe even other types.
2. I think it would be better to move the transformation from "enum" type to union type into the parser (see the comments).

jjtang1985:
I tested the types below by running the open api generator with the test-service.json spec, so we understand whether enum with certain types are supported by the open api generator.
- string: supported, see tests
- number: supported, see tests
- date: not supported, compilation error from the open api generator
- date-time: same as date
- boolean: supported, but the type is boolean instead of union type

marikaner:
> I tested the types below by running the open api generator with the test-service.json spec, so we understand whether enum with certain types are supported by the open api generator.
> 
> * string: supported, `enumString: 'a' | 'b'`
> * number: supported, `enumNumber: 1 | 2`
> * date: not supported, compilation error from the open api generator
> * date-time: same as date
> * boolean: supported, but the type is boolean instead of union type: `enumBoolean: boolean`

Thanks for this investigation, please add date / date-time to the openapi generator bug list (in the backlog).
"boolean" makes sense, we should also have the same behaviour (a boolean is a union type of true | false).

packages/openapi-generator/src/parser/parameters.spec.ts
```diff
@@ -27,6 +27,22 @@ describe('parseParameters', () => {
...
     };
+    const enumStringParam = {
+      in: 'query',
+      name: 'enumStringParam',
+      schema: {
+        type: 'string',
+        enum: ['value1', 'value2']
+      }
+    };
+    const enumNumberParam = {
```
marikaner, line 38: Maybe also add boolean here?

packages/openapi-generator/src/parser/parameters.ts
```diff
@@ -30,10 +30,38 @@ export function parseParameters(
...
 }
 
+function parseType(param: OpenAPIV3.ParameterObject, refs: $Refs) {
+  const originalType = resolveObject(param.schema, refs)?.type?.toString();
+  const tsType = getType(originalType);
+  const enumValue = resolveObject(param.schema, refs)?.enum;
+  if (enumValue && isValidEnumType(tsType)) {
+    return enumAsUnionType(tsType, enumValue, originalType);
+  }
+  return tsType;
```
marikaner, line 44: Consider a ternary:
```suggestion
  return enumValue && isValidEnumType(tsType) ? enumAsUnionType(tsType, enumValue, originalType) : tsType;
```

packages/openapi-generator/src/wrapper-files/api-file.spec.ts
```diff
@@ -15,6 +15,16 @@ describe('api-file', () => {
               name: 'limit',
               type: 'number'
             },
+            {
```
marikaner, line 18: I think this test is overkill, because you are not testing anything that the api-file module does. If you remember from our discussion on our [test strategy](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/adr/0010-test-strategy.md#unit-tests), we agreed that everything should be tested **once**. This is already tested in the parser, so I would rather like to remove this.

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -83,6 +83,7 @@ interface Parameter {
   type: string;
   name: string;
   required?: boolean;
+  enum?: string[];
```
marikaner, line 86: This is outdated, right?

tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
     "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
     "lib": [
-      "esnext"
+      "esnext",
+      "dom"
```
marikaner, line 8: Why?

test-packages/test-services/openapi/test-service/api.ts
```diff
@@ -12,14 +12,24 @@ export const TestServiceApi = {
     stringParameter?: string,
     integerParameter?: number,
     dollarParameter?: string,
-    dotParameter?: string
+    dotParameter?: string,
+    enumStringParameter?: 'value1' | 'value2',
+    enumInt32Parameter?: 1 | 2,
+    enumFloatParameter?: 1 | 2,
+    enumDoubleParameter?: 1 | 2,
```
marikaner, line 19: Consider whether we really need all three cases for number - seems trivial to me and it makes the definition of the testentity type more complex, maybe one or two are enough? I have no strong opinion here, up to you.

marikaner (commented):
I would like to make sure we don't test the same things twice in different locations. Other than that looks good to me.

tsconfig.json
jjtang1985: There was an issue when the dom was not added. Now it fixes by itself.

test-packages/test-services/openapi/test-service/api.ts
jjtang1985: I would like to keep two: `int` and `double` so we know both of these are tested, while `float` might be useless.

jjtang1985:
I opened an issue in the open api repo and added the link to the bug list.

marikaner (approved):
LGTM

PR #828, opened 2020-12-17
chore: Remove openapi tools json

marikaner:
Locks the cwd of the execution of the java openapi generator to the generator directory. Prevents creation of a new openapi-tools.json and thereby ensures usage of the version we specified.


FrankEssenberger (approved):
Great, than this is also properly fixed.

PR #829, opened 2020-12-17
feat: Add and use ErrorWithCause class

marikaner:
This adds the `ErrorWithCause` class and uses it throughout the SDK.
This is the [ADR](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/adr/0016-structure-of-error-object.md#option-b-include-the-original-error) for reference.

### DIFF
--- a/packages/util/src/error.ts
+++ b/packages/util/src/error.ts
@@ -23,7 +23,8 @@ export class ErrorWithCause extends Error {
    * @param message Error message.
    * @param cause Original error, causing this error.
    */
-  constructor(message: string, public cause?: Error) {
+  constructor(message: string, public cause: Error) {
+    // There is an issue with the prototype chain when extending from Error: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#support-for-newtarget
     super(message); // 'Error' breaks prototype chain here
     Object.setPrototypeOf(this, new.target.prototype); // restore prototype chain
     this.name = 'ErrorWithCause';
@@ -39,7 +40,7 @@ export class ErrorWithCause extends Error {
    * If there are multiple errors caused one by another, the root cause is the first error that occurred.
    * In case there is no root cause.
    */
-  get rootCause(): Error | undefined {
+  get rootCause(): Error {
     return isErrorWithCause(this.cause) ? this.cause.rootCause : this.cause;
   }
 }
@@ -49,8 +50,6 @@ export class ErrorWithCause extends Error {
  * @param err An error.
  * @returns Whether the given error is of type ErrorWithCause.
  */
-export function isErrorWithCause(
-  err: Error | undefined
-): err is ErrorWithCause {
+export function isErrorWithCause(err: Error): err is ErrorWithCause {
   return err?.name === 'ErrorWithCause';
 }
--- a/packages/util/src/error.ts
+++ b/packages/util/src/error.ts
@@ -15,7 +15,7 @@ export class ErrorWithCause extends Error {
 
     // Stack is a non-standard property according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types
     if (this.stack && cause?.stack) {
-      this.stack = this.stack + '\nCaused by:\n' + cause.stack;
+      this.stack = `${this.stack}\nCaused by:\n${cause.stack}`;
     }
   }
 
--- a/packages/util/src/error.ts
+++ b/packages/util/src/error.ts
@@ -1,5 +1,5 @@
 /**
- * @deprecated Since v1.34.0. Use `ErrorWithCause` class instead.
+ * @deprecated Since v1.34.0. Use [[ErrorWithCause]] instead.
  * Creates a new Error and adds the stack trace of the original error to the stack trace of the new one.
  * @param message Error message.
  * @param cause Original error, causing the new error.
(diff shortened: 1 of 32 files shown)


packages/util/src/error.ts
```diff
@@ -6,3 +6,24 @@ export function errorWithCause(message: string, cause: Error): Error {
...
+
+export class ErrorWithCause extends Error {
+  constructor(message: string, public cause?: Error) {
+    super(message); // 'Error' breaks prototype chain here
+    Object.setPrototypeOf(this, new.target.prototype); // restore prototype chain
+    this.name = 'ErrorWithCause';
+
+    // Stack is a non-standard property according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types
+    if (this.stack && cause?.stack) {
+      this.stack = this.stack + '\nCaused by:\n' + cause.stack;
```
florian-richter, line 18:
```suggestion
      this.stack = `${this.stack}\nCaused by:\n${cause.stack}`;
```

packages/util/src/error.ts
```diff
@@ -6,3 +6,24 @@ export function errorWithCause(message: string, cause: Error): Error {
   }
   return newError;
 }
+
+export class ErrorWithCause extends Error {
+  constructor(message: string, public cause?: Error) {
+    super(message); // 'Error' breaks prototype chain here
+    Object.setPrototypeOf(this, new.target.prototype); // restore prototype chain
+    this.name = 'ErrorWithCause';
```
florian-richter, line 14: A link to more information on the prototype problems would be helpful here

packages/util/src/error.ts
```diff
@@ -6,3 +6,24 @@ export function errorWithCause(message: string, cause: Error): Error {
...
+      this.stack = this.stack + '\nCaused by:\n' + cause.stack;
+    }
+  }
+
+  get rootCause(): Error {
+    return isErrorWithCause(this.cause) ? this.cause.rootCause : this.cause || this;
+  }
+}
+
+export function isErrorWithCause(err: Error | undefined): err is ErrorWithCause {
```
florian-richter, line 27:
```suggestion
export function isErrorWithCause(err?: Error): boolean {
```
I dont know this `:err is ErrorWithCause` syntax. What does this do?

florian-richter: Nevermind, I think I figured it out. Depending on the input, the return value will be known at compile time, right?

florian-richter (commented):
Naming is fine from my perspective. We definitely need documentation on ErrorWithCause though.

jjtang1985 (commented):
I'm fine with the naming.

packages/util/src/error.ts
marikaner: Yes. It is called a type guard.

packages/util/src/error.ts
marikaner: Done.

packages/util/src/error.ts
```diff
@@ -6,3 +13,43 @@ export function errorWithCause(message: string, cause: Error): Error {
...
+    }
+  }
+
+  /**
+   * Root cause of the error.
+   * If there are multiple errors caused one by another, the root cause is the first error that occurred.
+   * In case there is no root cause.
+   */
+  get rootCause(): Error {
+    return isErrorWithCause(this.cause) ? this.cause.rootCause : this.cause;
```
FrankEssenberger, line 44: Minor: will run in a loop if we create a cyclic chain. But unlikely.

FrankEssenberger: Question: Does JSON.stringify also evaluate the getter? This would be a nice in case somebody serializes such an error e.g. in a logger this will happen at a certain point right? Or how does our provided winston logger hanndle a `logger.error(err)` for this ErrorWithCause and some chain of errors inside.

FrankEssenberger (approved):
LGTM only minor comments

packages/util/src/error.ts
marikaner: > Minor: will run in a loop if we create a cyclic chain. But unlikely.

Good point, this is the same behaviour as in the old errorWithCause function and it should probably never happen.
The only way I can think of achieving this would be:
```ts
const cause = new ErrorWithCause('cause', new Error('root cause'));
const error = new ErrorWithCause('error', cause);
// Creates circular chain
error.cause = cause;
```

marikaner: > Question: Does JSON.stringify also evaluate the getter? This would be a nice in case somebody serializes such an error e.g. in a logger this will happen at a certain point right? Or how does our provided winston logger hanndle a `logger.error(err)` for this ErrorWithCause and some chain of errors inside.

I checked and those are two separate issues I think.

1. JSON.stringify gives me:
```ts
JSON.stringify(
    new ErrorWithCause('ErrorWithCause', new Error('root cause')),
    null,
    2
  )
// result
// {
//   "cause": {},
//   "name": "ErrorWithCause"
// }
```

There are two aspects to it.
* a) Yes, there is no rootCause on it, I could add it.
* b) Can you see what JSON.stringify does for Error? -.-
  An example:
  ```ts
  JSON.stringify(new Error('root cause'), null, 2)
  // result
  // {}
  ```
  Not sure whether we should do anything about this. On the one hand it's nice (and not difficult to implement) - on the other hand the behaviour is different from the standard error. Would you call JSON.stringify on all errors (Error and ErrorWithCause) you would get different results.

2. What does the logger do? It simply prints the error message. So for `new Error('root cause')` that would be `'root cause'`. This is the same `errorWithCause` and `ErrorWithCause`, e. g. `new ErrorWithCause('the error', new Error('root cause'))` results in `'the error'`. I guess that is not optimal, it might make sense to improve that, but I don't think this is related to this PR.

FrankEssenberger: Ok fine for me. Perhaps we make a item to the backlog to improve the logging if our `ErrorWithCause` is logged by our loggers.

FrankEssenberger: For the cyclic thing I would wait until this actually happens.

marikaner: Or even if error are logged for that matter...

PR #830, opened 2020-12-17
feat: [REST] Support inline schema

jjtang1985:
Support inline schema for:
- non array schema object
- array schema object

- [x] Discussion: whether rename the parameter name of the request body => use `body`


test-resources/odata-service-specs-e2e/swagger-yaml-service.yml
```diff
@@ -0,0 +1,55 @@
+info:
```
marikaner, line 1: I guess you committed this by mistake?

test-resources/odata-service-specs-e2e/test-service.json
```diff
@@ -0,0 +1,522 @@
+{
```
marikaner, line 1: Also by mistake?

marikaner (dismissed):
I left some comments.

marikaner (approved):
I think we still run into some issues in some cases that are not covered here. Let's merge for now and make a release and then later improve some additional cases.

PR #836, opened 2020-12-21
chore: Add the test select in expand

FrankEssenberger:
While investigating the expand/select alignment ticket for OData v2/v4 I found no test checking that a select within a expand is build correctly. I added a tests - if there are already tests which I missed feel free to remove them.


packages/core/src/odata-v2/request-builder/get-all-request-builder.spec.ts
```diff
@@ -39,6 +42,19 @@ describe('GetAllRequestBuilder', () => {
         .url(defaultDestination);
       expect(actual).toBe(expected);
     });
+
+    it('is built correctly select in expand', async () => {
```
marikaner, line 46: I am not sure that I fully understand the meaning of this, can you rephrase it to be a bit clearer?

packages/core/src/odata-v4/request-builder/get-all-request-builder.spec.ts
```diff
@@ -52,6 +53,17 @@ describe('GetAllRequestBuilder', () => {
     });
   });
 
+  it('builds correctly select withing expand.', async () => {
```
marikaner, line 56: 🤪

marikaner (changes requested):
The tests look fine, but I don't get the descriptions.

packages/core/src/odata-v4/request-builder/get-all-request-builder.spec.ts
```diff
@@ -52,6 +53,17 @@ describe('GetAllRequestBuilder', () => {
     });
   });
 
+  it('is built correctly for selects inside of an expand.', async () => {
```
marikaner, line 56: I will go ahead and approve already, but consider whether this can be still phrased differently as have discussed in our "test strategy" meeting, that "is built correctly" is not very speaking. I can see that there are many tests like this in this file, so I wouldn't be too strict about this.

marikaner (approved):
One minor comment, but can also be done in a bigger "offensive" later.

PR #848, opened 2020-12-23
chore: change internal logic, remove base suffix internally

tomfrenken:


marikaner (approved):
LGTM! I think we should also rename the files from -base to something else. But not on this PR ;)

Issue #849, opened 2020-12-23
Custom field in navigation of deep create possible?


stefania-santimbrean:
Hello,

I am doing a deep create using SAP Cloud SDK and in the navigation entity I have a custom field which is not saved.

Before `PurchaseOrder.requestBuilder().create(purchaseOrder).execute(dest);` is executed if I do `purchaseOrder.toPurchaseOrderItem[0].getCustomFields()` I can see my custom field. But when I check the Purchase Order in the SAP S/4HANA Cloud system it's not there.

It works if I create PurchaseOrderItem. I can see that the custom field is saved.

Could you please tell me if this functionality is supported or if I'm missing something? If there is already an issue/question which covers this scenario, please point me in the right direction. 

Thank you,
Best regards,
Stefania

marikaner:
Hey @stefania-santimbrean,

I tried to reproduce this and found that this in fact is a bug in our serialization. I will let you know once this is fixed.

Regards,
Marika

stefania-santimbrean:
Hello @marikaner ,

Thanks for taking a look at the issue!

Do you happen to know when the fix might be available? A general timeframe? I am asking so I know if to wait for the fix or try workarounds.

Best regards,
Stefania

marikaner:
Hey @stefania-santimbrean,

this is no promise, but we have discussed a possibility of making a new release this week. It depends a bit on how many changes will need to be done on the PR I opened. I will let you know as soon as I know more.

stefania-santimbrean:
Hello @marikaner ,

Thank you so much for the support!!

One more question on the topic: What about batch mode? Are the custom fields in the navigation entity supported when using batch mode?

I tried to use batch mode and the custom field is not saved. Will your fix also solve this in batch mode?

Best regards and thank you again,
Stefania

marikaner:
Hey @stefania-santimbrean,

we provided a fix in version 1.35.0 of the SDK. This should also solve the same issue in batch mode. Let me know in case you still encounter issues. I will go ahead and close this issue for now.

stefania-santimbrean:
@marikaner thank you! Deep create works with custom fields now perfectly!

PR #854, opened 2020-12-28
fix: Serialize custom fields on linked entities

marikaner:
Fix the serialization on linked entities. Related to #849.
When doing this I realized, that we have some issues with linked entities, making it hard to fix this without changing some internal handling of remote state and serialization. I ended up with an optional parameter for the serialization. The `diff` parameter indicates whether the whole entity or just the changed properties should be serialized.

Background: with this it is possible to serialize the entity using some functionality that the entity provides and serializing the diff only, by simply skipping the unchanged properties. Before we already passed the diff to this function, but this diff was no entity (although the type told us so - I fixed that). Therefore at the point where I wanted to serialize links the entity class information was lost.


packages/core/src/odata-v4/request-builder/create-request-builder.spec.ts
```diff
@@ -47,6 +47,8 @@ describe('CreateRequestBuilderV4', () => {
     testPostRequestOutcome(actual, entity.setOrInitializeRemoteState());
   });
 
+  // TODO: This test is quite useless and wrong. It does not fail even if we remove the multilink.
```
marikaner, line 50: This is not a nice comment, we should remove it ;)

marikaner: @jjtang1985 I am not sure what to do with this file. To be honest it seems to me that all the tests are quite useless. We are basically checking that the deserializer works. Which is already checked in the deserialize

packages/core/src/odata-v2/request-builder/update-request-builder.spec.ts
```diff
@@ -290,12 +290,11 @@ describe('UpdateRequestBuilder', () => {
     expect(actual['remoteState']).toEqual(entity);
   });
 
-  it('warns if navigaton properties are sent', async () => {
+  it.only('warns if navigaton properties are sent', async () => {
```
jjtang1985, line 293:
```suggestion
  it('warns if navigaton properties are sent', async () => {
```

packages/core/src/odata-v2/entity-serializer.spec.ts
```diff
@@ -111,43 +129,23 @@ describe('entity-serializer', () => {
...
-      ComplexTypeProperty: {
-        StringProperty: complexTypeProp.stringProperty,
-        BooleanProperty: complexTypeProp.booleanProperty,
-        Int16Property: complexTypeProp.int16Property
-      },
-      StringProperty: stringProperty
-    });
-  });
-
-  it('should serialize an entity with custom field', () => {
```
jjtang1985, line 137: I personally would keep these two tests.

packages/core/src/odata-v2/entity.spec.ts
```diff
@@ -1,114 +0,0 @@
-import { TestEntity } from '../../test/test-util/test-services/v2/test-service';
```
jjtang1985, line 1: I assume you move the tests from here to the common.

packages/core/src/odata-v4/request-builder/create-request-builder.spec.ts
jjtang1985: Would you fix this test or make a follow-up? 😄

packages/core/src/odata-v4/entity.spec.ts
```diff
@@ -55,8 +55,8 @@ describe('entity v4', () => {
       })
       .build();
   });
-  it('getCurrentMapKeys', () => {
-    const currentState = entity.getCurrentMapKeys();
+  it('asObject', () => {
```
jjtang1985, line 58: in general, `asObject` is not a v4 specific function. Should we keep it in the common or v2 folder?

packages/core/src/odata-v4/entity-serializer.spec.ts
```diff
@@ -117,4 +118,39 @@ describe('entity-serializer', () => {
...
+      )
+      .build();
+    expect(serializeEntityV4(testEntity, TestEntity)).toEqual({
+      to_SingleLink: {
+        custom: 'custom'
+      }
+    });
+  });
+
+  it('should consider only changed items when diff is true', () => {
```
jjtang1985, line 137: These are not v4 specific, right? Could we move them to v2/common then?

jjtang1985 (changes requested):
Looks good, please consider my comments.

packages/core/src/odata-v4/request-builder/create-request-builder.spec.ts
marikaner: The question is - what should be tested here? The v4 create request builder doesn't do anything, therefore I would say it is too trivial to be tested.

packages/core/src/odata-v2/entity-serializer.spec.ts
marikaner: I moved them to common.

packages/core/src/odata-v2/entity.spec.ts
marikaner: Yes.

packages/core/src/odata-v4/entity.spec.ts
marikaner: Yes, common makes more sense. Unfortunately there are so many untidy things around this, that I wasn't sure what to touch and what not... But I can move this test.

marikaner: Now, I remembered, why I didn't move it. There are clashes with some of the other tests and I want this to be done thoughtfully. I will create a ticket for this.

packages/core/src/odata-v4/request-builder/create-request-builder.spec.ts
marikaner: I now removed the test file, but also created a follow up ticket to ensure our odata tests are complete and correct ;)

packages/core/src/odata-v4/entity-serializer.spec.ts
marikaner: Also created follow up for this, as I think this is a bigger task, if we want to do it right ;)

jjtang1985 (approved):

PR #858, opened 2020-12-29
chore: change internal logic, remove V2 suffix internally

tomfrenken:


packages/core/src/odata-common/request/odata-create-request-config.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { TestEntity } from '../../../test/test-util/test-services/v2/test-service';
-import { oDataUriV2 } from '../../odata-v2/uri-conversion';
+import { oDataUri as oDataUriV2 } from '../../odata-v2/uri-conversion';
```
marikaner, line 2: This is not so great, but not your fault. We should consider whether we can test it differently... For now let's keep it as it.

packages/core/src/index.ts
```diff
@@ -6,12 +6,15 @@ export * from './odata-v2';
...
   DeleteRequestBuilder,
-  Entity,
+  UpdateRequestBuilder,
   GetAllRequestBuilder,
   GetByKeyRequestBuilder,
-  UpdateRequestBuilder
+  ODataFunctionImportRequestConfig,
+  Entity,
+  CustomField,
+  entityDeserializer,
```
marikaner, line 16: I think this was never exported as the v2 deserializer but the base one. Please check this.

marikaner (commented):
Just a minor question - I think the entityDeserializer was introduced after we already had v2 / v4. And I think there never was an `entityDeserializer` exposed as v2, only as base. Please check this and adjust if necessary. Otherwise this looks good to me.

packages/core/src/odata-common/request/odata-create-request-config.spec.ts
tomfrenken: I mean theoretically it is possible to directly import `oDataUriV2` since both `oDataUri` and `oDataUriV2` are exported, I just wasn't sure what the preferred syntax here was.

marikaner: Nah, it is more the fact that we import it here. But i get why it might be relevant. Still, later on this will most definitely become a problem, but don't worry about it. We will tackle it then I guess...

packages/core/src/index.ts
tomfrenken: Yeah I agree, but thinking about it a little bit more, with this explicit export that behaviour wouldn't change 🤔 unless I misunderstood something

marikaner: I think before this PR we have this situation:
```ts
import { entityDeserializer } from '@sap-cloud-sdk/core'; // entityDeserializer => src/odata-common/entity-deserializer
```

After this PR we have this situation:
```ts
import { entityDeserializer } from '@sap-cloud-sdk/core'; // entityDeserializer => src/odata-odata-v2/entity-deserializer
```

This is not the same as before, or did I miss anything?

tomfrenken: Oh true, I only saw the `EntityDeserializer` interface in odata-common and overlooked the  actual function 😅

tomfrenken: Alright, I've updated it.

marikaner (approved):
LGTM

PR #860, opened 2020-12-30
DRAFT: chore: change internal logic, remove V4 suffix internally

tomfrenken:


tomfrenken:
I'll close the PR for a bit until my tests pass locally again.

PR #861, opened 2020-12-30
chore: change internal logic, remove V4 suffix internally

tomfrenken:


packages/core/test/test-util/filter-factory.ts
```diff
@@ -1,6 +1,6 @@
 import { v4 as uuid } from 'uuid';
 import { or } from '../../src';
-import { all, any, filterFunctionsV4 } from '../../src/odata-v4';
+import { all, any, filterFunctions } from '../../src/odata-v4';
```
marikaner, line 3: Maybe you could do `filterFunctions as filterFunctionsV4`, just because in other places we do that too. This will make it easier to spot this later on, when we split the repos.

packages/core/src/index.ts
```diff
@@ -2,19 +2,48 @@ export * from './connectivity';
 export * from './header-util';
 export * from './http-client';
 export * from './odata-common';
-export { entityDeserializer } from './odata-common';
+// The explicit export is needed to guarantee backwards compatibility, it overrides the same named function in odata-v4.
```
marikaner, line 5: Maybe you can put the two comments together as one and phrase it something like: "... exports are needed ..." and "named exports in the other odata modules"

marikaner (approved):
Looks good to me, I left some minor comments, please take a look

packages/core/test/test-util/filter-factory.ts
tomfrenken: Good catch, updated it

packages/core/src/index.ts
tomfrenken: Alright, I've updated the comment and restructured the explicit exports to be adjacent to each other

PR #862, opened 2020-12-30
chore: Remove all apis that are not allowed to be here

marikaner:
Again, there were some SAP APIs slipped into this repository. This PR removes them.


tomfrenken (approved):
LGMT

PR #871, opened 2021-01-04
fix: missing templates used in the OpenAPI generator

jjtang1985:
This will fix the openapi generator for the missing templates.


marikaner (approved):
LGTM

PR #884, opened 2021-01-06
chore: Use shared eslint config and fix new eslint errors

florian-richter:
This was mostly easy fixes, but please keep an eye if some changes may be breaking (I removed some unnecessary parameters where I am unsure if they could be considered breaking changes in TypeScript?)

I commented some code that was unused but I was unsure if it should be preserved (e.g. unused loggers because we may refer to the logger names elsewhere?). Let me know if you would rather have the code removed or maybe if I should add a line above to explain why this is commented out.

There are 4 remaining lint errors that were difficult to remove.


packages/core/src/odata-common/filter/filter-lambda-expression.ts
```diff
@@ -13,10 +12,7 @@ export type FilterLambdaOperator = 'any' | 'all';
 /**
  * hidden
  */
-export class FilterLambdaExpression<
-  EntityT extends Entity,
-  FieldT extends FieldType
-> {
+export class FilterLambdaExpression<EntityT extends Entity> {
```
marikaner, line 15: I don't see an issue here, but @jjtang1985 can you double check?

packages/core/src/odata-common/request-builder/update-request-builder-base.ts
```diff
@@ -11,10 +7,11 @@ import { Selectable } from '../selectable';
...
 
-const logger = createLogger({
-  package: 'core',
-  messageContext: 'update-request-builder-v2'
-});
+// const logger = createLogger({
+//   package: 'core',
+//   messageContext: 'update-request-builder-v2'
+// });
+
```
marikaner, line 14:
```suggestion
```

marikaner, line 14:
```suggestion
```

packages/core/src/odata-common/uri-conversion/get-resource-path.ts
```diff
@@ -77,7 +77,7 @@ export function createGetResourcePathForKeys(
 
   function getNullishKeys(keys: Record<string, FieldType>): string[] {
     return Object.entries(keys)
-      .filter(([_, value]) => typeof value === 'undefined' || value === null)
+      .filter(([, value]) => typeof value === 'undefined' || value === null)
```
marikaner, line 80: I know this isn't part of this PR, but could you replace this with the `isNullish` function from `@sap-cloud-sdk/util`.
```suggestion
      .filter(([, value]) => isNullish(value))
```

packages/generator/src/edmx-parser/edmx-file-reader.ts
```diff
@@ -1,15 +1,15 @@
...
 import { getMergedPropertyWithNamespace } from './common';
 
-const logger = createLogger({
-  package: 'generator',
-  messageContext: 'edmx-edmx-parser'
-});
+// const logger = createLogger({
+//   package: 'generator',
+//   messageContext: 'edmx-edmx-parser'
+// });
```
marikaner, line 13:
```suggestion
```

packages/generator/src/edmx-to-vdm/common/complex-type.ts
```diff
@@ -23,10 +22,10 @@ import {
...
 import { enumTypeForName } from './entity';
 
-const logger = createLogger({
-  package: 'generator',
-  messageContext: 'complex-type'
-});
+// const logger = createLogger({
+//   package: 'generator',
+//   messageContext: 'complex-type'
+// });
```
marikaner, line 29:
```suggestion
```

packages/generator/src/entity/class.ts
```diff
@@ -221,7 +221,7 @@ function customField(
   };
 }
 
-function toJSON(entity: VdmEntity): MethodDeclarationStructure {
+function toJSON(/* entity: VdmEntity */): MethodDeclarationStructure {
```
marikaner, line 224:
```suggestion
function toJSON(): MethodDeclarationStructure {
```

packages/generator/src/imports.spec.ts
```diff
@@ -55,9 +51,9 @@ const oneToOneLink = {
   isCollection: false
 } as VdmNavigationProperty;
 
-const service = {
-  oDataVersion: 'v2'
-} as VdmServiceMetadata;
+// const service = {
+//   oDataVersion: 'v2'
+// } as VdmServiceMetadata;
```
marikaner, line 57:
```suggestion
```

packages/generator/src/imports.ts
```diff
@@ -40,7 +40,7 @@ export function externalImportDeclaration(
 
 export function coreImportDeclaration(
   namedImports: string[],
-  oDataVersion: ODataVersion
+  oDataVersion?: ODataVersion
```
marikaner, line 43:
```suggestion
```

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -1,15 +1,12 @@
...
+import { codeBlock, partition, unique } from '@sap-cloud-sdk/util';
 import {
   OpenApiDocument,
   OpenApiOperation,
   SchemaMetadata
 } from '../openapi-types';
-const logger = createLogger('openapi-generator');
+
+// const logger = createLogger('openapi-generator');
+
```
marikaner, line 9:
```suggestion
```

marikaner (commented):
There are still some checks failing according to GH actions. I would say, yes remove commented out code and the two places where you removed optional parameters, I think are not problematic. Looks good 👍

packages/core/src/connectivity/scp-cf/environment-accessor.ts
```diff
@@ -319,8 +319,7 @@ function matchingAudience(
 }
 
 function takeFirstAndWarn(
-  xsuaaInstances: Record<string, any>[],
-  token?: DecodedJWT
+  xsuaaInstances: Record<string, any>[]
```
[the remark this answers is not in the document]
FrankEssenberger: Yes if the token is not used and the method is private it can go.

packages/generator/src/imports.ts
florian-richter: This function is exported and therefore this could be a breaking change, right? (Assuming someone uses strict typescript it would no longer compile because it would not allow a function to be called with unnecessary arguments.

packages/core/src/odata-common/filter/filter-lambda-expression.ts
florian-richter: I guess it should be fine since it is `hidden`?

florian-richter: Although it is exposed as part of Filterable...

packages/core/src/connectivity/scp-cf/environment-accessor.ts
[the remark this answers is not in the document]
marikaner: Top

packages/generator/src/imports.ts
marikaner: But it is not exported externally (not from the root index.ts).

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -134,6 +134,7 @@ class DestinationFromServiceRetriever {
   private static async getProviderClientCredentialsToken(
     options: DestinationOptions
   ): Promise<string> {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
FrankEssenberger, line 137: Instead of removing the linter you could try: 
`const optionsWithoutJwt:Omit<DestinationOptions,'userJwt'> = options; `

packages/core/src/http-client/http-client.spec.ts
```diff
@@ -34,7 +36,7 @@ describe('generic http client', () => {
     sapClient: '001',
     proxyConfiguration: {
       headers: {
-        'Proxy-Authorization': 'youmaypass'
+        'Proxy-Authorization': proxyAuthorization
```
FrankEssenberger, line 39: what was the problem here?

packages/core/src/odata-common/uri-conversion/get-resource-path.ts
FrankEssenberger: @marikaner  has seen the `isNullish` very often in the currency conversion repo ;-)

packages/generator/src/imports.ts
FrankEssenberger: yes this can go.

FrankEssenberger (approved):
For me it is fine. I would put one thing for discussion. We have 3 warnings left regarding unused parameter. Two are just in test and one is a `VersionT = v2|v4` which we may need/use later. I would ignore these and set perhaps the lint level for this rule to error? I think we never want to get unused variables.

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
florian-richter: Problem is not the type but `userJwt` which is an unused variable that is needed to use this short snippet to copy an object with that one property removed. I am not aware of any shorter version to copy+remove property so I just kept it as is.

PR #888, opened 2021-01-07
chore: check api doc generation

jjtang1985:
Add a new check for the api doc generation so we detect potential issues before merging to master.


marikaner (approved):
LGTM

PR #889, opened 2021-01-07
chore: Fix docs for latest version of typedoc

marikaner:
Use the better support of es6 modules with typedoc 0.20. Remove lerna plugin and fix links.


packages/core/src/odata-common/filter/filterable.ts
```diff
@@ -27,7 +27,7 @@ export type Filterable<EntityT extends Entity> =
  *  .filter(and(filterExp1, filterExp2));
  * ```
  *
- * Note that the [[GetAllRequestBuilderV2.filter]]  and [[GetAllRequestBuilderV4.filter]] method take a rest parameter and thereby an array of filter expressions that are then combined conjunctively. As a consequence following is equivalent to the example above:
+ * Note that the [[GetAllRequestBuilder.filter | GetAllRequestBuilderV2.filter]]  and [[GetAllRequestBuilderV4.filter]] method take a rest parameter and thereby an array of filter expressions that are then combined conjunctively. As a consequence following is equivalent to the example above:
```
jjtang1985, line 30: `[[GetAllRequestBuilder.filter | GetAllRequestBuilderV2.filter]]` is the key to fix the reference?

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "sap-cloud-sdk",
+  "version": "1.34.0",
```
jjtang1985, line 3: Is this mandatory for the typedoc, since lerna is not used?

jjtang1985 (approved):
Looks good. I adjusted the script. Please have a check.

package.json
marikaner: Yes.

packages/core/src/odata-common/filter/filterable.ts
marikaner: Yes.

scripts/generate-docs.js
```diff
@@ -50,7 +50,7 @@ function adjustForGitHubPages() {
 }
 
 function adjustSearchJs(paths) {
-  const filtered = paths.filter(isSearchJson);
+  const filtered = paths.filter(isSearchJs);
```
marikaner, line 53: BTW, we might want to not filter it here, right?

PR #892, opened 2021-01-07
chore: include circular dependency check

FrankEssenberger:
Included the circular dependency check. Fix a few easy ones and for the ones not addressable without a breaking change I added an exclude.


packages/core/src/connectivity/scp-cf/user.ts
```diff
@@ -23,6 +22,23 @@ export interface User extends UserData {
...
+  decodedToken: DecodedJWT
+): Map<string, string[]> {
+  if (decodedToken[mappingUserFields.customAttributes.keyInJwt]) {
+    return readPropertyWithWarn(
+      decodedToken,
+      mappingUserFields.customAttributes.keyInJwt
+    ) as Map<string, string[]>;
+  }
+  return new Map<string, string[]>();
+}
```
marikaner, line 40: I know this was here before, but why is this a `Map`? and is it really a Map?

marikaner (commented):
Looks good, there is only this one thing that has already been there before, that surprises me. Do you know why we use a Map and whether this is correct?

packages/core/src/connectivity/scp-cf/user.ts
FrankEssenberger: I create a seperate ticket since there is also no test for the method. The ticket is then to add tests and see if we really need this Map<string,string[]> signature.

jjtang1985 (approved):
LGTM

PR #893, opened 2021-01-07
Parse input directory of OpenApi generator recursively

tomfrenken:
The changes make it possible that:

- you can specify a single file (instead of only directories)
- files of subdirectories are used for generation as well
- files that have the same name, e.g. `/test-service.json` and `/subDirectory/test-service.json` don't override each other


tomfrenken:
Note that the services are now generated sequentially instead of asynchronous, to check for duplicate directories - I am not sure if this is the best approach.

packages/openapi-generator/src/cli.ts
```diff
@@ -25,10 +25,11 @@ export = class GenerateOpenApiClient extends Command {
   ).version;
 
   static flags = {
-    inputDir: flags.string({
-      name: 'inputDir',
+    input: flags.string({
```
marikaner, line 28: I think that name makes more sense than the old one as it also allows to handle files (without directory). However, this now differs from the OData generator. We should consider renaming (and deprecating the old one) it there too. What do you think @FrankEssenberger @jjtang1985?

packages/openapi-generator/src/generator.ts
```diff
@@ -141,3 +110,96 @@ async function generateOpenApiService(
...
+async function duplicateServiceExists(
+  serviceName: string,
+  outputDir: string
+): Promise<boolean> {
+  const currentOutput: string[] = await readdir(outputDir);
+  return currentOutput.some(fileName => {
+    if (serviceName === fileName) {
+      return true;
+    }
+  });
```
marikaner, line 129: I think it is not a good idea to read the directory that we are currently writing to and derive names from this. In the worst case for many files that we write asynchronously, this might even lead to indeterministic behavior. 
Instead, we should use the uniquenamefinder, that we are using for the OData generator. I moved it to util, so it becomes available for both generators in this [PR](https://github.com/SAP/cloud-sdk-js/pull/906).

However, here a little idea for improvement for this kind of code:
```suggestion
  const currentOutput = await readdir(outputDir);
  return currentOutput.some(fileName => serviceName === fileName);
```

marikaner (changes requested):
Just a first glance, I will continue tomorrow.

packages/openapi-generator/src/generator.ts
```diff
@@ -141,3 +110,96 @@ async function generateOpenApiService(
...
+  await generateOpenApiService(convertedInputFilePath, serviceDir);
+  await generateSDKSources(serviceDir, parsedOpenApiDocument, options);
+}
+
+/**
+ * Recursively searches through a given input path and returns all file paths as a string array.
+ * @param input the path to the input directory.
+ * @returns all file paths as a string array.
+ */
+async function recursiveInputPathSearch(input: string): Promise<string[]> {
```
marikaner, line 190: You will also find this in the styleguide soon 👍
Functions should start with a verb. I think the fact that it is recursive does not need to be a part of the function name.
Consider `getInputFilePaths` or something in that direction.

packages/openapi-generator/src/generator.ts
```diff
@@ -141,3 +110,96 @@ async function generateOpenApiService(
...
+  await Promise.all(
+    directoryContents.map(async content => {
+      if (content.isDirectory()) {
+        recursiveInputPath = recursiveInputPath.concat(
+          await recursiveInputPathSearch(resolve(input, content.name))
+        );
+      } else {
+        recursiveInputPath.push(resolve(input, content.name));
+      }
+    })
```
marikaner, line 202: This `.map` currently returns an array of `Promise<void>` and adds something to the `recursiveInputPath` array. If you don't return anything, `.forEach` would be more appropriate.

That said, in this case `.map` goes into the right direction, but I think `.reduce` is the better choice.
Instead of adding something to an array, you should transform the list of `directoryContents` to a new array. As one entry in `directoryContents` can result in multiple (or no) entries in the result a `.map` won't work. Instead consider the following (I haven't tested it so there might be something wrong):
```suggestion
    const inputPaths = directoryContents.reduce(async (paths, directoryContent) => [
      ...paths,
      ...(content.isDirectory()
        ? await recursiveInputPathSearch(resolve(input, content.name))
        : resolve(input, content.name)
      ];
    }, [])
```

marikaner (commented):
Some more ;)

packages/openapi-generator/src/generator.ts
```diff
@@ -24,46 +24,15 @@ export async function generate(options: GeneratorOptions): Promise<void> {
...
-      openApiDocument,
-      serviceName
-    );
-
-    if (!parsedOpenApiDocument.operations.length) {
-      logger.warn(
-        `The given OpenApi specificaton does not contain any operations. Skipping generation for input file: ${filePath}`
-      );
-      return;
+  if (statSync(options.input).isFile()) {
```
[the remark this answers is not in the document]
tomfrenken: Yeah good catch, I was looking for a promise version but overlooked this one.

packages/openapi-generator/src/generator.ts
```diff
@@ -24,46 +24,15 @@ export async function generate(options: GeneratorOptions): Promise<void> {
...
+    const inputFilePath = options.input;
+    generateFromFile(inputFilePath, options);
+  } else {
+    const inputFilePaths = await recursiveInputPathSearch(options.input);
+    for (const filePath of inputFilePaths) {
+      await generateFromFile(filePath, options);
```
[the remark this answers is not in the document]
tomfrenken: I was using this approach before, however when i use `.forEach` instead of the `for( ... of ...)` loop, the files were not generated sequentially, which was an issue because then the `duplicateServiceExists` function wasn't properly working, however as you mentioned below, since I will probably remove the `duplicateServiceExists` function, I guess I should use it again

packages/openapi-generator/src/generator.ts
tomfrenken: I see, I changed it

packages/openapi-generator/src/generator.ts
```diff
@@ -170,11 +148,84 @@ function parseServiceName(filePath: string): string {
...
+}
+
 /**
  * Get the current SDK version from the package json.
  * @returns The SDK version.
  */
-export function getSDKVersion(): string {
-  return JSON.parse(readFileSync(resolve(__dirname, '../package.json'), 'utf8'))
-    .version;
+export async function getSDKVersion(): Promise<string> {
```
marikaner, line 227: I know you didn't implement this, but this is not following our coding conventions - it should be `getSdkVersion`. Could you please fix that?

packages/openapi-generator/src/generator.ts
```diff
@@ -29,56 +33,30 @@ const logger = createLogger('openapi-generator');
...
+    const inputFilePath = options.input;
+    generateFromFile(inputFilePath, options, vdmMapping, uniqueNameGenerator);
+  } else {
+    const inputFilePaths = await getInputFilePaths(options.input);
+    inputFilePaths.forEach(async filePath => {
+      await generateFromFile(
+        filePath,
+        options,
+        vdmMapping,
+        uniqueNameGenerator
```
marikaner, line 56: You can reduce the code a bit by moving the check whether it is a file to the `getInputFilePaths` function (see below).
```suggestion
  getInputFilePaths(options.input).forEach(async filePath => {
      await generateFromFile(
        filePath,
        options,
        vdmMapping,
        uniqueNameGenerator
      )
```

marikaner: Also, you pointed out very correctly, that forEach executes the functions asynchronously. Therefore the generated names could differ between multiple runs.

There are two possibilities to solve that (and you could use reduce, but nah) :
1. Execute this in a `for ... of` loop as you did before.
2. Generate the name before calling `generateFromFile` and pass the name to `generateFromFile` instead of the uniqueNameGenerator.

There are pros and cons for both options, but I think for now I would go with 1. and add a comment, so that nobody accidentally removes this (it is easy to miss this).

packages/openapi-generator/src/generator.ts
```diff
@@ -170,11 +148,84 @@ function parseServiceName(filePath: string): string {
...
+      const contentPath = resolve(input, directoryContent.name);
+      const newInputFilePaths = directoryContent.isDirectory()
+        ? await getInputFilePaths(contentPath)
+        : [contentPath];
+      return [...(await paths), ...newInputFilePaths];
+    },
+    Promise.resolve([])
+  );
+  return inputFilePaths;
+}
```
marikaner, line 221:
```suggestion
async function getInputFilePaths(input: string): Promise<string[]> {
  if(await lstat(options.input)).isFile()) {
    return [input];
  }
 
  const directoryContents = await readdir(input, { withFileTypes: true });
  return directoryContents.reduce(
    async (paths: Promise<string[]>, directoryContent) => [
      ...(await paths), ...(await getInputFilePaths(resolve(input, directoryContent.name);))
    ],
    Promise.resolve([])
  );
}
```

marikaner (changes requested):
Looks quite good, I think we can still remove some complexity.
Let's talk about the tests online.

packages/openapi-generator/src/generator.ts
tomfrenken: Implemented

jjtang1985:
Since the naming of the options was raised during the standup. Here are my personal opinion.
- The alias (`-i`) might be more important than the full name (`--input`), which is aligned between the odata generator and the openapi generator.
- I would make the options shorter and aligned. So I would prefer using `--input` and `--output` in the openapi generator for the alignment in this package and in the sdk 2.0, also rename the odata options with the same names.

marikaner (approved):
LGTM. As we have discussed tests already, I would propose to merge this already (as this is beta) and add tests with a separate PR.

marikaner:
> Since the naming of the options was raised during the standup. Here are my personal opinion.
> 
> * The alias (`-i`) might be more important than the full name (`--input`), which is aligned between the odata generator and the openapi generator.
> * I would make the options shorter and aligned. So I would prefer using `--input` and `--output` in the openapi generator for the alignment in this package and in the sdk 2.0, also rename the odata options with the same names.

I like that. `outputDir` is a bit more specific, but we can cover this by documentation.

marikaner:
> LGTM. As we have discussed tests already, I would propose to merge this already (as this is beta) and add tests with a separate PR.

@jjtang1985 @FrankEssenberger Please merge if you think that is ok, otherwise let's keep this open until next week and add tests here.

PR #896, opened 2021-01-08
chore: Styleguide

marikaner:
This is the initial version of the style guide. Please take some time to read it. There are points, that we do not consistently fulfill and there might be points that you might not agree with. If there are any, let's talk about them.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -1,79 +1,716 @@
-# SAP Cloud SDK style guide
+# SAP Cloud SDK style guide <!-- omit in toc -->
 
-This styleguide includes some basic rules to be applied when writing code in the SAP Cloud SDK. Most of the rules are being checked by lint, but not all of them can.
+This styleguide includes some basic rules to be applied when writing code in the SAP Cloud SDK.
+Most of the rules are being checked by lint, but not all of them can be checked.
+Those that are checked by lint are marked by the following symbol: ✓.
 
-## General
-
-### Whitespace
-
-* Use white
-
-### Tabs vs. spaces
-
-### Line Terminators
-
-
-### Comments
-
-### `null` vs. `undefined`
+#### Table of Contents
+- [Naming](#naming)
+  - [Use kebap case for file names](#use-kebap-case-for-file-names)
+  - [Use camel case for variable names ✓](#use-camel-case-for-variable-names-)
+  - [Use pascal case for classes ✓](#use-pascal-case-for-classes-)
+  - [Use pascal case for interface names and don't prefix them](#use-pascal-case-for-interface-names-and-dont-prefix-them)
+  - [Use pascal case for enums and enum properties](#use-pascal-case-for-enums-and-enum-properties)
+  - [Use verbs in function names](#use-verbs-in-function-names)
+- [Strings](#strings)
+  - [Use single quotes ✓](#use-single-quotes-)
+  - [String concatenation](#string-concatenation)
+- [Arrays](#arrays)
+  - [Use `.forEach` rather than `for`](#use-foreach-rather-than-for)
+- [`null` and `undefined`](#null-and-undefined)
+  - [Use optional parameters for optional values](#use-optional-parameters-for-optional-values)
+  - [Use `undefined` for explicit absence of values](#use-undefined-for-explicit-absence-of-values)
+  - [Use `null` for intentional explicit absence of values or if it is part of an external API](#use-null-for-intentional-explicit-absence-of-values-or-if-it-is-part-of-an-external-api)
+  - [Use truthy/falsy checks where possible](#use-truthyfalsy-checks-where-possible)
+- [White space](#white-space)
+  - [Indentation: Use 2 spaces, not tabs ✓](#indentation-use-2-spaces-not-tabs-)
+  - [Add one empty line before and after function definitions](#add-one-empty-line-before-and-after-function-definitions)
+- [Comments](#comments)
+  - [Document public API with TypeDoc comments](#document-public-api-with-typedoc-comments)
+  - [Use `@deprecated` tag for deprecation](#use-deprecated-tag-for-deprecation)
+  - [Don't reference types in parameters an type parameters](#dont-reference-types-in-parameters-an-type-parameters)
+  - [Use @returns if a function has a return value ✓](#use-returns-if-a-function-has-a-return-value-)
+- [Tests](#tests)
+  - [Use .spec in test file names](#use-spec-in-test-file-names)
+  - [Use it and describe notation](#use-it-and-describe-notation)
+  - [Put unit test files next to source files](#put-unit-test-files-next-to-source-files)
+- [Promises](#promises)
+  - [Use `async`/`await` rather than `.then`](#use-asyncawait-rather-than-then)
+- [Functions](#functions)
+  - [Use arrow functions for callbacks](#use-arrow-functions-for-callbacks)
+  - [Use function declarations to reference functions by name](#use-function-declarations-to-reference-functions-by-name)
 
 ## Naming
 
-### Files
+### Use kebap case for file names
+Use kebap case for all file names and directories, except for generated files and markdown files, that are not in the knowledge base.
 
-### Variables
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use camel case */
+someDir/someClass.ts
 
-### Classes and interfaces
+/* Don't use pascal case */
+SomeDir/SomeClass.ts
 
-### Functions
+/* Don't use snake case */
+some_dir/some_class.ts
+```
 
-booleans
+✅ Examples of **correct** code:
+```ts
+/* Use kebap case */
+some-dir/some-class.ts
+```
+
+### Use camel case for variable names ✓
+This is common in most JavaScript/TypeScript projects.
+This rule applies to **all** variables, even static ones or reusable constants.
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use uppercase snake case */
+const FOO_BAR = 'foo';
+
+/* Don't use snake case */
+const foo_bar = 'foo';
+
+/* Don't use different case for static variables */
+class FooBar {
+  static FOO_BAR;
+}
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use camel case */
+const fooBar = 'foo';
+
+/* Use different case for static variables */
+class FooBar {
+  static fooBar;
+}
+```
+
+### Use pascal case for classes ✓
+Pascal case is common in most JavaScript/TypeScript projects for classes.
+
+✅ Examples of **correct** code:
+```ts
+/* Use pascal case */
+class FooBar {
+  ...
+}
+```
+
+### Use pascal case for interface names and don't prefix them
+Pascal case is common in most JavaScript/TypeScript projects for interfaces.
+Don't prefix interfaces with the hungarian notation 'I' (or other prefixes).
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use 'I' prefix (or any other prefixes) */
+interface IFooBar {
+  ...
+}
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use pascal case without prefix */
+interface FooBar {
+  ...
+}
+```
+
+### Use pascal case for enums and enum properties
+Pascal case is common in most JavaScript/TypeScript projects for enums and enum properties.
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use upper case snake case for enums and enum properties */
+enum SomeEnum {
+  FOO_BAR,
+  BAR_FOO
+}
+
+/* Don't use camel case for enums and enum properties */
+enum SomeEnum {
+  fooBar,
+  barFoo
+}
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use pascal case */
+enum SomeEnum {
+  FooBar,
+  BarFoo
+}
+```
+### Use verbs in function names
+Every function should do something and that action should be described with a verb.
+The verb should fit the return type and value.
+An exception to this can be methods, that operate on an instance directly, e. g. `instance.toString()`, `instance.asObject()`.
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use functions without verb */
+function toFoo(): Foo {
+  ...
+}
+
+/* Don't use functions with misleading verbs */
+function isFoo(): void {
+  ...
+  throw Error(...);
+}
+
+/* Don't use camel case for enums and enum properties */
+enum SomeEnum {
+  fooBar,
+  barFoo
+}
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use functions with verbs */
+function transformToFoo(): Foo {
+  ...
+}
+
+/* Use functions with verbs that fit the return type */
+function isFoo(): boolean {
+  ...
+}
+
+/* Use functions with verbs that fit the return type */
+function validateFoo(): void {
+  ...
+  throw Error(...);
+}
+
+/* Exception, you can omit verbs in methods, that transform the instance that they are invoked on */
+class FooBar {
+  toString(): string {
+    ...
+  }
+
+  asObject(): Record<string, any> {
+    ...
+  }
+}
+```
 
 ## Strings
 
-### String concatenation
+### Use single quotes ✓
 
-### Single vs. double quotes
+Use single quotes, unless not possible otherwise, e. g. when your string
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use double quotes */
+const foo = "foo";
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use single quotes */
+const foo = 'foo';
+
+/* Use double quotes if the quoted string contains single quotes. */
+const foo = "'foo'";
+```
+
+### String concatenation
+When possible use template literals, this makes the code more readable in most cases.
+Feel free to concatenate strings with `+` in case it makes to code better readable.
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't concatenate strings with `+` */
+const foo = 'foo ' + bar + ' bar';
+
+/* Grey area: If template literals become too complex, don't use them. */
+const foo = `foo ${fn(`${bar}bar`)}`;
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Concatenate strings using template literals */
+const foo = `foo ${bar} bar`;
+
+/* Grey area: If template literals become too complex, feel free to concatenate with +. */
+const foo = `foo ${fn(bar +'bar')}`;
+```
 
 ## Arrays
 
-### `.forEach` vs `for ... in`
+### Use `.forEach` rather than `for`
+Use the functional language concepts of JavaScript in favor of the imperative ones, when possible.
+Use the correct function for your use case.
+In rare cases, when performance is of the essence, it makes sense to resort to the imperative `for` instead.
 
-### Type annotations
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use for ... of if you can use .forEach */
+for (const item of array) {
+  ...
+}
+
+/* Don't use .forEach if you can use .map */
+const result = [];
+array.forEach(item => {
+  result.push(item.property);
+});
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use .forEach if possible */
+array.forEach(item => {
+  ...
+});
+
+/* Use .map, if appropriate */
+const result = array.map(item => item.property);
+```
+
+
+## `null` and `undefined`
+
+### Use optional parameters for optional values
+Use optional parameters, if there is a default value or if it semantically makes sense to omit the parameters when consuming the API.
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use explicit undefined if the parameter is semantically optional */
+function foo(options: Options | undefined) {
+  ...
+}
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use optional parameter if the parameter is semantically optional */
+function foo(options?: Options) {
+  ...
+}
+
+/* Use default parameter if there is a default (making the parameter inherently optional for the consumer) */
+function foo(options: Options = {}) {
+  ...
+}
+```
+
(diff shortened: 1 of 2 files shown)


STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
+# SAP Cloud SDK style guide <!-- omit in toc -->
+
+This styleguide includes some basic rules to be applied when writing code in the SAP Cloud SDK.
```
artemkovalyov, line 3:
```suggestion
This style guide includes some basic rules to be applied when writing code in the SAP Cloud SDK.
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+
+✅ Examples of **correct** code:
+```ts
+/* Use undefined for non-existent values */
+function foo(obj: SomeType | undefined) {
+  ...
+}
+```
+
+### Use `null` for intentional explicit absence of values or if it is part of an external API
```
artemkovalyov, line 328:
```suggestion
### Use `null` for the intentional explicit absence of values or if it is part of an external API
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+  } else if (obj === undefined) {
+    ...
+  }
+}
+```
+
+### Use truthy/falsy checks where possible
+In most cases it is possible to check for truthyness/falsyness instead of explicitly comparing values.
+This should be used when possible, but carefully considered in cases where falsy values are valid and therefore semantically truthy values, e. g. 0, ''.
+Of course more fine granular checks should be applied if semantically needed.
```
artemkovalyov, line 358:
```suggestion
### Use truthy/falsy checks where possible
In most cases, it is possible to check for truthiness/falseness instead of explicitly comparing values.
This should be used when possible, but carefully considered in cases where falsy values are valid and therefore semantically truthy values, e. g. 0, ''.
Of course, more fine granular checks should be applied if semantically needed.
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+2. Description, that can extend over multiple lines.
+3. If necessary, one line per type parameter.
+4. One line per parameter.
+5. If necessary, one line for the description of the return value.
+
+There should be no empty lines in between, except if it is part of the description.
+
+### Use `@deprecated` tag for deprecation
+When deprecating public API, this is done through the TypeDoc comments.
+The first line of the comment should start with `@deprecated` followed by a note since which version this is deprecated (e. g. `Since v1.0.0.`) and a note what to use instead (or alternatively that it won't be replaced).
```
artemkovalyov, line 472:
```suggestion
### Document public API with TypeDoc comments
API that is exposed publicly, should be fully documented.
TypeDoc comments should start with `/**`.
Functions that return something should specify the return value with `@returns`.
Parameters should not have type annotations as those can be inherently retrieved from the TypeScript code.
Documentation comments should have the following structure:
1. If necessary, one line for deprecation.
2. Description, that can extend over multiple lines.
3. If necessary, one line per type parameter.
4. One line per parameter.
5. If necessary, one line for the description of the return value.

There should be no empty lines in between, except if it is part of the description.

### Use `@deprecated` tag for deprecation
When deprecating public API, this is done through the TypeDoc comments.
The first line of the comment should start with `@deprecated` followed by a note mentioning since which version this is deprecated (e. g. `Since v1.0.0.`) and a note of what to use instead (or alternatively that it won't be replaced).
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+/**
+ * @deprecated Since v1.2.3. Use [[foo]] instead.
+ */
+```
+
+### Don't reference types in parameters an type parameters
+In JSDoc it is common to specify the types of parameters and type parameters as JavaScript has not types.
+In TypeDoc this is not necessary, because TypeScript inherently has types and those should not differ between the implementation and documentation.
+A parameter or type parameter line should start with `@param` or `@typeparam`, followed by the (type) parameter name, a dash and the description.
+The description should end with a fullstop.
```
artemkovalyov, line 499:
```suggestion
### Don't reference types in parameters an type parameters
In JSDoc it is common to specify the types of parameters and type parameters as JavaScript has no types.
In TypeDoc this is not necessary, because TypeScript inherently has types and those should not differ between the implementation and documentation.
A parameter or type parameter line should start with `@param` or `@typeparam`, followed by the (type) parameter name, a dash, and the description.
The description should end with a full stop.
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+ */
+```
+
+### Use @returns if a function has a return value ✓
+Functions that return something should have an `@returns` statement in the documentation, followed by a description of the return value.
+The description should end with a fullstop.
+Similar to parameters, return types should not be referenced in the documentation.
+For consistency, `@return` should not be used.
+The only exception are asynchronous functions that return `Promise<void>`.
+They technically return a promise, which does not need an extra description.
```
artemkovalyov, line 533:
```suggestion
### Use @returns if a function has a return value ✓
Functions that return something should have an `@returns` statement in the documentation, followed by a description of the return value.
The description should end with a full stop.
Similar to parameters, return types should not be referenced in the documentation.
For consistency, `@return` should not be used.
The only exception is asynchronous functions that return `Promise<void>`.
They technically return a promise, which does not need an extra description.
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+src/
+  some-unit.ts
+  some-unit.spec.ts
+```
+
+## Promises
+
+### Use `async`/`await` rather than `.then`
+In most cases the more modern `async`/`await` notation, makes code a bit easier to read, therefore this is the preferred way.
+However, there are some cases where `.then` is equally as good and potentially even less error prone - in those cases both options are fine.
```
artemkovalyov, line 627:
```suggestion
However, there are some cases where `.then` is equally as good and potentially even less error-prone - in those cases both options are fine.
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+});
+```
+### Use function declarations to reference functions by name
+Arrow functions require less boiler plate code than function declarations, especially if the function returns something directly, e. g. in one-liners.
+For more complex functions, there is no significant difference.
+Arrow functions cannot be named and therefore have to be assigned to variables - much like function expressions.
+Function declarations are visually easier to differentiate from other variables than functions assigned to varibles.
+This becomes especially relevant, when exposing an API, as many IDEs visualize variables differently than functions, which in turn can improve the quality of autocompletion, when used correctly.
+
+An exception to this rule might be simple arrow functions that are only defined within the scope of another function.
```
artemkovalyov, line 691:
```suggestion
### Use function declarations to reference functions by name
Arrow functions require less boilerplate code than function declarations, especially if the function returns something directly, e. g. in one-liners.
For more complex functions, there is no significant difference.
Arrow functions cannot be named and therefore have to be assigned to variables - much like function expressions.
Function declarations are visually easier to differentiate from other variables than functions assigned to variables.
This becomes especially relevant when exposing an API, as many IDEs visualize variables differently than functions, which in turn can improve the quality of autocompletion when used correctly.

An exception to this rule might be simple arrow functions that are only defined within the scope of another function.
```

artemkovalyov (approved):
Looks good to me:) Let's publish a TS Handbook!
Consider some minor suggestions.

STYLEGUIDE.md
marikaner: I think the correct form is without the because this is a non-count word. I googled that and found english teacher melanie to the rescue: http://www.englishteachermelanie.com/grammar-when-not-to-use-the-definite-article/

STYLEGUIDE.md
marikaner: Oh good catch, but it still should be falsiness - I think there are a few more occurences, I will fix them with replace all.

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+const foo = `foo ${fn(`${bar}bar`)}`;
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Concatenate strings using template literals */
+const foo = `foo ${bar} bar`;
+
+/* Grey area: If template literals become too complex, feel free to concatenate with +. */
+const foo = `foo ${fn(bar +'bar')}`;
```
florian-richter, line 247:
```suggestion
const suffix = fn(bar + 'bar');
const foo = `foo ${suffix}`;
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+```
+
+## Arrays
+
+### Use `.forEach` rather than `for`
+Use the functional language concepts of JavaScript in favor of the imperative ones, when possible.
+Use the correct function for your use case.
+In rare cases, when performance is of the essence, it makes sense to resort to the imperative `for` instead.
```
florian-richter, line 255: I actually find it easier to read `for...of` loops and it is easier to use. `forEach` has pretty weird behavior (e.g. returns `undefined`) and can't easily be used with `async/await` (this might be worth a mention in this part of the style guide as well -> prefer `await Promise.all(array.map())`)

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use for ... of if you can use .forEach */
+for (const item of array) {
+  ...
+}
+
+/* Don't use .forEach if you can use .map */
```
florian-richter, line 264: Also worth a mention: `.filter` rather than `.forEach(i => i < 10 ? r.push(i) : undefined)` or `.reduce`.

And I would advice to use reduce only if it transforms to a single value (but I am guilty of using reduce for more complicated cases as well).

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+}
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use null if it semantically makes sense and is different from undefined */
+function foo(obj: SomeType | null | undefined) {
+  if (obj === null) {
+    ...
+  } else if (obj === undefined) {
```
florian-richter, line 349: Shouldn't this be `typeof obj === 'undefined'`?

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+In most cases it is possible to check for truthyness/falsyness instead of explicitly comparing values.
+This should be used when possible, but carefully considered in cases where falsy values are valid and therefore semantically truthy values, e. g. 0, ''.
+Of course more fine granular checks should be applied if semantically needed.
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use explicit comparison with undefined for objects */
+if(obj !== undefined) { ... }
+
+/* Don't use explicit comparison for array length (or other numbers)*/
```
florian-richter, line 365:
```suggestion
/* Don't use explicit comparison with undefined for objects */
if(obj !== undefined) { ... }

/* Don't use thruthy/falsey check for primitives */
function checkIfExist(obj: string | number) {
  if(obj) { ... }
}

/* Don't use explicit comparison for array length (or other numbers)*/
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+
+✅ Examples of **correct** code:
+```ts
+/* Add all deprecation information */
+/**
+ * @deprecated Since v1.2.3. Use [[foo]] instead.
+ */
+```
+
+### Don't reference types in parameters an type parameters
```
florian-richter, line 495:
```suggestion
### Don't reference types in parameters and type parameters
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+
+## Tests
+
+### Use .spec in test file names
+Use `.spec` not `.test` to differentiate test files.
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't use .test.ts */
+test-file.test.ts
```
florian-richter, line 570:
```suggestion
/* Don't use .test.ts */
test-file.test.ts

/* Don't put unit tests in a separate folder */
test/file.ts
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+    handleError(err);
+  }
+}
+
+/* Grey area: Feel free to use both variants when you have common error handling for multiple asynchronous calls */
+async function foo() {
+  return asyncWait1()
+    .then(() => asyncWait2())
+    .catch(err => handleError(err));
+}
```
florian-richter, line 660: Should this also mention the async/await + `.catch` version?

```ts
async function foo() {
  return await asyncWait1().catch(err => handleError(err));
}
```

STYLEGUIDE.md
florian-richter: I realized later that this is a duplicate for the next point in the style guide but I think it's best to include here as people may not read the guide front-to-back.

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+```ts
+/* Use pascal case */
+class FooBar {
+  ...
+}
+```
+
+### Use pascal case for interface names and don't prefix them
+Pascal case is common in most JavaScript/TypeScript projects for interfaces.
+Don't prefix interfaces with the hungarian notation 'I' (or other prefixes).
```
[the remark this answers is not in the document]
marikaner: Do you have an example where this would be the case? I think in most cases it makes sense to find a more general name for the interface or a more specific name for the class, but there are always exceptions to this rule. I'll add a sentence about this, let me know if you think there should be more.

marikaner, line 108:
```suggestion
Don't prefix interfaces with the hungarian notation 'I' (or other prefixes).
If you have a class with the same name as an interface, consider a more general name for the interface or a more specific name for the class - if they are different things it should be possible to name them differently.
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+    ...
+  }
+}
+```
+
+## Strings
+
+### Use single quotes ✓
+
+Use single quotes, unless not possible otherwise, e. g. when your string
```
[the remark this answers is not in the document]
marikaner: You think? 🤪

marikaner, line 211:
```suggestion
Use single quotes, unless not possible otherwise, e. g. when your string contains single quotes.
```

STYLEGUIDE.md
marikaner: This example was meant to produce a case where is a bit harder to read if there is a template string. Extracting it to a separate variable defeats the purpose a bit, because you could also write
```ts
fn(`${bar}bar`)
```

This was one of the rules where I had difficulties describing them, because the gray area is quite extensive and I wouldn't want to impose any hard rules.

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+### String concatenation
+When possible use template literals, this makes the code more readable in most cases.
+Feel free to concatenate strings with `+` in case it makes to code better readable.
+
+❌ Examples of **incorrect** code:
+```ts
+/* Don't concatenate strings with `+` */
+const foo = 'foo ' + bar + ' bar';
+
+/* Grey area: If template literals become too complex, don't use them. */
```
marikaner, line 237:
```suggestion
/* Gray area: If template literals become too complex, don't use them. */
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+const foo = `foo ${fn(`${bar}bar`)}`;
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Concatenate strings using template literals */
+const foo = `foo ${bar} bar`;
+
+/* Grey area: If template literals become too complex, feel free to concatenate with +. */
```
marikaner, line 246:
```suggestion
/* Gray area: If template literals become too complex, feel free to concatenate with +. */
```

STYLEGUIDE.md
marikaner, line 247:
```suggestion
const foo = `foo ${fn(bar + 'bar')}`;
```

STYLEGUIDE.md
marikaner: Excellent question. I always did (and still do this), but I recently thought to myself, why should we?
The reason we were doing this is, because you could potentially reassign undefined as this is not a reserved keyword. I think, we should just not allow using undefined as a variable name (and I think this rule would allow us to do this: https://eslint.org/docs/rules/id-denylist).

The second reason (that I just found out to be honest), is that the typeof check also works for undeclared variables. I believe that in TypeScript and JS strict mode this can never happen, right?

Therefore I don't see a good reason to make it more complex / more verbose.

What do you think? Am I missing anything?

STYLEGUIDE.md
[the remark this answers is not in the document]
florian-richter: I was thinking about something like "BusinessPartner" as an interface containing all fields that describe a BuPa coming from an API and "BusinessPartner" the class that is created by the SDK to easily use the BusinessPartner instance.

I think general name for the interface and specific for class is fine but maybe an example helps so people don't replace the prefix with a bad suffix or something else.

In the example `interface BusinessPartnerResponse` and `class BusinessPartner` would actually go against the guidance but might fit well.

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+    ...
+  } else if (obj === undefined) {
+    ...
+  }
+}
+```
+
+### Use truthy/falsy checks where possible
+In most cases it is possible to check for truthyness/falsyness instead of explicitly comparing values.
+This should be used when possible, but carefully considered in cases where falsy values are valid and therefore semantically truthy values, e. g. 0, ''.
```
marikaner, line 357:
```suggestion
This should be used when possible, but carefully considered in cases where falsy values are valid and therefore semantically truthy values, e. g. 0, ''.
Therefore, when checking for existence of primitives, don't use truthy/falsy checks.
```

STYLEGUIDE.md
marikaner: Meh, maybe if we move it around?

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use async/await as a default */
+async function getBar() {
+  return (await asyncWait()).bar;
+}
+
+/* Grey area: Feel free to use both variants when you have common error handling for multiple asynchronous calls */
```
marikaner, line 645:
```suggestion
/* Gray area: Feel free to use both variants when you have common error handling for multiple asynchronous calls */
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,716 @@
...
+async function foo() {
+  try {
+    await asyncWait1();
+    return await asyncWait2(); // If you want to handle the error here, this await is necessary!
+  } catch(err) {
+    handleError(err);
+  }
+}
+
+/* Grey area: Feel free to use both variants when you have common error handling for multiple asynchronous calls */
```
marikaner, line 655:
```suggestion
/* Gray area: Feel free to use both variants when you have common error handling for multiple asynchronous calls */
```

STYLEGUIDE.md
florian-richter: I just remembered that this used to be a problem in the past but I just looked it up and since ES5 you can't assign a value to `undefined` and accessing an non-existant global is not a problem in typescript so I think we can safely use `obj === undefined`. 👍

STYLEGUIDE.md
florian-richter: I would change it slightly 

```suggestion
Use single quotes, unless you can avoid escaping characters otherwise, e.g. double quotes when your string contains single quotes.
```

STYLEGUIDE.md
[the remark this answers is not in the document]
marikaner: Thanks for the clarification. Done.

STYLEGUIDE.md
marikaner: Done, added an extra section.

STYLEGUIDE.md
marikaner: I will resolve this for now, as I think the TOC should help here. I agree that things might be difficult to find, so maybe we can find better categories, but I wouldn't put different aspects into one rule.

STYLEGUIDE.md
```diff
@@ -0,0 +1,740 @@
...
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use .forEach if possible */
+array.forEach(item => {
+  ...
+});
+
+/* Use .map, if appropriate */
```
jjtang1985, line 296:
```suggestion
});

/* Exceptions: when conditional break is used */
for (let i=0;;i++) {
  if (i>10)
    break;
}

/* Use .map, if appropriate */
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,740 @@
...
+/* Don't use null for non-existent values */
+function foo(obj: SomeType | null) {
+  ...
+}
+```
+
+✅ Examples of **correct** code:
+```ts
+/* Use undefined for non-existent values */
+function foo(obj: SomeType | undefined) {
```
jjtang1985, line 346: why this is better than `function foo(obj?: SomeType) {`

jjtang1985 (approved):
Very good documentation. I left some minor comments.

STYLEGUIDE.md
marikaner: There is a section on optional values before. This section is about values that are not optional. The difference is that with optional parameters you can invoke the function like this `foo()`. If the parameter is not semantically optional this doesn't make sense and should not be possible.
Consider: 
```
function getFooFromBar(bar?: Bar) {
  return bar?.foo;
}

// This does not make sense
getFooFromBar();

// But this does
const bar = getBarFromSomewherePotentiallyUndefined();
getFooFromBar(bar);
```

I improved the description a bit, let me know if this is clearer or if we should improve the examples.

STYLEGUIDE.md
marikaner: @jjtang1985 I was thinking whether we can find a more plausible example, but I can't think of any case, where a for loop would be better than one of the array functions.

I would rather write your example from above as:
```
array.slice(0, 10).forEach(...)
```

jjtang1985: you are right. Let's update when I have more appropriate example.

jjtang1985 (approved):

PR #897, opened 2021-01-08
chore: add documentation on iss

FrankEssenberger:
Added documentation on the ISS field so it is used in the right way hopefully.

### DIFF
--- a/packages/core/src/connectivity/scp-cf/destination/destination-accessor-types.ts
+++ b/packages/core/src/connectivity/scp-cf/destination/destination-accessor-types.ts
@@ -29,9 +29,12 @@ export interface DestinationAccessorOptions {
   userJwt?: string;
 
   /**
-   * @hidden
+   * This property is only considered in case no userJwt is provided.
+   * It is meant for situations where you do not have a token e.g. background processes.
+   * The value for iss is the issuer field of a JWT e.g. https://<your-subdomain>.localhost:8080/uaa/oauth/token'
+   *
+   * ATTENTION: If this property is used no validation of the provided subdomain value is done in contrast to the JWT case.
+   * So be careful that the used value is not manipulated and breaks the tenant isolation of your application.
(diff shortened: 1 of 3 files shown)


FrankEssenberger:
I will add a test

packages/core/src/connectivity/scp-cf/destination/destination-accessor-types.ts
```diff
@@ -29,9 +29,12 @@ export interface DestinationAccessorOptions {
   userJwt?: string;
 
   /**
-   * @hidden
+   * This property is only considered in case no userJwt is provided.
+   * It is meant for situations where you do not have a token e.g. background processes.
+   * The value for iss is the issuer field of a JWT e.g. https://<your-subdomain>.localhost:8080/uaa/oauth/token'
+   *
+   * ATTENTION: If this property is used no validation of the provided subdomain value is done in contrast to the JWT case.
```
marikaner, line 36: I found this sentence a bit slippery, not sure if this is an improvement:
```suggestion
   * ATTENTION: If this property is used, no validation of the provided subdomain value is done. This is differs from how the `userJwt` is handled.
```

marikaner (approved):
Now that I see it, theoretically it can never happen, that userJwt and iss are set at the same time - this sounds as if we actually needed a conditional type here. I guess that would produce a breaking change as you already mentioned, but do you think for v2 that could be a relevant (and simple) improvement? If so maybe add a comment and/or a ticket.

FrankEssenberger:
> Now that I see it, theoretically it can never happen, that userJwt and iss are set at the same time - this sounds as if we actually needed a conditional type here. I guess that would produce a breaking change as you already mentioned, but do you think for v2 that could be a relevant (and simple) improvement? If so maybe add a comment and/or a ticket.

Yes you are right. I created ticket for 2.0.

PR #898, opened 2021-01-08
chore: Resolve axios to 0.21.1 for all dependencies as well

marikaner:
As in the title.


FrankEssenberger (approved):
LGTM - so the vulnerability is fixed

jjtang1985 (approved):

PR #899, opened 2021-01-10
feat: OpenAPI generator support package.json generation and service mapping

jjtang1985:
- [x] added an option to generate `package.json`
- [x] added an option to provide service mapping info


packages/openapi-generator/src/generator.ts
```diff
@@ -20,6 +27,10 @@ const logger = createLogger('openapi-generator');
  * @param options Options to configure generation.
  */
 export async function generate(options: GeneratorOptions): Promise<void> {
+  options.serviceMapping =
+    options.serviceMapping ||
```
FrankEssenberger, line 31: In the description you state:  By default it will be saved to/read from the input directory as "service-mapping.json".
So I would assume that the `options.serviceMapping` points to a file containing the service mapping and if this is not there it reads it from the file `service-mapping.json`. Compare also to the OData part and perhaps adjust the description if the current behavior the one also used in OData.

packages/openapi-generator/src/service-mapping.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+/**
+ * Read the service mapping from a given path provided in the options.
+ * @param options The options provided from the cli command.
+ * @returns The parsed service mapping object.
+ */
+export function readServiceMapping(options: GeneratorOptions): VdmMapping {
+  return (
+    (options.serviceMapping &&
```
FrankEssenberger, line 19: I find this hard to read, also because I am not used to return (A && B). I would assume that the method return true or false but it obviously must return the readJson result or {}

packages/openapi-generator/src/generator.ts
```diff
@@ -20,6 +27,10 @@ const logger = createLogger('openapi-generator');
  * @param options Options to configure generation.
  */
 export async function generate(options: GeneratorOptions): Promise<void> {
+  options.serviceMapping =
```
FrankEssenberger, line 30: I would not adjust the original object but use a variable which the path to the file which is then passed to readServiceMapping. Or even do the fallback path inside the readServiceMapping method?

packages/openapi-generator/src/generator.ts
```diff
@@ -141,3 +169,12 @@ async function generateOpenApiService(
 function parseServiceName(filePath: string): string {
   return parse(filePath).name.replace(/-openapi$/, '');
 }
+
+/**
+ * Get the current SDK version from the package json.
+ * @returns The SDK version.
+ */
+function getSDKVersion(): string {
```
FrankEssenberger, line 177: if we enable `"resolveJsonModule": true` in the tsconfig we can just use `import {version} from './package.json'`

packages/openapi-generator/src/cli.ts
```diff
@@ -41,9 +41,30 @@ export = class GenerateOpenApiClient extends Command {
...
       default: false,
       required: false
+    }),
+    versionInPackageJson: flags.string({
+      name: 'versionInPackageJson',
+      description:
+        'By default, when generating package.json file, the generator will set a version by using the generator version. It can also be set to a specific version.',
+      required: false
+    }),
+    generatePackageJson: flags.boolean({
```
FrankEssenberger, line 55: Make a compatibility note for all files which are created now and were not created in the past.

packages/openapi-generator/src/parser/document.spec.ts
```diff
@@ -16,7 +16,17 @@ describe('parseOpenApiDocument', () => {
...
+      input,
+      'TestService',
+      'openapi/test-service.json',
+      {
+        'test-service': {
+          npmPackageName: '@sap/cloud-sdk-openapi-vdm-test-service',
+          directoryName: 'test-service'
+        }
+      }
+    );
```
FrankEssenberger, line 29: Could you make the assertion that the used package name and directory are the ones given in the mapping config.

packages/openapi-generator/src/service-mapping.ts
jjtang1985: when the option is provided, returns the content of the file, otherwise, `{}` is the default value.

packages/openapi-generator/src/parser/document.spec.ts
FrankEssenberger: Also could you add a test that the service-mapping.json is written after one run. I think this is the case for the OData generator.

FrankEssenberger (changes requested):
In general all good. Have a look at my comments.

packages/openapi-generator/src/parser/document.spec.ts
jjtang1985: As discussed, as initial version we only rely on the user input and will not write the file.

jjtang1985:
@FrankEssenberger , please review after the adjustment.

packages/openapi-generator/src/generator.ts
FrankEssenberger: as discussed this leads to other problems so we go for the read method.

CHANGELOG.md
```diff
@@ -14,11 +14,11 @@
 
 ## Compatibility Notes
 
--
+- [openapi-generator] The `package.json` is generated by default.
 
 ## New Functionality
 
--
+- [openapi-generator] Support `package.json` generation and the service mapping configuration.
```
FrankEssenberger, line 21: Mention the service mapping option.

packages/util/src/fs.ts
```diff
@@ -16,3 +16,15 @@ export function findProjectRoot(path: string, lastPath: string = path): string {
...
 }
+
+/**
+ * Read a JSON file from the file system.
+ * @param path The path to the JSON file.
+ * @returns An object parsed from the JSON file.
+ */
+export function readJSON(path: PathLike): { [key: string]: any } {
+  if (existsSync(path)) {
+    return JSON.parse(readFileSync(path, 'utf8'));
```
FrankEssenberger, line 27: add a log statement if nothing is found

FrankEssenberger (approved):
LGTM only two minor comments

PR #903, opened 2021-01-11
chore: Let pipeline fail on failing licence check

FrankEssenberger:
Let the licence check fail in the pipeline. The unhandled promise did not stop the pipeline in the used node version. This will change in the future but to be save I introduced a `process.exit(1)`. I also went for TS since I need types :-).

The licence checker looks at all dependencies on root level in the node modules. These are not all direct but I think it is saver when we check more. In case we get problems we can still adjust the tested dependencies.


scripts/check-licenses.js
```diff
@@ -20,7 +20,7 @@ async function getLicenses() {
     licenseChecker.init(
       {
         start: path.resolve(__dirname, '..'),
-        direct: true,
+        direct: false,
```
jjtang1985, line 23: I guess this means we also check transitive dependencies.
The original ideas was to only check the direct dependencies.
Question: do we have to check transitive dependencies?

jjtang1985 (commented):
`Python-2.0` is the new license detected that is not in the "allow list".
I guess we can also add this license to the "allow list"?

scripts/check-licenses.js
FrankEssenberger: this was just a test but true or false do not really make a difference. The problem is that dependencies get hoisted up in the node_modules and are found there. In the end this is not bad, because we do not get wrong licences even if they come in from an other dependency. However, if they are labled wrong this becomes an issues. In the corrent case the `python-2.0` licence is fine so I just added it.

scripts/check-licenses.ts
```diff
@@ -0,0 +1,80 @@
...
+    notAllowedDependencies.forEach(notAllowedDependency => {
+      logger.error(
+        `Not allowed license ${notAllowedDependency[1].licenses} found for dependency: ${notAllowedDependency[0]}.`
+      );
+    });
+    logger.error(
+      'Check if the faulty licences are in the FLOSS list: https://en.wikipedia.org/wiki/Category:Free_and_open-source_software_licenses and update the check script accordingly.'
+    );
+
+    process.exit(1);
```
jjtang1985, line 76: I assume this is THE change that applied to the old js file.

jjtang1985 (approved):
LGTM

scripts/check-licenses.ts
FrankEssenberger: yes this makes it fail.

PR #906, opened 2021-01-12
chore: Expose Unique name finder in utils

marikaner:
I fixed a minor bug in the name finder and moved it to util, so it available to both generators. 
I also went a little crazy and improved it a bit, I hope you like it.


CHANGELOG.md
```diff
@@ -18,7 +18,7 @@
 
 ## New Functionality
 
--
+- [util] Expose `UniqueNameFinder` class that stores used names and generates a unique name for each given string.
```
```suggestion
- [util] Expose `UniqueNameGenerator` class that stores used names and generates a unique name for each given string.
```

packages/generator/src/reserved-words.ts
```diff
@@ -71,4 +61,15 @@ export const reservedServiceKeywords: string[] = [
   'Link',
   'Time',
   'TimeField'
-];
+] as const;
```
FrankEssenberger, line 64: just out of curiosity what is the `as const` doing? The type is redundant, that is why you removed it?

FrankEssenberger (approved):
LGTM. Please answer the comments since I am interested :-)

packages/util/src/unique-name-generator.ts
```diff
@@ -0,0 +1,214 @@
+/**
```
FrankEssenberger, line 1: Since the file was moved, what was the small error you fixed?

packages/util/src/unique-name-generator.ts
```diff
@@ -0,0 +1,214 @@
...
+ * Holds state on already used names and provides new names if there are naming conflicts.
+ */
+export class UniqueNameGenerator {
```
FrankEssenberger, line 4: What was the reason you went from `finder` to `generator`? I have no strong opinion just want to now the reason.

packages/generator/src/reserved-words.ts
marikaner: 1. const makes this readonly - that avoids us accidentally adding anything to it (because this is an array and if you pass it is passed by reference). I am not 100% sure that this is the best way to achieve this, btw.
(so the type here becomes `readonly string[]`)
2. The type was `Set` before and it was used inconsistently. I don't see a good reason for it to be a set and made it an array instead (and then I left out the type annotations ;) )

packages/util/src/unique-name-generator.ts
marikaner: Sorry, yes you cannot see this anymore... -.-

To generate the final name, we internally get the name without index using a regex. This regex had `'_'` hardcoded as a separator. As you can set different separators, the following test would fail for `'-'`:
```ts
it('should add -2 if -1 is already used', () => {
    const generator = new UniqueNameGenerator('-', ['MyClass-1']);
    expect(generator.generateUniqueName('MyClass-1')).toBe('MyClass-2');
});
```
The result was `'MyClass-1_1'` before.

Now that I think of it, this technically is a breaking change. `-` is used for directory names and npm package names, but for existing projects those respect the `service-mapping.json`. We might want to mention it in the compatibility notes...

packages/util/src/unique-name-generator.ts
marikaner: I liked the notion of a generator more, because `finder` implies that it already exists and only has to be filtered from something (like Array.find).
What we do is "get" the index based on some logic and "create" a new name using this. I then thought of the concept of generators in Python and decided that this is close enough.
As this is now public API, I just found it a bit more on the point.

I am open to discuss this, though. Changing it back or to something else isn't a big deal.

packages/util/src/unique-name-generator.ts
FrankEssenberger: yes

Issue #870, opened 2021-01-04
using a destination with authentication type OAuth2ClientCredentials does not use the defined scope (which is required)


blaart:
We are using a token server of Microsoft for a user token grant request.
the url is like: https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token

See also https://github.com/SAP/cloud-sdk-js/issues/816

As 816 is solved I now can test the call to microsoft, but receive an error as the scope is required. scope in my case is defined in the destination as an additional parameter. it should be added in the body (like grant_type, client_id en client_secret, which  are indeed already filled).

jjtang1985:
Hi @blaart , I assume you are using the `additional properties` from the CF for your `additional parameters` in the request body. Could you please confirm that you are using the `scope` as an additional property saved in the destination on the CF and the the mandatory parameter is also called `scope` that you need in the request body? Of course, this should be considered as general issue instead of a specific key.

blaart:
Hi @jjtang1985, you are correct, we use an additional property in the CF destination. Both the CF additional property and the parameter in the body are specified as 'scope'

jjtang1985:
Thank you @blaart for your quick answer. I'll let you know once there are some updates.

FrankEssenberger:
Hello and happy new year,

I investigated the issue a bit and I will post my findings here so that we have all information together.

I directly found an suboptimal coding that for the `OAuth2ClientCredentials` flow the  find destination endpoint is not triggered. Only this call to the destination service will trigger the OAuth flow and lead to a destination with the `authToken` property filled with the access token see [here](https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/4e1d742a3d45472d83b411e141729795.html). Currently the returned destination has no authToken and later when we build the header the token is added by [hand](https://github.com/SAP/cloud-sdk-js/blob/6063f24ad0aa2cea0c1de98876045a2a4cd0177f/packages/core/src/connectivity/scp-cf/authorization-header.ts#L246)

[...]

PR #910, opened 2021-01-13
fix: use destination service for OAuth2ClientCredentials flow

FrankEssenberger:
As reported in #870 the SDK did not consider the scope on the destination object. The reason for this was a manual implementation of the token for the OAuth2ClientCredentials flow.

This PR delegates the task to the destination service.


blaart:
Hi Frank, thanks for picking this up. When reviewed and merged I'm happy to (canary) test this in our setup!

packages/core/src/connectivity/scp-cf/authorization-header.ts
```diff
@@ -238,12 +222,11 @@ async function getAuthenticationRelatedHeaders(
       return {};
     case 'OAuth2SAMLBearerAssertion':
     case 'OAuth2UserTokenExchange':
+    case 'OAuth2ClientCredentials':
       return headerFromTokens(
-        destination.authentication,
+        destination.authentication as any,
```
jjtang1985, line 227: Do we have to do so?

packages/core/src/connectivity/scp-cf/destination/destination-accessor-authentication-type.spec.ts
```diff
@@ -181,6 +183,68 @@ describe('authentication types', () => {
...
+
+    const httpMocks = [
+      mockInstanceDestinationsCall(nock, [], 200, providerServiceToken),
+      mockSubaccountDestinationsCall(
+        nock,
+        oauthClientCredentialsMultipleResponse,
+        200,
+        providerServiceToken
+      ),
+      mockSingleDestinationCall(
```
jjtang1985, line 233: Please correct me, I saw the `xxxSingleResonse` and `xxxMultipleResonse` are mocked together in multiple places.
I guess this is because we as sdk are doing optimisation and fetch all the potential destinations from a provider/subscriber token and cache them.
If I'm an application developer, I can avoid the additional call and just ge the single response, right?

jjtang1985 (approved):
LGTM, I left a question and a minor comment.

packages/core/src/connectivity/scp-cf/destination/destination-accessor-authentication-type.spec.ts
FrankEssenberger: No you can't. I have already created a refactoring ticket for 2.0 to get rid of this premature optimization. Since for all OAuth flows you need to call the find destination by name anyhow the gain of fetching all destinations first is not big and it is making the code unnecessary complicated.

PR #913, opened 2021-01-14
fix: openapi cli

jjtang1985:
The openapi cli fails with the error ``TypeError: Class constructor Command cannot be invoked without 'new'``, due to the incompatibility of ES5 and oclif (ES2017).

The internal generator e2e tests are added.


marikaner (approved):
LGTM

PR #914, opened 2021-01-14
feature:  compilation of OpenApi clients

FrankEssenberger:
The TS client is transpiled to JS with a default tsconfig.json. The flag `generateJS` has default `true`. I did this in analogy to the OData generator but this is open for discussion. 

Since the tests were flacky when they did the client creation in the pipeline I checked the generated file in. So please ignore the `dist` folders in the `test-services` directories for this PR.


FrankEssenberger:
Fun fact. The `rmdir` command behaves different under ubuntu. If the folder is not there it fails. I added a check for that. Good that we have the matrix tests.

packages/openapi-generator/src/generator.ts
```diff
@@ -30,30 +32,34 @@ const logger = createLogger('openapi-generator');
...
   const inputFilePaths = await getInputFilePaths(options.input);
-  inputFilePaths.forEach(async inputFilePath => {
+
+  const promises = inputFilePaths.map(async inputFilePath => {
     const uniqueServiceName = uniqueNameGenerator.generateAndSaveUniqueName(
       parseServiceName(inputFilePath)
     );
 
-    await generateFromFile(
+    return generateFromFile(
```
jjtang1985, line 55: I guess it is intended to `await` as the original implementation, since running in parallel will result in race condition in terms of the unique service name check

packages/openapi-generator/src/wrapper-files/create-file.ts
```diff
@@ -17,10 +17,15 @@ export async function createFile(
   content: string,
   overwrite: boolean
 ): Promise<void> {
-  return writeFile(join(directoryPath, fileName), wrapContent(content), {
-    encoding: 'utf8',
-    flag: overwrite ? 'w' : 'wx'
-  });
+  const addHeader = !fileName.endsWith('.json');
```
jjtang1985, line 20: I have similar [fix](https://github.com/SAP/cloud-sdk-js/pull/918).

packages/openapi-generator/src/cli.ts
```diff
@@ -60,12 +60,26 @@ export = class GenerateOpenApiClient extends Command {
...
       name: 'serviceMapping',
       description:
         'Configuration file to ensure consistent names between multiple generation runs with updated / changed metadata files. By default it will be read from the input directory as "service-mapping.json".',
       parse: input => resolve(input),
       required: false
+    }),
+    tsConfig: flags.string({
+      name: 'tsConfig',
+      description:
+        'tsconfig.json file to overwrite the default tsconfig.json. Existing keys will be overwritten new key will be added.".',
```
jjtang1985, line 80: what is the default tsconfig?

package.json
```diff
@@ -34,7 +34,7 @@
     "test:e2e": "yarn e2e-tests test",
     "generate:test-services:v2": "yarn generate v2 && yarn lint:fix && yarn --frozen-lockfile",
     "generate:test-services:v4": "yarn generate v4 && yarn lint:fix && yarn --frozen-lockfile",
-    "generate:test-services:openapi": "yarn generate openapi && yarn lint:fix && yarn --frozen-lockfile",
+    "generate:test-services:openapi": "n && yarn lint:fix && yarn --frozen-lockfile",
```
jjtang1985, line 37: ;)

packages/util/src/fs.ts
```diff
@@ -35,3 +37,42 @@ export function readJSON(path: PathLike): { [key: string]: any } {
...
+    })
+    .catch(err => {
+      console.error('Error in generation catched' + JSON.stringify(err))
+      logger.error(`Error: Failed to generate js files: ${err}`);
+      process.exit(1);
+    });
+}
+
+export function deleteDirectory(path: string): Promise<void> {
+  function callBackk(
```
jjtang1985, line 63:
```suggestion
  function callBack(
```

packages/util/src/fs.ts
```diff
@@ -35,3 +37,42 @@ export function readJSON(path: PathLike): { [key: string]: any } {
...
+  ) {
+    if (err) {
+      rej(`Error in deleting: ${path} with ${err.message}`);
+    } else {
+      res();
+    }
+  }
+
+  return new Promise<void>((res, rej) => {
+    rmdir(path, { recursive: true }, err => callBackk(err, res, rej));
```
jjtang1985, line 76:
```suggestion
    rmdir(path, { recursive: true }, err => callBack(err, res, rej));
```

packages/openapi-generator/src/wrapper-files/create-file.ts
FrankEssenberger: we take your way and let the developer decide if the header should be added or not.

packages/openapi-generator/src/generator.ts
FrankEssenberger: I saw in the test, that the generation that the await did not have the desired effect. This was just a test. I made a Promise.all which really resolves once all are finished.

packages/openapi-generator/src/cli.ts
FrankEssenberger: It is some config I have copied from the OData case.

.gitignore
```diff
@@ -2,6 +2,7 @@ node_modules
 !.github/actions/**/node_modules
 .vscode
 dist
+!/test-packages/test-services/openapi/**/dist
```
jjtang1985, line 5: I'll ignore this change and assume all the files under the dist folder will be removed.

packages/openapi-generator/src/cli.ts
```diff
@@ -60,12 +60,26 @@ export = class GenerateOpenApiClient extends Command {
       default: true,
       required: false
     }),
+    generateJs: flags.boolean({
+      name: 'generateJs',
+      description:
+        'By default, the generator will also generate transpiled .js, .js.map, .d.ts and .d.ts.map files. When set to false, the generator will only generate .ts files.',
```
jjtang1985, line 66:
```suggestion
        'By default, the generator will also generate transpiled .js, .js.map, .d.ts and .d.ts.map files. When setting to false, the generator will only generate .ts files.',
```

packages/openapi-generator/src/cli.ts
```diff
@@ -60,12 +60,26 @@ export = class GenerateOpenApiClient extends Command {
...
       name: 'serviceMapping',
       description:
         'Configuration file to ensure consistent names between multiple generation runs with updated / changed metadata files. By default it will be read from the input directory as "service-mapping.json".',
       parse: input => resolve(input),
       required: false
+    }),
+    tsConfig: flags.string({
+      name: 'tsConfig',
+      description:
+        'tsconfig.json file to overwrite the default tsconfig.json.".',
```
jjtang1985, line 80: double quotes occurs only once

jjtang1985 (commented):
I would "approve" with minor comments. But what about the `dist` folder?

.gitignore
FrankEssenberger: I try to make the test self-contained again. I hope this is stable....

packages/openapi-generator/src/cli.ts
```diff
@@ -60,12 +60,26 @@ export = class GenerateOpenApiClient extends Command {
...
       name: 'serviceMapping',
       description:
         'Configuration file to ensure consistent names between multiple generation runs with updated / changed metadata files. By default it will be read from the input directory as "service-mapping.json".',
       parse: input => resolve(input),
       required: false
+    }),
+    tsConfig: flags.string({
+      name: 'tsConfig',
+      description:
+        'tsconfig.json file to overwrite the default "tsconfig.json.".',
```
jjtang1985, line 80:
```suggestion
        'tsconfig.json file to overwrite the default "tsconfig.json".',
```

packages/openapi-generator/src/generator.spec.ts
```diff
@@ -1,7 +1,47 @@
-import { getSdkVersion } from './generator';
+import { resolve } from 'path';
+import { existsSync, promises } from 'fs';
+import { getSdkVersion, generate } from './generator';
 
-describe('generator', () => {
-  it('getSdkVersion returns a valid stable version', async () => {
+const { rmdir } = promises;
+
+xdescribe('generator', () => {
```
jjtang1985, line 7: I assume the tests are still useful as local tests.

packages/openapi-generator/src/generator.ts
```diff
@@ -166,7 +183,7 @@ async function generateFromFile(
   try {
     openApiDocument = await convertOpenApiSpec(filePath);
   } catch (err) {
-    logger.error(
+    await logger.error(
```
jjtang1985, line 186: maybe leftover?

test-packages/test-services/openapi/swagger-yaml-service/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "swagger-yaml-service",
+  "version": "1.37.0",
```
jjtang1985, line 3: This version and the sdk version in line 26 rely on the current version, which is dynamic.
Every release will push changes like [this](https://github.com/SAP/cloud-sdk-js/commit/f9ad1fc1323f84c9b261e4cbf8bc2953c9a8f71d) automatically via this [script](https://github.com/SAP/cloud-sdk-js/blob/main/scripts/after-bump.js)
Without changing the script, the build of this commit will fail.
So we either adjust the script to update this json file or provide more parameters to the generator to generate static versions.
I would vote for static versions if the package json is just used for the e2e tests.

test-packages/test-services/openapi/test-service/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "test-service",
+  "version": "1.37.0",
```
jjtang1985, line 3: Same for this.

jjtang1985 (changes requested):
Please consider my comments.

packages/openapi-generator/src/wrapper-files/tsconfig-json.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+import { GeneratorOptions } from '../options';
+import { tsconfigJson } from './tsconfig-json';
+
+describe('tsconfigJson', () => {
+  it('contains the needed dom lib', () => {
+    const tsConfig = JSON.parse(tsconfigJson({} as GeneratorOptions));
+    expect(tsConfig.compilerOptions.lib).toContain('dom');
+  });
+
+  it('returns the custom ts config content', () => {
```
marikaner, line 12: This test will actually do fs and relies on a directory that is not part of this package. The fs-mock package that Tom introduced is perfect to mock this away and provide some dummy file content. I think this would be the better maintainable and more performant approach.

packages/openapi-generator/src/generator.spec.ts
FrankEssenberger: yes and I have a follow up ticket to bring them back. I think they are good to have beside the E2E tests we will also have soon.

packages/openapi-generator/src/generator.ts
FrankEssenberger: yes, I found that the logger is not synchronous, but the await did also not help. Only console.error directly put the message out.

test-packages/test-services/openapi/swagger-yaml-service/package.json
FrankEssenberger: yes good point. I will add one more flag to the generator to pin the version.

FrankEssenberger: ah there is already the `  versionInPackageJson` flag which I can use.

FrankEssenberger: and for the test we have at the moment the package.json can go way completely.

packages/openapi-generator/src/generator.spec.ts
```diff
@@ -1,7 +1,47 @@
...
+  const input = resolve(
```
marikaner, line 8: Here again, I would use mock-fs. Although I am not sure how it handles writing.

packages/openapi-generator/src/generator.spec.ts
```diff
@@ -1,7 +1,47 @@
...
+  it('getSDKVersion returns a valid stable version', async () => {
     expect((await getSdkVersion()).split('.').length).toBe(3);
   });
+
+  it('should transpile the generated sources', () => {
+    const jsApiFile = resolve(testServicePath, 'dist', 'api.js');
+    expect(existsSync(jsApiFile)).toBe(true);
+  });
+
+  it('should create package.json', () => {
```
marikaner, line 38:
```suggestion
  it('should create a package.json', () => {
```

packages/openapi-generator/src/generator.spec.ts
```diff
@@ -1,7 +1,47 @@
...
+    const packageJson = resolve(testServicePath, 'package.json');
+    expect(existsSync(packageJson)).toBe(true);
+  });
+
+  it('should create tsconfig.json', () => {
```
marikaner, line 43:
```suggestion
  it('should create a tsconfig.json', () => {
```

packages/openapi-generator/src/generator.ts
```diff
@@ -30,30 +32,34 @@ const logger = createLogger('openapi-generator');
  * Generates files using the OpenApi Generator CLI and wraps the resulting API in an SDK compatible API.
  * @param options Options to configure generation.
  */
-export async function generate(options: GeneratorOptions): Promise<void> {
+export async function generate(options: GeneratorOptions): Promise<void[]> {
```
marikaner, line 35: Does that make sense? What does is mean?

packages/openapi-generator/src/generator.ts
```diff
@@ -30,30 +32,34 @@ const logger = createLogger('openapi-generator');
...
   options.serviceMapping =
     options.serviceMapping ||
     resolve(options.input.toString(), 'service-mapping.json');
 
   if (options.clearOutputDir) {
-    await rmdir(options.outputDir, { recursive: true });
+    if (existsSync(options.outputDir)) {
```
marikaner, line 41: I am pretty sure this check is not necessary. The `recursive` option should already handle this.

marikaner, line 41: I am pretty sure this check is not necessary. The `recursive` option should already handle this.

packages/openapi-generator/src/generator.ts
```diff
@@ -30,30 +32,34 @@ const logger = createLogger('openapi-generator');
...
+      await rmdir(options.outputDir, { recursive: true });
+    }
   }
 
   const vdmMapping = readServiceMapping(options);
   const uniqueNameGenerator = new UniqueNameGenerator('-');
   const inputFilePaths = await getInputFilePaths(options.input);
-  inputFilePaths.forEach(async inputFilePath => {
+
+  const promises = inputFilePaths.map(async inputFilePath => {
```
marikaner, line 50: With this change you don't need the async anymore.

packages/openapi-generator/src/generator.ts
```diff
@@ -130,10 +146,11 @@ async function generateOpenApiService(
       `Successfully generated a client using the OpenApi generator CLI ${response.stdout}`
     );
   } catch (err) {
-    throw new ErrorWithCause(
-      'Could not generate the OpenApi client using the OpenApi generator CLI.',
-      err
-    );
+    // The npx creates strange error objects, hence we add the method here:
```
marikaner, line 149: The npx?

marikaner (changes requested):
Overall looks good, I mostly have questions and minor suggestions, but with the tests I really think mock-fs would be awesome.

packages/util/src/fs.ts
```diff
@@ -35,3 +36,15 @@ export function readJSON(path: PathLike): { [key: string]: any } {
   logger.warn(`File "${path}" does not exist, return empty object.`);
   return {};
 }
+
+export function transpileDirectory(path: string): Promise<void> {
+  logger.info(`Transpiling files in the directory: ${path} started.`);
```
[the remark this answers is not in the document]
FrankEssenberger: set to deub

packages/util/src/fs.ts
```diff
@@ -35,3 +36,15 @@ export function readJSON(path: PathLike): { [key: string]: any } {
...
+  return execa('tsc', { cwd: path })
+    .then(() => {
+      logger.info(`Transpiling files in directory: ${path} finished.`);
+    })
+    .catch(err => {
+      logger.error(`Error: Failed to generate js files: ${err}`);
+      process.exit(1);
+    });
```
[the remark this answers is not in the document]
FrankEssenberger: execa is not declared as `async` funciton so when I used await it was complaining.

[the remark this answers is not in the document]
marikaner: Are you sure, that it wasn't complaining that `transpileDirectory` has to be async? Because you should be able to even do things like this: `await 3;`.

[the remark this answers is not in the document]
marikaner: Are you sure, that it wasn't complaining that `transpileDirectory` has to be async? Because you should be able to even do things like this: `await 3;`.

packages/openapi-generator/src/generator.ts
marikaner: I just saw your funfact comment and crosschecked with [the documentation](https://nodejs.org/api/fs.html#fs_fs_rmdir_path_options_callback). According to this it should not break. Is it possible that there was another error?

FrankEssenberger: duplicate

packages/openapi-generator/src/generator.ts
FrankEssenberger: This is a result of the Promise.all(promises from the N spech files), which results in a single promise with an array giving the results. I added now a `.then(()=>undefined)` to get a single promise out of it.

packages/util/src/fs.ts
```diff
@@ -35,3 +36,15 @@ export function readJSON(path: PathLike): { [key: string]: any } {
   logger.warn(`File "${path}" does not exist, return empty object.`);
   return {};
 }
+
+export function transpileDirectory(path: string): Promise<void> {
```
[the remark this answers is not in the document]
FrankEssenberger: or we make it hidden because it is not for public use?

packages/openapi-generator/src/wrapper-files/tsconfig-json.ts
```diff
@@ -0,0 +1,31 @@
...
+      moduleResolution: 'node',
+      esModuleInterop: true,
+      inlineSources: false,
+      strict: true
+    },
+    include: ['**/*.ts'],
+    exclude: ['dist/**/*', 'test/**/*', '**/*.spec.ts', 'node_modules/**/*']
+  };
+  if (options.tsConfig) {
+    tsConfig = readJSON(options.tsConfig) as any;
```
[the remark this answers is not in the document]
FrankEssenberger: I first had the merge version. But since both objects can have subobjects a proper merge is not so simple. I think if your are at the step, that you want to adjust the default tsconfig I would say you either take the existing one as a starting point or you are able to understand the compile error and adjust the config accordingly.

packages/util/src/fs.ts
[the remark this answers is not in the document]
FrankEssenberger: yes you are right

packages/util/src/fs.ts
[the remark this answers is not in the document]
FrankEssenberger: I added a documentation.

packages/openapi-generator/src/wrapper-files/tsconfig-json.spec.ts
FrankEssenberger: Then it would be symmetric to the OData case

FrankEssenberger:
So I adjusted the code:
- The openApi are now generated and transpiled like the OData ones in the `generate-test-services.ts` script. 
- The openApi are then also copied to the core with the local core as reference like the OData ones (in case we want to do some mock tests later for the openApi request builders as well
- The `generate-test-services.ts` has been adjusted to really fail if something goes wrong in one of the steps. In the past the promises were not handled. Now it is linear (await) but we should see if some errors appear there.

packages/openapi-generator/src/wrapper-files/tsconfig-json.ts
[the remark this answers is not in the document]
marikaner: I think that makes sense until proven otherwise :)

packages/openapi-generator/src/wrapper-files/tsconfig-json.spec.ts
marikaner: I agree that this is rather an E2E test.
And you are right, I think execa would require different handling if combined with mock-fs. Then again, execa is has nothing to do with the module at hand (tsconfig-json.ts) and therefore should not be of concern for this spec.
I don't fully understand the dist dir thing, but let's discuss this separately.

packages/openapi-generator/src/generator.ts
```diff
@@ -42,18 +44,20 @@ export async function generate(options: GeneratorOptions): Promise<void> {
...
 
-    await generateFromFile(
+    return generateFromFile(
       inputFilePath,
       options,
       vdmMapping,
       uniqueServiceName
     );
   });
+  return Promise.all(promises).then(() => undefined);
```
marikaner, line 60: I still think the code would be a little nicer. The reason is, that a function that has the return type void is a function that returns nothing. In my eyes not returning anything comes more natural than return `undefined`. Consider this example as a non-async case, that we would definitely not implement like this:
```ts
function setSomeGlobalVariable(val: any): void {
  globalVariable = val;
  return undefined;
}
```
```suggestion
  await Promise.all(promises);
```

marikaner (approved):
I left one minor suggestion for the code, other than that LGTM / covered by follow up ;)

packages/openapi-generator/src/generator.ts
FrankEssenberger: yes agreed. I think your changes does implicitly the same, but to use the feature that if nothing said it is done automatically is more elegant.

FrankEssenberger:
Marika approved.

PR #918, opened 2021-01-15
fix: remove copyright from json

jjtang1985:
Since json does not support comments, the copyright comments make it invalid.


FrankEssenberger (approved):
LGTM I like the approach with a boolean flag. I also had this first, but did not want to change so much.

PR #922, opened 2021-01-18
fix: Allow duplicate and no operationIds

marikaner:
This should allow us to generate for erroneous specifications as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,12 +22,12 @@
 
 ## Improvements
 
--
+- [openapi-generator] Fix generation for duplicate operation names.
 
 ## Fixed Issues
 
 - [openapi-generator] Remove the copyright comments from the generated package.json.
-
+- [util] Don't fail on undefined template arguments when using `codeBlock`.
 
 # 1.36.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@
 
 ## Improvements
 
-- [openapi-generator] Fix generation for duplicate operation names.
+- [openapi-generator] Improve generation for duplicate operation names and non existing `operationId`s.
 
 ## Fixed Issues
 
--- a/packages/openapi-generator/src/document-converter.ts
+++ b/packages/openapi-generator/src/document-converter.ts
@@ -3,7 +3,13 @@ import { parse } from 'path';
 import { OpenAPIV3 } from 'openapi-types';
 import { convert } from 'swagger2openapi';
 import { load } from 'js-yaml';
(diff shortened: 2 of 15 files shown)


packages/openapi-generator/src/parser/operation.ts
```diff
@@ -15,15 +14,14 @@ export function parseOperation(
...
-  const operationName = parseOperationName(operation, pattern, method);
 
   return {
     ...operation,
     pattern,
     method,
     requestBody,
     parameters,
-    operationName
+    operationName: operation.operationId!
```
jjtang1985, line 24: how about rename it to `operationId`?

packages/openapi-generator/src/document-converter.ts
```diff
@@ -73,13 +79,168 @@ export function convertDocToGlobalTag(
...
+
+const nameMapping = {
+  get: 'get',
+  post: 'create',
+  patch: 'update',
+  put: 'update',
+  delete: 'delete',
+  head: 'getHeadersFor',
+  options: 'getOptionsFor',
+  trace: 'trace'
```
jjtang1985, line 245: I guess you did a manual test with the openapi generator and made this alignment.
Could we add a follow up ticket, so we can have a test that checking this alignment or maybe it's covered by some tests implicitly?

jjtang1985 (approved):
LGTM.

packages/openapi-generator/src/document-converter.ts
marikaner: What exactly do you mean? This is used for the names of the operationIds if they are not given and there is test for some of those (I didn't go through all of them because that would be trivial). Or do you mean an alignment to something else?

CHANGELOG.md
```diff
@@ -22,12 +22,12 @@
 
 ## Improvements
 
--
+- [openapi-generator] Fix generation for duplicate operation names.
```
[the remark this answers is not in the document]
marikaner: The openAPI generator actually "fixes" them too. Now that I already built it, I wouldn't want to remove it ;) But I'll improve the message a bit as this is more an improvement than a fix.

marikaner, line 25:
```suggestion
- [openapi-generator] Improve generation for duplicate operation names and non existing operation ids.
```

packages/openapi-generator/src/parser/operation.ts
marikaner: Actually you are totally right, and with this I could even make the type a bit nicer.

jjtang1985 (approved):

PR #927, opened 2021-01-20
chore: create more openapi tests

tomfrenken:
- Now includes tests for the `getInputFilePaths` function
- Added mock-fs to the dev dependencies


packages/openapi-generator/src/generator.ts
```diff
@@ -201,16 +201,16 @@ async function generateFromFile(
...
  * @returns all file paths as a string array.
  */
-async function getInputFilePaths(input: string): Promise<string[]> {
+export async function getInputFilePaths(input: string): Promise<string[]> {
   if ((await lstat(input)).isFile()) {
     return [input];
   }
 
-  const directoryContents = await readdir(input, { withFileTypes: true });
+  const directoryContents = await readdir(input);
```
marikaner, line 209: I now wonder why nothing is failing. This should not work... Because later execution needs the file type...

marikaner (changes requested):
This should fail, if it doesn't let's take a look at why...

packages/openapi-generator/src/generator.ts
tomfrenken: Well, previously the filetype was derived from the `fs.Dirent` object I believe, but since then `lstat()` was used, so we have a stats object already, which is why it is not failing.

tomfrenken: Also, `lstat` was used instead of the `Dirent`, because of consistency if I remember

packages/openapi-generator/src/generator.spec.ts
```diff
@@ -1,7 +1,29 @@
...
+        'test-service.txt': 'file content here',
+        'empty-dir': {},
+        'sub-dir': {
+          'test-service.txt': 'another fake service',
+          'sub-directory-service.txt': 'just to add some more'
+        }
+      }
+    });
+
+    expect((await getInputFilePaths('/path/to/test/dir')).sort()).toEqual(
```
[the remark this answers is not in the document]
tomfrenken: Hmm true, I guess I could just sort the test array by hand and don't need to use the `sort` method.

marikaner (approved):
LGTM

PR #938, opened 2021-01-26
feat: Support readme generation and copying additional files

jjtang1985:
Added the same options like odata generator: `additionalFiles` and `writeReadme`


CHANGELOG.md
```diff
@@ -18,7 +18,9 @@
 
 ## New Functionality
 
-- [openApi-generator] Support the generation of transpiled JavaScript OpenAPI clients with the new flags: `generateJs` and `tsConfig`.
+- [openapi-generator] Support the generation of transpiled JavaScript OpenAPI clients with the new flags: `generateJs` and `tsConfig`.
+- [openapi-generator] Add options `writeReadme` for generating readme files.
```
marikaner, line 22:
```suggestion
- [openapi-generator] Add option `writeReadme` for generating readme files.
```

packages/core/test/test-util/test-services/openapi/swagger-yaml-service/README.md
```diff
@@ -0,0 +1,16 @@
+# swagger-yaml-service
+
+This package contains the OpenAPI VDM for the SwaggerYamlServiceApi.
```
marikaner, line 3: Are we speaking of VDM in the context of OpenApi (and generally non-s4)?
```suggestion
This package contains the OpenAPI client for the SwaggerYamlServiceApi.
```

packages/core/test/test-util/test-services/openapi/swagger-yaml-service/test-license
```diff
@@ -0,0 +1,35 @@
+SAP DEVELOPER LICENSE AGREEMENT
```
marikaner, line 1: Maybe use a dummy license file (or any other file, doesn't have to have a notion of license), because this might be confusing.
Acutally, adding a changelog is enough in my opinion.

packages/util/src/generator/readme.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
marikaner, line 1: Generally I agree with reducing code duplication, but in this case I am not so sure. Up to now, all functionality from util is generic and could potentially also be used by our users. This is very specific and nobody apart from us will use it. In my opinion this is a constant, that is used in two places rather than a reusable function, that should be shared publicly.
I think as of today, I would rather duplicate the code in the two generators. If we encounter more of these specific code duplications, I would even consider creating a separate package for it. @FrankEssenberger what is your opinion?

packages/openapi-generator/src/wrapper-files/readme.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @experimental This API is experimental and might change in newer versions. Use with caution.
+ * Generate the readme for an openapi client.
+ * @param openApiDocument Parsed service.
+ * @returns The readme contents.
+ */
+export function readme(openApiDocument: OpenApiDocument): string {
+  return codeBlock`# ${openApiDocument.npmPackageName}
+
+This package contains the OpenAPI VDM for the ${openApiDocument.apiName}.
```
marikaner:
```suggestion
This package contains the OpenAPI client for the ${openApiDocument.apiName}.
```

packages/openapi-generator/src/generator.ts
```diff
@@ -239,3 +253,34 @@ export async function getSdkVersion(): Promise<string> {
     await readFile(resolve(__dirname, '../package.json'), 'utf8')
   ).version;
 }
+
+function generateAdditionalFiles(
+  additionalFiles: string,
+  serviceDir: string
+): Promise<void[]> {
+  logger.info(
```
marikaner, line 261: Isn't this more debug than info?

packages/openapi-generator/src/wrapper-files/create-file.ts
```diff
@@ -41,3 +41,21 @@ ${content}
...
+ * @param toDirectoryPath Path of the directory to write to.
+ * @param overwrite Whether or not existing files should be overwritten.
+ */
+export async function copyFile(
+  fromPath: string,
+  toFileName: string,
+  toDirectoryPath: string,
+  overwrite: boolean
+): Promise<void> {
+  const fileContent = readFileSync(resolve(fromPath), { encoding: 'utf8' });
```
marikaner, line 59: I would really like to avoid using sync calls where possible. The async methods don't block the execution and I think we can boost our performance with this (especially when generating many services).
Also, why don't we use the fs.copyFile method?

packages/openapi-generator/src/generator.ts
```diff
@@ -239,3 +253,34 @@ export async function getSdkVersion(): Promise<string> {
     await readFile(resolve(__dirname, '../package.json'), 'utf8')
   ).version;
 }
+
+function generateAdditionalFiles(
```
marikaner, line 257:
```suggestion
function copyAdditionalFiles(
```

marikaner (dismissed):
I left some comments. I would still be interested in Frank's take on the util question.

packages/openapi-generator/src/generator.ts
jjtang1985: I don't have strong opinion and I'm fine with both. This is just an alignment with the odata generator.

packages/util/src/generator/readme.ts
jjtang1985: I think eventually we need a `generator-common` or `generator-util` package, for the time being, I would also like to prepare it. Therefore, a new folder is created in the `util`, which is called `generator` and will hold all the common functions. Without this, it's not clear when do we need a new package.

packages/core/test/test-util/test-services/openapi/swagger-yaml-service/test-license
jjtang1985: I used the `test-license` as the file name, but you are right, the content might also be confusing.

packages/util/src/generator/readme.ts
jjtang1985: Discussed with @FrankEssenberger , we will create a new package for that. I created a follow up.

packages/openapi-generator/src/wrapper-files/create-file.ts
```diff
@@ -41,3 +41,28 @@ ${content}
...
+ */
+export async function copyFile(
+  src: string,
+  dest: string,
+  overwrite: boolean
+): Promise<void> {
+  if (!overwrite && (await existsSync(dest))) {
+    return;
+  }
+  fsCopyFile(src, dest, err => {
```
marikaner, line 60: Isn't there a promises version of this function? With a Promise you can wait for the copy to finish - currently the end is sometime, but nobody knows when.

packages/openapi-generator/src/wrapper-files/readme.ts
```diff
@@ -0,0 +1,39 @@
...
+import { OpenApiDocument } from '../openapi-types';
+
+/**
+ * @experimental This API is experimental and might change in newer versions. Use with caution.
+ * Generate the readme for an openapi client.
+ * @param openApiDocument Parsed service.
+ * @returns The readme contents.
+ */
+export function readme(openApiDocument: OpenApiDocument): string {
+  return codeBlock`# ${openApiDocument.npmPackageName}
```
marikaner, line 11: Do you need the codeBlock here? It mostly takes care of indentation, so might be useful, just doesn't seem that this is used here.

packages/openapi-generator/src/wrapper-files/create-file.ts
```diff
@@ -41,3 +41,28 @@ ${content}
...
 }
+
+/**
+ * @experimental This API is experimental and might change in newer versions. Use with caution.
+ * Copy a file from a given path.
+ * @param src Path to the source file.
+ * @param dest Path to the destination file
+ * @param overwrite Whether or not existing files should be overwritten.
+ */
+export async function copyFile(
```
marikaner, line 52: Maybe the file-name needs to be changed? `create-file` does not imply copying. Maybe 'fs'?

packages/util/src/generator/index.ts
```diff
@@ -0,0 +1 @@
+export * from './unique-name-generator';
```
marikaner, line 1: I think there is no need to move it into a separate directory anymore as there will be a different package for generators.

scripts/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# Change Log
+
+All notable changes to this module will be documented centrally in the [SAP Cloud SDK release notes](https://sap.github.io/cloud-sdk/docs/js/release-notes-sap-cloud-sdk-for-javascript-and-typescript/).
```
marikaner, line 3: This is not a script - I would rather put it into test-resources.

marikaner (changes requested):
Looks good, but I think some files should be placed somewhere else / renamed.

scripts/CHANGELOG.md
jjtang1985: moved

packages/core/test/test-util/test-services/openapi/test-service/test-license
```diff
@@ -0,0 +1 @@
+test license
```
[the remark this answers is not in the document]
jjtang1985: use the same files as the odata generator test

packages/openapi-generator/src/wrapper-files/create-file.ts
jjtang1985: renamed

packages/openapi-generator/src/wrapper-files/create-file.ts
jjtang1985: 👍 switched to the promises version

packages/openapi-generator/src/wrapper-files/fs.spec.ts
```diff
@@ -1,5 +1,5 @@
 import * as fs from 'fs';
-import { createFile } from './create-file';
+import { createFile } from './fs';
```
marikaner, line 2: Does it make sense to add a test for `copyFile`?

packages/openapi-generator/src/generator.ts
```diff
@@ -76,6 +79,8 @@ async function generateSDKSources(
   await createFile(serviceDir, 'api.ts', apiFile(openApiDocument), true);
   await createFile(serviceDir, 'index.ts', indexFile(), true);
   if (options.generatePackageJson) {
+    logger.info(`Generating package.json in ${serviceDir}.`);
```
marikaner, line 82: Minor: Isn't that rather debug level

packages/generator/src/service/readme.ts
```diff
@@ -32,6 +32,11 @@ function businessDocumentationLine(service: VdmServiceMetadata): string[] {
     : [];
 }
 
+// TODO 1728 move to a new package for reduce code duplication.
+/**
+ * Build the helpful links section of the readme file.
+ * @returns The content of the section.
+ */
```
FrankEssenberger, line 39: Perhaps you can use the content of the generic readme I introduced in this PR: https://github.com/SAP/cloud-sdk-js/pull/974
 also for the readme section.

FrankEssenberger: We can also mention this in the ticket 1728 to do that.

packages/openapi-generator/src/generator.ts
```diff
@@ -239,3 +253,33 @@ export async function getSdkVersion(): Promise<string> {
     await readFile(resolve(__dirname, '../package.json'), 'utf8')
   ).version;
 }
+
+async function copyAdditionalFiles(
```
marikaner, line 257: Isn't this also duplicate? (add a todo?)

packages/openapi-generator/src/generator.spec.ts
```diff
@@ -56,4 +56,16 @@ describe('generator', () => {
     const tsconfig = resolve(testServicePath, 'tsconfig.json');
     expect(existsSync(tsconfig)).toBe(true);
   });
+
+  it('should create changelog and license files', () => {
+    const changelog = resolve(testServicePath, 'CHANGELOG.md');
```
FrankEssenberger, line 61: I wouls make two tests. One for the CHANGELOG and one for the additional file.

packages/openapi-generator/src/generator.ts
```diff
@@ -99,6 +105,14 @@ async function generateSDKSources(
     );
     await transpileDirectory(serviceDir);
   }
+
+  if (options.additionalFiles) {
```
FrankEssenberger, line 109: I see new `CHANGELOG.md` files but I do not see an option, are they copied via the additional files I guess? Perhaps you can adjust the PR title.

FrankEssenberger (approved):
LGTM only a few minor comments.

packages/openapi-generator/src/wrapper-files/fs.ts
```diff
@@ -1,7 +1,8 @@
-import { promises } from 'fs';
+import { promises, existsSync } from 'fs';
 import { join } from 'path';
 import { codeBlock } from '@sap-cloud-sdk/util';
 const { writeFile } = promises;
+const fsCopyFile = promises.copyFile;
```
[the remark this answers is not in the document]
jjtang1985: No, it won't work.
The `copyFile` function is created in this file, so I have to rename it, making it hard to merge to the previous line.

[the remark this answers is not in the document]
marikaner: `const { copyFile: fsCopyFile, writeFile } = promises;`

packages/generator/src/service/readme.ts
jjtang1985: created a follow up

packages/openapi-generator/src/generator.ts
jjtang1985: I'm fine with `debug`, but currently most of the cases we use `info`, perhaps we should reconsider all the log level again for alignment.

packages/openapi-generator/src/wrapper-files/fs.spec.ts
jjtang1985: added tests

PR #941, opened 2021-01-26
fix: test service spec error detected by the validator

jjtang1985:
The validation step was skipped. I fixed one spec error detected by the validator.


FrankEssenberger (dismissed):
LGTM. Only one thing you could do - check if the generator fails with exit code != 0 in case of a faulty spec file.

FrankEssenberger:
Only one thing you could do - check if the generator fails with exit code != 0 in case of a faulty spec file.

jjtang1985:
@FrankEssenberger , I skipped the validation again and only kept the fix for the test service spec after discussed with @marikaner , because this might be related to a decision about whether we allow duplicate `operationId` in the spec file.

test-resources/openapi-service-specs/swagger-yaml-service.yml
```diff
@@ -41,7 +41,8 @@ paths:
       responses:
         '200':
           description: Success
-          type: string
+          schema:
+            type: string
```
jjtang1985, line 45: the real fix

marikaner: Good catch, I though this was 2.0 syntax, but I crosschecked the docs and it isn't ;)

marikaner (approved):
LGTM

PR #961, opened 2021-02-02
chore: fix the unhandledRejection warning when generating test services

jjtang1985:
- Fix the warning:
`UnhandledPromiseRejectionWarning: Error: Reading test service file 'openapi' failed: Error: EISDIR: illegal operation on a directory` by applying a filter `dirent.isFile()`.
- The script will fail in there future instead of having a warning.

### DIFF
--- a/scripts/generate-test-services.ts
+++ b/scripts/generate-test-services.ts
@@ -85,12 +85,14 @@ async function generateTestServicesWithLocalCoreModules(
   }
 
   (await readServiceDirectories()).forEach(serviceDirectory =>
-    readServiceDirectory(serviceDirectory).then(files =>
-      files.forEach(file =>
-        readServiceFile(serviceDirectory, file).then(data => {
-          replaceWithLocalModules(serviceDirectory, file, data);
-        })
-      )
+    readServiceDirectory(serviceDirectory).then(dirents =>
+      dirents
+        .filter(dirent => dirent.isFile())
+        .forEach(dirent =>
+          readServiceFile(serviceDirectory, dirent.name).then(data => {
+            replaceWithLocalModules(serviceDirectory, dirent.name, data);
+          })
+        )
     )
   );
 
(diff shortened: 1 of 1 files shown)


scripts/generate-test-services.ts
```diff
@@ -85,12 +85,14 @@ async function generateTestServicesWithLocalCoreModules(
...
 
   (await readServiceDirectories()).forEach(serviceDirectory =>
-    readServiceDirectory(serviceDirectory).then(files =>
-      files.forEach(file =>
-        readServiceFile(serviceDirectory, file).then(data => {
-          replaceWithLocalModules(serviceDirectory, file, data);
-        })
-      )
+    readServiceDirectory(serviceDirectory).then(dirents =>
+      dirents
```
FrankEssenberger, line 89: `dirents` are either files or directories? I do not really like the name. Perhaps `serviceDirectoryContent`?

scripts/generate-test-services.ts
```diff
@@ -85,12 +85,14 @@ async function generateTestServicesWithLocalCoreModules(
...
+        .filter(dirent => dirent.isFile())
+        .forEach(dirent =>
```
FrankEssenberger, line 91: And here since you filtered these are only `serviceSourceFiles` right?

scripts/generate-test-services.ts
```diff
@@ -132,6 +134,11 @@ async function generateTestServicesWithLocalCoreModules(
 }
 
 async function generateAll(): Promise<void> {
+  process.on('unhandledRejection', (reason, promise) => {
```
FrankEssenberger, line 137: I assume this was needed, but I am wondering why the catch in line 182 did not handle it.

FrankEssenberger (approved):
LGTM: Only two minor questions and comments

scripts/generate-test-services.ts
jjtang1985: Discussed and added a comment.

scripts/generate-test-services.ts
jjtang1985: Added the `fs.Direent` type name for less confusing.

PR #974, opened 2021-02-04
chore: improve readme.md in the packages

FrankEssenberger:
Improved the README.md files in the packages.
I will try to automatically put this with some placeholder like oclif does it.
Since a few of the old script files are JS and the new ones are TS I wanted to make the `util.js` als a TS file.

The only non TS part is the github action so I transpiled the script for that use case. I think you will never touch this really.


FrankEssenberger:
Please go ahead and review.

scripts/util.ts
```diff
@@ -0,0 +1,44 @@
...
+    cwd: resolve(__dirname, '../')
+  });
+  return JSON.parse(response.stdout) as LernaModule[];
+}
+
+export async function getNonTestLernaModules(): Promise<LernaModule[]> {
+  const response = await execa('lerna', ['list', '--json', '-a'], {
+    cwd: resolve(__dirname, '../')
+  });
+  return (JSON.parse(response.stdout) as LernaModule[]).filter(
```
jjtang1985, line 34:
```suggestion
  getAllLernaModules().filter(
```

jjtang1985 (changes requested):
My major concern is the `util.ts`, which is used by the js scripts.

packages/core/README.md
```diff
@@ -4,13 +4,18 @@
...
+
+## Installation
 ```
 $ npm install @sap-cloud-sdk/core
 ```
 
-### Usage
-The core is the heart of the SAP Cloud SDK and contains the functionality that is essential to every project powered by the SDK. Any OData client built by the SAP Cloud SDK, be it the VDM or clients built by the generator are using the core. We recommend to install this in addition to your clients.
+## Usage
+
```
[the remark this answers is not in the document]
FrankEssenberger: I am not sure what you mean. Shout we do have no empty line after or before the `## Usage`?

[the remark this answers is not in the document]
jjtang1985: It's hard to see the difference from my suggested change, but you added a new empty line (line 13) after the `## Usage`. I guess we should delete it.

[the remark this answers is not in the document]
FrankEssenberger: ok there is a single empty line now so I guess this is fine.

FrankEssenberger:
> My major concern is the `util.ts`, which is used by the js scripts.

I changed all script files to `ts` now. Only in the github action this is not supported so I transpiled the one small `getChangelog` method. I executed the transpiled method and it gave the expected result so I hope I did not break something.

.github/actions/get-changelog/action.yml
```diff
@@ -5,4 +5,4 @@ outputs:
     description: 'The current changelog'
 runs:
   using: 'node12'
-  main: 'index.js'
+  main: 'transpiled/.github/actions/get-changelog/index.js'
```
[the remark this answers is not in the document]
marikaner: Also, it would be great if we wouldn't transpile .d.ts files

packages/analytics/README.md
```diff
@@ -1,5 +1,39 @@
...
 # cloud-sdk-analytics
 
 This package contains the code we use for collecting usage data.
 Please note that usage analytics is disabled by default, i.e. you need to explicitly opt-in for data to be collected.
-For more information on how to opt-in and which data is collected, please refer to [this document](https://github.com/SAP/cloud-sdk-cli/blob/main/usage-analytics.md).
+For more information on how to opt-in and which data is collected, please refer to [this document](https://github.com/SAP/cloud-sdk-js/blob/main/packages/analytics/usage-analytics.md).
+
+
+<!-- This block is inserted by scripts/copy-generic-readme.ts and not oclif like the commands block --> 
+<!--genericPart-->
```
marikaner, line 11: Maybe instead of an additional comment give it a more speaking name? E. g. `<!-- sap-cloud-sdk-common-readme -->`. In addition you could call the script `replace-common-readme.ts` to correspond to this.

marikaner: BTW, I think an additional comment can make sense to warn us not to change the generated part. In that case, it should be part of what is generated. But I think this might be overkill.

packages/openapi-generator/README.md
```diff
@@ -1,17 +1,18 @@
-## SAP Cloud SDK OpenAPI Client Generator (Beta)
+<a href="https://sap.com/s4sdk"><img src="https://help.sap.com/doc/2324e9c3b28748a4ae2ad08166d77675/1.0/en-US/logo-with-js.svg" alt="SAP Cloud SDK for JavaScript Logo" height="122.92" width="226.773"/></a>
```
marikaner, line 1: BTW this could also be replaced, right? `<!-- sap-cloud-sdk-logo -->`

packages/openapi-generator/package.json
```diff
@@ -38,8 +38,9 @@
     ]
   },
   "scripts": {
-    "compile": "yarn tsc -b && oclif-dev readme && cp -r -f src/templates dist",
+    "compile": "yarn tsc -b && cp -r -f src/templates dist",
     "prepare": "yarn compile",
+    "readme": "oclif-dev readme",
```
marikaner, line 43: This should be run automatically somewhere, so that we can't forget to update it.

scripts/GENERIC-README-PART.md
```diff
@@ -0,0 +1,27 @@
+## Support
```
marikaner, line 1: Minor: Consider renaming it to `COMMON-README.md`. "Generic" has a slightly different tone.

marikaner (changes requested):
I absolutely love the idea of replacing the common parts in the readme!
I have some minor renaming suggestions and a few suggestions regarding the transpilation of the get-changelog action.

packages/openapi-generator/README.md
FrankEssenberger: yes good idea.

scripts/GENERIC-README-PART.md
FrankEssenberger: yes

packages/openapi-generator/package.json
FrankEssenberger: yes. I will include it in the postinstall step, ok?

packages/analytics/README.md
FrankEssenberger: I added a `Do not adjust it manually part`

.github/actions/get-changelog/action.yml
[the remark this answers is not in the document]
FrankEssenberger: I removed the .d.ts files. However I can not get lesser deep folder structure. If I put the root dir to the action, it is es desired from the folder structure but the compiler complains because not all source files are under root. The `rootDirs` make the import easier but they do not change the generated result. What would work is making a proper module out of the script files folder, but this seems a bit too much effort.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sap-cloud-sdk",
-  "version": "1.37.1",
+  "version": "1.37.0",
```
jjtang1985, line 3:
```suggestion
  "version": "1.37.1",
```

jjtang1985 (changes requested):
👍 for switching to `.ts` files. Only missing one package json version update.

jjtang1985 (approved):

packages/openapi-generator/package.json
marikaner: I think that should be ok

.github/actions/get-changelog/action.yml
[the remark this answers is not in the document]
marikaner: I agree. Still, consider transpiling it to `dist`.

packages/analytics/README.md
```diff
@@ -1,5 +1,42 @@
...
 # cloud-sdk-analytics
 
 This package contains the code we use for collecting usage data.
 Please note that usage analytics is disabled by default, i.e. you need to explicitly opt-in for data to be collected.
-For more information on how to opt-in and which data is collected, please refer to [this document](https://github.com/SAP/cloud-sdk-cli/blob/main/usage-analytics.md).
+For more information on how to opt-in and which data is collected, please refer to [this document](https://github.com/SAP/cloud-sdk-js/blob/main/packages/analytics/usage-analytics.md).
+
+
+<!-- This block is inserted by scripts/copy-generic-readme.ts and not oclif like the commands block. Do not adjust it manually. --> 
+<!--sapCloudSdkCommonReadme-->
```
marikaner, line 14: What do you think about moving the comment into the generated block? Because currently I think you are adding it manually everywhere, which might be a bit annoying.
```suggestion
<!--sapCloudSdkCommonReadme-->
<!-- This block is inserted by scripts/copy-generic-readme.ts and not oclif like the commands block. Do not adjust it manually. --> 
```

scripts/util.ts
```diff
@@ -0,0 +1,41 @@
...
+    cwd: resolve(__dirname, '../')
+  });
+  return JSON.parse(response.stdout) as LernaModule[];
+}
+
+export async function getNonTestLernaModules(): Promise<LernaModule[]> {
+  return (await getAllLernaModules()).filter(
+    module => !module.location.includes('test-packages')
+  );
+}
```
marikaner, line 34: Consider renaming:
```suggestion
export async function getProductiveLernaModules(): Promise<LernaModule[]> {
  return (await getAllLernaModules()).filter(
    module => !module.location.includes('test-packages')
  );
}
```

scripts/replace-common-readme.ts
```diff
@@ -0,0 +1,64 @@
+import { resolve } from 'path';
+import { promises, readFileSync } from 'fs';
+import { EOL } from 'os';
+import { createLogger } from '@sap-cloud-sdk/util';
+import { getNonTestLernaModules } from './util';
+
+const startTagCommonReadme = '<!--sapCloudSdkCommonReadme-->';
+const endTagCommonReadme = '<!--sapCloudSdkCommonReadmeStop-->';
+const startTagLogo = '<!--sapCloudSdkLogo-->';
+const endTagLogo = '<!--sapCloudSdkLogoStop-->';
```
marikaner, line 10: Very minor, I would add spaces and use kebapcase here:
```suggestion
const startTagCommonReadme = '<!-- sap-cloud-sdk-common-readme -->';
const endTagCommonReadme = '<!-- sap-cloud-sdk-common-readme-stop -->';
const startTagLogo = '<!-- sap-cloud-sdk-logo -->';
const endTagLogo = '<!-- sap-cloud-sdk-logo-stop -->';
```

marikaner (approved):
I left you some more comments, but other than that we're good to go.

PR #978, opened 2021-02-06
fix: shared entity type as return type of function/action

jjtang1985:
When an entity type is shared by multiple entity sets and is used as return type of a(n) function/action import, we pick the first entity set currently, which is wrong.

In this PR, an error will be thrown for this case as a known issue.
In the future, using `executeRaw` is the recommended approach.

- [x] ~~As always, the vdm is adjusted manually first, once approved, the generator will be adjusted.~~

~~This PR should be merged after https://github.com/SAP/cloud-sdk-js/pull/1067.~~


packages/core/src/odata-v4/request-builder/function-import-request-builder.spec.ts
```diff
@@ -17,4 +23,16 @@ describe('FunctionImportRequestBuilder', () => {
       `NonNullableStringParam='${params.nonNullableStringParam}'`
     );
   });
+
+  it('throws an error when shared entity type is used as return type', async () => {
```
jjtang1985, line 27: Test for showing the error message.

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
```diff
@@ -87,6 +89,46 @@ export function testFunctionImportEntityReturnTypeCollection(parameters: TestFun
...
+ * @param parameters - Object containing all parameters for the function import.
+ * @returns A request builder that allows to overwrite some of the values and execute the resultng request.
+ */
+export function testFunctionImportSharedEntityReturnType(parameters: TestFunctionImportSharedEntityReturnTypeParameters): FunctionImportRequestBuilderV4<TestFunctionImportSharedEntityReturnTypeParameters, any> {
+  const params = {
+
+  }
+
+  return new FunctionImportRequestBuilderV4('/sap/opu/odata/sap/API_TEST_SRV', 'TestFunctionImportSharedEntityReturnType', (data) => throwErrorWhenReturnTypeIsUnionType('TestFunctionImportSharedEntityReturnType'), params);
+}
```
jjtang1985, line 110: The function import adjusted manually, where the return type is set to `any` and the deserialise function is replaced by a function that throws an error: `throwErrorWhenReturnTypeIsUnionType`;

test-resources/odata-service-specs/v4/API_TEST_SRV/API_TEST_SRV.edmx
```diff
@@ -224,6 +239,10 @@
           <NavigationPropertyBinding Path="to_OtherMultiLink" Target="A_TestEntityMultiLink"/>
           <NavigationPropertyBinding Path="to_SingleLink" Target="A_TestEntitySingleLink"/>
         </EntitySet>
+        <EntitySet Name="A_TestEntitySharesEntityType1" EntityType="API_TEST_SRV.A_TestEntityTypeToBeShared" sap:content-version="1">
+        </EntitySet>
+        <EntitySet Name="A_TestEntitySharesEntityType2" EntityType="API_TEST_SRV.A_TestEntityTypeToBeShared" sap:content-version="1">
+        </EntitySet>
```
jjtang1985, line 245: These two entity sets share the same entity type `API_TEST_SRV.A_TestEntityTypeToBeShared`

test-resources/odata-service-specs/v4/API_TEST_SRV/API_TEST_SRV.edmx
```diff
@@ -202,6 +209,14 @@
         <ReturnType Type="Collection(API_TEST_SRV.A_TestEntityType)"/>
       </Function>
 
+      <Function Name="TestFunctionImportSharedEntityReturnType">
+        <ReturnType Type="API_TEST_SRV.A_TestEntityTypeToBeShared"/>
```
jjtang1985, line 213: The shared entity type is used as the return type.

jjtang1985 (commented):
Please only review the `function-import.ts` and the test for now.

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
jjtang1985: @marikaner , as discussed, the error is thrown via the client code.

FrankEssenberger:
> In this PR, an error will be thrown for this case as a known issue.
> In the future, using `executeRaw` is the recommended approach.
> 
> As always, the vdm is adjusted manually first, once approved, the generator will be adjusted.

I like this approach because it will keep the PRs small. However, we do a generate:all in the pipeline and a git diff. Will this then break the pipeline? 


Regarding the content: Fine for me, but I would like to discuss if it makes sense to generate a function import which will throw an exception when called? Would it not be better to not generate the faulty function import at all?

jjtang1985:
@FrankEssenberger 
> I like this approach because it will keep the PRs small. However, we do a generate:all in the pipeline and a git diff. Will this then break the pipeline?
This is the first step of the PR, because generating a client with a function import that will throwing an error is still a open topic that needs to be aligned before adjusting the generator. Once this is aligned, I'll adjust the generator.

>Regarding the content: Fine for me, but I would like to discuss if it makes sense to generate a function import which will throw an exception when called? Would it not be better to not generate the faulty function import at all?
The most important thing is we should provide the `executeRaw` as the work around for the users, which is not there.
When implementing the `executeRaw`, the function import will be reused next time for both `execute` (using deserialiser => see error) and `executeRaw` (ignoring deserialiser => no error)

Fortunately, there is even no known metadata that makes it happen.

FrankEssenberger:
@Junjie but when we merge it, will the generate:all and git diff not break the pipeline?

FrankEssenberger:
So after a discussion, I am fine with the current state i.e. throwing an exception. The final solution will be, that the exception provides some guidance to use the execute raw.

jjtang1985:
@marikaner , one failing test is related to [this line](https://github.com/SAP/cloud-sdk-js/blob/aa7df2c020a9080b8bcab9f7a32ef3f33986af5e/test-resources/odata-service-specs/v2/API_TEST_SRV/API_TEST_SRV.edmx#L224), where duplicate entity sets are detected. Do you remember whether 
- we should treat them as same entity sets and ignore one of the duplicate one (I guess this one)
or
- we make them as case sensitive and two entity sets will be generated.

marikaner:
> @marikaner , one failing test is related to [this line](https://github.com/SAP/cloud-sdk-js/blob/aa7df2c020a9080b8bcab9f7a32ef3f33986af5e/test-resources/odata-service-specs/v2/API_TEST_SRV/API_TEST_SRV.edmx#L224), where duplicate entity sets are detected. Do you remember whether
> 
> * we should treat them as same entity sets and ignore one of the duplicate one (I guess this one)
>   or
> * we make them as case sensitive and two entity sets will be generated.

This fix was built because of an issue that the C4C colleagues had, ignoring one of them would break this behavior again, so I would strongly advise against this. What is wrong with the previous fix using the unique name generator?

jjtang1985:
> This fix was built because of an issue that the C4C colleagues had, ignoring one of them would break this behavior again, so I would strongly advise against this. What is wrong with the previous fix using the unique name generator?

The unique name generator is fine, but we then have one entity type (`A_TestEntityType`) that maps to two entity sets (`A_Testentity` and `A_TestEntity`).
In this PR, this entity type cannot be used as return type of a function import.
OK, I'll adjust the test service and keep supporting all the existing strategy for C4C.

jjtang1985:
@marikaner @FrankEssenberger , please review .

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
```diff
@@ -87,6 +87,46 @@ export function testFunctionImportEntityReturnTypeCollection(parameters: TestFun
...
+ *
+ * @param parameters - Object containing all parameters for the function import.
+ * @returns A request builder that allows to overwrite some of the values and execute the resultng request.
+ */
+export function testFunctionImportSharedEntityReturnType(parameters: TestFunctionImportSharedEntityReturnTypeParameters): FunctionImportRequestBuilderV4<TestFunctionImportSharedEntityReturnTypeParameters, any> {
+  const params = {
+
+  }
+
+  return new FunctionImportRequestBuilderV4('/sap/opu/odata/sap/API_TEST_SRV', 'TestFunctionImportSharedEntityReturnType', (data) => throwErrorWhenReturnTypeIsUnionType(data, 'TestFunctionImportSharedEntityReturnType'), params);
```
marikaner, line 107: Now that I see this, I actually have another idea, that might be better or might be overkill. Maybe the `.execute` method should not exist if this cannot be deserialized. That way we wouldn't have to throw an error, it just wouldn't exist.
I think an even better (and easier) solution might be, that is exists, but it does not exist on the type, so that you get an error at design time and in case you are using js you get a meaningful error at runtime (as it is now). What do you think?

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
```diff
@@ -87,6 +87,46 @@ export function testFunctionImportEntityReturnTypeCollection(parameters: TestFun
...
+ */
+export interface TestFunctionImportSharedEntityReturnTypeParameters {
+}
+
+/**
+ * Test Function Import Shared Entity Return Type.
+ *
+ * @param parameters - Object containing all parameters for the function import.
+ * @returns A request builder that allows to overwrite some of the values and execute the resultng request.
+ */
```
marikaner, line 101: I think it would be helpful if the comment would also contain some information on this.

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
```diff
@@ -87,6 +87,46 @@ export function testFunctionImportEntityReturnTypeCollection(parameters: TestFun
...
+export function testFunctionImportSharedEntityReturnType(parameters: TestFunctionImportSharedEntityReturnTypeParameters): FunctionImportRequestBuilderV4<TestFunctionImportSharedEntityReturnTypeParameters, any> {
```
marikaner, line 102: See comment below
```suggestion
export function testFunctionImportSharedEntityReturnType(parameters: TestFunctionImportSharedEntityReturnTypeParameters): Exclude<FunctionImportRequestBuilderV4<TestFunctionImportSharedEntityReturnTypeParameters, any>, 'execute> {
```

packages/generator/src/action-function-import/request-builder-arguments.ts
```diff
@@ -9,13 +10,21 @@ export function getRequestBuilderArgumentsBase(
...
+      ]
+    : [
+        `'${service.servicePath}'`,
+        `'${actionFunctionImport.originalName}'`,
+        `(data) => ${responseTransformerFunctionName(
+          actionFunctionImport.returnType,
+          service.oDataVersion
+        )}(data, ${actionFunctionImport.returnType.builderFunction})`,
+        'params'
+      ];
```
marikaner, line 29: How about this:

```suggestion
  const transformer = actionFunctionImport.returnType.returnTypeCategory === VdmReturnTypeCategory.ENTITY_NOT_DESERIALIZABLE ? `(data) => throwErrorWhenReturnTypeIsUnionType(data, '${actionFunctionImport.originalName}')` : `(data) => ${responseTransformerFunctionName(
          actionFunctionImport.returnType,
          service.oDataVersion
        )}(data, ${actionFunctionImport.returnType.builderFunction})`;
  return [
        `'${service.servicePath}'`,
        `'${actionFunctionImport.originalName}'`,
        transformer,
        'params'
      ];
```

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -67,7 +67,7 @@ function parseReturnTypes(
   }
 
   const entity = findEntityType(returnType, entities);
-  if (entity) {
+  if (entity.length) {
```
marikaner, line 70: Minor `entity` might not be the best name here.

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
marikaner: Another thought: this actually executes the request before throwing the error. I don't have a good idea at the moment, but if we know that this will happen here already, it doesn't seem appropriate.
The only valid idea that I have right now (and don't like too much) would be to have a `NoExecuteFnImportRB` base class, that either has no `.execute` function or maybe a protected one, that can still be invoked at runtime, but shows you an error at design time.

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -150,15 +150,27 @@ function getEdmReturnType(
...
+        builderFunction: first(entities)!.className,
+        isMulti: isCollection,
+        isCollection
+      }
+    : {
+        returnTypeCategory: VdmReturnTypeCategory.ENTITY_NOT_DESERIALIZABLE,
+        returnType: 'any',
+        builderFunction: '',
+        isMulti: false,
+        isCollection: false
```
marikaner, line 172: The correct value for `isCollection` should be passed to the function already

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -150,15 +150,27 @@ function getEdmReturnType(
...
+    ? {
+        returnTypeCategory: VdmReturnTypeCategory.ENTITY,
+        returnType: first(entities)!.className,
+        builderFunction: first(entities)!.className,
+        isMulti: isCollection,
+        isCollection
+      }
+    : {
+        returnTypeCategory: VdmReturnTypeCategory.ENTITY_NOT_DESERIALIZABLE,
+        returnType: 'any',
```
marikaner, line 169: Maybe `void` is the better return type here, given that nothing will ever be returned ;)

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -150,15 +150,27 @@ function getEdmReturnType(
...
+        builderFunction: '',
```
marikaner, line 170: This seems a bit "workaroundy"...

packages/generator/src/vdm-types.ts
```diff
@@ -131,7 +131,8 @@ export enum VdmReturnTypeCategory {
   ENTITY,
   COMPLEX_TYPE,
   EDM_TYPE,
-  VOID
+  VOID,
+  ENTITY_NOT_DESERIALIZABLE
```
marikaner, line 135: `ENTITY_NOT_DESERIALIZABLE` does not really sound like a category to me... See the overall comment from my review.

marikaner (dismissed):
I am a bit worried about the approach to send the request first and only then throw an error. I have the impression, that going through the return type category is a workaround to make this problem fit into what is already there, which actually leads to some more workarounds throughout the code. I mentioned some ideas directly in the code, but I don't think that those are the ultimate solutions. We might want to discuss this.

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
FrankEssenberger: > I think an even better (and easier) solution might be, that is exists, but it does not exist on the type, so that you get an error at design time

Can you explain a bit how this would work if we want to ship pregenerated clients, which contain this ERROR in typescript. Would this not lead to problems?

jjtang1985: I guess [this](https://www.typescriptlang.org/tsconfig#stripInternal) is the answer.

jjtang1985: I tested, with this config (annotation in the `.ts` file and enable it in the tsconfig), it can generate a function import, where
- the `.d.ts` does not contain the `execute` function.
- the `.js` contains the `execute` function.

FrankEssenberger: Ah, this could be a good options to remember, but than people would still not know why the execute is not there? We can do it like this, but if we get questions like `where is my function import` we should reconsider.

FrankEssenberger:
I will take one more look after @marikaner comments are settled. I think the `@internal` annotation is a nice feature to remember.

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
jjtang1985: btw, there is also a warning about this option
> This is an internal compiler option; use at your own risk

marikaner: Compiler option? Huh? 🤔

marikaner: Still the question remains, should we throw the error before executing?

packages/core/test/test-util/test-services/v4/test-service/function-imports.ts
jjtang1985: As discussed, this will be the chosen API design.

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -89,15 +89,15 @@ function findEdmType(returnType: string): string | undefined {
 function findEntityType(
   returnType: string,
   entities: VdmEntity[]
-): VdmEntity | undefined {
+): VdmEntity[] {
```
marikaner, line 92: The name of this function should probably also be adjusted (findEntityTypes).

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -150,15 +150,26 @@ function getEdmReturnType(
...
 ): VdmFunctionImportReturnType {
-  return {
-    returnTypeCategory: VdmReturnTypeCategory.ENTITY,
-    returnType: entity.className,
-    builderFunction: entity.className,
-    isMulti: isCollection,
-    isCollection
-  };
+  if (!entities.length) {
+    throw Error('Cannot get entity return type from an empty collection.');
```
marikaner, line 156: "collection" might be misleading here. Consider something in this direction: `Could not get entity return type for function import. No matching entity types found.`

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -150,15 +150,26 @@ function getEdmReturnType(
...
+        builderFunction: first(entities)!.className,
+        isMulti: isCollection,
+        isCollection
+      }
+    : {
+        returnTypeCategory: VdmReturnTypeCategory.ENTITY_NOT_DESERIALIZABLE,
+        returnType: 'void',
+        isMulti: isCollection,
+        isCollection
+      };
```
marikaner, line 172: I am still not a fan of `VdmReturnTypeCategory.ENTITY_NOT_DESERIALIZABLE` because strictly speaking this is not a return type category.

I am not sure that this would be an improvement, but have you considered the following options and could they make things simpler down the line?

Option 1:
Could we just return undefined here, and use this information down the line?

Option 2:
Not sure if this would even work, but could we use something like this:
```ts
{
  returnTypeCategory: VdmReturnTypeCategory.NEVER,
  returnType: 'never',
  isMulti: false,
  isCollection: false
 };
```

packages/generator/src/action-function-import/request-builder-arguments.ts
```diff
@@ -9,13 +10,32 @@ export function getRequestBuilderArgumentsBase(
...
+  actionFunctionImport: VdmFunctionImport | VdmActionImport,
+  service: VdmServiceMetadata
+): string {
+  if (
+    actionFunctionImport.returnType.returnTypeCategory ===
+    VdmReturnTypeCategory.ENTITY_NOT_DESERIALIZABLE
+  ) {
+    return `(data) => throwErrorWhenReturnTypeIsUnionType(data, '${actionFunctionImport.originalName}')`;
+  }
+  if (actionFunctionImport.returnType.builderFunction) {
```
marikaner, line 32: This check is only here because we had to make the `builderFunction` optional. The error before can actually never happen, right?

If we would make the `actionFunctionImport.returnType` optional we might be able to avoid this. What do you think?

packages/generator/src/action-function-import/request-builder-arguments.ts
```diff
@@ -9,13 +10,32 @@ export function getRequestBuilderArgumentsBase(
...
+
+function getTransformer(
+  actionFunctionImport: VdmFunctionImport | VdmActionImport,
+  service: VdmServiceMetadata
+): string {
+  if (
+    actionFunctionImport.returnType.returnTypeCategory ===
+    VdmReturnTypeCategory.ENTITY_NOT_DESERIALIZABLE
+  ) {
+    return `(data) => throwErrorWhenReturnTypeIsUnionType(data, '${actionFunctionImport.originalName}')`;
```
marikaner, line 30: Minor: Maybe it would be a good idea to not go through the core to call `throwErrorWhenReturnTypeIsUnionType` but just throw an error here directly. I don't have any good arguments for this in mind, though.

marikaner (changes requested):
I have some questions, but I think this looks quite good. I am still not a fan of `VdmReturnTypeCategory.ENTITY_NOT_DESERIALIZABLE` - I gave some context and ideas in the comments. But I can also see that it might not be possible to have an "easy" solution here, so let's just discuss and make a decision.

test-resources/odata-service-specs/v4/API_TEST_SRV/API_TEST_SRV.edmx
```diff
@@ -224,6 +247,10 @@
           <NavigationPropertyBinding Path="to_OtherMultiLink" Target="A_TestEntityMultiLink"/>
           <NavigationPropertyBinding Path="to_SingleLink" Target="A_TestEntitySingleLink"/>
         </EntitySet>
+        <EntitySet Name="A_TestEntitySharesEntityType1" EntityType="API_TEST_SRV.A_TestEntityTypeToBeShared" sap:content-version="1">
```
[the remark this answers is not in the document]
jjtang1985: `A_TestEntityWithSharedEntityType1`

packages/generator/src/vdm-types.ts
```diff
@@ -117,21 +117,27 @@ export type VdmActionFunctionImportReturnType =
...
-  builderFunction: string;
+  builderFunction?: string;
   returnType: string;
   /**
    * @deprecated Since v1.22.0. Use `isCollection` instead.
    */
   isMulti?: boolean;
   isCollection: boolean;
   returnTypeCategory: VdmReturnTypeCategory;
+  unsupportedFunction?: VdmUnsupportedFunction;
```
marikaner, line 128: The name of the property, as well as the name of the enum imply that this represents a function or a type of function. You called it "reason" yesterday. How about `neverReason` (or `unsupportedReason`)?

marikaner (approved):
I left one comment. Other than that LGTM.

PR #979, opened 2021-02-07
feat: support using tags for splitting OpenAPIs by default

jjtang1985:
As default behaviour, the OpenAPI generator should split all the APIs by using `tags` information. The number of generated api files should equal to the number of tags.

~As always, I manually adjusted the generated clients. Once approved, I'll adjust the generator.~


test-resources/openapi-service-specs/test-service.json
```diff
@@ -11,15 +11,23 @@
   },
   "tags": [
     {
-      "name": "someTag",
-      "description": "Some test tag"
+      "name": "entity",
```
jjtang1985, line 14: test service spec adjustment

test-packages/test-services/openapi/test-service/default-api.ts
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved.
+ *
+ * This is a generated file powered by the SAP Cloud SDK for JavaScript.
+ */
+import { OpenApiRequestBuilder } from '@sap-cloud-sdk/core';
+import { DefaultApi } from './openapi';
+
+export const TestServiceDefaultApi = {
```
jjtang1985, line 9: default api

test-packages/test-services/openapi/test-service/entity-api.ts
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved.
+ *
+ * This is a generated file powered by the SAP Cloud SDK for JavaScript.
+ */
+import { OpenApiRequestBuilder } from '@sap-cloud-sdk/core';
+import { EntityApi } from './openapi';
+import { TestEntity } from './openapi/model';
+
+export const TestServiceEntityApi = {
```
jjtang1985, line 10: entity api

test-packages/test-services/openapi/test-service/test-case-api.ts
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved.
+ *
+ * This is a generated file powered by the SAP Cloud SDK for JavaScript.
+ */
+import { OpenApiRequestBuilder } from '@sap-cloud-sdk/core';
+import { TestCaseApi } from './openapi';
+import { SimpleTestEntity } from './openapi/model';
+
+export const TestServiceTestCaseApi = {
```
jjtang1985, line 10: test case api

jjtang1985 (commented):
Please review the clients that are adjusted manually.

jjtang1985:
This PR is ready to be reviewed.

packages/openapi-generator/src/parser/document.ts
```diff
@@ -41,13 +42,27 @@ export function parseAllOperations(
...
+/**
+ * Collect all the tags used by given operations.
+ * @param operations The given operations.
+ * @returns An array that holds the unique tags.
+ */
+export function collectTags(operations: OpenApiOperation[]): string[] {
+  const tags: string[] = [];
+  operations.forEach(o => {
+    const tagsOfOperation = o.tags?.length ? o.tags : ['default'];
+    tagsOfOperation.forEach(tag => tags.push(tag));
```
FrankEssenberger, line 65: you could use:     `['c'].push(...['a','b'])` and I guess the most JS like way would be to use the `reduce` method to collect all tags in one list @marikaner what do you say as the JS native:  :-).

packages/openapi-generator/src/generator.ts
```diff
@@ -101,6 +102,26 @@ async function generateSDKSources(
...
+  serviceDir: string,
+  openApiDocument: OpenApiDocument
+) {
+  return openApiDocument.tags.map(tag => {
+    const matchedOperations = openApiDocument.operations.filter(o =>
+      o.tags?.includes(tag)
+    );
+    const fileName = `${voca.kebabCase(tag + 'Api')}.ts`;
+
+    createFile(
```
FrankEssenberger, line 115: This could be dangerous in case I await the function I get the list of files, but the actial creation is not awaited. Also add the return type to the method signature.

packages/openapi-generator/src/wrapper-files/index-file.ts
```diff
@@ -1,13 +1,21 @@
...
+ * @param apis The generated api files.
  * @returns The index file contents.
  */
-export function indexFile(): string {
+export function indexFile(apis: string[]): string {
   return codeBlock`
 export * from './openapi/model';
-export * from './api';
+${apis
+  .filter(api => api.endsWith('.ts'))
```
FrankEssenberger, line 13: can it be that non ts files are in the list? From the usage in generator.ts it does not look like it.

packages/openapi-generator/src/wrapper-files/index-file.ts
```diff
@@ -1,13 +1,21 @@
...
+  .map(api => exportAll(first(api.split('.ts'))!))
```
FrankEssenberger, line 14: The first(api.split('.ts')) is to remove the `.ts` on the filename right? This seems a bit hacky. I would make the exportAll method a bit more intelligent and allow as input `file` and `file.ts` and do a `replace(/$\.ts/,"") on it so that it is clear what is done.

=> question what would happen if someone would have a tag `ts.LaLA`. Is this not allowed? In general we should check that our file generation also works for that odd value.

FrankEssenberger (changes requested):
Looks good. See my minor comments. I have one general question. I did not see the change in the generator which changed the behavior from `all in default one api file` to seperate files per tag and only the ones without to default. Or was this changed already in an other PR?

packages/openapi-generator/src/generator.ts
jjtang1985: I'll refactor it

packages/openapi-generator/src/wrapper-files/index-file.ts
jjtang1985: I'll refactor it

packages/openapi-generator/src/wrapper-files/index-file.ts
jjtang1985: I'll create a test case for this.

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -2,46 +2,48 @@ import {
...
-  OpenApiOperation,
-  SchemaMetadata
-} from '../openapi-types';
+import { OpenApiOperation, SchemaMetadata } from '../openapi-types';
 const logger = createLogger('openapi-generator');
 /**
  * @experimental This API is experimental and might change in newer versions. Use with caution.
- * Get the content of the SAP Cloud SDK API wrapper.
- * @param openApiDocument Parsed service.
+ * Get the content of the SAP Cloud SDK API wrapper for a specific Api.
```
marikaner, line 12:
```suggestion
 * Get the file contents for an API wrapper file.
```

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -2,46 +2,48 @@ import {
...
+ * @param serviceName The service name.
+ * @param apiName The name of the Api.
```
marikaner, line 14:
```suggestion
 * @param apiName The name of the API.
```

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -2,46 +2,48 @@ import {
...
-  SchemaMetadata
-} from '../openapi-types';
+import { OpenApiOperation, SchemaMetadata } from '../openapi-types';
 const logger = createLogger('openapi-generator');
 /**
  * @experimental This API is experimental and might change in newer versions. Use with caution.
- * Get the content of the SAP Cloud SDK API wrapper.
- * @param openApiDocument Parsed service.
+ * Get the content of the SAP Cloud SDK API wrapper for a specific Api.
+ * @param serviceName The service name.
```
marikaner, line 13:
```suggestion
 * @param serviceName The name of the service.
```

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -2,46 +2,48 @@ import {
...
+ * @param apiName The name of the Api.
+ * @param operations All operations belong to this Api.
```
marikaner, line 15:
```suggestion
 * @param operations All operations, that belong to the given API.
```

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -51,22 +53,30 @@ function getRequestBodyReferenceTypes(
...
+ * @param operations The given operation list.
+ * @param apiNamePascal The Api that holds the operations.
  * @returns All operations as a string.
  */
-function getOperations(openApiDocument: OpenApiDocument): string {
-  return openApiDocument.operations
-    .map(operation => getOperation(operation))
+function getOperations(
+  operations: OpenApiOperation[],
+  apiNamePascal: string
```
marikaner, line 63:
```suggestion
  apiName: string
```

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -51,22 +53,30 @@ function getRequestBodyReferenceTypes(
...
 /**
  * Get the string representation of one operation.
  * @param operation Operation to serialize.
+ * @param apiNamePascal The Api that holds the operations.
  * @returns The operation as a string.
  */
-function getOperation(operation: OpenApiOperation): string {
+function getOperation(
+  operation: OpenApiOperation,
+  apiNamePascal: string
```
marikaner, line 78:
```suggestion
  apiName: string
```

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -51,22 +53,30 @@ function getRequestBodyReferenceTypes(
 }
 
 /**
- * Get all operation representations for the given service.
- * @param openApiDocument Parsed service.
+ * Get all operation representations for the given operations that belong to a specific Api.
+ * @param operations The given operation list.
+ * @param apiNamePascal The Api that holds the operations.
```
marikaner, line 58:
```suggestion
 * Get all operation representations for the given operations that belong to a specific API.
 * @param operations All operations that belong to the given API.
 * @param apiName The name of the API.
```

packages/openapi-generator/src/generator.ts
```diff
@@ -101,6 +102,26 @@ async function generateSDKSources(
   }
 }
 
+async function createApis(
+  serviceDir: string,
+  openApiDocument: OpenApiDocument
+) {
```
marikaner, line 108: Add a return type

packages/openapi-generator/src/document-converter.ts
```diff
@@ -15,15 +15,15 @@ const { readFile } = promises;
...
+ * Documents are expected to be formatted as JSON and compliant with OpenAPI version 3.
  * @param filePath File content of the original spec.
  */
 export async function convertOpenApiSpec(
   filePath: string
 ): Promise<OpenAPIV3.Document> {
   const file = await parseFileAsJson(filePath);
   const openApiDocument = await convertDocToOpenApiV3(file);
-  return convertDocToGlobalTag(convertDocToUniqueOperationIds(openApiDocument));
+  return convertDocToUniqueOperationIds(openApiDocument);
```
marikaner, line 26: Luckily you didn't delete the `convertDocToGlobalTag` method yet. I think it would be a good idea, to add the 'default' tag, where there are no tags or only an empty list of tags.

packages/openapi-generator/src/generator.ts
```diff
@@ -101,6 +102,26 @@ async function generateSDKSources(
...
+  return openApiDocument.tags.map(tag => {
+    const matchedOperations = openApiDocument.operations.filter(o =>
+      o.tags?.includes(tag)
+    );
```
marikaner, line 112: What happens if the operation has no tags? Please see my comment in document-converter regarding this.

marikaner, line 112: Please, don't call it `o` ;)
```suggestion
    const matchedOperations = openApiDocument.operations.filter(operation =>
      operation.tags?.includes(tag)
    );
```

marikaner: Consider adjusting the operation type to `tags: string[]` instead of `tags?: string[]` based on the comment above.

marikaner (changes requested):
Looks mostly good already. Mostly minor things.

packages/openapi-generator/src/generator.ts
jjtang1985: @marikaner , the default tag is added when tags are missing.
However, I still use the original openapi type, which contains `tags?: string[]`, but I alway use `!` to tell "I'm sure now `tags` is there."
I know it's confusing since `tags` is not optional anymore, but if you change the type, you have to differentiate between the old one (before convert) and new one (after convert) and really create a new type.
Since this is the first time we see the use case of the new type, if this can be avoided, I would still keep using the original type.
What's your feeling?

jjtang1985:
I adjusted it based on the comments.
There is still a open topic regarding a potential new type: https://github.com/SAP/cloud-sdk-js/pull/979#discussion_r575125662

packages/openapi-generator/src/wrapper-files/index-file.ts
FrankEssenberger: Also cosider my comment about the `split`. A replace is closer to the thing really done.

jjtang1985: The split is not needed any more, I removed it.

packages/openapi-generator/src/generator.ts
jjtang1985: discussed

FrankEssenberger:
@junjie could you comment on my question regarding the where is the generator changed so that we do not have a single default tag. Was this already done before? Or did we manualy introduce the tag but than this must have been removed somewhere. I just could not see it.

packages/openapi-generator/src/document-converter.ts
```diff
@@ -69,24 +71,60 @@ export async function convertDocToOpenApiV3(
...
 
 /**
- * Workaround for OpenAPI generation to build one and only one API for all tags.
- * Modify spec to contain only one 'default' tag.
+ * Modify spec to contain the 'default' tag when no tags are defined.
  * @param openApiDocument OpenAPI JSON document.
  * @returns The modified document.
  */
-export function convertDocToGlobalTag(
+export function convertDocWithDefaultTag(
```
jjtang1985, line 78: @FrankEssenberger , regarding the change for the global tag handling.

FrankEssenberger: ah there it is thanks.

FrankEssenberger (approved):

PR #985, opened 2021-02-08
chore: Move transpilation to generators

marikaner:
Having `execa` in utils breaks the frontend e2e tests. I was able to fix that to a certain degree, but this will also break for users who use the SDK in the frontend. In order to avoid that, I thought it might be nice, to for now just duplicate this one function and later on put this into a separate library as discussed.

Let me know if this makes sense for you.


jjtang1985 (approved):
Just fix the lint and go ahead.

FrankEssenberger:
As discussed with @jjtang1985 we will introduce later a new package to handle the shared generator methods.

PR #988, opened 2021-02-09
fix: Improve swagger path generation to actually check for file paths

marikaner:
This is a further improvement, because the old version actually fails in some cases (e. g. if swagger enhancement is enabled and there is a .xml, but no .json file, we currently return the original .xml filepath. Later on JSON.parse generation will fail).

### DIFF
--- a/packages/generator/src/input-path-provider.ts
+++ b/packages/generator/src/input-path-provider.ts
@@ -1,5 +1,5 @@
-import { lstatSync, PathLike, readdirSync } from 'fs';
-import { join, extname } from 'path';
+import { lstatSync, PathLike, readdirSync, existsSync } from 'fs';
+import { join, extname, parse } from 'path';
 
 const validFileExtensions = ['.edmx', '.xml'];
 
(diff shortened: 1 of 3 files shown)


packages/generator/src/input-path-provider.ts
```diff
@@ -30,9 +30,12 @@ export function inputPaths(
   });
 }
 
-export function swaggerPathForEdmx(edmxPath: PathLike): PathLike {
-  const extension = new RegExp(`${extname(edmxPath.toString())}\$`);
-  return edmxPath.toString().replace(extension, '.json');
+export function swaggerPathForEdmx(edmxPath: PathLike): PathLike | undefined {
```
jjtang1985, line 33: So, this means, only use the json when it exists, which I checked;)

jjtang1985 (approved):
GOGOGO

packages/generator/src/input-path-provider.ts
marikaner: Correct! + we also check for .JSON ;)

PR #1001, opened 2021-02-12
chore: Reduce the usage of Mac OS in the tests

jjtang1985:
It happened 5+ times (on Feb.12th) that the GH Actions cost 20+ minute, because of the waiting time of the Mac OS containers.
I guess with more people working on it, it's harder to get the containers.

In this PR, I would like to minimise the usage of the Mac OS container and only keep using one.

I disabled the combination of Mac OS and node 10, since we still have two cases below:
- Mac OS and node 12
- ubuntu and node 10 

Let me know your opinion.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,9 @@ jobs:
       matrix:
         os: [macos-latest, ubuntu-latest]
         node-version: [10.x, 12.x]
+        exclude:
+          - os: macos-latest
+            node: 10.x
     steps:
       - uses: actions/checkout@v1
       - name: Get yarn cache directory path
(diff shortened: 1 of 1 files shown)


FrankEssenberger:
For me it is fine to just use one mac container. In general I would prefer to have also one windows container in the mix, because mac should be more like unix so, but windows is real different.

jjtang1985:
> In general I would prefer to have also one windows container in the mix, because mac should be more like unix so, but windows is real different.

For windows, we have a separate ticket, because we have to fix something before enabling. Also, windows is slower than Mac OS and ubuntu, which needs to be evaluated again.

FrankEssenberger (approved):
LGTM

marikaner (approved):
Agreed.

PR #1014, opened 2021-02-17
chore: Add tests for stack logging

FrankEssenberger:
We have tests for the exception logger. Also if you call `logger.error(Error)` it will log the stack now. Documentation will come.


florian-richter (approved):
Looks good to me.

Issue #931, opened 2021-01-21
Cannot create entity due to "`entity.getCustomFields` is not a function


marikaner:
**Describe the bug**
When creating a custom client and executing create requests the error above occurs, while this has worked before.

marikaner:
Requested previous and current version from user.

blaart:
hi I have the same problem (for me first time using this). I use v1.37.0.
I get this on this call on the .create:
await OrderSet
      .requestBuilder()
      .create(orderSet)
      .execute(..)

marikaner:
Thanks @blaart, we will investigate this and try to solve it as soon as possible. Does your `orderSet` instance contain navigation properties?

blaart:
Thanks, @marikaner, the orderSet has several navigation properties (some in an array)

blaart:
Same problem with .update by the way

ja-ne:
Hi, I can reproduce this for versions >1.34.0 as well. Is there a workaround other than leaving out navigation properties (I've only tested with array navigation properties)?

Thanks in advance.

marikaner:
Hey @blaart and @ja-ne,

I am on it, but I am currently unable to reproduce this problem, therefore I am also not able to give you any good advice for a workaround (maybe stay with a version < 1.34).

Could you please provide a small example on how you are building the entity that you want to create?

marikaner:
Ok, I just confirmed a suspicion.

It seems that you are not building the entities as intended, with the entity builder:

```ts
const navProperty = SomeNavProperty().builder().someProperty('navProperty').build();
const myEntity = MyEntity.builder().someNavProperty(navProperty).build();
```

Instead are trying to use an object:
```ts
const myEntity = {
  someNavProperty: {
    someProperty: 'navProperty'
  }
};
```

Is this possible? If yes, can you give me a bit more context why you are doing it this way? It seems that this is not a single case, therefore I would be interested, whether we could / should improve our usability here.

ja-ne:
Hi @marikaner,
thanks for the quick response.

You are right, I do not use the entity builder as you described but the `.fromJson` method of the parent entity.

For example:
```
//entityJson comes from somewhere else
const entityJson = {
  firstProperty: 'firstValue',
  toNavigationProperty: [{
      "navigationProperty": 'value'
    }]
};
const entity: MyEntity= MyEntity.builder().fromJson(entityJson);
```

[...]

PR #1018, opened 2021-02-18
fix: Fix entityBuilder.fromJson() to accept actual JSON

marikaner:
Related to #931. Transforms JSON to an entity without assuming instances of entity for navigation properties.


packages/core/src/odata-common/entity-builder.ts
```diff
@@ -49,24 +63,44 @@ export class EntityBuilder<EntityT extends Entity, JsonT> {
...
+      ([key]) => typeof entityBuilder[key] === 'function'
+    );
+
+    entityEntries.forEach(([key, value]) => {
+      const propertyValue = isNavigationProperty(key, entityConstructor)
+        ? buildNavigationPropertyFromJson(key, value, entityConstructor)
+        : value;
+
+      entityBuilder[key](propertyValue);
+    }, entityBuilder);
```
jjtang1985, line 81: question:
Can I remove the second parameter `entityBuilder` of the `forEach` in line 81?

packages/core/src/odata-common/entity-builder.ts
```diff
@@ -82,3 +116,26 @@ export class EntityBuilder<EntityT extends Entity, JsonT> {
 function toClassName(entityName: string) {
   return entityName.substr(entityName.indexOf('_') + 1);
 }
+
+function buildNavigationPropertyFromJson<EntityT extends Entity>(
+  key: string,
+  value: FromJsonType<unknown>,
+  entityConstructor: Constructable<EntityT>
+) {
```
jjtang1985, line 124: minor: add a return type

packages/core/src/odata-common/entity-builder.ts
```diff
@@ -49,24 +63,44 @@ export class EntityBuilder<EntityT extends Entity, JsonT> {
...
-      if (typeof entityBuilder[key] === 'function') {
-        entityBuilder[key](value);
-      }
-      if (key === '_customFields') {
-        entityBuilder.withCustomFields(value as Record<string, any>);
-      }
-    });
+    const entityConstructor = this._entityConstructor;
+
+    const [entityEntries, customEntries] = partition(
```
jjtang1985, line 70: Instead of `Partial `, `FromJsonType` is used, which is more complicated but should be more precise.
I see the type of `entityEntries` has been changed from `[string, unkonwn][]` to `[string, any][]`.
Could you please share more insights about the benefit of the new type?

packages/core/src/odata-common/entity-builder.ts
```diff
@@ -1,12 +1,28 @@
-import { createLogger } from '@sap-cloud-sdk/util';
+import {
```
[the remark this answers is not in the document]
marikaner: Done - I was waiting for the release ;)

packages/core/src/odata-common/entity-builder.ts
```diff
@@ -1,12 +1,28 @@
...
+  [key: string]: any;
+} & {
+  [P in keyof T]?: T[P] extends (infer U)[]
+    ? U extends Record<string, any>
+      ? FromJsonType<U>[]
+      : T[P]
+    : T[P] extends Record<string, any>
+    ? FromJsonType<T[P]>
+    : T[P];
+};
```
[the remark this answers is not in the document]
marikaner: Are you able to fix that in lint? ;)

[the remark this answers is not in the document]
marikaner: duplicate

packages/core/src/odata-common/entity-builder.spec.ts
```diff
@@ -38,40 +40,145 @@ describe('EntityBuilder', () => {
...
-    entityJson['_customFields'] = { additionalField: 'someAdditionalValue' };
-    const actual = builder.fromJson(entityJson);
-    expect(actual.stringProperty).toBe(entityJson.stringProperty);
-    expect(actual.getCustomFields()).toEqual(entityJson['_customFields']);
-    expect(actual.toSingleLink.stringProperty).toBe(
-      linkedEntity.stringProperty
-    );
-    expect(actual.toSingleLink.getCustomFields()).toEqual(
-      linkedEntity.getCustomFields()
-    );
```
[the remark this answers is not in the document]
marikaner: Yes.

packages/core/src/odata-common/entity-builder.ts
marikaner: The `Partial` type was just completely wrong ;)
1. Not recursive:
```ts
const bupaJson = {
  firstName: 'Peter',
  to_Adress: [
    city: 'Berlin'
  ]
}
```

If you try to assign it to a variable with `Partial<BusinessPartner>` this will not work, because `Partial` is not recursive and you have to specify every property of the address, that is not optional.

2. Did not allow `_customFields`
There was an implementation to account for `_customFields`, but according to the type, you weren't able to set those.
I deprecated this behavior and instead allowed to set unknown properties directly.

packages/core/src/odata-common/entity-builder.ts
marikaner, line 81:
```suggestion
    });
```

packages/core/src/odata-common/entity-builder.ts
marikaner: Done

packages/core/src/odata-common/entity-builder.ts
```diff
@@ -1,12 +1,28 @@
...
 import type { Constructable, Entity } from './entity';
-import { toStaticPropertyFormat } from './name-converter';
+import { isNavigationProperty } from './properties-util';
 
 const logger = createLogger({
   package: 'core',
   messageContext: 'entity-builder'
 });
 
+type FromJsonType<T> = {
```
FrankEssenberger, line 14: could you add some explaining text which this type does. My conditional type magic is not so powerful to grasp this quicly.

packages/core/src/odata-common/entity-builder.spec.ts
```diff
@@ -38,40 +40,145 @@ describe('EntityBuilder', () => {
...
     const actual = TestEntity.builder()
-      .dateTimeProperty(expected.dateTimeProperty)
+      .dateProperty(expected.dateProperty)
       .decimalProperty(expected.decimalProperty)
       .build();
 
     expect(expected).toEqual(actual);
   });
+
+  describe('fromJson', () => {
```
FrankEssenberger, line 57: The `fromJson` method is called recursivly via the `buildNavigationPropertyFromJson` could you also add a test for this case in other words at least two levels of nested json properties.

packages/core/src/odata-common/entity-builder.ts
```diff
@@ -82,3 +116,30 @@ export class EntityBuilder<EntityT extends Entity, JsonT> {
...
+        buildSingleEntityFromJson(item, linkedEntityConstructor)
+      )
+    : buildSingleEntityFromJson(value, linkedEntityConstructor);
+}
+
+function buildSingleEntityFromJson<LinkedEntityT extends Entity>(
+  json: FromJsonType<unknown>,
+  linkedEntityConstructor: Constructable<LinkedEntityT>
+): LinkedEntityT {
+  return json instanceof linkedEntityConstructor
```
FrankEssenberger, line 142: Is this correct? From a quick look I would guess that if it is a `linkedEntityConstructor ` you can not return it directly but need to call a builder?

packages/core/src/odata-common/entity-builder.spec.ts
```diff
@@ -38,40 +40,145 @@ describe('EntityBuilder', () => {
...
+    it('should build an entity from json', () => {
```
FrankEssenberger, line 58: Could you perhaps also add one E2E test with the custom fields for the `GET` and `CREATE/UPDATE` case so that we see, that serialization/desrialization works end to end.

FrankEssenberger (changes requested):
Only minor comment and question.

packages/core/src/odata-common/entity-builder.ts
marikaner: Done.

packages/core/src/odata-common/entity-builder.spec.ts
marikaner: What case would be covered, that currently isn't?
Exactly because the function is called recursively, it is called again and therefore if it works on the first level, it works on all levels. Maybe I am missing something?

marikaner: Will add as discussed

packages/core/src/odata-common/entity-builder.spec.ts
marikaner: Not necessary, as serializers are not called for custom fields.

packages/core/src/odata-common/entity-builder.spec.ts
marikaner: Done

packages/core/src/odata-common/entity-builder.ts
marikaner: It is a little awkward, but we are looking at `instanceof` and the constructor is basically the class.

packages/core/src/odata-common/entity-builder.ts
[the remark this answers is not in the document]
jjtang1985:
```
/* eslint-disable */
/* eslint-enable */
```

jjtang1985 (approved):
LGTM. There is a minor comment about lint.

FrankEssenberger (approved):
LGTM

jjtang1985:
@marikaner would you like to fix the merge conflict and merge?

PR #1038, opened 2021-02-23
feat: support custom api names and operation names

jjtang1985:
The following OpenAPI extensions are supported to define api names and operation names:
- `x-sap-cloud-sdk-api-name`
- `x-sap-cloud-sdk-operation-name`

#### api name logic
`x-sap-cloud-sdk-api-name` exists?
- Y. Use it.
- N. `tags` exists?
    - Y. Use `tags`.
    - N. Use `defaultApi`

#### operation name logic
`x-sap-cloud-sdk-operation-name` exists?
- Y. Use it.
- N. `operationId` exists?
    - Y. Use `operationId`.
    - N. Generate operation name by using `path` and `method`.


packages/openapi-generator/src/openapi-types.ts
```diff
@@ -24,6 +24,7 @@ export interface OpenApiOperation extends OpenAPIV3.OperationObject {
   path: string;
   requestBody?: OpenApiRequestBody;
   parameters: OpenApiParameter[];
+  'x-sap-cloud-sdk-operation-name'?: string;
```
marikaner, line 27: I think this should be done in `convertDocToUniqueOperationIds`, where we already parse the operationIds. That way you wouldn't have to touch any code down the line.

packages/openapi-generator/src/document-converter.ts
```diff
@@ -75,47 +75,32 @@ export async function convertDocToOpenApiV3(
...
-export function convertDocWithDefaultTag(
+export function convertDocWithApiNameTag(
   openApiDocument: OpenAPIV3.Document
 ): OpenAPIV3.Document {
-  const tag = 'default';
+  const defaultTag = 'default';
 
-  executeForAllOperationObjects(openApiDocument, operation => {
-    operation.tags = addGlobalTagToOperationWhenNoTagsAreUsed(
+  executeForAllOperationObjects(openApiDocument, (operation,path, method, extensionApiName) => {
```
marikaner, line 83: @florian-richter Why isn't the linter complaining about this? Is there something broken? Could you check this? (There are more issues in this file)
```suggestion
  executeForAllOperationObjects(openApiDocument, (operation, path, method, extensionApiName) => {
```

packages/openapi-generator/src/document-converter.ts
```diff
@@ -75,47 +75,32 @@ export async function convertDocToOpenApiV3(
...
+    operation.tags = extensionApiName? [extensionApiName]:addGlobalTagToOperationWhenNoTagsAreUsed(
```
marikaner, line 84: The way the tags are created here, seems slightly more complex than necessary. `addGlobalTagToOperationWhenNoTagsAreUsed` is only used here and would be easier to read if combined with the code here. Consider adding something like `getTags(operation.tags, extensionApiName)`.

packages/openapi-generator/src/document-converter.ts
marikaner, line 83: With four arguments I think it makes sense to transform this to an object - that way you won't have to pass unused parameters as in this case.

packages/openapi-generator/src/wrapper-files/api-file.ts
```diff
@@ -85,7 +85,7 @@ function getOperation(operation: OpenApiOperation, apiName: string): string {
 
   return codeBlock`
 ${
-  operation.operationId
+  operation['x-sap-cloud-sdk-operation-name'] || operation.operationId
```
marikaner, line 88: This should not be necessary.

marikaner (changes requested):
I would like to propose some general changes before looking at details, as you will find in the comments ;)

FrankEssenberger:
I will get involved once @marikaner changes are included and discussed and we have already some convergence.

packages/openapi-generator/src/openapi-types.ts
jjtang1985: I guess this is the major "general change".
Initially, I also think the `operation-name`/`operationId` should be treated like the `api-name`/`tags`, where we overwrite `tags` by using `api-name`.
However, from the openapi perspective, `operationId` should be unique, while there are no constraints about `tags`.
This means the example below will lead to a new spec, which is invalid, because `op1` will be used by 2 `operationId`s.
```
- api-name: apiA
  - operation-name: op1
  - operation-name: op2
- api-name: apiB
  - operation-name: op1
```
Now we disabled the validation of the openapi generator, which might not be supported in the future.
Only if when we get rid of the openapi generator and use our own solution, then we are safe.

jjtang1985: As discussed, we use `operationId` for now for simplicity. We'll face this when the validation is enabled.

packages/openapi-generator/src/document-converter.ts
```diff
@@ -75,54 +75,39 @@ export async function convertDocToOpenApiV3(
...
-function addGlobalTagToRootTags(
-  tags: OpenAPIV3.TagObject[] | undefined,
-  globalTag: string
-): OpenAPIV3.TagObject[] {
-  return tags?.find(tag => tag.name === globalTag)
-    ? tags
-    : tags?.length
-    ? tags.concat({ name: globalTag })
-    : [{ name: globalTag }];
+function collectTag(
```
marikaner, line 92:
```suggestion
function collectTags(
```

marikaner (approved):
LGTM, just one minor remark ;)

PR #1041, opened 2021-02-24
chore: commit openapi jar

jjtang1985:
The jar file needs to be downloaded on the fly when necessary. The URL (maven central) for downloading the jar file is not configurable. This is the solution when downloading from external target URL is not allowed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,9 @@ packages/**/*.jar
 *.jar
 local-docs/_site
 
+# openapi generator
+!**/5.0.0.jar
+
 sqlite.db
 sqlite.db-journal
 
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -28,6 +28,9 @@ packages/**/*.jar
 *.jar
 local-docs/_site
 
+# openapi generator
+!**/5.0.0.jar
```
marikaner, line 32: This is here because of the jetty mock server, right? Do we still have this mock server? If not we should remove the gitignore entries.

marikaner (approved):
Looks good to me, but please consider my comment about the gitignore ;)

PR #1042, opened 2021-02-24
yarn2 migration

tomfrenken:
With this PR we migrate from yarn to yarn2 (aka. yarn berry)

Changes are:
- workspaces no longer have shared dependencies, therefore we need to use scripts in the format of 
`"foo:bar": "cd $INIT_CWD && bar"` to access binaries we want to share (in this case. `bar`), `bar` in this case needs to be added to the devDependencies in the root `package.json`.
- the audit-ci tool is being replaced with `yarn npm audit` since it doesn't support yarn2
- `npx` commands are being replaced with `yarn` commands as they lead to unexpected/unsupported behaviour in yarn(2)
- the cli now uses the ./dist folder instead of the ./src folder (previously used the ./lib folder and was then fixed with ./src, but should be ./dist instead)
- the `yarn version` command is fixed and now uses the .ts file instead of the .js file (probably already fixed)
- arbitrary pre/post scripts are no longer supported, if they are not part of the yarn2 [lifecycle](https://yarnpkg.com/advanced/lifecycle-scripts) they have to be executed together

All scripts have been tested (except nightly CLI tests, as they aren't working on `main` either), example of all tests passing:
https://github.com/tomfrenken/cloud-sdk-js/runs/1971924849
Note that the canary release is only failing because I have (fortunately) no access to the sdk package on npmjs


tomfrenken:
I need to update the yarn.lock first before I open the pull request again

PR #1043, opened 2021-02-24
Yarn2 migration

tomfrenken:
With this PR we migrate from yarn to yarn2 (aka. yarn berry)

Changes are:
- workspaces no longer have shared dependencies, therefore we need to use scripts in the format of 
`"foo:bar": "cd $INIT_CWD && bar"` to access binaries we want to share (in this case. `bar`), `bar` in this case needs to be added to the devDependencies in the root `package.json`.
- the audit-ci tool is being replaced with `yarn npm audit` since it doesn't support yarn2
- `npx` commands are being replaced with `yarn` commands as they lead to unexpected/unsupported behaviour in yarn(2)
- the cli now uses the ./dist folder instead of the ./src folder (previously used the ./lib folder and was then fixed with ./src, but should be ./dist instead)
- the `yarn version` command is fixed and now uses the .ts file instead of the .js file (probably already fixed)
- arbitrary pre/post scripts are no longer supported, if they are not part of the yarn2 [lifecycle](https://yarnpkg.com/advanced/lifecycle-scripts) they have to be executed together

All scripts have been tested (except nightly CLI tests, as they aren't working on `main` either), example of all tests passing:
https://github.com/tomfrenken/cloud-sdk-js/runs/1972197128
Note that the canary release is only failing because I have (fortunately) no access to the sdk package on npmjs


package.json
```diff
@@ -30,17 +30,17 @@
   ],
   "scripts": {
     "postinstall": "yarn compile && yarn run readme",
-    "test": "yarn workspaces run test",
+    "test": "yarn workspaces foreach -v run test",
```
FrankEssenberger, line 33: why verbose?

tomfrenken: Well it looks better in the CLI, no other reason than that

packages/cli/package.json
```diff
@@ -17,7 +17,7 @@
     "sap-cloud-sdk": "./bin/run"
   },
   "publishConfig": {
-    "access": "private"
+    "access": "restricted"
```
marikaner, line 20: Why? This should not be released yet and therefore stay private.

packages/cli/package.json
```diff
@@ -65,7 +65,7 @@
     "typescript": "^4.0.3"
   },
   "oclif": {
-    "commands": "./src/commands",
+    "commands": "./dist/commands",
```
marikaner, line 68: I guess this makes sense, but are you sure and was this on purpose?

packages/cli/src/utils/scaffold.ts
```diff
@@ -67,7 +67,7 @@ export async function buildScaffold(
   };
 
   await execa(
-    'npx',
+    'yarn',
```
marikaner, line 70: I think this won't work on the user side, if they don't have yarn. We should stay with `npx` here.

packages/generator/src/service/package-json.ts
```diff
@@ -29,8 +29,8 @@ export function packageJson(
           url: ''
         },
         scripts: {
-          compile: 'npx tsc',
-          doc: 'npx typedoc',
+          compile: 'yarn tsc',
+          doc: 'yarn typedoc',
```
marikaner, line 33: Here again, I think we should stay with npx - If a user generates a TS version of the client and tries to transpile it, whithout yarn it won't work.

package.json
```diff
@@ -63,7 +63,11 @@
     "e2e-tests": "yarn workspace @sap-cloud-sdk/e2e-tests",
     "nightly-tests": "yarn workspace @sap-cloud-sdk/nightly-tests",
     "integration-tests": "yarn workspace @sap-cloud-sdk/integration-tests",
-    "type-tests": "yarn workspace @sap-cloud-sdk/type-tests"
+    "type-tests": "yarn workspace @sap-cloud-sdk/type-tests",
+    "root:tsc": "cd $INIT_CWD && tsc",
+    "root:jest": "cd $INIT_CWD && jest",
+    "root:depcheck": "cd $INIT_CWD && depcheck",
+    "root:ts-node": "cd $INIT_CWD && ts-node"
```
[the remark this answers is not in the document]
marikaner: I slightly lean towards adding dependencies to all packages.

marikaner (changes requested):
We probably shouldn't use yarn for any user facing scripts. I found some minor issues, but I am especially interested in the discussion on the root dev scripts.

package.json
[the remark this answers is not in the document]
jjtang1985: notes:
> $INIT_CWD will always point to the directory running the script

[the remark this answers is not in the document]
jjtang1985: | |`cd $INIT_CWD && tsc`|add all root dependencies to sub-packages|
|-|-|-|
|dependency size|+|-|
|the structure of the shared dependencies|-|+|

For me, if the two major differenced listed above are the main concerns, I would vote for user experience (use the current workaround), but perhaps we need to investigate, if we add all the root dependencies to the sub-modules, how big are the additional dependencies. 1MB, for example, for the core, might not be good.

package.json
tomfrenken: As per documentation it `Prefixes each output line with the name of the originating workspace` but it also adds colors to differentiate the workspaces 🌈

package.json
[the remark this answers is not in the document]
marikaner: @jjtang1985 Please be aware that this applies only to dev dependencies, where only the binary is used. If I am not mistaken they will be hoisted to the root and linked to the packages when added as dev dependencies - @tomfrenken please correct me if not.

.gitignore
```diff
@@ -34,3 +34,10 @@ sqlite.db-journal
 test-packages/integration-tests/test/auth-flows/vcap-services.json
 test-packages/integration-tests/test/auth-flows/user-access-token.json
 test-packages/integration-tests/test/auth-flows/systems.json
+
+.yarn/*
+!.yarn/releases
+!.yarn/plugins
+!.yarn/sdks
+!.yarn/versions
+.pnp.*
```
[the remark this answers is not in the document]
tomfrenken: Ignoring the .pnp.* is the [recommended](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored) approach when not using the [zero installs](https://yarnpkg.com/features/zero-installs) mode of yarn, I assume we could also not ignore the `.pnp.*` but then we would automatically switch to `zero installs` when we try to use pnp

package.json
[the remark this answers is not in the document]
tomfrenken: Indeed this only applies to devDependencies, we don't share regular dependencies in the root, and yes those dependencies would be hoisted, so as long as the workspaces use the same versions the size should not change

packages/cli/README.md
```diff
@@ -49,7 +49,7 @@ EXAMPLE
   $ sap-cloud-sdk add-approuter
 ```
 
-_See code: [src/commands/add-approuter.ts](https://github.com/SAP/cloud-sdk-js/blob/v1.38.0/src/commands/add-approuter.ts)_
+_See code: [dist/commands/add-approuter.ts](https://github.com/SAP/cloud-sdk-js/blob/v1.38.0/dist/commands/add-approuter.ts)_
```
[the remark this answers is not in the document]
tomfrenken: Good catch, I think this happened while trying out the `yarn doc` 😅

packages/cli/package.json
tomfrenken: According to the 
Screenshot 2021-03-01 at 10 56 14

tomfrenken: I just saw this in my IDE and thought `private` doesn't work

tomfrenken: I think `restricted` is doing what `private` should be doing https://docs.npmjs.com/cli/v7/using-npm/config#config-settings

packages/cli/package.json
tomfrenken: Yes this was on purpose, earlier when I encountered the issue where it was still  `./lib/commands` I discussed it with @florian-richter and concluded that the path should be the path to the compiled folder not the source (`./disc` is the folder the `tsconfig` compiles to)

package.json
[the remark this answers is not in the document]
jjtang1985: @tomfrenken , does the committed jar file bring some conflicts with yarn2?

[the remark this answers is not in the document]
tomfrenken: Hmm I tested it last week and all the test were passing, but now I see that `yarn check:test-services` seems to fail, if I test it locally I receive alot of eslint errors, I'll look into it, but I'm not sure if this is caused by the `.jar` file 🤔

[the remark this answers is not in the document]
tomfrenken: Okay it works with the jar 👌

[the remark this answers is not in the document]
FrankEssenberger: I would go for the declare it in each package. However, I would like to have a check that on each package for the same dev dependency the same version is use. Is there a flag to enforce such a thing. Lerna had it :-).

packages/analytics/package.json
```diff
@@ -23,12 +23,12 @@
   ],
   "repository": "github:SAP/cloud-sdk-js",
   "scripts": {
-    "compile": "yarn tsc -b",
+    "compile": "yarn root:tsc -b",
```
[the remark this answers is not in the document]
FrankEssenberger: As mentioned above I like self contained package files but enforecement on versions.

test-packages/e2e-tests/package.json
```diff
@@ -12,9 +12,9 @@
     "start": "pm2 start server.js || pm2 restart server.js",
     "stop": "pm2 stop server.js",
     "deploy": "cds deploy --to sqlite",
-    "test": "yarn jest --runInBand",
-    "test:local": "yarn jest",
-    "check:dependencies": "depcheck . --ignores='express,sqlite3,@sap/cds'"
+    "test": "yarn pretest && yarn root:jest --runInBand && yarn posttest",
```
FrankEssenberger, line 15: why all this pretest and post is needed? Is this not executed autotically?

FrankEssenberger: If so I am afraid we missed it somewhere else.

package.json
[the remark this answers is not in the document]
tomfrenken: You can actually enforce it, but it's rather complex, they call it [constrains](https://yarnpkg.com/features/constraints) basically you write constrains in the language `Prolog` ...

FrankEssenberger (changes requested):
LGTM - we can merge once we decided on a devDep way of life. Only minor things besides that. Thanks for migrating @tomfrenken

test-packages/e2e-tests/package.json
tomfrenken: Only arbitrary pre-&post scripts are no longer running by default, [life-cycle scripts](https://yarnpkg.com/advanced/lifecycle-scripts) however still work - this was the only occurrence of a script that isn't part of the lifecycle

packages/cli/package.json
jjtang1985: I guess this is npm version 7 specific, but we are using yarn.
With yarn1, here is the [doc](https://classic.yarnpkg.com/en/docs/package-json/#toc-private), and I don't find it in the [migration guide](https://yarnpkg.com/getting-started/migration).

jjtang1985: I found the [yarn2 doc](https://yarnpkg.com/configuration/manifest#private).

tomfrenken: True that's a good catch!

tomfrenken: Actually, now that I look at it, we use the private flag in line 7, so I think the whole publishConfig is an npm artifact that is redundant/deprecated anyway

tomfrenken:
@FrankEssenberger @marikaner I think you have to re-review so I can merge (the requested changes are blocking it currently) I've created a follow-up ticket to create self-contained packages in the backlog.

FrankEssenberger (approved):
LGTM with the follow up task

marikaner (approved):

PR #1044, opened 2021-02-24
Fix: function/action import false positive warning message

jjtang1985:
This PR includes the changes:
- fix the false positive warning
- add more function import tests
- add more function/action import in the cap service, which we don't support for the time being.
    - bound function/action
    - action/function with entity as parameters


packages/generator/src/edmx-to-vdm/v4/action-import.ts
```diff
@@ -57,7 +57,7 @@ function joinActionImportData(
     []
   );
 
-  if (actionImportsWithoutActions) {
+  if (actionImportsWithoutActions.length) {
```
FrankEssenberger, line 60: why an empty array is truthy? Who does that. Zero is falsy so we are fine now.

packages/generator/src/edmx-to-vdm/v4/function-import.ts
```diff
@@ -60,7 +60,7 @@ function joinFunctionImportData(
     []
   );
 
-  if (functionImportsWithoutFunctions) {
+  if (functionImportsWithoutFunctions.length) {
```
FrankEssenberger, line 63: optional: add spyOn in some tests to see that the warn is done as expected.

FrankEssenberger (approved):
LGTM, consider my comments.

Issue #1037, opened 2021-02-23
CSRF token validation failed - SAP Business Application Studio destination with on-premise destination


maxstreifeneder:
**Describe the bug**
Similar issue as #617 from my point of view. 

Debugging a CAP + Cloud SDK app in BAS using a destination with type on-premise fails with "CSRF Token validation failed" (OData read/select works) while the deployed app in CF works fine. 

Happy to provide more insights/guidance/explanation if necessary. Thanks already in advance :-)

**To Reproduce**
1. ```git clone https://github.tools.sap/D063937/cloud-sdk-destination-issue.git``` in BAS
2. ```npm install```
3. ```mbt build && f deploy mta_archives/test_1.0.0.mtar```
4. Create a run configuration of type CAP and connect to the destination (should start the built in proxy)
    ![image](https://user-images.githubusercontent.com/36535298/108909070-5f9da800-7624-11eb-8892-7cbba87ce543.png)

5. Bind the services from the previously deployed MTA (connectivity, destination, xsuaa) to your local app. 
6. merge the destination + VCAP_services information into one .env file (usually different files) and double check that the .env file in the **.vscode/launch.json**is properly set. 
7. Start the run configuration 
    ![image](https://user-images.githubusercontent.com/36535298/108909657-19951400-7625-11eb-9108-210a9b818d3d.png)

8. Go to /test/requests.http and send the request to fire the _getAllBupa_ function
    -> Business Partner from S/4 on prem will be listed in the console

9. send the request to fire the _insertBupa__ function. This won't work. 
10. Do the same with the URL for the deployed app and watch the logs with **cf logs test-srv** - you will see that the CSRF token handle works perfectly as expected. 

**Expected behavior**
Cloud SDK takes care of the full CSRF token handling, no matter if there's a proxy of BAS between the connectivity the service and the locally running app in BAS or not. 

**Screenshots**

**Used Versions:**
- node: v10.23.1
 - npm: 6.14.10
 - SAP Cloud SDK version:  "@sap/cloud-sdk-vdm-business-partner-service": "^1.24.0",

**Code Examples**
https://github.tools.sap/D063937/cloud-sdk-destination-issue/blob/master/srv/catalog-service.js#L10-L18

**Log file**

Log output in BAS: 

```
[cds] - GET /test/insertBupa()
[...]

PR #1056, opened 2021-03-01
fix: Fix Redirect issue in latest BAS on CF

FrankEssenberger:
Relates to #1037 and changes the logic that first the call is made against the URL with a `/` in the end.

### DIFF
--- a/packages/core/src/connectivity/scp-cf/authorization-header.ts
+++ b/packages/core/src/connectivity/scp-cf/authorization-header.ts
@@ -233,7 +233,7 @@ async function getAuthenticationRelatedHeaders(
       return headerForPrincipalPropagation(destination);
     default:
       throw Error(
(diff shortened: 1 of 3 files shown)


jjtang1985 (approved):
Looks good.
With slash and without slashes are different URLs. Different service might treat them differently like:
- treat them with the same behaviour by redirecting from one to another
- treat them completely different with different behaviour

So far, all the issues come from on-promise system that we want to fix, which might not behave the same as the cloud services or other OData Services.
Therefore, we either try A then switch to B like this PR, or make a new option for the generator, so that every service can map to it's csrf token url.

I guess this is a quick win. If there are some complaints about the performance due to the additional requests, we might switch to the mapping.

FrankEssenberger:
@jjtang1985 I checked the cloud system and there both versions work with and wouthout slash. I assume that only the onPremise is broken in a sense that it likes the `/` in the end. But we will keep and eye on this.

Issue #682, opened 2020-11-06
Calling function import from SAP Cloud SDK returns an empty object


miyasuta:
**Describe the bug**
Calling a function import from SAP Cloud SDK returns an empty object, even though the call is successful.
You can reproduce the issue by using the following git repository.
https://github.com/miyasuta/sdk-function-import-test

**To Reproduce**
1. Develop a custom odata service which includes a function import.
2. In a Cloud SDK app, generate an odata client using below command as explained [here](https://developers.sap.com/tutorials/cloudsdk-js-generator.html).
`sap-cloud-sdk generate-odata-client -i service-specification/ -o odata-client/`
3. Call the function import from the Cloud SDK app.
4. When the function import gets called, it falls into then() block, meaning that the call was successful.
However, the response object is empty.

**Expected behavior**
Response object should be filled with returned data.

**Screenshots**
Calling the OData function import directly
![image](https://user-images.githubusercontent.com/39408125/98360018-7667c100-206c-11eb-8b0b-8f270dd95fe9.png)

Calling function import from Cloud SDK app (delivery data should be returned between {})
![image](https://user-images.githubusercontent.com/39408125/98360064-8e3f4500-206c-11eb-8c73-008e2678384e.png)

**Used Versions:**
 - node version via `node -v`: 12.16.1
 - npm version via `npm -v`: 6.14.4
 - SAP Cloud SDK version you used as dependency: 1.31.0

**Code Examples**
My controller code is as below.

```
import { Controller, Get } from '@nestjs/common';
import { functionImports } from 'odata-client/catalog-service';


@Controller('get-delivery')
export class GetDeliveryController {
    @Get()
    getDeliveryInfo() {
        return new Promise((resolve, reject) => {
            functionImports.getDeliveryDate({})
            .execute({
                url: 'http://localhost:4004'                
            })
            .then(res => {
                resolve('Function call successful: ' + JSON.stringify(res))
            })
            .catch(error => {
                resolve('Error occurred: ' + JSON.stringify(error))
            })
        })
    }
}
```

**Log file**
No error has occurred.

**Additional context**
None

**Please check the following points**
- [X ] Have you provided detailed description about your issue?
- [X ] Have you shared all the necessary information to help the cloud-sdk developer reproduce the issue?
- [ X] Have you provided the expected behaviour?
- [ X] Have you attached some useful screenshots?
- [ X] Have you provided the node/npm/cloud-sdk versions?
[...]

PR #1067, opened 2021-03-03
chore: support `executeRaw` that returns raw response without deserialization

jjtang1985:
[ADR](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/implementation-documentation/api-improvements.md#execute-raw)

### open discussion
For update request builder, we have some optimization that returns original entities without sending the request if no non-key properties are sent, because nothing should be updated. In this case, we have alternatives:
- send request anyways and return raw response, making `execute` and `executeRaw` different
- do not send request and align with the `execute`

I chose the second one and return `undefined`, since there is no request/response.

Feel free to suggest better naming.


packages/core/src/http-client/http-client.spec.ts
```diff
@@ -306,4 +306,41 @@ describe('generic http client', () => {
...
+          a: 'a',
+          b: 'b'
+        })
+        .reply(200, { res: 'ult' }, { sharp: 'header' });
+
+      const config = {
+        method: HttpMethod.GET,
+        url: '/api/entity',
+        params: {
+          a: 'a',
```
FrankEssenberger, line 333: The parameters are not asserted - what do they add to the test? Matching the mock implicitly tests they the parameters are considered, but this done by axios.

packages/core/src/odata-common/request-builder/get-by-key-request-builder.ts
```diff
@@ -68,19 +69,33 @@ export abstract class GetByKeyRequestBuilder<
...
   }
+
+  /**
+   * Execute request and return the request and the raw response.
+   *
+   * @param destination - Destination to execute the request against
+   * @param options - Options to employ when fetching destinations
+   * @returns A promise resolving to an [[HttpRequestAndResponse]].
+   */
+  async executeRaw(
```
FrankEssenberger, line 92: Can you check if you can move this method in the `MethodRequestBuilder` to have higher reuse. The build method is already there.

packages/core/src/http-client/http-client.ts
```diff
@@ -187,6 +197,13 @@ function executeWithAxios(request: HttpRequest): Promise<HttpResponse> {
   return axios.request({ ...getAxiosConfigWithDefaults(), ...request });
 }
 
+function executeWithAxiosReturnRequestAndResponse(request: HttpRequest): Promise<HttpRequestAndResponse> {
+  const merged: HttpRequest = { ...getAxiosConfigWithDefaults(), ...request };
```
FrankEssenberger, line 201: could you use the executeHttpRequest method here.

packages/core/src/http-client/http-client-types.ts
```diff
@@ -74,3 +74,8 @@ export interface HttpResponse extends KnownHttpResponseFields {
 export interface HttpReponse extends KnownHttpResponseFields {
   [otherKey: string]: any;
 }
+
+export interface HttpRequestAndResponse{
```
FrankEssenberger, line 78: Discussion @marikaner @jjtang1985 @florian-richter: What if we would also include a typed data object, then people can use the request and response but also have a typed data in the executeRaw response.

packages/core/src/http-client/http-client.ts
```diff
@@ -126,6 +126,16 @@ export async function buildAxiosRequestConfig<T extends HttpRequestConfig>(
...
 
+/**
+ * Builds a [[DestinationHttpRequestConfig]] for the given destination, merges it into the given requestConfig
+ * and executes it (using Axios). The request is also included in the object returned.
+ *
+ * @param destination - A destination or a destination name and a JWT.
+ * @param requestConfig - Any object representing an HTTP request.
+ * @returns An [[HttpRequestAndResponse]].
+ */
+export const executeHttpRequestReturnRequestAndResponse = execute(executeWithAxiosReturnRequestAndResponse);
```
FrankEssenberger, line 137: Why do want to offer this method? I guess I missed the use case - sorry for that. I understand that on the odata requests it is good for people to get the actual request and response if needed. But if people are using the `executeHttpRequest` method they have their request and response, or is the request also adjusted from axios but then I still have my reference to req if I just called `res = executeHttp(req)` I have the res and req pair.

packages/core/src/http-client/http-client.spec.ts
jjtang1985: I'll simplify the test.

packages/core/src/http-client/http-client-types.ts
```diff
@@ -74,3 +74,8 @@ export interface HttpResponse extends KnownHttpResponseFields {
...
+  request: HttpRequest;
+  response: HttpResponse;
```
FrankEssenberger, line 80: Seconde discussion @marikaner @jjtang1985 @florian-richter related to the one above. For the opanApi client we return an `AxiosResponse` if you execute `executeRaw` which has a data field and sone other information. I wonder if we could align the two `executeRaw`. For me it is an implementation detail that we use axios so I would rather have a return type which could orientate at the `AxiosResponse` but perhaps also includes the request.

FrankEssenberger (dismissed):
In general looks good. Only minor comments but a bigger discussion on the return type in alignment with the openApi executeRaw. In the ADR I could not see a clear decision on the options.

packages/core/src/http-client/http-client.ts
jjtang1985: I can extract a function for getting the merged request, but since this is used in the return object, the method call is not avoidable.

packages/core/src/http-client/http-client-types.ts
jjtang1985: For the missing alignment, I was about to add a ticket to add the original request to the openapi one as well.

packages/core/src/http-client/http-client-types.ts
jjtang1985: I have some use cases where we can not build the type data (return entity from deserialisation), which makes this ticket in the input queue.
- function/action import with entity type as return type, that matches to multiple entity sets. We recommend to use `executeRaw`, since deserialisation is not possible.
- this [os issue](https://github.com/SAP/cloud-sdk-js/issues/682) where special response structure does not meet our expectation and therefore deserialisation is not possible.

I think typically, the best practise is to use `execute` to get an entity. But for the cases shown above or for debugging purpose, we can use `executeRaw`.

packages/core/src/http-client/http-client.ts
jjtang1985: ##### use case
This is purely for sdk internal usage.

##### background
The `executeHttpRequest` below contains two part `execute(destination, requestConfit)` and `executeWithAxios`, where
- `execute(destination, requestConfit)` builds a final request from destination and requestConfig and use the executor(currently only `axios`) to make the request.
- `executeWithAxios` uses `axios` to execute the final request
```
export const executeHttpRequest = execute(executeWithAxios)
```

The structure is nice, because it provides the flexibility when we want to support a second executor.

##### challenge
Given that the final request has to be build by us (currently in the `execute`), if we want to keep the structure, we can only touch the `executeWithAxios`. So I created a new variant.

##### alternatives
We can refactor the `executeHttpRequest`
```
export const executeHttpRequest = executeWithAxios(buildRequest(destination, requestConfit))
```
Then we have below when there is a second executor
```
executeWithSomeExecutor(buildRequest(destination, requestConfit))
```
Here is a new version for `executeHttpRequestReturnRequestAndResponse`, where we need to build the request first to avoid building twice, so the actual code is a bit more.
```
export const executeHttpRequestReturnRequestAndResponse = {
  request: buildRequest(destination, requestConfit)
  response: executeWithAxios(buildRequest(destination, requestConfit))
}
```

FrankEssenberger: ok understood.

packages/core/src/http-client/http-client-types.ts
FrankEssenberger: Ok I got the point, that the `executeRaw` is also our fallback if something goes wrong with the deserialization. But I do not see the `action-function-import-request-builder`. Is it missing?

packages/core/src/http-client/http-client-types.ts
FrankEssenberger: Fine, I guess it is conceptional not possible to remove the deserializing of the response like in the `executeRaw`, because this is done within the generated client where we can not do much?

jjtang1985: Although we cannot influence the underlying openapi magic, but it does not really do so much. For the deserialization it simply returns the code below with the return type `AxiosPromise<Array<TestEntity>>>`
```
return axios.request(axiosRequestArgs);
```
So with the current version, we are still able to achieve the `executeRaw`.

packages/core/src/http-client/http-client-types.ts
jjtang1985: Good catch, I'll add it.

packages/core/src/odata-v2/request-builder/update-request-builder.ts
```diff
@@ -70,6 +71,28 @@ export class UpdateRequestBuilder<EntityT extends Entity>
...
+    destination: Destination | DestinationNameAndJwt,
+    options?: DestinationOptions
+  ): Promise<HttpRequestAndResponse| undefined> {
+    if (this.isEmptyObject(this.requestConfig.payload)) {
+      logger.info('The request is not executed because no properties of the entity are modified.');
+      return;
+    }
+
+    const request = await this.build(destination, options);
+    warnIfNavigation(request, this._entity, this._entityConstructor);
```
jjtang1985, line 92: This is the additional warning.

packages/core/src/odata-common/request-builder/get-by-key-request-builder.ts
jjtang1985: Currently, there is a technical difficulty, where update request builder v2 needs a check for additional warning messages, which is injected between the two steps of the `executeRaw` below:
- building the final request
- execute it
See comment in the `packages/core/src/odata-v2/request-builder/update-request-builder.ts`

FrankEssenberger (approved):
FIne for me with the follow up on adjusting the openApi execute raw also to contain request object.

PR #1068, opened 2021-03-03
fix: Proxy and client cert authentication

FrankEssenberger:
I realized while working on BAS that client certificate authentication is not working when also proxies are used.
I adjusted our code that the proxy settings are also passed to the http-agent. Unfortunately also the proxy agent itself does not forward the presented certificates to the target system. So they stay on the proxy and you get a 401.

I checked by locally adjusting the node modules that this PR https://github.com/TooTallNate/node-https-proxy-agent/pull/111 fixes the issue, but the repo seems to be not very active. We can merge the code on our side but it will not work until we get an update on the node-https-proxy and this could take a while...

### DIFF
--- a/packages/core/src/connectivity/scp-cf/proxy-util.ts
+++ b/packages/core/src/connectivity/scp-cf/proxy-util.ts
@@ -270,6 +270,14 @@ export function proxyAgent(
     throw new Error('Proxy config must not be undefined.');
   }
 
+  if(options?.host){
+    logger.warn(`The agent options you passed to the proxy agent creation contains the host "${options.host}" which will overwrite the host from the proxy config.`)
+  }
+
+  if(options?.port){
+    logger.warn(`The agent options you passed to the proxy agent creation contains the port "${options.host}" which will overwrite the port from the proxy config.`)
+  }
+
   const agentConfig = {
     host: proxyConfig.host,
     protocol: proxyConfig.protocol,
--- a/packages/core/src/connectivity/scp-cf/proxy-util.ts
+++ b/packages/core/src/connectivity/scp-cf/proxy-util.ts
@@ -271,11 +271,11 @@ export function proxyAgent(
   }
 
   if(options?.host){
-    logger.warn(`The agent options you passed to the proxy agent creation contains the host "${options.host}" which will overwrite the host from the proxy config.`)
+    logger.warn(`The agent options you passed to the proxy agent creation contains the host "${options.host}" which will overwrite the host from the proxy config.`);
   }
 
   if(options?.port){
-    logger.warn(`The agent options you passed to the proxy agent creation contains the port "${options.host}" which will overwrite the port from the proxy config.`)
+    logger.warn(`The agent options you passed to the proxy agent creation contains the port "${options.host}" which will overwrite the port from the proxy config.`);
   }
 
   const agentConfig = {
--- a/packages/core/src/connectivity/scp-cf/proxy-util.ts
+++ b/packages/core/src/connectivity/scp-cf/proxy-util.ts
@@ -275,7 +275,7 @@ export function proxyAgent(
   }
 
   if(options?.port){
-    logger.warn(`The agent options you passed to the proxy agent creation contains the port "${options.host}" which will overwrite the port from the proxy config.`);
+    logger.warn(`The agent options you passed to the proxy agent creation contains the port "${options.port}" which will overwrite the port from the proxy config.`);
   }
 
   const agentConfig = {
--- a/packages/core/src/connectivity/scp-cf/proxy-util.ts
+++ b/packages/core/src/connectivity/scp-cf/proxy-util.ts
@@ -1,3 +1,4 @@
+import { AgentOptions } from 'https';
 import { HttpProxyAgent } from 'http-proxy-agent';
 import { HttpsProxyAgent } from 'https-proxy-agent';
 import { createLogger } from '@sap-cloud-sdk/util';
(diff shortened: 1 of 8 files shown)


test-packages/integration-tests/test/auth-flows/auth-flow.spec.ts
```diff
@@ -125,6 +125,13 @@ describe('OAuth flows', () => {
     expect(response.status).toBe(200);
   }, 60000);
 
+  it('ClientCertificate: Fetches the certificate and uses it',async ()=>{
```
jjtang1985, line 128: This will again be disabled for the automated tests.
Should we consider adding an e2e test for it?

packages/core/src/connectivity/scp-cf/proxy-util.ts
```diff
@@ -266,18 +270,21 @@ export function proxyAgent(
     throw new Error('Proxy config must not be undefined.');
   }
 
+  const agentConfig = {
+    host: proxyConfig.host,
+    protocol: proxyConfig.protocol,
+    port: proxyConfig.port,
+    ...options
```
jjtang1985, line 277: Should we add a warning if `options` contain some keys like `host` to make our life easy when debugging?

packages/core/src/http-client/http-agent.ts
```diff
@@ -29,10 +29,12 @@ export function getAgentConfig(
...
     ? AgentType.PROXY
     : AgentType.DEFAULT;
+  // eslint-disable-next-line @typescript-eslint/no-shadow
+  const certificateOptions = getCertificateOption(destination);
   if (agentType === AgentType.PROXY) {
-    return createProxyAgent(destination);
+    return createProxyAgent(destination,certificateOptions);
   }
-  return createDefaultAgent(destination);
+  return createDefaultAgent(destination,certificateOptions);
```
jjtang1985, line 37: @marikaner , I found missing spaces;) I guess the lint config should be aligned. Have we reached some decision?

packages/core/src/http-client/http-agent.ts
```diff
@@ -29,10 +29,12 @@ export function getAgentConfig(
   const agentType = destination.proxyConfiguration
     ? AgentType.PROXY
     : AgentType.DEFAULT;
+  // eslint-disable-next-line @typescript-eslint/no-shadow
+  const certificateOptions = getCertificateOption(destination);
```
jjtang1985, line 33: I don't see shadow name. Could you please share a bit details?

packages/core/src/http-client/http-agent.ts
```diff
@@ -77,6 +73,31 @@ const certificateOptions = (destination: Destination) => (
...
+function getCertificateOption(destination: Destination): Record<string, any>  {
+  // http case: no certificate needed
+  if(getProtocolOrDefault(destination) === Protocol.HTTP){
+  if (destination.isTrustingAllCertificates) {
+    logger.warn('"isTrustingAllCertificates" is not available for HTTP.');
+  }
+    return {};
+  }
+  // https case: get certificate options
+  if (destination.isTrustingAllCertificates) {
```
jjtang1985, line 92: Without linting, nested `if-else` is terrible.

jjtang1985 (changes requested):
Looks good.
My concern would be the linting and automated tests.

packages/core/src/http-client/http-agent.ts
marikaner: I'll create a ticket to solve this ;)

packages/core/src/http-client/http-agent.ts
FrankEssenberger: for me I got a lint error that the variable in the if block below is shadowed - I do not understand that either.

packages/core/src/http-client/http-agent.ts
FrankEssenberger: I agree. I did execute `yarn lint:fix` but it did not fix it. Also the pipeline is happy - why is that?

packages/core/src/connectivity/scp-cf/proxy-util.ts
FrankEssenberger: good idea. I do that.

test-packages/integration-tests/test/auth-flows/auth-flow.spec.ts
FrankEssenberger: yes forgot to disable it. There is an E2E test in the pipeline and also some mocked tests - I just needed the real deal to see that the fix is working E2E.

packages/core/src/http-client/http-agent.ts
FrankEssenberger: I can not find the ticket? Is it in our backlog or in this repo?

packages/core/src/connectivity/scp-cf/proxy-util.ts
```diff
@@ -266,18 +270,29 @@ export function proxyAgent(
     throw new Error('Proxy config must not be undefined.');
   }
 
+  if(options?.host){
+    logger.warn(`The agent options you passed to the proxy agent creation contains the host "${options.host}" which will overwrite the host from the proxy config.`);
+  }
+
+  if(options?.port){
+    logger.warn(`The agent options you passed to the proxy agent creation contains the port "${options.host}" which will overwrite the port from the proxy config.`);
```
jjtang1985, line 278:
```suggestion
    logger.warn(`The agent options you passed to the proxy agent creation contains the port "${options.port}" which will overwrite the port from the proxy config.`);
```

jjtang1985 (approved):
Looks good, with a minor comment.
Linting will be handled in a follow up ticket.

jjtang1985:
For the failing unit test, if you need further review, just let me know.

packages/core/src/http-client/http-agent.ts
jjtang1985: https://github.wdf.sap.corp/MA/scp-rocket/issues/1768

PR #1069, opened 2021-03-04
chore: Remove transpliation artifacts from unit tests

marikaner:
This removes files that should not be present in the unit tests.


scripts/generate-test-services.ts
```diff
@@ -42,7 +42,7 @@ const generatorConfigOpenApi: GeneratorOptions = {
   input: path.resolve('test-resources', 'openapi-service-specs'),
   outputDir: path.resolve('test-packages', 'test-services', 'openapi'),
   clearOutputDir: true,
-  generateJs: true,
```
FrankEssenberger, line 45: Hm, I enabled this to see if also the transpiling in other words correct code is generated. I think since the objects are used in so many tests we can be sure that it generate valide code. We also have the E2E tests.

scripts/generate-test-services.ts
```diff
@@ -170,7 +170,10 @@ async function generateAll(): Promise<void> {
   }
 
   if (arg === 'openapi' || arg === 'rest' || arg === 'all') {
-    await generateOpenApi(generatorConfigOpenApi);
+    await generateOpenApi({
+      ...generatorConfigOpenApi,
+      generateJs: true
+    });
```
FrankEssenberger, line 176: Question: why we want the JS here?

FrankEssenberger (changes requested):
Just a minor question

scripts/generate-test-services.ts
marikaner: We generate the services twice, once in the core unit tests and once in the test services package. The unit tests should not contain any transpiled code, but the e2e tests should. At least that way it would be consistent with OData. (And that is the reason for this PR). I hope this also answers the second question ;)

scripts/generate-test-services.ts
marikaner: See [here](https://github.com/SAP/cloud-sdk-js/pull/1069#discussion_r592326926).

FrankEssenberger (approved):
LGTM

Issue #939, opened 2021-01-26
Batch DELETE throws batch-response-parser error


stefania-santimbrean:
Hello,

I am doing a batch request which contains DELETE requests for PurchaseOrderItems using the following code:

```
public async deletePurchaseOrderItems(purchaseOrderId: string, purchaseOrderItemIds: string[]): Promise<void> {
    try {
        const deleteRequests = purchaseOrderItemIds.map(poi =>
            PurchaseOrderItem.requestBuilder().delete(purchaseOrderId, poi)
        )
        // Execute the batch request
        await batch(changeset(...deleteRequests)).execute(Constants.DESTINATION);
        this.logger.debug(`Purchase Order ${purchaseOrderId} items deleted`);
    } catch (err) {
        this.logger.debug(err);
    }
}
```
The request does what I need, which is to mark for deletion the items in the Purchase Order, but the logs show an error:
![image](https://user-images.githubusercontent.com/74544010/105845156-ff92f200-5fe2-11eb-9a65-23abcbbed18a.png)

Could you please help me figure out where is the error coming from an how to handle it?

jjtang1985:
Hi @stefania-santimbrean , thank you for approaching us.

From the log statement, I assume this is either because of the unexpected response format from the service or there is a bug from the sdk.
Is it possible to provide the response so that we can understand better?

Please be aware, I see logs with both `error` and `warn` levels, which might come from different cases.
Therefore, if you see batch response with different format, please share them together.

Thank you very much.
Junjie

stefania-santimbrean:
Hello @jjtang1985 ,

Thanks for responding so quick!

I changed the code to:
```
// Execute the batch request
const response = await batch(changeset(...deleteRequests)).execute(Constants.DESTINATION);
this.logger.debug(response);
```
The logged response object looks like this:
```
[
  {
    "responses": [
      {
        "body": {},
        "httpCode": 204
      },
      {
        "body": {},
        "httpCode": 204
      },
      {
        "body": {},
        "httpCode": 204
      },
      {
        "body": {},
        "httpCode": 204
      },
      {
        "body": {},
        "httpCode": 204
      },
      {
        "body": {},
        "httpCode": 204
      },
      {
        "body": {},
        "httpCode": 204
      },
      {
        "body": {},
        "httpCode": 204
      },
      {
        "body": {},
        "httpCode": 204
      },
      {
        "body": {},
        "httpCode": 204
      },
      {
        "body": {},
[...]

PR #1074, opened 2021-03-05
fix: Batch delete bug

tomfrenken:
This PR is related to issue #939 where errors were thrown when a batch response was parsed and the response contained an empty body.


packages/core/src/odata-common/request-builder/batch/batch-response-parser.ts
```diff
@@ -122,7 +122,13 @@ export function splitChangeSetResponse(changeSetResponse: string): string[] {
  * @returns A list of sub responses represented as strings.
  */
 export function splitResponse(response: string, boundary: string): string[] {
-  const parts = response.split(`--${boundary}`).map(part => part.trim());
```
tomfrenken, line 125: This was causing the bug, it was trimming every part of a batch response, which is exactly what we want, except when the body is empty. When the body of the response is empty the `parseResponseBody()` function, which is used by the `parseResponseData()` function, is trying to parse the last line of the response with `JSON.parse()`, which results in an error because, with an empty body, the last line is not valid JSON, with my fix the last line is now an empty line if the response has no body.

packages/core/src/odata-common/request-builder/batch/batch-response-parser.spec.ts
```diff
@@ -258,4 +261,63 @@ describe('batch response parser', () => {
...
+        'content-transfer-encoding: binary',
+        '',
+        'HTTP/1.1 204 No Content',
+        'Content-Length: 0',
+        'dataserviceversion: 2.0',
+        '',
+        '{"valid": "json"}',
+        '',
+        '',
+        '--3B17E95920A7FAF8BCB7495D043515001',
```
jjtang1985, line 284:
```suggestion
        '--3B17E95920A7FAF8BCB7495D043515001',
```
I tried to build a batch request like below by using our cap e2e application, it seems there is no empty lines between the response body and the next boundary.
```
batch(
      getAll,
      changeset(create,deleteRequestBuilder)
    )
```

packages/core/src/odata-common/request-builder/batch/batch-response-parser.ts
```diff
@@ -122,7 +122,14 @@ export function splitChangeSetResponse(changeSetResponse: string): string[] {
...
+  const parts = response.split(`--${boundary}`).map(part => {
+    let trimmedPart = part.trim();
+    if(trimmedPart.includes('No Content')){
```
jjtang1985, line 127: To be safe, maybe apply a stricter rule like `includes('204 No Content')` or maybe a line starts with `HTTP` and end with `No Content` (example: `HTTP/1.1 204 No Content`)
Also, add a comment about why the empty line matters.

jjtang1985 (changes requested):
Nice catch about the missing empty line.
Please consider my comments and update the changelog.

packages/core/src/odata-common/request-builder/batch/batch-response-parser.ts
tomfrenken: Yeah good idea, I wasn't sure if only 204 responses can have empty bodies, that's why I didn't specify it further, I've updated the tests & the function - thanks for the feedback.

packages/core/src/odata-common/request-builder/batch/batch-response-parser.ts
```diff
@@ -122,7 +122,14 @@ export function splitChangeSetResponse(changeSetResponse: string): string[] {
...
 export function splitResponse(response: string, boundary: string): string[] {
-  const parts = response.split(`--${boundary}`).map(part => part.trim());
+  const newLineSymbol = detectNewLineSymbol(response);
+  const parts = response.split(`--${boundary}`).map(part => {
+    let trimmedPart = part.trim();
+    if(trimmedPart.includes('204 No Content')){
+      trimmedPart += newLineSymbol;
+    }
+    return trimmedPart;
+  });
```
marikaner, line 132: Do we even need the trimming here?

marikaner, line 132:
```suggestion
  const parts = response.split(`--${boundary}`).map(part => {
    const trimmedPart = part.trim();
    return trimmedPart.includes('204 No Content')) ? `${trimmedPart}${newLineSymbol}` : trimmedPart;
  });
```

marikaner: With a ternary this might look a little nicer, other than that from my side, lgtm ;)

tomfrenken: agreed, I should be using ternary more frequent 🤔

jjtang1985 (approved):

tomfrenken:
*it was failing because of a typo, already fixed it 👍

PR #1083, opened 2021-03-09
fix: `executeRaw` returns enhanced `HttpResponse`

jjtang1985:
This PR will "fix" the return type of the `executeRaw`, so it returns `AxiosResponse`, which aligns with the openapi request builder.

### how does request look like in the response
The `AxiosResponse` contains an attribute `config: AxiosRequestConfig;`, which is typed and contains everything about the original request.

### how about other clients if I don't want to use axios
- This [page](https://nodejs.libhunt.com/axios-alternatives) shows some major `axios` alternatives. I would say `axios` would be our first choice maybe at least until the end of this year, given that [request](https://www.npmjs.com/package/request) is deprecated.
- In order to be a valid candidate in terms of [the existing structure of our generic http client](https://github.com/SAP/cloud-sdk-js/blob/main/packages/core/src/http-client/http-client.ts#L201), the lib should support the usage like `lib.someFunction(request);`. The usage of `superagent`, e.g., can be found below needs additional effort from the sdk side to build such function call. Without good reasons, these kind of libs would not be our first choice.
```
superagent
  .post('/api/pet')
  .send({ name: 'Manny', species: 'cat' }) // sends a JSON post body
  .set('X-API-Key', 'foobar')
  .set('accept', 'json')
  .end((err, res) => {
    // Calling the end function will send the request
  });
```

### in case we would like to use a second client, how about the return type of `executeRaw`, which is bound to `axios` now
- Personally, I don't see this will happen before the next year, meaning, we can release 3.0 sdk with breaking changes.
- [Option A] we have multiple `execute` function with client option, so different `execute` returns different response type. This will make all of them independent, so no breaking changes.
- [Option B] we build our own response type instead of using the return type from client (e.g., `AxiosResponse`), so we use different client based on user input, but always returns the same data type. This is also doable, but the additional effort would be, we map the client response to sdk response and convert it. I would leave it until we find a proper candidate, because although the common key of the request is quite obvious (baseUrl, path, method, header, body...) so the sdk response type is almost there, but the mapping is not. See some examples below for some comparison between `axios` and `superagent`. Since the mapping is not clear for the second candidate, I would leave it as it is to avoid this unnecessary additional effort for now.

| |`axios`|`superagent`|
|-|-|-|
|base url|`baseUrl`|`host`|
|protocol|`baseUrl`|`protocol`|
|request body|`data`|`_data`|


jjtang1985:
@marikaner @FrankEssenberger , as discussed today, I investigated a bit and this PR is my proposal for the alignment of the odata executeRaw and openapi executeRaw. Feel free to approach me if you have questions.

packages/core/src/http-client/http-client-types.ts
```diff
@@ -74,8 +74,3 @@ export interface HttpResponse extends KnownHttpResponseFields {
 export interface HttpReponse extends KnownHttpResponseFields {
   [otherKey: string]: any;
 }
-
-export interface HttpRequestAndResponse{
```
FrankEssenberger, line 78: Should we only deprecate? Was this exported to root level so some could use it? If we remove it then add a compatibility node.

FrankEssenberger (approved):
LGTM consider my comment regarding compatability. I like the quick win with the axios, then it is also aligned with the openApi client.

packages/core/src/http-client/http-client-types.ts
jjtang1985: This interface is still under the canary version. I guess it's fine.

marikaner: Agreed.

packages/core/src/http-client/http-client.ts
```diff
@@ -201,11 +201,8 @@ function executeWithAxios(request: HttpRequest): Promise<HttpResponse> {
   return axios.request(mergeRequestWithAxiosDefaults(request));
 }
 
-function executeWithAxiosReturnRequestAndResponse(request: HttpRequest): Promise<HttpRequestAndResponse> {
-  const merged = mergeRequestWithAxiosDefaults(request);
-  return axios.request(merged).then(
-    res => ({ request: merged, response: res })
-  );
+function executeWithAxiosReturnAxiosResponse(request: HttpRequest): Promise<AxiosResponse> {
```
marikaner, line 204: I am not a fan of this function. It is an exact duplication of the function above and the only difference is, that we are using a different return type. I don't think that the name `executeHttpRequestReturnAxiosResponse` should be used in the public API.
In my opinion we should consider one of two options:
1. Either enrich the `HttpResponse` type with some more information. We could use [this](https://github.com/axios/axios#response-schema) as a basis. The downside here is, that this is not possible without breaking changes. Therefore the added properties would have to be optional or we would have to think of something different.
2. Use `AxiosResponse` in executeWithAxios directly. This should work OOB because `AxiosResponse` is compatible with `HttpResponse`. I am not a big fan of using another tool's API in our own, because they would have to install axios to reference the type.

I don't have a strong opinion here, but maybe @FrankEssenberger has something to add? And maybe @mr-flannery might have some feedback as well?

marikaner (changes requested):
Hey @jjtang1985, I like the resulting API for executeRaw, but I am not a fan of the API in http-client. I would like to discuss whether we should go for one of the two options I mentioned.

packages/core/src/http-client/http-client.ts
jjtang1985: As discussed with Marika. I would go with the option 1 and add an optional property `request` to our interface `HttpResponse`.
- no breaking changes
- since it's compatible with `AxiosReponse`, one can cast it to `AxiosResponse` on demand
- the ugly function will disappear, which does not do anything but having a different return type.

CHANGELOG.md
```diff
@@ -18,7 +18,7 @@
 
 ## New Functionality
 
-- [core] Support `executeRaw` in all `RequestBuilder`, which returns request and raw response without deserialization.
+- [core] Support `executeRaw` in all `RequestBuilder`, which returns `DefaultHttpResponse` that contains raw response and original request.
```
```suggestion
- [core] Support `executeRaw` in all `RequestBuilder`, which returns `DefaultHttpResponse` that contains the raw response and original request.
```

packages/core/src/http-client/http-client-types.ts
```diff
@@ -75,7 +75,7 @@ export interface HttpReponse extends KnownHttpResponseFields {
   [otherKey: string]: any;
 }
 
-export interface HttpRequestAndResponse{
-  request: HttpRequest;
-  response: HttpResponse;
+export interface DefaultHttpResponse extends HttpResponse {
+  request: any;
```
marikaner, line 79: Why not: `HttpRequest`?
```suggestion
  request: HttpRequest;
```

packages/core/src/http-client/http-client-types.ts
```diff
@@ -75,7 +75,7 @@ export interface HttpReponse extends KnownHttpResponseFields {
...
+  [otherKey: string]: any;
```
marikaner, line 80: Is this necessary?
```suggestion
```

packages/core/src/http-client/http-client-types.ts
```diff
@@ -75,7 +75,7 @@ export interface HttpReponse extends KnownHttpResponseFields {
   [otherKey: string]: any;
 }
 
-export interface HttpRequestAndResponse{
-  request: HttpRequest;
-  response: HttpResponse;
+export interface DefaultHttpResponse extends HttpResponse {
```
marikaner, line 78: Should we also add `request?: HttpRequest` to the original `HttpResponse`?

packages/core/src/http-client/http-client.ts
```diff
@@ -201,11 +202,8 @@ function executeWithAxios(request: HttpRequest): Promise<HttpResponse> {
   return axios.request(mergeRequestWithAxiosDefaults(request));
 }
 
-function executeWithAxiosReturnRequestAndResponse(request: HttpRequest): Promise<HttpRequestAndResponse> {
-  const merged = mergeRequestWithAxiosDefaults(request);
-  return axios.request(merged).then(
-    res => ({ request: merged, response: res })
-  );
+function executeWithAxiosReturnDefaultResponse(request: HttpRequest): Promise<DefaultHttpResponse> {
```
marikaner, line 205: Now this function should not be needed anymore. We should change `executeWithAxios` to `function executeWithAxios(request: HttpRequest): Promise<HttpResponse>`

marikaner (changes requested):
I still think we don't need two functions, other than that looks good!

packages/core/src/http-client/http-client-types.ts
marikaner: That would force you to follow the HttpRequest API.

packages/core/src/http-client/http-client-types.ts
jjtang1985: Because the request does not really match `HttpRequest`.
The url (`response.request.url`) and the request body (`response.request.data`) cannot be found from the return object.

packages/core/src/http-client/http-client-types.ts
jjtang1985: good catch

packages/core/src/http-client/http-client.ts
jjtang1985: you mean change to this `function executeWithAxios(request: HttpRequest): Promise<DefaultHttpResponse>`?

packages/core/src/http-client/http-client.ts
```diff
@@ -122,20 +122,10 @@ export async function buildAxiosRequestConfig<T extends HttpRequestConfig>(
  *
  * @param destination - A destination or a destination name and a JWT.
  * @param requestConfig - Any object representing an HTTP request.
- * @returns An [[HttpResponse]].
+ * @returns Promise resolving to an [[HttpResponse]].
```
marikaner, line 125:
```suggestion
 * @returns A promise resolving to an [[HttpResponse]].
```

marikaner (approved):
Looks awesome to me.

PR #1088, opened 2021-03-10
fix: dependabot with yarn2

tomfrenken:


.github/dependabot.yml
```diff
@@ -4,5 +4,7 @@ updates:
   directory: "/"
   schedule:
     interval: daily
-    time: "01:00"
+    time: "09:14"
```
jjtang1985, line 7:
```suggestion
    time: "08:40"
```
I would like to see whether this works and don't want to wait for another 34 min;)

jjtang1985:
> Scheduled workflows run on the latest commit on the default or base branch.

I guess without merging to the main branch, the schedule won't work.
Here is the [doc](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events).

tomfrenken:
Indeed, I've tested it on forks this far, but working with dependabot and Github actions on a fork is rather difficult because forks can't use secrets, and also the default token `GITHUB_TOKEN` which you usually use to authenticate a workflow has read-only permissions on a fork, so I wasn't able to push with a Github action to a branch.

tomfrenken:
Also, the when you schedule a specific time, you are essentially still put in a queue before dependabot actually starts working, which delays it pretty much

tomfrenken:
Okay I ran a modified version to be able to run on forks here:
https://github.com/tomfrenken/cloud-sdk-js/pull/1
seems to be working as intended

jjtang1985 (approved):
LGTM.
For the existing dependabot PR, please consider some follow up tests by e.g., making some dummy commits without real changes.

jjtang1985:
Example git command will be `git commit --allow-empty -m "test"`.

PR #1094, opened 2021-03-11
Revert "Yarn2 migration"

jjtang1985:
Reverts SAP/cloud-sdk-js#1043


jjtang1985:
As the fix is still in progress, I would revert it as discussed.

PR #1096, opened 2021-03-11
fix: gh action authentication issue

tomfrenken:
This fix tries to use a generic user instead of the bot user in an attempt to use the right secrets, it works on a branch, but we can only be sure that it really works after merging (previous fixes also worked on branches ...)


tomfrenken:
After fruther investgation, since march 01, 2021 dependabot pull-requests can't access secrets 🙄

PR #1097, opened 2021-03-11
chore: generate sdk metadata

FrankEssenberger:
This PR creates a first approach to create the SDK metadata. I started for the OData case. We should of course also do something for the openApi. For this a shared generator package would be good. Are there any plans yet?

The metadata consist of the header and the client part. The client has a general part and one for the pregenerated lib. I created a BLI for the follow up tasks.


scripts/generate-test-services.ts
```diff
@@ -35,7 +35,8 @@ const generatorConfigOData = {
   generateCSN: false,
   // Unnecessary options
   sdkAfterVersionScript: false,
-  s4hanaCloud: false
+  s4hanaCloud: false,
+  generateSdkMetadata: false
```
jjtang1985, line 39: minor: default is false.

packages/generator/src/service/package-json.ts
```diff
@@ -1,15 +1,14 @@
 export function packageJson(
   npmPackageName: string,
-  versionInPackageJson: string | undefined,
-  generatorVersion: string,
+  version: string,
```
jjtang1985, line 3: The idea of two parameters is you can generate a client with `1.0.0` version but uses dependency `sdk-core` version `2.0.0`.

packages/generator/test/test-util/create-generator-options.ts
```diff
@@ -18,6 +18,7 @@ export function createOptions(
     generateJs: false,
     sdkAfterVersionScript: false,
     generateCSN: false,
+    generateSdkMetadata:false,
```
jjtang1985, line 21: minor: by default it's false

packages/generator/src/sdk-metadata/links.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { executeHttpRequest } from '@sap-cloud-sdk/core';
+import { ErrorWithCause } from '@sap-cloud-sdk/util';
+import { getLinks } from './links';
+
+describe('sdk metadata links',()=>{
```
jjtang1985, line 5: This is an e2e test, 10 sec is still fine. If this becomes longer or unstable, we might want to move to nightly job.

packages/generator/src/sdk-metadata/generation-and-usage.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+  it('creates generic usage example',async()=>{
+    await expect(getGenericUsage()).resolves.toMatchSnapshot();
+  });
+
+  it('creates api specific usage for entity',async()=>{
+    await expect(getApiSpecificUsage({ npmPackageName: '@sap/dummy-package', entities : [{ className:'DummyClass' }] } as any)).resolves.toMatchSnapshot();
+  });
+
+  it('gives instruction with working link',async ()=> {
+    const url = getGenerationDocumentation().match(/(http.*)/)![1];
```
jjtang1985, line 14: minor: I would create a const which only returns a url and test that url to avoid using regex, which might be buggy (e.g., `http http://www.example.com`)

packages/generator/src/sdk-metadata/code-samples/generic-get-all-code-sample.ts
```diff
@@ -0,0 +1,3 @@
+import { TestEntity } from '@sap-cloud-sdk/core/test/test-util/test-services/v2/test-service';
```
jjtang1985, line 1: The code example should only import from root index
```suggestion
import { TestEntity } from '@sap-cloud-sdk/core'
```

packages/generator/src/sdk-metadata/pregenerated-lib.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  it('returns description of the service',()=>{
+    expect(getServiceDescription(service,createOptions())).toMatchSnapshot();
+  });
+
+  it('returns a timestamp in unix format',()=>{
+    expect(getTimeStamp()).toMatch(/\/Date\(\d{13,13}\)\//);
+  });
+
+  it('checks if there is a client published',async ()=>{
```
jjtang1985, line 40: Could we mark these newly added e2e tests with e.g. tags [E2E] so once we want to take further actions, we can easily search for them.

test-resources/odata-service-specs/index.ts
```diff
@@ -0,0 +1,3 @@
+/* Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. */
+
+export const oDataServiceSpecs = __dirname;
```
jjtang1985, line 3: could you please share the use case of this?

packages/generator/src/sdk-metadata/sdk-metadata-types.ts
```diff
@@ -0,0 +1,42 @@
...
+  support: UrlString;
+  apiHubTutorial: UrlString;
+  generationManual: UrlString;
+}
+
+export interface GenerationAndUsage {
+  successfulGenerationVerified: boolean; // we have tested the API spec with the generator -> true, false if not tested
+  generationSteps: string; // "some-step<br />another step <br />another step"
+  apiSpecificUsage: string;
+  genericUsage: string; // import { <ServiceName> } from @sap/cloud-sdk-vdm-<service-name>-service<br /><ServiceName>.requestBuilder()<br />.<operationName>()<br />.execute(destination);
```
jjtang1985, line 37: I see lots of types are using `string`, but typically there are line breaks `<br />` used.
Could we use `string[]` and `join` later on?

jjtang1985 (changes requested):
Looks good, please consider my comments.

test-resources/odata-service-specs/index.ts
FrankEssenberger: I used it in the getTestService to read a service spec. I found many read commands in our tests like: 
```
 edmxPath:
              '../../test-resources/odata-service-specs/v2/API_TEST_SRV/API_TEST_SRV.edmx'
```

I would like to replace them all in a seperate PR because using the variable makes refactoring easier.

scripts/generate-test-services.ts
FrankEssenberger: True but it is not optional so I made it optional as well.

packages/generator/src/service/package-json.ts
FrankEssenberger: oh, I misssed that thanks for finding the mistake.

packages/generator/src/sdk-metadata/links.spec.ts
FrankEssenberger: I had the same thought - I add an [E2E] to all tests making calls to the internet.

packages/generator/src/sdk-metadata/sdk-metadata-types.ts
FrankEssenberger: This type is meant a bit as the schema for the final web service. It is not optimal to use TS for that but it was so convinient because you can refactor a lot. I will create a new dummy type like `MultiLineText` to indicate that the field contains a text with multiple lines.

jjtang1985 (approved):
LGTM. Only some minor comments.

PR #1099, opened 2021-03-11
fix: dependabot write privileges

tomfrenken:
With this change, dependable receives write privileges and can therefore push to our repository.


tomfrenken:
Closed until security concerns are resolved.

PR #1109, opened 2021-03-12
feat: Custom OpenAPI generator

marikaner:
This is the basis for a custom OpenAPI generator. The code is neither polished nor tested (and tests will be failing). But it passes the E2E tests and except for documentation and bugs should be complete from my point of view.

I think it is too early for thorough reviews, but maybe you have general feedback or ideas that you want to have in here.


packages/core/test/test-util/test-services/openapi/test-service/model/complex-test-entity.ts
```diff
@@ -0,0 +1,37 @@
...
+            'item'?: string;
+          } | Record<string, any>[];
+      'uniqueItemsProperty'?: Set<string>;
+      'requiredPropertiesProperty'?: {
+            'optionalProperty'?: string;
+            'requiredProperty': string;
+          } | Record<string, any>;
+      'enumProperty'?: 'one' | 'two';
+      'oneOfProperty'?: SimpleTestEntity | TestEntity;
+      'allOfProperty'?: SimpleTestEntity & {
```
jjtang1985, line 21: this type might be too tolerant? Because it accepts {abc: ''}. I guess you have special reasons for `| Record<string, any>`?

packages/openapi-generator/src/file-serializer/index-file.ts
```diff
@@ -0,0 +1,34 @@
...
+    ...(openApiDocument.schemas.length ? ['model'] : [])
+  ];
+  return codeBlock`
+    ${exportAllFiles(files)}
+  `;
+}
+
+export function modelIndexFile(openApiDocument: OpenApiDocument): string {
+  return codeBlock`
+    ${exportAllFiles(openApiDocument.schemas.map(schema => schema.name))}
```
FrankEssenberger, line 28: Does this model work if there are no schemas - see internal issue on services with no schemas.

packages/openapi-generator/src/generator.ts
```diff
@@ -75,10 +74,23 @@ async function generateSDKSources(
   openApiDocument: OpenApiDocument,
   options: GeneratorOptions
 ): Promise<void> {
+  // TODO: This isn't really "request builder" anymore
   logger.info(`Generating request builder in ${serviceDir}.`);
   // TODO: what about overwrite?
+  if (openApiDocument.schemas.length) {
+    const modelDir = resolve(serviceDir, 'model');
+    await createInterfaceFiles(modelDir, openApiDocument);
```
FrankEssenberger, line 82: we have an inconsistent notion of `model`, `interfaces` and `schemas` which I think are roughly the same. From the old api generator the word `model` is familiar on typescript they are `interfaces` and in the open api spec they are called `schemas`. I think I would use a a folders `schemas` and use the word schema in the code. On the source code level they are of course `interfaces`.

FrankEssenberger: but this is up for discussio.

packages/openapi-generator/src/generator.ts
```diff
@@ -75,10 +74,23 @@ async function generateSDKSources(
...
+    await createFile(
+      modelDir,
+      'index.ts',
+      modelIndexFile(openApiDocument),
+      true
```
FrankEssenberger, line 87: General: Should the overwrite flag not be taken from the options?

packages/openapi-generator/src/generator.ts
```diff
@@ -75,10 +74,23 @@ async function generateSDKSources(
   openApiDocument: OpenApiDocument,
   options: GeneratorOptions
 ): Promise<void> {
+  // TODO: This isn't really "request builder" anymore
   logger.info(`Generating request builder in ${serviceDir}.`);
   // TODO: what about overwrite?
+  if (openApiDocument.schemas.length) {
+    const modelDir = resolve(serviceDir, 'model');
+    await createInterfaceFiles(modelDir, openApiDocument);
+    await createFile(
```
FrankEssenberger, line 83: Perhaps you can create a follow up refactoring task, since I realized a few things which were not touched in this PR but are related and I would like to keep it nice from the start :-):
- Many methods have to many arguments e.g. `createFile()` One way would be that the `modelIndexFile` of anything else return not just a string but an interface containing also the `fileName` here 'index.ts' as well as the path where it is written.
- Also the `generateFromFile` has to many arguments. An the options come often together with the serviceName and the openpi content. Go over the code find common patterns and create perhaps a combined speaking object.

FrankEssenberger: - The input parameter of `generateFromFile` is just called `filePath` perhaps something like `inputFilePaht` is better.
- Also the `generateFromFile` does also a conversion and writing to a standard file I would draw this step out in a separate method.

So just create a follow up with `refactor method signatures and flow....`

packages/openapi-generator/src/file-serializer/api-file.ts
```diff
@@ -0,0 +1,59 @@
...
+      ]),
+    []
+  );
+}
+
+function hasNotSchemaInOperations(operations: OpenApiOperation[]): boolean {
+  return operations.reduce((foundNotSchema, operation) => {
+    const requestBodySchema = operation.requestBody?.schema;
+    return (
+      foundNotSchema || (!!requestBodySchema && hasNotSchema(requestBodySchema))
```
FrankEssenberger, line 39: I find the boolean here hard to graps. Can you move it down to the hasNotSchema. Also I find the [some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) method on the array better in a case if you want to check if at least one element fullfills boolean. What do you think?

packages/openapi-generator/src/model.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+            anyOf: [{ $ref: 'ref3' }]
+          },
+          { $ref: 'ref3' }
+        ]
+      })
+    ).toStrictEqual(['ref1', 'ref2', 'ref3']);
+  });
+
+  it('collects one reference if the schema is a reference', () => {
+    expect(collectRefs({ $ref: 'test' })).toStrictEqual(['test']);
```
jjtang1985, line 121: maybe use a long name as parameter, so it's clear what will be returned. (`#/components/schemas/typeName` v.s. `typeName`)
```
     {
        $ref: '#/components/schemas/typeName'
      }
```

packages/openapi-generator/src/parser/responses.ts
```diff
@@ -0,0 +1,30 @@
+import { OpenAPIV3 } from 'openapi-types';
+import { $Refs } from '@apidevtools/swagger-parser';
+import { OpenApiSchema } from '../openapi-types';
+import { resolveObject } from './refs';
+import { parseMediaType } from './media-type';
+
+export function parseResponses(
+  responses: OpenAPIV3.ResponsesObject | undefined,
+  refs: $Refs
+): OpenApiSchema {
```
jjtang1985, line 10: 👍 nice return type

packages/openapi-generator/src/parser/operation-naming.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+  ] as OperationInfo[]);
+
+  expect(uniqueOperations.map(({ operation }) => operation)).toEqual([
+    expect.objectContaining({
+      operationId: 'nameFromExtension',
+      summary: 'operation3'
+    }),
+    { operationId: 'getX', summary: 'operation1' },
+    { operationId: 'getX', summary: 'operation2' }
```
jjtang1985, line 41: duplicate `operationId`

packages/openapi-generator/src/parser/operation-naming.ts
```diff
@@ -0,0 +1,178 @@
...
+  return [
+    ...renamedOperationsWithExtensions,
+    ...namedOperations,
+    ...renamedUnnamedOperations
+  ];
+}
+
+function validateUniqueExtensions(operations: OperationInfo[]): void {
+  operations.reduce((uniqueOperationNames: string[], operationInfo) => {
+    const operationName = renameOperationWithExtension(operationInfo);
```
jjtang1985, line 68: I would not expect the the operations are modified (renamed) in a `validate` function. Maybe use something like "rename"/"format".

packages/openapi-generator/src/generator.ts
```diff
@@ -254,7 +209,7 @@ async function generateFromFile(
     convertedInputFilePath,
     JSON.stringify(openApiDocument, null, 2)
   );
-  await generateOpenApiService(convertedInputFilePath, serviceDir);
+  // await generateOpenApiService(convertedInputFilePath, serviceDir);
```
jjtang1985, line 212: TODO

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -0,0 +1,85 @@
...
+  bodyOrResponseObject:
+    | OpenAPIV3.RequestBodyObject
+    | OpenAPIV3.ResponseObject
+    | undefined
+): OpenApiSchema | undefined {
+  const allMediaTypes = getMediaTypes(bodyOrResponseObject);
+  if (allMediaTypes.length) {
+    const jsonMediaType = parseApplicationJsonMediaType(bodyOrResponseObject);
+
+    if (!jsonMediaType) {
```
jjtang1985, line 40: I'm afraid log every occurrence might be too much.

jjtang1985 (commented):
In general, looks very promising.
I suggest we try to get this merged as soon as possible with important points, so that we can work on the follow ups and other openapi tickets in parallel.

packages/openapi-generator/src/file-serializer/imports.ts
```diff
@@ -0,0 +1,16 @@
...
+  const relevantImports = imports.filter(({ names }) => names.length);
+  return relevantImports
+    .map(
+      ({ names, moduleIdentifier }) =>
+        codeBlock`import { ${names.join(', ')} } from '${moduleIdentifier}';`
+    )
+    .join('\n');
+}
+
+export interface Import {
```
FrankEssenberger, line 13: Minor: I think this is a follow up later optimizaiton issue, but we could add a `isTypeImport` boolean, which is a switch that the import is serialized to `import` or `type import` this should be used whereever possible to avoid cyclic dependencies.

FrankEssenberger, line 13: I know the Import is simple but a little Doc can not hurt.

packages/openapi-generator/src/file-serializer/index-file.ts
```diff
@@ -0,0 +1,34 @@
+import { codeBlock, kebabCase } from '@sap-cloud-sdk/util';
+import { OpenApiDocument } from '../openapi-types';
+
+/**
+ * @experimental This API is experimental and might change in newer versions. Use with caution.
+ * Generate the index file to expose the SAP Cloud SDK API and models.
+ * @param openApiDocument Parsed service.
+ * @returns The index file contents.
+ */
+export function exportAllFiles(fileNames: string[]): string {
```
FrankEssenberger, line 10: I think this method is only used here, so no need to export it.

packages/openapi-generator/src/file-serializer/operation.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+        }
+      ],
+      response: { type: 'string' },
+      // eslint-disable-next-line no-template-curly-in-string
+      pathTemplate: 'test/${id}/${subId}'
+    };
+    expect(serializeOperation(operation)).toMatchInlineSnapshot(`
+      "getFn: (id: string, subId: string, queryParameters?: {'limit'?: number}) => new OpenApiRequestBuilder<string>(
+        'get',
+        \`test/\${id}/\${subId}\`,
```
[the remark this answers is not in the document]
FrankEssenberger: I was also wondering about this point. My opinion on this is that if you have a simple asertion like for the imports it is better to have it visible in the code, because I do not have to go to the snapshop file. So two  or three lines of content. Here it is long engough to use the mathc snapshot I would say.

packages/openapi-generator/src/model.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+  if (isReferenceObject(schema)) {
+    return [schema.$ref];
+  }
+  return Object.values(schema)
+    .filter(value => typeof value === 'object')
+    .reduce((refs, value) => unique([...refs, ...collectRefs(value)]), []);
+}
+
+export function hasNotSchema(schema: OpenApiSchema): boolean {
```
FrankEssenberger, line 26: Two questions:
- What happens if in a `schema` an other schema is referenced? I think this is possible and we than have to follow the reference and keep on checking if there is the not property.
- If the reference problem is real, we need to consider cyclic cases.

FrankEssenberger: see related comment in the schema.

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -10,3 +22,124 @@ export function isArraySchemaObject(
 ): obj is OpenAPIV3.ArraySchemaObject {
   return obj?.type === 'array';
 }
+
+export function parseSchema(
```
FrankEssenberger, line 26: General concern: References. In the spec it is stated that references are simple in a sense that they do not extend an schema if there are `sibling` properties. This makes it simpleer for us, puh.

https://swagger.io/docs/specification/using-ref/

However, we should add tests for $ref cases which could be chained. This is allowed I think.

packages/openapi-generator/src/generator.ts
FrankEssenberger: After reading a bit I understand the distinction a bit better like model for parisng cause related to openApi lagnuage, schema for evalutaiton cause spec is evaluated and we write than interface. Nevertheless I found it hard to find the where this is done place.

packages/core/test/test-util/test-services/openapi/test-service/model/complex-test-entity.ts
marikaner: Additional properties are only used in objects, that is correct. In this case, the type is correct, because one of the types given is an object.

packages/openapi-generator/src/file-serializer/index-file.ts
marikaner: Currently this is prevented by only ever calling this if there are schemas. Do you think this is enough?

packages/openapi-generator/src/generator.ts
marikaner: I like this, done.

packages/openapi-generator/src/generator.ts
marikaner: Yes, but it is not implemented yet.

packages/openapi-generator/src/generator.ts
marikaner: Makes sense

packages/openapi-generator/src/file-serializer/api-file.ts
marikaner: OMG, that's so much better!

packages/core/src/openapi/types.ts
```diff
@@ -45,3 +45,10 @@ export type UnPromisify<T> = T extends Promise<infer U> ? U : T;
  * @typeparam T Type of an instance created by the constructor.
  */
 export type ConstructorType<T> = new (...args: any[]) => T;
+
+/**
+ * A type to represent to disallow a given type (DisallowedT).
+ * @typeparam T Type to test against the disallowed type.
+ * @typeparam DisallowedT The disallowed type.
+ */
+export type Except<T, DisallowedT> = T extends DisallowedT ? never : T;
```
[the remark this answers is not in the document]
marikaner: Good catch!

packages/core/src/openapi/openapi-request-builder.spec.ts
```diff
@@ -1,59 +1,58 @@
...
-    const response = await requestBuilder.executeRaw(destination);
-
-    expect(response).toEqual(rawResponse);
+  it('executeRaw executes a request without parameters', () => {
+    const requestBuilder = new OpenApiRequestBuilder('get', '/test');
+    requestBuilder.executeRaw(destination);
+    expect(httpClient.executeHttpRequest).toHaveBeenCalledWith(destination, {
+      method: 'get',
+      url: '/test',
+      headers: {},
```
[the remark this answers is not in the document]
marikaner: Yes

packages/core/test/test-util/test-services/openapi/test-service/entity-api.ts
```diff
@@ -4,68 +4,57 @@
...
+          queryParameters
+        }
   ),
-  updateEntityWithPut: (args?: {
-    body?: TestEntity[]
-  }) => new OpenApiRequestBuilder<EntityApi, 'updateEntityWithPut'>(
-    EntityApi,
-    'updateEntityWithPut',
-    args?.body
+  updateEntityWithPut: (body: TestEntity[] | undefined) => new OpenApiRequestBuilder<any>(
```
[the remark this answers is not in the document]
marikaner: That was not my understanding. Can you elaborate why?
If there is a known return type we pass this, if not we use any. The only question where I wasn't sure after our discussion is what if there is 'application/json' and other media types.

packages/openapi-generator/src/model.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+                schema: { not: { type: 'number' } }
+              }
+            ]
+          }
+        ]
+      })
+    ).toBe(true);
+  });
+
+  it('returns true if the root schema is not schema', () => {
```
FrankEssenberger, line 61: Add test if the not is in a referenced schema.

packages/openapi-generator/src/parser/operation-naming.ts
```diff
@@ -0,0 +1,178 @@
...
+  return namedOperations.reduce(
+    (
+      partitionedOperations: {
+        uniqueOperationNames: string[];
+        operationsWithUniqueNames: OperationInfo[];
+        operationsWithDuplicateNames: OperationInfo[];
+      },
+      operationInfo
+    ) => {
+      if (
```
FrankEssenberger, line 120: this reduce is to complicated. Could you try to restructure the code by introducind a type for the 
```
{
        uniqueOperationNames: string[];
        operationsWithUniqueNames: OperationInfo[];
        operationsWithDuplicateNames: OperationInfo[];
}
```

introducing helper methods because I got lost in the `if` `followed by multiple `||`

packages/openapi-generator/src/parser/operation-naming.ts
```diff
@@ -0,0 +1,178 @@
...
+ * Partition operations into a list of unique names that do not need to be changed and a list of operations where the name is duplicate or subject to change and therefore potentially duplicate.
+ * @param namedOperations All operations that have an operationId
+ * @returns an object containing the unique operation names and operations with (potentially) duplicate names
+ */
+function partitionNamedOperationsToUniqueAndDuplicate(
+  namedOperations: OperationInfo[]
+): {
+  uniqueOperationNames: string[];
+  operationsWithUniqueNames: OperationInfo[];
+  operationsWithDuplicateNames: OperationInfo[];
```
FrankEssenberger, line 109: The name `operationsWithDuplicateNames` does not really reflect the doc: `name is duplicate or subject to change and therefore potentially duplicate` So either make a split lie duplicate, potential duplicate ro rename the property.

packages/openapi-generator/src/parser/api.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @param refs List of crossreferences that can occur in the document.
+ * @returns A flat list of parsed APIs.
+ */
+export function parseApis(
+  document: OpenAPIV3.Document,
+  refs: $Refs
+): OpenApiApi[] {
+  const operationsByApis = getOperationsByApis(document);
+
+  return Object.entries(operationsByApis).map(([name, operations]) => ({
```
FrankEssenberger, line 22: could you add the type of `operations`? I think it is already `OperationsInfo`.

packages/openapi-generator/src/parser/operation-info.ts
```diff
@@ -0,0 +1,9 @@
+import { OpenAPIV3 } from 'openapi-types';
+import { OperationNameExtended } from '../extensions';
+import { Method } from '../openapi-types';
+
+export interface OperationInfo {
+  pathPattern: string;
```
FrankEssenberger, line 6: I could not find out what the `pathPattern` is. Could you add a doc with perhaps an example. Also the whole operation-info.ts feels a bit detached. So I would perhaps call it operation-types and include all operation related types?

packages/core/src/openapi/openapi-request-builder.spec.ts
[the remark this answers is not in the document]
marikaner: Done.

packages/openapi-generator/src/parser/api.ts
```diff
@@ -0,0 +1,94 @@
...
+    name,
+    operations: ensureUniqueOperationIds(
+      nameOperations(operations)
+    ).map(({ pathPattern, method }) =>
+      parseOperation(
```
FrankEssenberger, line 27: consider reduing the number of arguments by 1 -> passing the operationsOptions containing patter and method.

packages/core/test/test-util/test-services/openapi/test-service/entity-api.ts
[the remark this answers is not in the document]
marikaner: @FrankEssenberger @jjtang1985 
So we talked about some, but not all cases for the response / request body types:
* only 'application/json' => return type based on schema
* only unknown type, e. g. 'application/xml' => return type any
* multiple types including 'application/json' => ??

packages/openapi-generator/src/parser/responses.ts
marikaner: Right? It makes things sooooo much easier! ;)

packages/openapi-generator/src/parser/operation-naming.spec.ts
marikaner: True, they are renamed in the next step. I know this is a little unusual, but I found it easier to understand when reading the code, than doing this in one step.

packages/openapi-generator/src/parser/operation-info.ts
```diff
@@ -0,0 +1,21 @@
...
+   */
+  pathPattern: string;
+  /**
+   * Original operation as given in the OpenAPI document.
+   */
+  operation: OpenAPIV3.OperationObject & OperationNameExtended;
+  /**
+   * HTTP method for the operation as given in the OpenAPI document.
+   */
+  method: Method;
```
FrankEssenberger, line 20: consider also adding the OpenAPIV3.PathItemObject in the Operation info. The information are passe doften together.

packages/openapi-generator/src/parser/operation-info.ts
jjtang1985: I guess this is just the path like `/some-service/get-somthing/{id}`.
I guess it's named like this because the openapi type call it pattern
```
    interface PathsObject<T extends {} = {}> {
        [pattern: string]: PathItemObject<T> | undefined;
    }
```

packages/openapi-generator/src/parser/media-type.ts
marikaner: True, should we do this as debug only? This is the same for the type parsing btw, I would do the same there.

packages/core/test/test-util/test-services/openapi/test-service/entity-api.ts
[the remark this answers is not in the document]
jjtang1985: The fact is that not all the specs provide enough info about the media type as a contract. 
E.g., the WF does not provide any info for return type.

- I would return `any` in case there is not clear info, because it's really `any`. 
- We should only give a type, once the spec gives clear information and/or users provide some input based on their knowledge.

So, without supporting the user import, the user can still cast the response. As sdk, to be safe, I would only give you the type, when only `application/json` is mentioned. For the rest, return `any`. Maybe even for this PR, just return any.

packages/openapi-generator/src/parser/media-type.ts
marikaner: Done.

packages/openapi-generator/src/file-serializer/imports.ts
marikaner: I take it you would like to do this for the interface types, right?

marikaner: done.

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -1,50 +1,141 @@
...
-  method: Method,
   refs: $Refs
 ): OpenApiOperation {
-  const operation = getOperation(pathItem, method);
-  // TODO: What does the OpenApi generator do in this case?
   const requestBody = parseRequestBody(operation.requestBody, refs);
-  const parameters = parseParameters(operation, refs);
+  const response = parseResponses(operation.responses, refs);
+  const relevantParameters = getRelevantParameters(
+    [...(pathItem.parameters || []), ...(operation.parameters || [])],
```
FrankEssenberger, line 31: What happens with duplicates entires? Can this happen? Is this a problem? Are the parameters the query or path parameters?

FrankEssenberger: forget the question just read the code below and the implementation.

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -1,50 +1,141 @@
...
+    .join('/');
+}
+
+export function parsePathParameters(
+  pathParameters: OpenAPIV3.ParameterObject[],
+  pathPattern: string
+): OpenApiParameter[] {
+  const sortedPathParameters = sortPathParameters(pathParameters, pathPattern);
+  const nameGenerator = new UniqueNameGenerator('', [
+    'body',
```
FrankEssenberger, line 123: I guess these are the names of the arguments in the method so protected due to our generation wrapper, right? However what about other TypeScript protected words like `const` I guess they should also be avoided? So perhaps we can reuse/copy the TS protected words and add the openApi specific ones.

packages/openapi-generator/src/model.ts
marikaner: I think cyclic references should not be an issue, because we never resolve the final schemas (as in components/schema). When there is a schema we just parse the reference string.
The `hasNotSchema` function recursively calls itself for all values in an object, so that other referenced schemas are covered.

packages/openapi-generator/src/file-serializer/schema.spec.ts
```diff
@@ -0,0 +1,173 @@
+import { serializeSchema } from './schema';
```
FrankEssenberger, line 1: I know the readme.ts is not touched but could we have a follow up that we use some shared content in both generators and consider the `COMMON-README-PART.md` as a starting point? We should than also generate this shared file via code and check each link in our E2E test that it is at least still present.

packages/openapi-generator/src/file-serializer/schema.ts
```diff
@@ -0,0 +1,103 @@
...
+  isAnyOfSchema,
+  isNotSchema
+} from '../model';
+
+/**
+ * Serialize a schema.
+ * @param schema Parsed schema to be serialized.
+ * @returns The serialized schema
+ */
+export function serializeSchema(schema: OpenApiSchema): string {
```
FrankEssenberger, line 25: I think the name could be improved. This transforms to string true so it is serializing the object to a format which can be stored in a file. However it is also serializing to TypeScript format so I would like to see this mentioned.

packages/openapi-generator/src/file-serializer/schema.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (isReferenceObject(schema)) {
+    return parseTypeNameFromRef(schema);
+  }
+  if (isArraySchema(schema)) {
```
FrankEssenberger, line 29: Architecture question:
Does it make sense to define a serializer interface which standardises all shema types and related serializing methods. In this method one would than just loop over all check and method pairs (isSomething,serialiazeSomething). Doing so we could potentially serialize letter to different languages?

packages/openapi-generator/src/file-serializer/schema.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  if (isAnyOfSchema(schema)) {
+    return schema.anyOf.map(type => serializeSchema(type)).join(' | ');
+  }
+
+  if (isNotSchema(schema)) {
+    return codeBlock`Except<any, ${serializeSchema(schema.not)}>`;
+  }
+
+  return getType(schema.type);
```
FrankEssenberger, line 57: Should we add a `isType` check here and throw an exception if no match was found to ensure we are not missing a schema type.

packages/openapi-generator/src/file-serializer/api-file.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Serialize an API representation to a string representing the resulting API file.
+ * @param api Represenation of an API.
+ * @returns The serialized API file contents.
+ */
+export function apiFile(api: OpenApiApi): string {
+  const imports = serializeImports(getImports(api));
+  return codeBlock`
+${imports}
+
+export const ${api.name} = {
```
FrankEssenberger, line 17: could you include the TypeScript specific part in a `serializeOperations` methods. It would be nice to see via the POC what are the serialization things we need:
- sechma
- import
- operations
- ...

FrankEssenberger: perhaps we sort these concrete serialization function in a folder called ts-seriliazer just for future things which might come.

packages/openapi-generator/src/file-serializer/interface-file.ts
```diff
@@ -0,0 +1,37 @@
...
+  `;
+}
+
+function getImports(schema: OpenApiSchema): Import[] {
+  const refImports = collectRefs(schema).map(ref => ({
+    names: [parseTypeNameFromRef(ref)],
+    typeOnly: true,
+    moduleIdentifier: `./${parseFileNameFromRef(ref)}`
+  }));
+  const coreImportNames = hasNotSchema(schema) ? ['Except'] : [];
```
FrankEssenberger, line 32: The `Except` is the only core dependency so line 34 makes no sense with no names given in the case of hasNotSchema === false.

packages/openapi-generator/src/file-serializer/interface-file.ts
```diff
@@ -0,0 +1,37 @@
...
+export function interfaceFile({ name, schema }: OpenApiNamedSchema): string {
+  const imports = serializeImports(getImports(schema));
+
+  return codeBlock`
+    ${imports}
+    export type ${name} = ${serializeSchema(schema)};
+  `;
+}
+
+function getImports(schema: OpenApiSchema): Import[] {
```
FrankEssenberger, line 26: The `getImports` method feels a bit hacky. What about that the `serializeSchema` does not only returns a TypeScript sting but also a list all depedencies which are required but this string to work.

packages/openapi-generator/src/parser/api.ts
marikaner: done

packages/openapi-generator/src/parser/operation-info.ts
marikaner: Yes and also because it contains a pattern ;) I saw that @jjtang1985 used `path` in some other places. I am now using `pathPattern` consistently everywhere, but I don't have a strong opinion on this, we can also call it `path`. Do you have preferences?

@FrankEssenberger I updated the documentation in the mean time, if it is better now. I think `operation-types` does not really reflect what was meant.

packages/openapi-generator/src/parser/api.ts
marikaner: Done, this was actually even a bug.

packages/core/src/openapi/openapi-request-builder.spec.ts
```diff
@@ -1,59 +1,58 @@
-import { AxiosRequestConfig, AxiosResponse } from 'axios';
-import { OpenApiRequestBuilder } from './openapi-request-builder';
+jest.mock('../http-client');
```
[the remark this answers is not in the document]
marikaner: Good catch, especially, because I removed it with some refactoring 🙈

packages/openapi-generator/src/model.ts
marikaner: Clarified, fixed through type imports

packages/openapi-generator/src/file-serializer/interface-file.ts
marikaner: in TODOs

packages/openapi-generator/src/file-serializer/interface-file.ts
marikaner: Discussed, rather minor and brings other complexity.

marikaner:
Follow up:
* Ticket for separate serializer with interfaces for potential different languages
* Ticket for schema validation

TODO:
- [ ] getImports for interface file should not add empty names
- [ ] Pathitem parameters on OperationInfo

packages/core/src/http-client/http-client-types.ts
```diff
@@ -7,7 +7,7 @@ export interface DestinationHttpRequestConfig {
   httpsAgent?: http.Agent;
 }
 
-type Method =
+export type Method =
```
marikaner, line 10:
```suggestion
/**
 * HTTP Methods supported by the http-client.
 */
export type Method =
```

package.json
```diff
@@ -38,7 +38,7 @@
     "test:nightly": "yarn nightly-tests test",
     "generate:test-services:v2": "yarn generate v2 && yarn lint:fix && yarn --frozen-lockfile",
     "generate:test-services:v4": "yarn generate v4 && yarn lint:fix && yarn --frozen-lockfile",
-    "generate:test-services:openapi": "yarn generate openapi && yarn lint:fix && yarn --frozen-lockfile",
+    "generate:test-services:openapi": "yarn generate openapi",
```
marikaner, line 41:
```suggestion
    "generate:test-services:openapi": "yarn generate openapi && yarn lint:fix && yarn --frozen-lockfile",
```

packages/openapi-generator/src/parser/operation-info.ts
marikaner: Discussed, that pathPattern is legit.

packages/openapi-generator/src/parser/operation-info.ts
marikaner: Added parameters instead, as discussed.

packages/openapi-generator/src/parser/operation.ts
marikaner: Done.

packages/openapi-generator/src/file-serializer/schema.ts
marikaner: We always serialize to any for unknown types.

packages/openapi-generator/src/file-serializer/schema.ts
marikaner: As discussed, will be considered for a followup.

packages/openapi-generator/src/generator.ts
marikaner: Done.

packages/core/src/openapi/types.ts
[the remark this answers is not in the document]
marikaner: So I found that it works, but I would have to use generics (and drag it through the whole code).
I would suggest to just use any at the moment. What do you think?

packages/core/test/test-util/test-services/openapi/test-service/entity-api.ts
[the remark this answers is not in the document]
marikaner: This is already implemented as you described 😊 In case there is application/json and others I return `ParsedType | any` which is any, but it is nice to have for reference.

packages/openapi-generator/src/parser/operation-naming.ts
marikaner: This is separate. We only get the name and add it to the list of unique names.

packages/openapi-generator/src/generator.ts
marikaner: Done

packages/openapi-generator/src/parser/schema.ts
marikaner: As we never resolve schema references, this should not be a concern. Discussed.

packages/openapi-generator/src/model.spec.ts
marikaner: Same as above.

packages/openapi-generator/src/parser/operation-naming.ts
marikaner: Improved by a lot ;)

packages/openapi-generator/src/file-serializer/schema.spec.ts
marikaner: We already have tickets for this, I improved the descriptions.

packages/openapi-generator/src/file-serializer/schema.ts
marikaner: I created a ticket for this.

packages/openapi-generator/src/file-serializer/api-file.ts
marikaner: I assume this is covered by the serializer interface for which I created a ticket?

marikaner:
@FrankEssenberger @jjtang1985 I incorporated all your feedback, kindly recheck.

packages/openapi-generator/src/parser/operation-info.ts
```diff
@@ -0,0 +1,25 @@
+import { OpenAPIV3 } from 'openapi-types';
+import { OperationNameExtended } from '../extensions';
+import { Method } from '../openapi-types';
+
+/**
+ * Represents an object holding all relevant information for operation parsing.
+ */
```
FrankEssenberger, line 7:
```suggestion
 * The OpenApi definitions looks like:
 "my/path": {
    parameter: {...}
    get: {...}
    post {...}
    ...
 }
 for each operation like GET,POST,... the path e.g.  "my/path" is the same and stored under `pathPattern`. 
The object containing all the operations is call pathItem and may contain parameters which apply to all operations. These parameter are hold by `pathItemParameters`.
 */ 
```

test-resources/openapi-service-specs/test-service.json
```diff
@@ -26,9 +26,7 @@
   "paths": {
     "/entities": {
       "get": {
-        "tags": [
-          "entity"
-        ],
+        "tags": ["entity"],
```
FrankEssenberger, line 29: Since this has been a problem with the old openapi can you also include a real simple service with no schema at all and just a single get to see that also this leads to a compiling example.

marikaner: Sure

marikaner: done.

knowledge-base/adr/0018-openapi-generator.md
```diff
@@ -0,0 +1,45 @@
+# OpenApi Client generator
+
+## Status
+
+proposed
```
FrankEssenberger, line 5: If we merge this now should we go to `accepted`?

packages/core/test/test-util/test-services/openapi/swagger-yaml-service/schema/test-entity.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved.
+ *
+ * This is a generated file powered by the SAP Cloud SDK for JavaScript.
+ */
+    
+    export type TestEntity = {
+      'stringProperty'?: string;
+      'integerProperty'?: number;
+    } | Record<string, any>;
```
FrankEssenberger, line 10: Question: I thought we agreed on not to add this allowed property.

packages/core/test/test-util/test-services/openapi/test-service/schema/complex-test-entity.ts
```diff
@@ -0,0 +1,37 @@
...
+    export type ComplexTestEntity = {
+      'referenceProperty'?: SimpleTestEntity;
+      'arrayProperty'?: {
+            'item'?: string;
+          } | Record<string, any>[];
+      'uniqueItemsProperty'?: Set<string>;
+      'requiredPropertiesProperty'?: {
+            'optionalProperty'?: string;
+            'requiredProperty': string;
+          } | Record<string, any>;
```
FrankEssenberger, line 18: See comment above.

packages/openapi-generator/src/model.ts
```diff
@@ -0,0 +1,146 @@
+import { kebabCase, last, pascalCase, unique } from '@sap-cloud-sdk/util';
```
FrankEssenberger, line 1: this is moere a `schema-util` instead of a model?

packages/openapi-generator/src/openapi-types.ts
```diff
@@ -77,3 +82,85 @@ export interface SchemaMetadata {
   isInnerTypeReferenceType: boolean;
   arrayLevel?: number;
 }
+
+// export type OpenApiSchemaObject =
```
FrankEssenberger, line 86: Dead code

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -0,0 +1,85 @@
...
+      logger.debug(
+        "Could not parse media type, because it is not 'application/json'. Generation will continue with 'any'. This might lead to errors at runtime."
```
FrankEssenberger, line 42: Should we give here a hint that if you want to make a call and it is really not application/json use `executeRaw`? Or should we add this warning at runtime? Before the crash will most likely appear.

packages/openapi-generator/src/parser/operation-info.ts
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Original operation as given in the OpenAPI document.
+   */
+  operation: OpenAPIV3.OperationObject & OperationNameExtended;
+  /**
+   * HTTP method for the operation as given in the OpenAPI document.
+   */
+  method: Method;
+  /**
+   * Parameters from the path items.
```
FrankEssenberger, line 31:
```suggestion
   * Parameters from the path item.
```

packages/core/src/openapi/openapi-request-builder.ts
```diff
@@ -1,42 +1,35 @@
...
    */
   constructor(
-    private apiConstructor: ConstructorType<ApiT>,
-    public fn: FnT,
-    ...args: ParametersType<ApiT, FnT>
-  ) {
-    this.args = args;
-  }
+    public method: Method,
+    private pathPattern: string,
```
FrankEssenberger, line 25: I think you wanted to change this to a list of parameters and the path building is done in the builder, or is this a follow up?

FrankEssenberger (approved):
LGTM I added a few final questions remarks. Note that due to the big number of changes we have to trust our tests and improve perhaps over time.

knowledge-base/adr/0018-openapi-generator.md
marikaner: Good point. Maybe we should also write some more things down. I will do this on a separate PR.

packages/core/test/test-util/test-services/openapi/swagger-yaml-service/schema/test-entity.ts
marikaner: You probably mean the additional property? We agreed to do it for objects. additionalproperty is by default true in those cases.

packages/openapi-generator/src/model.ts
marikaner: yeah you're right. I tried to avoid "util" but it is what it is, why lie about it ;)

jjtang1985 (approved):
Looks promising with follow ups. Would you like to update the changelog in the next PRs?

packages/openapi-generator/src/openapi-types.ts
marikaner: Removed

packages/openapi-generator/src/parser/media-type.ts
marikaner: Actually they can also use execute, we don't parse anything, we just return the response.data. I would rather see that in the API doc of the function. because they have to set the header themselves.

packages/openapi-generator/src/parser/operation-info.ts
marikaner: It is always one 🧐

packages/core/src/openapi/openapi-request-builder.ts
marikaner: No I did it, that is why it is not the path pattern again, not the path template.

packages/core/test/test-util/test-services/openapi/swagger-yaml-service/schema/test-entity.ts
FrankEssenberger: ah ok

PR #1110, opened 2021-03-12
yarn2 migration (after dependabot issue has been resolved)

tomfrenken:
This PR saves the state of the yarn2 migration without the redundant workaround.
It should be merged only after dependabot supports yarn2.


tomfrenken:
Will be closed as it doesn't look like yarn2 will ever be supported with dependabot.

PR #1121, opened 2021-03-16
chore: reduce hard coded paths

FrankEssenberger:
In the sdk metadata PR I introduced a `index.ts` giving the location of the OData service specs. Using this file one can get rid of the hard coded file paths which makes the code more resilient to refactoring in the future. Also due to operating system differences in the file paths it is always better to use `resolve` to build paths for read methods.


jjtang1985 (approved):
LGTM

PR #1127, opened 2021-03-18
fix: write edmx to input dir

FrankEssenberger:
I adjusted the meta data generation so the the output is written next to the edmx input. Since this is only done when you activate the hidden flag `generateSdkMetadata` I think it is ok to write something in the input dir.


packages/generator/src/sdk-metadata/sdk-metadata-types.ts
```diff
@@ -68,13 +68,13 @@ export interface GenerationAndUsage {
   links: Links;
 }
 
-type UrlString = string;
+export type UrlString = string;
```
marikaner, line 71: Maybe I have missed something here, but why?

marikaner (approved):
LGTM, I left a question that is not related to this PR.

packages/generator/src/sdk-metadata/sdk-metadata-types.ts
FrankEssenberger: In the type everything is a string. I wanted to get some means to distinguis a URL, from a time from a multi line code block. @marikaner does the `JSON.stringify()` call the `toString()` on each attribute so that one could use the real type like URL, datatime, codeblock and implement a `toString()`?

PR #1128, opened 2021-03-18
improve pipeline

tomfrenken:
## Changes
- `yarn install --frozen-lockfile` is now only run if changes to the `yarn.lock` were made
- we don't cache the `yarn` cache anymore, instead, we cache `node_modules` and our transpiled typescript (e.g. `**/dist`)
- if the `yarn.lock` didn't change, we only compile the packages that were changed, with `yarn lerna run compile --since origin/<branch>` (e.g. `origin/main`)
- update all standard actions to `@v2` from `@v1`

## Benefits
- Average 1min faster, because `yarn install` becomes redundant in most cases
- Especially faster when multiple build workflows are queued, 5x`yarn install` less/workflow results in a lot less resource usage

## Findings
- The macOS environment is very flaky, which results in random test durations, this is apparently due to GitHub limitations
- `build once` (caching the entire repo after `yarn install`) is good from a theoretical standpoint, but takes a ridiculous amount of time and is therefore not recommended, [example](https://github.com/SAP/cloud-sdk-js/actions/runs/665749539)
- caching the yarn cache as we did so far accelerates `yarn install` a lot, but can be slightly improved because with the current method it is still necessary to run `yarn install` every run on every job

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,8 +20,8 @@ jobs:
           - os: macos-latest
             node-version: 10.x
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
       - uses: actions/cache@v2
@@ -44,8 +44,8 @@ jobs:
   checks:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
         with:
           node-version: '12.x'
       - uses: actions/cache@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -20,19 +20,23 @@ jobs:
...
         id: cache
         with:
-          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          path: |
+            node_modules
+            */*/node_modules
+            **/dist
           key: ${{ matrix.os }}-${{ hashFiles('**/yarn.lock') }}
       - run: yarn install --frozen-lockfile
+        if: steps.cache.outputs.cache-hit != 'true'
```
FrankEssenberger, line 37: minor: I would use `== 'false'`  or whit do you compare to != true?

tomfrenken: `steps.cache.outputs.cache-hit` returns a boolean so yes, false would be the cleaner approach you are correct, thanks for the feedback

FrankEssenberger (approved):
LGTM - I have a few general questions:
- A few times there is a repetition like `if cache do that Ii no chache do that` Is there a way in this workflow to make a `method` out of this or some custon step?
- Also a question: Is the actions/cache well document or should we add some documentation for our use case here? or is it so simple vanilla basic that we do not need that?

tomfrenken:
Regarding the repetition, I didn't find a way to run steps by default (such as the setup/checkout/cache actions) however it would be possible to create our own [custom action](https://docs.github.com/en/actions/creating-actions) that would work as a wrapper I believe 🤔 - that just takes more effort to make

PR #1131, opened 2021-03-19
Fix release build

jjtang1985:
This PR should fix the 3 failing tests on the [release build](https://github.com/SAP/cloud-sdk-js/runs/2146291094?check_suite_focus=true)
See comments below for details.

### DIFF
--- a/packages/generator/src/generator.spec.ts
+++ b/packages/generator/src/generator.spec.ts
@@ -17,7 +17,7 @@ describe('generator', () => {
       const project = await generateProject(
         createOptions({
           inputDir:resolve(oDataServiceSpecs,'v2','API_TEST_SRV'),
-          outputDir: 'common-copy-additional-files',
+          forceOverwrite: true,
           additionalFiles: '../../test-resources/*.md'
         })
       );
@@ -35,7 +35,7 @@ describe('generator', () => {
       const project = await generateProject(
         createOptions({
           inputDir:resolve(oDataServiceSpecs,'v2','API_TEST_SRV'),
-          outputDir: 'common-api-hub-metadata',
+          forceOverwrite: true,
           generateSdkMetadata:true
         })
       );
--- a/packages/generator/src/generator.spec.ts
+++ b/packages/generator/src/generator.spec.ts
@@ -17,6 +17,7 @@ describe('generator', () => {
       const project = await generateProject(
         createOptions({
           inputDir:resolve(oDataServiceSpecs,'v2','API_TEST_SRV'),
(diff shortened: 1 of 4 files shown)


packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
```diff
@@ -39,6 +39,6 @@ exports[`sdk-metadata generates the header content 1`] = `
 Object {
   "name": "API_TEST_SRV",
   "type": "odata",
-  "version": "1.39.0",
+  "version": "1.0.0",
```
jjtang1985, line 42: This snapshot relies on the current version, which will be changed after a release.

packages/generator/src/generator.spec.ts
```diff
@@ -34,6 +35,7 @@ describe('generator', () => {
       const project = await generateProject(
         createOptions({
           inputDir:resolve(oDataServiceSpecs,'v2','API_TEST_SRV'),
+          outputDir: 'common-api-hub-metadata',
```
jjtang1985, line 38: This will fix the `file already exists` error

scripts/generate-test-services.ts
```diff
@@ -33,7 +33,7 @@ const generatorConfigOData = {
   generateTypedocJson: false,
   generatePackageJson: false,
   generateCSN: false,
-  generateSdkMetadata: true,
+  generateSdkMetadata: false,
```
jjtang1985, line 36: This might be a quick fix.

Currently, the `test service version check` (the checks job instead of tests job) failed due to changed value of `version` and `generatedAt`, which should be dynamic.
There are some  other related issues. With option `true`, it will generate some api hub metadata and the files are not committed. However, the `.gitignore` does not contain the generated metadata files, which is confusing.
If the generated metadata has to be there for some reasons, we need to introduce additional logic to support static value for `version` and `generatedAt`.

packages/generator/src/generator.spec.ts
marikaner: Wouldn't it be easier to enable the overwrite flag?

jjtang1985: I'm fine with both. Setting output dir allows you to check multiple generated files.

marikaner (approved):

scripts/generate-test-services.ts
FrankEssenberger: yes, it was a mistake from my side to set this to true.

PR #1133, opened 2021-03-19
fix: pipeline-improvement

tomfrenken:
- Fixes the branch reference (previous one relied on `pull_request` event, which passes the pipeline in a pull request, but fails when pushing to main
- Change `== 'false'` to `!= 'true'`, `steps.outputs.cache.cache-hit` doesn't return `false` when no cache is hit, instead it returns `null`, already reported [bug](https://github.com/actions/cache/issues/533)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
             **/dist
           key: ${{ matrix.os }}-${{ hashFiles('**/yarn.lock') }}
       - run: yarn install --frozen-lockfile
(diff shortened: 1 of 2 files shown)


tomfrenken:
The reason == 'false' doesn't work: https://github.com/actions/cache/issues/533

FrankEssenberger (approved):
Sorry for pointing in the wrong direction with the != true versus false....

tomfrenken:
I don't think you are at fault, `== 'false'` should be the right way to do it, it's just undocumented/bugged that it has a different behaviour.

PR #1135, opened 2021-03-22
Fix: reverted ts keywords conflicts

jjtang1985:
This PR should fix the issues:
- [x] when ts keywords are used as operation name e.g., `export`, no compilation errors should occur.
- [x] when `Api` is used as suffix as api name e.g., `AbcApi`, `abc-api.ts` is generated instead of `abc-api-api.ts`.


test-resources/openapi-service-specs/test-service.json
```diff
@@ -464,6 +464,30 @@
         }
       }
     },
+    "/test-cases/reserved-words": {
+      "get": {
+        "tags": [
+          "test case"
+        ],
+        "operationId": "export",
```
jjtang1985, line 472: test case

test-resources/openapi-service-specs/test-service.json
```diff
@@ -584,7 +608,37 @@
...
         "responses": {
-          "201": {
+          "204": {
+            "description": "no content"
+          }
+        }
+      }
+    },
+    "/test-cases/extension/with-Api-suffix": {
+      "x-sap-cloud-sdk-api-name": "extensionApi",
```
jjtang1985, line 618: test case

jjtang1985:
I'll close this, since the issue should be fixed.

Issue #1102, opened 2021-03-12
One-to-one navigation properties with value null


ja-ne:
Hello cloud-sdk-js,

I have a question concerning one-to-one navigation properties. Depending on the data they can be null although this is not specified in their type definition.

Example (@sap-cloud-sdk/core 1.39.0, @sap/cloud-sdk-op-vdm-business-partner-service 1.24.0):
```
const destination = getDestinationFromSomewhere();
const result = await BusinessPartner
	.requestBuilder()
	.getAll()
	.execute(destination);
```
	
If customer/supplier is not maintained the corresponding one-to-one navigation properties `toCustomer` and `toSupplier` are null which does not match their type definition:

```
/**
* One-to-one navigation property to the [[Customer]] entity.
*/
toCustomer: Customer;
/**
* One-to-one navigation property to the [[Supplier]] entity.
*/
toSupplier: Supplier;
```

This may cause problems when using the result in the `BusinessPartner.builder().fromJson` method because it would only accept `Customer`/`Supplier` or `undefined`.

Could you please have a look at this?

Thanks in advance.

jjtang1985:
Hi @ja-ne ,

I guess this is a bug.
Does it block your current development?
Are you working on some live system or maybe PoC project?

Best reagards,
Junjie

ja-ne:
Hi @jjtang1985,

thanks for the answer. I'm working on a live system.

I use the `.fromJson` method to generically handle many different entities. The entity data come from somewhere else so I do not know which properties and values will be there.

I could write some code to ignore one-to-one navigation properties of value null but this would mean going through all the different entities and their navigation properties and checking if null is a valid value for each.

jjtang1985:
Hi @ja-ne ,

thank you for answering.

@artemkovalyov , FYI.

FrankEssenberger:
Is it ok if we close this issue and inform update you when we piccked the item from the backog?

ja-ne:
Sure, that's ok, thank you.

jjtang1985:
@FrankEssenberger as discussed, we'll keep it opened until we solve the issue.

jjtang1985:
@artemkovalyov the user mentioned [here](https://github.com/SAP/cloud-sdk-js/issues/1129) that this ticket should be treated as high priority.

FrankEssenberger:
PR for the issue: https://github.com/SAP/cloud-sdk-js/pull/1141/

jjtang1985:
[...]

PR #1141, opened 2021-03-23
Fix: allow setting null for one-to-one navigation properties

jjtang1985:
This PR aims to fix the errors below so one can set null value to one-to-one navigation properies.

```
TestEntity.builder().toSingleLink(null); //compilation error
TestEntity.builder().fromJson({ toSingleLink: null }); //compilation error
TestEntity.builder().fromJson({ toSingleLink: null } as any); //runtime error
```

- [x] update changelog with compatibility notes


packages/core/src/odata-common/entity-builder.spec.ts
```diff
@@ -206,9 +206,19 @@ describe('EntityBuilder', () => {
...
+        expectedEntity.toSingleLink!.getCustomFields()
       );
     });
+
+    it('should build an entity from json with one-to-one navigation properties being null', () => {
+      const entity = TestEntity.builder().fromJson({
+        toSingleLink: null
+      });
+      const expectedEntity = TestEntity.builder()
+        .toSingleLink(null)
```
FrankEssenberger, line 219: undefined should also work right? Add a similar test for undefined if this is the case.

packages/core/src/odata-v2/entity-serializer.spec.ts
```diff
@@ -148,4 +148,16 @@ describe('entity-serializer', () => {
...
   });
+
+  it('should serialize null value', () => {
+    const testEntity = TestEntity.builder()
+      .stringProperty(null)
+      .toSingleLink(null)
+      .build();
+
+    expect(serializeEntity(testEntity, TestEntity)).toEqual({
+      StringProperty: null,
```
FrankEssenberger, line 159: same here test for undefined?

packages/core/test/test-util/test-services/v2/test-service/TestEntity.ts
```diff
@@ -126,7 +126,7 @@ export class TestEntity extends EntityV2 implements TestEntityType {
   /**
    * One-to-one navigation property to the [[TestEntitySingleLink]] entity.
    */
-  toSingleLink!: TestEntitySingleLink;
+  toSingleLink?: TestEntitySingleLink | null;
```
FrankEssenberger, line 129: Question: you could also removed the `?` and make it `TestEntitySingleLink | null | undefined`? I wonder how we do it mainly in our code?

FrankEssenberger:
closes: #1102

packages/core/src/odata-common/entity-builder.spec.ts
```diff
@@ -206,9 +206,19 @@ describe('EntityBuilder', () => {
       const entity = TestEntity.builder().fromJson(entityJson);
 
       expect(entity).toStrictEqual(expectedEntity);
-      expect(entity.toSingleLink.getCustomFields()).toEqual(
-        expectedEntity.toSingleLink.getCustomFields()
+      expect(entity.toSingleLink!.getCustomFields()).toEqual(
```
FrankEssenberger, line 209: This could be a breaking change right? If someone accesses the `toSingleLink` it could be silently undefined in the past. Now we get a compile error. I think we should add a compatablity node in the changelog.md.

FrankEssenberger (changes requested):
LGTM, only concern is the potential breaking change but I would go for it and add a competability node.

packages/core/test/test-util/test-services/v2/test-service/TestEntity.ts
marikaner: There is a difference between those two approaches and we use it depending on which makes more sense. If it is optional you can omit the parameter, if it can be undefined you can set it to undefined, but not omit it. Here I think optional is better.

jjtang1985: For `toSingleLink`, typically, there are three cases:
- an object, as the one-to-one link.
- `null`, because one would like to explicitly set the value to `null`.
- `undefined`, if the value is not set. So it's default value.

If I can choose, I would rather remove the `?`, since as a user, I don't see the use case for setting a `undefined`.
The major reason, why I have to add `?` is to make the compiler happy. Otherwise there will be an error `TS2564: Property 'toSingleLink' has no initializer and is not definitely assigned in the constructor.`

Since `undefined` is not a valid use case, I would not add it to the type and only mark it as optional with `?`.
Here is a [style guide](https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-optional-parameters-for-optional-values)

test-packages/type-tests/test/v2/entity-builder.spec.ts
```diff
@@ -19,3 +19,155 @@ builder.fromJson({ stringProperty: '1' });
...
+
+// $ExpectType TestEntity
+builder.fromJson({ stringProperty: null });
+
+// $ExpectError
+builder.fromJson({ stringProperty: 1 });
+
+// <!--------- single link starts --------->
+// $ExpectType TestEntity
+builder.fromJson({
```
jjtang1985, line 31: > Wouldn't this result in an endless loop for null and undefined?

I guess this is the answer?

packages/core/test/test-util/test-services/v2/test-service/TestEntity.ts
marikaner: But given the user sets this value to undefined, what is the difference between undefined and null?
I believe what it is (and should be) is that with undefined if you send it towards a system it is ignored, but with null it is explicitly "deleted". We should also mention that in the changelog.

marikaner (commented):
Looks good, I still have some questions, but I don't see anything to worry about.

packages/core/src/odata-common/entity-builder.ts
```diff
@@ -17,16 +17,17 @@ const logger = createLogger({
...
-      ? FromJsonType<U>[]
-      : JsonT[P]
-    : JsonT[P] extends Record<string, any>
-    ? FromJsonType<JsonT[P]>
-    : JsonT[P];
+    ? (U extends Record<string, any>
+      ? FromJsonType<U>[] // one-to-many navigation properties
+      : JsonT[P]) // collection type
+    : (JsonT[P] extends Record<string, any> | null | undefined
+      ? FromJsonType<JsonT[P]> | null | undefined // one-to-one navigation properties
```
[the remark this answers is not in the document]
jjtang1985: complex type comment and double check null value with complex type.

packages/generator/src/entity/class.ts
```diff
@@ -111,8 +111,8 @@ function navProperty(
   }
   return {
     kind: StructureKind.Property,
-    name: navProp.instancePropertyName + '!',
-    type: entity.className + (navProp.isCollection ? '[]' : ''),
+    name: navProp.instancePropertyName + (navProp.isCollection ? '!' : '?'),
+    type: entity.className + (navProp.isCollection ? '[]' : ' | null'),
```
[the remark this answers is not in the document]
jjtang1985: This is not covered by the documentation, so I think there should not be a constraint to mark one-to-one navigation as mandatory, which might block some user actions.
Here is a hint, where for odata v2, `Multiplicity="0..1"` is the syntax for making a one-to-one relation. I guess then optional is valid for odata v2.

packages/core/src/odata-common/entity-builder.ts
```diff
@@ -17,16 +17,17 @@ const logger = createLogger({
  * It also allows setting unknown properties, which will be treated as custom fields.
  * @typeparam JsonT JSON type of the entity
  */
+// prettier-ignore
```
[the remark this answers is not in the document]
jjtang1985: As discussed, I'll keep the structure but remove the parantheses.

CHANGELOG.md
```diff
@@ -26,7 +27,8 @@
 
 ## Fixed Issues
 
--
+- [odata-generator] Fix a type error of one-to-one navigation properties, so they can set `null` as valid values.
```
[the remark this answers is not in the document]
jjtang1985: Basically, we treat `null` and `undefined` as the same, but I would not advertise `undefined` and the user should use `null`, which is valid for JSONs in the scenarios:
- an input that come from a system (`undefined` is not valid)
- a handwritten JSON

jjtang1985:
@marikaner , I updated the complex info in the `FromJsonType` type and added tests for `undefined`.

FrankEssenberger (approved):
LGTM

PR #1143, opened 2021-03-24
chore: Add an option to generic http client for handling csrf token.

jjtang1985:
The generic http clint now accepts an optional option object. It allows users to delegate the csrf token handling to the SDK.
- As default, no csrf token is handled automatically so odata/CAP are not affected. (`{fetchCsrfToken: false}`)
- For openapi, we can enable it by passing the option `{fetchCsrfToken: true}`.
- Odata can be migrated and the option will be set to `{fetchCsrfToken: true}`.
- The default value of the option will be changed from `false` to `true` after `sdk 2.0`, so user will benefit from this new feature by default.
- This option also provides the odata/openapi request builders additional flexibility so they can support new api like below, the request builders can simply pass the config to the generic http client.
```
TestEntity.requestBuilder()
  .create(
    entity
  )
  .disableCsrfTokenHandling() // this can then pass the the option "{fetchCsrfToken: false}"
  .execute(destination);
```

I would like to merge this PR, execute the automated e2e tests (which is done manually for the time being), then continue the odata migration.

- [x] update changelog


packages/core/src/http-client/http-client-types.ts
```diff
@@ -75,3 +75,7 @@ export interface HttpResponse extends KnownHttpResponseFields {
 export interface HttpReponse extends KnownHttpResponseFields {
   [otherKey: string]: any;
 }
+
+export interface HttpRequestOptions{
```
jjtang1985, line 79: object is better for extension in the future for e.g., cookie/cache

packages/core/src/http-client/http-client.ts
```diff
@@ -122,9 +125,16 @@ export async function buildAxiosRequestConfig<T extends HttpRequestConfig>(
  *
  * @param destination - A destination or a destination name and a JWT.
  * @param requestConfig - Any object representing an HTTP request.
+ * @param options - An [[HttpRequestOptions]] of the http request for configuring e.g., csrf token delegation.
  * @returns A promise resolving to an [[HttpResponse]].
  */
-export const executeHttpRequest = execute(executeWithAxios);
+export function executeHttpRequest<T extends HttpRequestConfig>(
```
jjtang1985, line 131: I added the return type and make it as function for better readability.

packages/core/src/http-client/http-client.ts
```diff
@@ -213,3 +223,53 @@ export function getAxiosConfigWithDefaultsWithoutMethod(): Omit<
     httpsAgent: new https.Agent()
   };
 }
+
+function getDefaultHttpRequestOptions(): HttpRequestOptions {
+  // TODO: 2.0 change to true
```
marikaner, line 228: Can you add this information to the api doc as well?

marikaner (approved):
LGTM, just one minor comment and I have one question:

* This could be done on a separate PR: What I would also see as part of the ticket, is to use this functionality in the core instead of fetching the token there.
* Have you compared to the code from core? We had some issues with on premise systems and proxies, I don't think this is considered here?
* Don't forget to update the documentation afterwards ;)

packages/core/src/http-client/http-client.ts
jjtang1985: I guess you mean the js doc?

jjtang1985:
> Have you compared to the code from core? We had some issues with on premise systems and proxies, I don't think this is considered here?

The on premise proxies are done in some existing functions, which is reused. So I think we are safe. But anyway this should be tested in the automated e2e tests before releasing.

PR #1150, opened 2021-03-26
chore: Generalize naming in the OpenAPI generator

marikaner:
I carved out the duplicated/not yet formatted parts of the operation naming, so that I can reuse them for property naming in the schemas. I think it makes sense to merge this separately.
I also preserved the order of the operations as they come from the specification. This is probably not necessary, but makes a nicer API.


packages/openapi-generator/src/parser/api.ts
```diff
@@ -22,8 +23,13 @@ export function parseApis(
   return Object.entries(operationsByApis).map(
     ([name, operations]: [string, OperationInfo[]]) => ({
       name,
-      operations: ensureUniqueOperationIds(
-        nameOperations(operations)
+      operations: ensureUniqueNames(
+        nameOperations(operations),
```
FrankEssenberger, line 27: Question: What would happen if the operations have names which are protected keywords in typescipt? I did not see a check in the `ensureUniqueNames` and `nameOperations` should one add the check in the `ensureUniqueNames` in the `UniqueNameFinder`?

packages/openapi-generator/src/parser/unique-naming.ts
```diff
@@ -0,0 +1,80 @@
...
+export function ensureUniqueNames<ItemT>(
+  items: ItemT[],
+  getName: (item: ItemT) => string = getNameDefault,
+  setName: (item: ItemT, name: string) => void = setNameDefault,
+  formatName: (name: string) => string = camelCase
+): ItemT[] {
+  const uniqueItems = getCorrectlyNamedItems(items, getName, formatName);
+
+  const nameGenerator = new UniqueNameGenerator(
+    '',
```
FrankEssenberger, line 21: Pass here protected TypeScript words?

FrankEssenberger (changes requested):
Please consider my comment about the key words and also I would know where the `x-sap-...` keys are considered in the naming process cause I did not see it.

FrankEssenberger:
I found it in the `nameOperations` that the `x-sap-cloud-sdk-operation-name` is used. How do we handle a conflict if a `x-sap-cloud-sdk-operation-name` clashed with an other name in the API or with a keyword? Should we throw an execption because the intention of this property was to use this name?

FrankEssenberger (approved):
My concerns are not an issue since as property names everything is possible. However, we discussed to potentially make the renaming an exception where you need a flag to allow this. We will create a BLI to discuss the behavior.

marikaner:
> My concerns are not an issue since as property names everything is possible. However, we discussed to potentially make the renaming an exception where you need a flag to allow this. We will create a BLI to discuss the behavior.

Done

PR #1151, opened 2021-03-26
feat: Use CSRF token in OpenAPI request builder by default

marikaner:
Fetches a CSRF token before making the modifying request.

### DIFF
--- a/packages/core/src/openapi/openapi-request-builder.ts
+++ b/packages/core/src/openapi/openapi-request-builder.ts
@@ -59,6 +59,7 @@ export class OpenApiRequestBuilder<ResponseT = any> {
         params: this.parameters?.queryParameters,
         data: this.parameters?.body
       },
+      // TODO: Remove this in v 2.0, when this becomes true becomes the default
       { fetchCsrfToken }
     );
   }
--- a/packages/core/src/openapi/openapi-request-builder.ts
+++ b/packages/core/src/openapi/openapi-request-builder.ts
@@ -47,13 +47,20 @@ export class OpenApiRequestBuilder<ResponseT = any> {
   async executeRaw(
     destination: Destination | DestinationNameAndJwt
   ): Promise<HttpResponse> {
-    return executeHttpRequest(destination, {
-      method: this.method,
-      url: this.getPath(),
-      headers: this.customHeaders,
-      params: this.parameters?.queryParameters,
-      data: this.parameters?.body
-    });
+    const fetchCsrfToken = ['post', 'put', 'patch', 'delete'].includes(
+      this.method.toLowerCase()
+    );
+    return executeHttpRequest(
+      destination,
+      {
+        method: this.method,
+        url: this.getPath(),
+        headers: this.customHeaders,
+        params: this.parameters?.queryParameters,
+        data: this.parameters?.body
+      },
+      { fetchCsrfToken }
+    );
   }
 
   /**
--- a/test-packages/e2e-tests/openapi.js
+++ b/test-packages/e2e-tests/openapi.js
@@ -38,7 +38,17 @@ async function createApi() {
   entities[0].keyProperty = 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa';
 
   api.register({
-    getAllEntities: (c, req, res) => res.status(200).json(entities),
+    getAllEntities: (c, req, res) => {
+      const hasCsrfFetchHeader = Object.entries(c.request.headers).some(
+        ([key, value]) =>
+          key.toLowerCase() === 'x-csrf-token' &&
+          value.toLowerCase() === 'fetch'
+      );
+      if (hasCsrfFetchHeader) {
+        return res.status(200).set('x-csrf-token', 'e2e-test-token').end();
+      }
+      return res.status(200).json(entities);
+    },
     countEntities: (c, req, res) => res.status(200).json(entities.length),
     getEntityByKey: (c, req, res) => {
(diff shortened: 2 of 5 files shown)


test-resources/openapi-service-specs/test-service.json
```diff
@@ -24,6 +24,7 @@
     }
   ],
   "paths": {
+
```
jjtang1985, line 27: minor: delete?

packages/core/src/openapi/openapi-request-builder.ts
```diff
@@ -47,13 +47,20 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
+      destination,
+      {
+        method: this.method,
+        url: this.getPath(),
+        headers: this.customHeaders,
+        params: this.parameters?.queryParameters,
+        data: this.parameters?.body
+      },
+      { fetchCsrfToken }
+    );
```
jjtang1985, line 63:
```suggestion
    return executeHttpRequest(
      destination,
      {
        method: this.method,
        url: this.getPath(),
        headers: this.customHeaders,
        params: this.parameters?.queryParameters,
        data: this.parameters?.body
      },
      { fetchCsrfToken: true }
    );
```
No, you don't have to consider the method, the `http-client` will ignore get.

test-packages/e2e-tests/openapi.js
```diff
@@ -38,7 +38,17 @@ async function createApi() {
   entities[0].keyProperty = 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa';
 
   api.register({
-    getAllEntities: (c, req, res) => res.status(200).json(entities),
+    getAllEntities: (c, req, res) => {
+      const hasCsrfFetchHeader = Object.entries(c.request.headers).some(
```
jjtang1985, line 42: nice custom code 👍

packages/core/src/openapi/openapi-request-builder.ts
jjtang1985: See [this](https://github.com/SAP/cloud-sdk-js/blob/main/packages/core/src/http-client/http-client.ts#L254) logic, which is used for deciding whether fetching the token is necessary.

jjtang1985 (changes requested):
Look good. Please consider my comments.

test-resources/openapi-service-specs/test-service.json
marikaner, line 27:
```suggestion
```

marikaner: yes

test-packages/e2e-tests/openapi.js
marikaner: 😄

packages/core/src/openapi/openapi-request-builder.ts
marikaner: Hmm, then we might want to reconsider this logic... If I set this to true as a user, I would expect it to be done. However, I understand this might not make a lot of sense...
Independently, I would rather want to do it "right" here, as this is not obvious from outside... I will add a comment though, to remove it in v 2.0.

packages/core/src/openapi/openapi-request-builder.ts
```diff
@@ -47,13 +47,20 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
+    return executeHttpRequest(
+      destination,
+      {
+        method: this.method,
+        url: this.getPath(),
+        headers: this.customHeaders,
+        params: this.parameters?.queryParameters,
+        data: this.parameters?.body
+      },
+      { fetchCsrfToken }
```
marikaner, line 62:
```suggestion
      // TODO: Remove this in v 2.0, when this becomes true becomes the default
      { fetchCsrfToken }
```

packages/core/src/openapi/openapi-request-builder.ts
jjtang1985: You can do "right" here, since there might be some changes based on our discussion.

I'll write my thoughts here.
Currently, the OData request has a contract to handle the CSRF token automatically. However, there are also some hidden "optimisations", where the token request is skipped:
- when method is get (this logic has been moved from odata to generic http client)
- when `x-csrf-token` key exists in the header (this logic remains in the reused function, which is not touched)

In the long term, this option is set to `true` as default value, the user should care about the value and just delegate it to the SDK, which is exactly the same behaviour as the current OData. Otherwise, most of the users have to check the method and pass the right option (`true` of `false`) to the SDK.
So, "delegation" means the intelligent SDK will decide whether it needs to fetch the CSRF token for you. I'll document the option with details, since it's a public api.

I know this might solve 99% of the use cases and the corner case would be "my get request also need a csrf token".
So far, we never receive any feedback for this use case and the current SDK is not capable for this case anyways, I would not consider this case.
I would introduce a new flag `forceFetchCsrfToken` when we receive some requests from the users.

jjtang1985 (approved):

packages/core/src/openapi/openapi-request-builder.ts
marikaner: I think `forceFetchCsrfToken` makes it more complicated than necessary, `fetchCsrfToken` should already be a "force" in my opinion.

Based on our discussion, I think we could also do the following (in version 2.0):
* `fetchCsrfToken` is not set, do the sdk default with optimizations
* `fetchCsrfToken` is false, don't fetch the csrf token ever
* `fetchCsrfToken` is true, fetch the csrf token, always

PR #1152, opened 2021-03-26
Fix: force compilation

jjtang1985:
The compilation fo the `main` branch was skipped withe the option `--since`. Without the compilation, packages like `type-tests`/`integration-tests` use the old version (cache) of the `core`, making the failing pipeline.


marikaner (approved):
👍

PR #1153, opened 2021-03-26
chore: odata requests use csrf token handling from http client

jjtang1985:
Since the generic http client is able to handle the csrf token, I would like remove the csrf token handling from odata and adopt this new feature.

**Backgroud**
For odata, we use the service url `/sap/opu/odata/sap/API_BUSINESS_PARTNER` for fetching the csrf token, which is a "working" service that returns metadata. So we can use it to get a `2xx` response and pick the token from the header.
However, for openapi, there is no such a "safe working service" for every system and it's not guaranteed that a spec contains at least one `get` method.
Therefore, we cannot rely on any "working" get operations.
Actually, for fetching a csrf token, even you get a response with `405`, for example, if the authentication passed, you still can get the token from the header.

### Alternatives of OData
- Still pass the safe url `/sap/opu/odata/sap/API_BUSINESS_PARTNER` to the generic http client so we don't change the behaviour, which needs an additional option of the generic http client.
- Use the same url for fetching the csrf token as the real request. (I choose this one)

- [x] move `csrf-token-header.ts` from `scp-cf` to `http-client`
- [x] add a test in the `auth-flow-spec.ts` for non-get request to cover the csrf token.
- [x] grooming => tests
- [ ] adjust changelog?

what's next
- support "disabling csrf token handling" in the request builder as requested.


packages/core/src/http-client/csrf-token-header.ts
```diff
@@ -58,7 +62,10 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
   return executeHttpRequest(destination, appendSlash(axiosConfig))
     .then(response => response.headers)
     .catch(e1 => {
-      logger.error(
+      if (hasCsrfToken(e1)) {
+        return e1.response.headers;
+      }
+      logger.warn(
```
FrankEssenberger, line 68: I know the message was already there but coud you correct it to something like:
`'Initial try to fetch CSRF token failed - retry without slash at end : ${URL where the request goes to.}'`

packages/core/src/http-client/csrf-token-header.ts
```diff
@@ -58,7 +62,10 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
   return executeHttpRequest(destination, appendSlash(axiosConfig))
     .then(response => response.headers)
     .catch(e1 => {
-      logger.error(
+      if (hasCsrfToken(e1)) {
```
FrankEssenberger, line 65: Perhaps give a comment or a documentation on the `hasCsrfToken` method that the tkoen can be present even if the request has failed.

packages/core/src/http-client/csrf-token-header.ts
```diff
@@ -67,14 +74,29 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
...
           );
+          // todo suggest to disable csrf token handling when the api is implemented
+          return {};
         });
     });
 }
 
+function hasCsrfToken(e1): boolean {
+  return (
+    e1.isAxiosError &&
```
FrankEssenberger, line 94: why the check on `isAxiosError` is relevant?

packages/core/src/odata-common/request/odata-request.spec.ts
```diff
@@ -50,6 +52,7 @@ describe('OData Request', () => {
           ODataDeleteRequestConfig,
           destination
         );
+        request.config.fetchCsrfToken = false;
```
FrankEssenberger, line 55: In your initial description you mention something about disabling the csrf handling. This would mean to add a method like `disableCSRF` on the request builder because the line above is just a work around right?

jjtang1985: I want to make a new PR for a new API of the request builder, which will then control the `request.config.fetchCsrfToken`.

packages/core/src/http-client/csrf-token-header.ts
jjtang1985: Because the type of error is any, only if this is an axios error, you can expect the token is located here:
```
error.response.headers['x-csrf-token']
```
This is not good, because axios is bound. 
Alternatives would be, create an abstraction `parseErrorToResponse(error)` in general to allow different errors, and use `parseAxiosErrorToResponse(error)` for this case, which will be similar to the `executeHttpRequest`, where call `executeWithAxios` internally.

FrankEssenberger: ah ok.

FrankEssenberger (approved):
LGTM

packages/core/src/connectivity/scp-cf/legacy-csrf-token-header.ts
```diff
@@ -1,5 +1,5 @@
 import { ODataRequest, ODataRequestConfig } from '../../odata-common';
-import { buildCsrfHeaders } from './csrf-token-header';
+import { buildCsrfHeaders } from '../../http-client/csrf-token-header';
```
marikaner, line 2:
```suggestion
import { buildCsrfHeaders } from '../../http-client';
```

marikaner, line 2:
```suggestion
import { buildCsrfHeaders } from '../../http-client';
```

packages/core/src/http-client/csrf-token-header.ts
```diff
@@ -67,14 +74,29 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
...
+              e2
+            )
           );
+          // todo suggest to disable csrf token handling when the api is implemented
+          return {};
         });
     });
 }
 
+function hasCsrfToken(e1): boolean {
```
marikaner, line 92:
```suggestion
function hasCsrfToken(error: Error): boolean {
```

packages/core/src/http-client/csrf-token-header.ts
```diff
@@ -58,7 +62,10 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
   return executeHttpRequest(destination, appendSlash(axiosConfig))
     .then(response => response.headers)
     .catch(e1 => {
-      logger.error(
+      if (hasCsrfToken(e1)) {
+        return e1.response.headers;
+      }
+      logger.warn(
         new ErrorWithCause(
           'Initial try to fetch CSRF token failed - retry without slash at ',
```
marikaner, line 70: I know this was here before, but it seems that there is a part of the sentence missing?
```suggestion
          'First attempt to fetch CSRF token failed. Retrying without trailing slash.',
```

packages/core/src/http-client/csrf-token-header.ts
marikaner, line 65: Minor, this was here before, but consider using `error` or `error1` as a variable name.
Also, as we do the same thing twice, execute + return headers + check whether there is csrf token this should be a separate function.

packages/core/src/http-client/csrf-token-header.ts
```diff
@@ -67,14 +74,29 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
...
         .catch(e2 => {
-          throw new ErrorWithCause(
-            'Also second try to fetch SRF token failed - No CSRF token fetched.',
-            e2
+          if (hasCsrfToken(e2)) {
+            return e2.response.headers;
+          }
+          logger.warn(
+            new ErrorWithCause(
+              'Also second try to fetch CSRF token failed - No CSRF token fetched.',
```
marikaner, line 82:
```suggestion
              'Second attempt to fetch CSRF token failed. No CSRF token fetched.',
```

packages/core/src/http-client/csrf-token-header.ts
marikaner: From the conversation below I guess this suggestion is incorrect, the type should probably be `any`?
Please just assign the correct type in the signature.

packages/core/src/http-client/csrf-token-header.ts
marikaner: But assuming it is `any` the result of

```ts
    !!e1.response?.headers?.['x-csrf-token']
```

will be the same, right? And if the error is not an axios error but has the csrf token on it, we want this to be true or don't we?

packages/core/src/http-client/csrf-token-header.ts
```diff
@@ -67,14 +74,29 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
...
+          return {};
         });
     });
 }
 
+function hasCsrfToken(e1): boolean {
+  return (
+    e1.isAxiosError &&
+    e1.response?.headers &&
+    e1.response.headers['x-csrf-token']
```
marikaner, line 96: Independent of the conversation above, you just need one line:
```suggestion
    e1.response?.headers?.['x-csrf-token']
```

packages/core/src/http-client/http-client.ts
```diff
@@ -262,14 +262,12 @@ export function shouldHandleCsrfToken(
   );
 }
 
-export const xCsrfTokenHeaderKey = 'x-csrf-token';
```
marikaner, line 265: Although I think this really wasn't necessary, technically this is a breaking change...
Consider just deprecating it (and turning off the linter for this).
However, I am not really worried that this was used by anyone ;)
So we could also just mention this in the changelog as wrongfully exposed variable or so...

packages/core/src/http-client/csrf-token-header.spec.ts
```diff
@@ -5,10 +5,10 @@ import {
   defaultDestination,
   defaultHost,
   mockHeaderRequest
-} from '../../../test/test-util/request-mocker';
-import { createCreateRequest } from '../../../test/test-util/create-requests';
+} from '../../test/test-util/request-mocker';
+import { createCreateRequest } from '../../test/test-util/create-requests';
```
marikaner, line 9: Minor: Maybe those can be combined?

packages/core/src/http-client/csrf-token-header.spec.ts
```diff
@@ -5,10 +5,10 @@ import {
...
+import { Destination } from '../connectivity/scp-cf/destination';
```
marikaner, line 10:
```suggestion
import { Destination } from '../connectivity/scp-cf';
```

packages/core/src/odata-common/header-builder.spec.ts
```diff
@@ -112,63 +109,6 @@ describe('Header-Builder', () => {
     expect(headers['SAP-Connectivity-SCC-Location_ID']).toBe('Potsdam');
   });
 
-  it('Adds location id headers to the fetch csrf token request if there is a cloudConnectorLocationId in the destination', async () => {
```
marikaner, line 115: Have these tests moved or are they obsolete?

packages/core/src/odata-common/request/odata-request.spec.ts
marikaner: Just FMI: Why are we setting this to false here?

packages/core/src/odata-v2/request-builder/function-import-request-builder.spec.ts
```diff
@@ -169,6 +168,10 @@ describe('FunctionImportRequestBuilder', () => {
   });
 
   it('return undefined or throw in failure case', async () => {
+    nock(defaultHost)
+      .get(`${serviceUrl}/TestFunctionImportNoReturnType?$format=json`)
```
marikaner, line 172: I would prefer it if we would use the `query` option of nock. This makes it in dependent of the order of the parameters and I find it better to read (see the test above).
`.query({ $format: 'json' })`

marikaner: I saw this in the other spec files as well.

marikaner (dismissed):
Looks very good to me, I have some comments on code improvements and a few questions.

packages/core/src/odata-common/header-builder.spec.ts
jjtang1985: I re-wrote the tests.

packages/core/src/http-client/http-client.ts
jjtang1985: This line was added recently and was not released. Worst case is, this PR misses the next release and people may start using it. I would take this risk as breaking changes;)

packages/core/src/http-client/csrf-token-header.ts
jjtang1985: will use `error: any`

packages/core/src/http-client/csrf-token-header.ts
jjtang1985: Will use `error1`.

I know we can make a function for reusing with the parameters:
- a function to control the request url (with/without trailing slash)
- a warning message (first attempt/second attempt)
- a flag to control whether the next attempt is needed

However, as a work around for the known issue of the axios, I would make it more readable instead of more abstract.
Also, when debugging, I found the current structure is easier, because you know it's first attempt or second within a second.

packages/core/src/odata-common/request/odata-request.spec.ts
jjtang1985: See [line 51](https://github.com/SAP/cloud-sdk-js/blob/migrate-odata-csrf/packages/core/src/odata-common/request/odata-request.spec.ts#L51), where it calls `createRequestWithHeaders`.
The `createRequestWithHeaders` contains a mock:
```
jest.spyOn(req, 'headers').mockResolvedValue({});
```
This mock fits the old implementation and skips the csrf token fetching.
Now we migrate to the new version, where the `http-client` holds the logic of fetching the token.
I would like to keep using the same behaviour (do not consider csrf token) and make it more explicit.

packages/core/src/http-client/csrf-token-header.ts
jjtang1985: I personally would stick to "only consider axios" for the reasons below:
- This workaround is purely developed for the axios issue. There might be some conflicts when used with other client.
- > And if the error is not an axios error but has the csrf token on it, we want this to be true or don't we?
  - If the request uses axios, I would not expect this can happen, because it should be an axios error.
  - If the request does not use axios, I would not know where can I find the token. Therefore, I would like to see some real use case and then improve it.

packages/core/src/http-client/csrf-token-header.ts
```diff
@@ -55,26 +60,52 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
...
-      logger.error(
+    .catch(error1 => {
+      if (hasCsrfToken(error1)) {
+        return getResponseHeadersFromAxiosError(error1);
+      }
+      logger.warn(
         new ErrorWithCause(
-          'Initial try to fetch CSRF token failed - retry without slash at ',
-          e1
+          `First attempt to fetch CSRF token failed with the url: ${requestConfigWithTrailingSlash.url}. Retrying without trailing slash.`,
```
marikaner, line 72:
```suggestion
          `First attempt to fetch CSRF token failed with the URL: ${requestConfigWithTrailingSlash.url}. Retrying without trailing slash.`,
```

packages/core/src/http-client/csrf-token-header.ts
```diff
@@ -55,26 +60,52 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
...
-          throw new ErrorWithCause(
-            'Also second try to fetch SRF token failed - No CSRF token fetched.',
-            e2
+        .catch(error2 => {
+          if (hasCsrfToken(error2)) {
+            return getResponseHeadersFromAxiosError(error2);
+          }
+          logger.warn(
+            new ErrorWithCause(
+              `Second attempt to fetch CSRF token failed with the url: ${requestConfigWithOutTrailingSlash.url}. No CSRF token fetched.`,
```
marikaner, line 85:
```suggestion
              `Second attempt to fetch CSRF token failed with the URL: ${requestConfigWithOutTrailingSlash.url}. No CSRF token fetched.`,
```

packages/core/src/http-client/csrf-token-header.ts
```diff
@@ -55,26 +60,52 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
...
+}
+
+function hasCsrfTokenInAxiosError(error: AxiosError): boolean {
+  return error.response?.headers?.['x-csrf-token'];
+}
+
+function getResponseHeadersFromAxiosError(error: AxiosError) {
+  return error.response!.headers;
+}
+
```
marikaner, line 108: This is still very dependent on axios and I don't see why it needs to be.
Consider the code below. It does the same thing but there is no notion of axios.
Also, it makes the flow a little less redundant (accessing the headers and then checking it once).
```ts
function hasCsrfToken(headers: Record<string, any>): boolean {
  return !!headers.['x-csrf-token'];
}

// Non-2xx responses can contain valid csrf tokens in their headers.
function getResponseHeadersFromError(error: any): Record<string, any> {
  return error.response?.headers || {};
}

// Usage
const headers = getResponseHeadersFromError(errorX);
if (hasCsrfToken(headers)) {
  return headers;
}
...
```

PR #1156, opened 2021-03-29
chore: Generate docs for OpenApis clients

FrankEssenberger:
Create also docs for the OpenApi clients. This is part of the GA for the OpenApi client.


packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,71 @@
+import { EOL } from 'os';
```
marikaner, line 1: We should probably replace all occurences of `\n` with this?

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,71 @@
...
+  OpenApiRequestBody
+} from '../openapi-types';
+import { serializeSchema } from './schema';
+
+export function apiDocumentation(
+  api: OpenApiApi,
+  document: OpenApiDocument
+): string {
+  return documentationBlock`
+  Representation of the ${api.name} API.
```
marikaner, line 17: `api.name` already ends with `Api`. So it would always be "Representation of the SomeApi API."
But now thinking of it, I don't have a better idea. If you don't have either feel free to just close this one.

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,71 @@
...
+  OpenApiDocument,
+  OpenApiOperation,
+  OpenApiParameter,
+  OpenApiRequestBody
+} from '../openapi-types';
+import { serializeSchema } from './schema';
+
+export function apiDocumentation(
+  api: OpenApiApi,
+  document: OpenApiDocument
```
marikaner, line 14: Maybe if would suffice to just pass the service name? Also easier for testing.

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,71 @@
...
+): string {
+  return documentationBlock`
+  Representation of the ${api.name} API.
+  This API is part of the ${document.serviceName} service.
+  
+  This API client has been created automatically using the SAP Cloud SDK - do not edit manually.
```
marikaner, line 20: This is very useful, but consider putting it on top of the file. I think we already have something along with the copyright.

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,71 @@
...
+  const signature: string[] = [];
+  if (operation.pathParameters.length > 0) {
+    signature.push(...getSignatureOfPathParameters(operation.pathParameters));
+  }
+  if (operation.requestBody) {
+    signature.push(getSignatureOfBody(operation.requestBody));
+  }
+  if (operation.queryParameters.length > 0) {
+    signature.push(
+      '@param queryParameters Optional object containing the query parameters.'
```
marikaner, line 34: This is not necessarily optional. Maybe just remove the optional here?

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,71 @@
...
+  `;
+}
+
+function getSignatureOfPathParameters(
+  parameters: OpenApiParameter[]
+): string[] {
+  return parameters.map(
+    (parameter, i) =>
+      `@param ${parameter.name} ${
+        parameter.description || `Path parameter number ${i + 1}`
```
marikaner, line 56: Not sure that the number is so helpful here. Maybe instead we could mention the original name of the parameter?

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,71 @@
...
+      }`
+  );
+}
+
+function getSignatureOfBody(body: OpenApiRequestBody): string {
+  return `@param body ${
+    body.required ? 'Object' : 'Optional object'
+  } containing the request body of type '${serializeSchema(body.schema)}'`;
```
marikaner, line 64: The fact that it is optional should be visible in the type. Same applies to the type (if it is an inline type this can be a long string. I would just leave this information out. 

The request body also has a description. I think we should add it to the parsed object and use it here as well. (https://swagger.io/specification/) <-- Request Body

```suggestion
  return `@param body Request body.`;
```

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+function getOperationDescriptionText(operation: OpenApiOperation): string {
```
marikaner, line 67: Operations also have descriptions.

marikaner (commented):
I just took a brief look and it mostly looks good. My main comments are:
* We should parse the description of the objects.
* I didn't see any documentation for the schema, please add it unless I missed it.
* It is not necessary to mention types or "optional" in the docs as this is covered automatically by typedoc.

packages/openapi-generator/src/file-serializer/docs.ts
FrankEssenberger: This depends I would say. In Regex the `\s` or `\n` should work on any os. However, if you build a multi line string or so you should use the `EOL` in my opinion. I would create a follow up ticket for this?

packages/openapi-generator/src/file-serializer/docs.ts
FrankEssenberger: included the descirption

packages/openapi-generator/src/file-serializer/docs.ts
FrankEssenberger: I was considering that but I put the if block in the `getOperationDescriptionText` text.

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,71 @@
...
+import {
+  OpenApiApi,
+  OpenApiDocument,
+  OpenApiOperation,
+  OpenApiParameter,
+  OpenApiRequestBody
+} from '../openapi-types';
+import { serializeSchema } from './schema';
+
+export function apiDocumentation(
```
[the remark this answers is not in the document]
FrankEssenberger: If it is only used from the file serializer objects I would not do it because we should export as little as possible right?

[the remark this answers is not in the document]
FrankEssenberger: but on the other hand the root index only exports the generator so I guess you are right.

packages/util/src/documentation-block.spec.ts
```diff
@@ -0,0 +1,90 @@
+import { EOL } from 'os';
+import { documentationBlock } from './documentation-block';
+
+describe('documentation-block', () => {
+  it('creates empty documentation block', () => {
+    const actual = documentationBlock``;
+    expect(actual).toBe(`/**${EOL} * ${EOL}*/`);
```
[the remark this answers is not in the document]
FrankEssenberger: hm, this is a good question. Do you expect empty or empty documentBlock @marikaner what is your opinion?

packages/openapi-generator/src/file-serializer/docs.ts
marikaner, line 20: What I meant, was that one occurence probably is enough.

packages/util/src/documentation-block.spec.ts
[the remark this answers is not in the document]
marikaner: No strong opinion, but I would also say no content, no comment.

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,83 @@
...
+
+function getOperationDescriptionText(
+  operation: OpenApiOperation,
+  operationResponseType: string
+): string {
+  if (operation.description) {
+    return operation.description;
+  }
+
+  return `Makes a ${operation.method} request to the '${operation.pathPattern}' endpoint and returns a '${operationResponseType}'`;
```
marikaner, line 82: Here again, this does not execute the request yet, we should make this clear here as well.

marikaner: Also, if possible I would really avoid mentioning types in the description (unless helpful), because TypeDoc gives you that for free and if might not be so easy to serialize the types (I haven't checked what happens if the response type is an inline object).

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,83 @@
...
+    (parameter, i) =>
+      `@param ${parameter.name} ${
+        parameter.description ||
+        `Path parameter with the original name ${parameter.originalName}`
+      }`
+  );
+}
+
+function getSignatureOfBody(body: OpenApiRequestBody): string {
+  return `@param body ${body.description || 'Request body'}`;
```
marikaner, line 71: For consistency: We always (as it is not enforced maybe rather most of the time) make a fullstop at the end of docs in our code. We do this in the generated odata code as well. We should consider this here as well - in OData we even have a function that checks whether given descriptions end with `.` and otherwise appends it.

packages/openapi-generator/src/file-serializer/docs.ts
```diff
@@ -0,0 +1,83 @@
...
+import type {
+  OpenApiApi,
+  OpenApiNamedSchema,
+  OpenApiObjectSchemaProperty,
+  OpenApiOperation,
+  OpenApiParameter,
+  OpenApiRequestBody
+} from '../openapi-types';
+
+export function apiDocumentation(api: OpenApiApi, serviceName: string): string {
```
marikaner, line 12: Rather minor: I am not a fan of the additional `doc.ts` file as I don't see much shared functionality/reuse between/of the functions in here. It also contradicts the existing structure a little. I believe that all the functions here will be called from one module each, so maybe we should we just put it there, close to where it is needed?

packages/openapi-generator/src/file-serializer/api-file.ts
```diff
@@ -1,23 +1,30 @@
...
+import { apiDocumentation } from './docs';
 
 /**
  * Serialize an API representation to a string representing the resulting API file.
  * @param api Represenation of an API.
+ * @param document Representation of the Document used the generated documentation.
  * @returns The serialized API file contents.
  */
-export function apiFile(api: OpenApiApi): string {
+export function apiFile(api: OpenApiApi, document?: OpenApiDocument): string {
```
marikaner, line 23: I thought I already commented this... 🤔

1. Why is this optional?
2. Maybe we should just pass the service name instead of the whole document?

packages/util/src/documentation-block.ts
```diff
@@ -0,0 +1,86 @@
...
+  return strings.map(str => str.replace(/\n/g, `${EOL} * `));
+}
+
+function removeWhiteSpaceAroundNewLine(strings: string[]): string[] {
+  return strings.map(str => str.replace(/ *\n */g, `${EOL}`));
+}
+
+function replaceAllWhiteSpacesBySingleOne(strings: string[]): string[] {
+  return strings.map(str => str.replace(/ +/g, ' '));
+}
```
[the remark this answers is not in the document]
FrankEssenberger: When I was doing the tests I had problems with uninteded white spaces. I will try to remove this.

packages/util/src/documentation-block.ts
```diff
@@ -0,0 +1,86 @@
...
+  ...args: string[]
+): string {
+  if (isCodeBlockEmpty(strings.raw, args)) {
+    return '';
+  }
+
+  let adjustedStrings = strings.raw as string[];
+  adjustedStrings = removeSpaceNewLineStartAndEnd(adjustedStrings);
+  adjustedStrings = replaceAllWhiteSpacesBySingleOne(adjustedStrings);
+  adjustedStrings = removeWhiteSpaceAroundNewLine(adjustedStrings);
```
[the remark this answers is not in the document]
FrankEssenberger: The first method is removed and since it zipps the strings are always first and last potentially empty if you start with an empty argument. But I wanted to keep the arguments more or less as they are because they are not used for formatting.

packages/util/src/documentation-block.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    expect(actual).toBe('');
+
+    const empty = `
+       
+    `;
+    actual = documentationBlock`   
+    ${empty}
+    
+     `;
+    expect(actual).toBe('');
```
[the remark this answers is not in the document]
FrankEssenberger: ok

packages/openapi-generator/src/file-serializer/docs.ts
FrankEssenberger: I agree totally. However, It is nice if you have the tests for the docs in one file because it is easier to see them all. I will refactor.

packages/openapi-generator/src/file-serializer/docs.ts
FrankEssenberger: yes I mixed this up. I forgot about the execute() method. Have a look at the new text.

packages/openapi-generator/src/file-serializer/docs.ts
FrankEssenberger: So I would keep the docs.spec.ts?

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -117,16 +125,19 @@ function parseObjectSchema(
...
+  schema: OpenAPIV3.NonArraySchemaObject,
+  refs: SwaggerParser.$Refs
 ): OpenApiObjectSchemaProperty[] {
   return Object.entries(schema.properties || {}).reduce(
     (props, [propName, propSchema]) => [
       ...props,
       {
-        schema: parseSchema(propSchema),
+        schema: parseSchema(propSchema, refs),
+        description: resolveObject(propSchema, refs)?.description,
```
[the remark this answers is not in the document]
FrankEssenberger: We go for option 1 since the description on the referenced schema does not really make senese to put there.

jjtang1985 (approved):
LGTM.

packages/openapi-generator/src/file-serializer/api-file.ts
```diff
@@ -11,17 +12,20 @@ import { Import, serializeImports } from './imports';
...
+    ${api.operations
+      .map(operation => serializeOperation(operation))
+      .join(',\n')}
+  };`;
 
-export const ${api.name} = {
-  ${api.operations.map(operation => serializeOperation(operation)).join(',\n')}
-};
-`;
+  return [imports, apiDoc, apiContent].join(EOL);
```
marikaner, line 28: You wanted to adjust this on a separate PR, right?

marikaner (approved):
LGTM!

PR #1159, opened 2021-03-30
chore: Add OpenAPI naming ADR

marikaner:
Documentation of our decisions regarding OpenAPI naming.

### DIFF
--- a/knowledge-base/adr/0019-openapi-generator-naming.md
+++ b/knowledge-base/adr/0019-openapi-generator-naming.md
@@ -56,14 +56,14 @@ Examples:
 
 Operations are named based on their given `operationId`.
 If no `operationId` is given the name is parsed from the method and the path under which the operation occured.
-Example: 
+Example:
 
 - path: `entity/{id}`
 - method: `get`
 - resulting name: `getEntityById`
-Users can set specific operation names using the `x-sap-cloud-sdk-operation-name` extension.
-This extension can be set on the operation only.
-Names will be transformed to camel case, duplicate names will cause generation to fail.
+  Users can set specific operation names using the `x-sap-cloud-sdk-operation-name` extension.
+  This extension can be set on the operation only.
+  Names will be transformed to camel case, duplicate names will cause generation to fail.
 
 ### Property and query parameter naming
 
@@ -71,6 +71,8 @@ Keep the names as provided per the specification.
 No camel case renaming is done to avoid the original name corner cases discussed above.
 
 ## Consequences
+
+As the OpenAPI generator has not been released yet, we can safely change the naming strategies.
 
 ### API naming
 
--- a/knowledge-base/adr/0019-openapi-generator-naming.md
+++ b/knowledge-base/adr/0019-openapi-generator-naming.md
@@ -1,0 +1,86 @@
+# Naming of generated APIs, operations, properties and query parameters (OpenAPI generator)
+
+## Status
+
+decided
+
+## Context
+
+The OpenAPI generator has to generate names for APIs, their operations and schemas.
+We initially used the file name to generate an API name, but found that for bigger APIs this is would result in very complex APIs, therefore we need a more sensible default.
+In addition API owners whish to be able to customize names of APIs and operations.
+To follow naming conventions in JavaScript/TypeScript we planned on formatting API names as pascal case, operation names as camel case and property and parameter names as camel case.
+For properties and parameters we would have to resolve the original names to use them in the requests.
+We found that for most cases this is straight forward, but there are corner cases that require a different/more complex architecture.
+
+### Issues with properties and parameters
+
+Given the following two types and their original properties, if those were used in combination, e. g. `allOf`, there is no possibility to set both values.
+
+```ts
+type A {
+  someProperty: string; // original name: 'some_property`
+}
+
+type A {
+  someProperty: string; // original name: 'some-property`
+}
+```
+
+This could be solved by making names unique throughout the whole service.
+However this again would lead to less convenient names, e.g. `name`, `name1`, `name2`.
+
+## Decisions
+
+### API naming
+
+APIs are named based on the first tag, if there are any.
+If there is no tag, a "default" tag is used.
+To allow customizing of APIs, we introduced the `x-sap-cloud-sdk-api-name` extension, which takes precedence over the default name generation.
+This extension can be set on on multiple levels:
+
+- operation, causing this operation to be treated as part of the given API.
+- path, causing all operations under this path to be treated as part of the given API.
+- document root, causing all operations in the document to be treated as part of the given API.
+
+All names are transformed to pascal case (`SomeApi`).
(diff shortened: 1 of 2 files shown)


knowledge-base/adr/0019-openapi-generator-naming.md
```diff
@@ -0,0 +1,86 @@
...
+Operations are named based on their given `operationId`.
+If no `operationId` is given the name is parsed from the method and the path under which the operation occured.
+Users can set specific operation names using the `x-sap-cloud-sdk-operation-name` extension.
+This extension can be set on the operation only.
+Names will be transformed to camel case, duplicate names will cause generation to fail.
+
+### Property and query parameter naming
+
+Keep the names as provided per the specification.
+No camel case renaming is done.
```
FrankEssenberger, line 65:
```suggestion
No camel case renaming is done to avoid the original name corner cases discussed above.
```

knowledge-base/adr/0019-openapi-generator-naming.md
```diff
@@ -0,0 +1,86 @@
...
+- `my` => `MyApi`
+- `my-api-api` => `MyApiApi`
+
+### Operation naming
+
+Operations are named based on their given `operationId`.
+If no `operationId` is given the name is parsed from the method and the path under which the operation occured.
+Users can set specific operation names using the `x-sap-cloud-sdk-operation-name` extension.
+This extension can be set on the operation only.
+Names will be transformed to camel case, duplicate names will cause generation to fail.
```
FrankEssenberger, line 60: I wonder if the reformatting to camel case for the API and Operation is good? What are the values for the `operationId` and `tag` in general? Would perhaps be more consistent to also keep the names?

FrankEssenberger: Also mention the `strictNames` generator flag somewhere in the document.

knowledge-base/adr/0019-openapi-generator-naming.md
```diff
@@ -0,0 +1,86 @@
...
+- path, causing all operations under this path to be treated as part of the given API.
+- document root, causing all operations in the document to be treated as part of the given API.
+
+All names are transformed to pascal case (`SomeApi`).
+API names are formed by removing "api" suffix (case independant and only if available) and adding "Api".
+Examples:
+
+- `my-api` => `MyApi`
+- `my` => `MyApi`
+- `my-api-api` => `MyApiApi`
```
FrankEssenberger, line 52: this contradicts the: 
> by` removing "api" suffix (case independant and only if available)

knowledge-base/adr/0019-openapi-generator-naming.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Operation naming
+
+Operations are named based on their given `operationId`.
+If no `operationId` is given the name is parsed from the method and the path under which the operation occured.
```
FrankEssenberger, line 57: Could you perhaps given an example here of path, method etc and what will be the name generated from that.

FrankEssenberger (changes requested):
Please consider my comments.

knowledge-base/adr/0019-openapi-generator-naming.md
```diff
@@ -0,0 +1,86 @@
...
+If there is no tag, a "default" tag is used.
+To allow customizing of APIs, we introduced the `x-sap-cloud-sdk-api-name` extension, which takes precedence over the default name generation.
+This extension can be set on on multiple levels:
+
+- operation, causing this operation to be treated as part of the given API.
+- path, causing all operations under this path to be treated as part of the given API.
+- document root, causing all operations in the document to be treated as part of the given API.
+
+All names are transformed to pascal case (`SomeApi`).
+API names are formed by removing "api" suffix (case independant and only if available) and adding "Api".
```
marikaner, line 47:
```suggestion
API names are formed by appending an "Api" suffix.
If the original name ends with an "api" suffix (case independent), the original suffix is removed prior to appending "Api".
```

knowledge-base/adr/0019-openapi-generator-naming.md
marikaner: But also 
> and adding "Api"

I improved the text a little, I hope it is more on point now?

knowledge-base/adr/0019-openapi-generator-naming.md
marikaner: As this is just the ADR, I would like to leave those kinds of details for the documentation portal. Would that be ok for you?

knowledge-base/adr/0019-openapi-generator-naming.md
marikaner:
```suggestion
If no `operationId` is given the name is parsed from the method and the path under which the operation occured.
Example: 

- path: `entity/{id}`
- method: `get`
- resulting name: `getEntityById`
```

knowledge-base/adr/0019-openapi-generator-naming.md
marikaner: > I wonder if the reformatting to camel case for the API and Operation is good? What are the values for the `operationId` and `tag` in general? Would perhaps be more consistent to also keep the names?

For APIs we have to do this, as tags can be anything, e. g. also containing spaces and APIs are technically just variables and they have to have a valid name. Also, there is nothing comparable to operationId, that would allow extracting technical names for our purposes. Is is a little different with the extensions. Here we could consider using them as is and throwing an error if they are incorrect. 
For operations it is a little different, those could technically keep their names (even if there are spaces) and the `operationId`s are intended to be used for technical names.
However, this might lead to APIs that on the first glance look a little weird, e.g.:

```ts
const SomeApi = {
  a: () => 1,
  'a-2': () => 2
};

SomeApi.a();

// meh
SomeApi['a-2']();
```

The second invocation here is a little unusual. This totally works, but for me personally this is a bit awkward for functions.
The same can occur when accessing properties in objects (e.g. request bodies):

```ts
const body = {
  b: 1,
  'b-2': 2
}

const someVal1 = body.b; 
const someVal2 = body['b-2']; 
```

But in my personal opinion this is less awkward.

Still, this is a valid question, the best way to decide this would be to ask users...

knowledge-base/adr/0019-openapi-generator-naming.md
```diff
@@ -0,0 +1,86 @@
...
+This extension can be set on the operation only.
+Names will be transformed to camel case, duplicate names will cause generation to fail.
+
+### Property and query parameter naming
+
+Keep the names as provided per the specification.
+No camel case renaming is done.
+
+## Consequences
+
```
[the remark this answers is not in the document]
marikaner: I added the second sentence, but not the first one as this is the `Consequences` section and the sentence speaks about the `Decision` secsion.

marikaner:
@FrankEssenberger Please take another look 😊

marikaner:
Can be merged as discussed.

PR #1160, opened 2021-03-30
fix: Fix schema parsing to use unique pascal case names for schemas

marikaner:
This adds the parsed name of a reference in the schema, so that sub-sequential parsing is not necessary during serialization. This ensures correct naming.


test-resources/openapi-service-specs/test-service.json
```diff
@@ -661,6 +679,9 @@
       "OtherSimpleTestEntity": {
         "$ref": "#/components/schemas/SimpleTestEntity"
       },
+      "simple-test-entity«WITH_Symbols»": {
```
jjtang1985, line 682: May be add a test with a schema "simple<test<entity<WITH_Symbols<" so the generated type will be the same as this instance, although it seems to be tested in unit test.

packages/openapi-generator/src/schema-util.spec.ts
```diff
@@ -100,21 +115,22 @@ describe('collectRefs', () => {
               {
                 name: 'refProperty',
                 required: false,
-                schema: { $ref: 'ref1' }
+                schema: { $ref: 'ref1', schemaName: 'Ref1' }
               }
             ]
           },
-          { $ref: 'ref2' },
+          { $ref: 'ref2', schemaName: 'Ref2' },
```
jjtang1985, line 122: Please correct me. I guess `$ref` to `schemaName` should be many-to-one relation? If so, could we make a test to see the result?
I know this case is added in the `test-service.json`, but also want to see the behaviour of this function as unit test.

packages/openapi-generator/src/schema-util.ts
```diff
@@ -117,16 +147,24 @@ export function isNotSchema(obj: any): obj is OpenApiNotSchema {
   return obj?.not;
 }
 
-// TODO: handle duplicates after pascal case => my-type, MyType
```
jjtang1985, line 120: I guess I must miss the test, could you please share the reference of the test?

packages/openapi-generator/src/schema-util.ts
```diff
@@ -8,24 +8,54 @@ import {
...
  */
-export function collectRefs(schema: OpenApiSchema | undefined): string[] {
+export function collectRefs(
+  schema: OpenApiSchema | undefined
+): OpenApiReferenceSchema[] {
+  return getUniqueRefs(collectAllRefs(schema));
+}
+
+/**
+ * Collect all reference schemas within a schema.
```
jjtang1985, line 27: The descriptions of these two method are the same, and the function names do not really tell the differences as they are both plural. 
Maybe add a keyword `unique` to the description and the name of the `collectRefs`?

packages/openapi-generator/src/schema-util.ts
```diff
@@ -8,24 +8,54 @@ import {
...
+/**
+ * Reduce a list of reference schemas to a list of unique reference schemas, based on $ref.
+ * @param refs List of reference schemas.
+ * @returns List of unique reference schemas.
+ */
+export function getUniqueRefs(
+  refs: OpenApiReferenceSchema[]
+): OpenApiReferenceSchema[] {
+  return refs.reduce((uniqueRefs: OpenApiReferenceSchema[], collectedRef) => {
+    if (!uniqueRefs.some(ref => ref.$ref === collectedRef.$ref)) {
```
jjtang1985, line 54: I guess for the example below, one schema will be filtered in this function?
```ts
[
  { $ref: 'Person', schemaName: 'Student'},
  { $ref: 'Person', schemaName: 'Teacher'}
]
```

test-resources/openapi-service-specs/test-service.json
marikaner: Done.

packages/openapi-generator/src/schema-util.spec.ts
marikaner: Maybe I am misunderstanding, but each ref is resolved to a unique name (this is the whole point of this PR ;) ).
So no, it is a one-to-one relation.

packages/openapi-generator/src/schema-util.ts
marikaner: Added this to the docs

packages/openapi-generator/src/parser/document.ts
```diff
@@ -25,8 +27,11 @@ export async function parseOpenApiDocument(
   const document = (await parse(clonedContent)) as OpenAPIV3.Document;
   const refs = await resolve(document);
   const originalFileName = removeFileExtension(basename(filePath));
+  const schemaInfo = parseSchemaInfo(document);
+  const schemaRefMapping = parseSchemaRefMapping(schemaInfo);
```
FrankEssenberger, line 31: I know you will not like this idea, because it it not 100% JavaScripty, but I am posting it anyway. So we could have a class like `ParsedSchemaManager`. This class knows all the parsed schemas and the mapping with names, refs  or other information we would need later potentially like some description in the docs. Then we pass this `ParsedSchemaManager` instance around. And call the `getOrginalName` `getReference` on this object later on.

packages/openapi-generator/src/schema-util.spec.ts
jjtang1985: Let's have a call.

test-resources/openapi-service-specs/test-service.json
```diff
@@ -661,6 +679,12 @@
...
+        "$ref": "#/components/schemas/SimpleTestEntity"
+      },
+      "simple-test-entity<WITH_Symbols>": {
```
jjtang1985, line 685: example

packages/openapi-generator/src/schema-util.spec.ts
jjtang1985: done

test-resources/openapi-service-specs/test-service.json
jjtang1985:
```
{ 
  $ref: '#/components/schemas/simple-test-entity«WITH_Symbols»', // by definition, this is unique
  schemaName: 'simpleTestEntityWithSymbols1' // this is generated to be unique
}
```

jjtang1985 (approved):
LGTM.
Please go ahead and make follow-ups when necessary, so we can run the generator test again.

packages/openapi-generator/src/parser/document.ts
marikaner: Actually, I expected a suggestion for a class from you, but not for this ;)
Let's discuss options in person, I would do this on a separate PR anyways.

PR #1162, opened 2021-03-30
Implement jwt token

tomfrenken:
- Implements `jwtBearerToken` and `jwtBearerTokenGrant` see api notes [here](https://docs.cloudfoundry.org/api/uaa/version/74.23.0/index.html#jwt-bearer-token-grant)
- Changes default authorization workflow from `userApprovedServiceToken` to `jwtBearerToken`
- Deprecates `userApprovedServiceToken`, `refreshTokenGrant` and `userTokenGrant`


packages/core/src/connectivity/scp-cf/xsuaa-service.ts
```diff
@@ -93,18 +94,19 @@ export function userTokenGrant(
 }
 
 /**
+ * @deprecated Since v1.XX.X Use [[jwtBearerTokenGrant]] instead.
```
FrankEssenberger, line 97: I think the next version will be 1.41.0

packages/core/src/connectivity/scp-cf/xsuaa-service.ts
```diff
@@ -292,5 +330,6 @@ function getInstanceCircuitBreaker(breaker?: any | undefined): any {
 const grantTypeMapper = {
   user_token: 'User token',
   refresh_token: 'Refresh token',
-  client_credentials: 'Client credentials'
+  client_credentials: 'Client credentials',
+  'urn:ietf:params:oauth:grant-type:jwt-bearer': 'JWT token'
```
FrankEssenberger, line 334: can you use the value of the GrantType enum here as key? @marikaner is this possible, then you have the string only in one place?

FrankEssenberger (approved):
Looks very good. All the discussed points are there, please consider only the two small comments I gave.

PR #1173, opened 2021-04-06
chore: User EOL

FrankEssenberger:
We use for all generation `\n` using a constant with explanation called `unixEOL`.
For the web related text as in batch we use a constant `webEOL`.
In this PR we also reenabled the windows tests and I excluded the `Readme.md` files from the linter because the git diff script was not happy because changes were induced in the Readme.md file if you generate it but do not run the linter.


jjtang1985 (approved):
LGTM.
By the way, the line break was the reason why windows tests were disabled from the matrix.

FrankEssenberger:
The usage analytics test fails on windows. the findConfigPath() fails

test-packages/integration-tests/test/analytics.spec.ts
```diff
@@ -11,14 +11,9 @@ describe('usage analytics', () => {
...
-    __dirname
-      .split(sep)
-      .reduce((prev, curr) =>
-        prev.endsWith('integration-tests') ? prev : `${prev}${sep}${curr}`
-      ),
-    'sap-cloud-sdk-analytics.json'
-  );
+  // The cwd is different on windows and mac hence the path is set here like this
+  // The findConfigPath calls resolve() which relies on the cwd.
+  const rootConfigPath = resolve(process.cwd(), 'sap-cloud-sdk-analytics.json');
```
jjtang1985, line 16: > The usage analytics test fails on windows. the findConfigPath() fails the resolve() does returns the root dir on windows but the integration path on mac.
process.cwd() is different on mac and windows.

But `resolve(process.cwd(), 'sap-cloud-sdk-analytics.json');` returns the same correct path for both windows and Mac?

jjtang1985 (approved):
Nice to have windows tests.

marikaner (approved):
I probably wouldn't have replaced it everywhere (e. g. logs), but LGTM.

FrankEssenberger:
@marikaner let's discuss on monday where I missed it.

PR #1174, opened 2021-04-07
fix: dist cache overlap

tomfrenken:
## Changes:
- Removes `dist` from the cache in the pipeline
- Transpiles entire code instead of difference to main

## Reason:
If a previous `build` workflow had the same `yarn.lock` and `os`, its `dist` folder got used in the following `build` workflow, which leads to issues if the code had any bugs and worked on different packages than the following `build` workflow (e.g. triggered by a different PR), since only the difference to `origin/main` got transpiled, which can (and did!) lead to an unwanted cache overlap.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,6 @@ jobs:
           path: |
             node_modules
             */*/node_modules
-            **/dist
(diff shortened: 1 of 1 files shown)


marikaner (approved):
LGTM

PR #1175, opened 2021-04-07
chore: use common copyright header

FrankEssenberger:
While working on the docs I realized that the copy right header was there twice and only in one case with the year beeing auto filled. This PR uses a central place and updates the clients to have the new year.

For the review only consider these three files: 

image

### DIFF
--- a/packages/util/src/constants.ts
+++ b/packages/util/src/constants.ts
@@ -1,1 +1,13 @@
+import { codeBlock } from './code-block';
+
 export const VALUE_IS_UNDEFINED = 'VALUE_IS_UNDEFINED';
+
+export function getCopyrightHeader(): string {
+  return codeBlock`
+/*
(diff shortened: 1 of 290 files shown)


packages/util/src/constants.ts
```diff
@@ -1 +1,13 @@
+import { codeBlock } from './code-block';
+
 export const VALUE_IS_UNDEFINED = 'VALUE_IS_UNDEFINED';
+
+export function getCopyrightHeader(): string {
```
marikaner, line 5: We have a ticket in the backlog to have a generator-util at some point in time. Until then we have duplicated all code an added a todo. I would suggest to do the same here, because otherwise we will have to make breaking changes to move this to the new package.

marikaner (changes requested):
That totally makes sense, but please check my one comment.

marikaner (approved):
As discussed, please duplicate. Other than that looks good to me!

PR #1182, opened 2021-04-08
feat: custom axios options

jjtang1985:
- Support custom axios options for all request builders (OData + OpenAPI).
- There is a list of disallowed keys that cannot be customised.


packages/core/src/http-client/http-request-config.ts
```diff
@@ -0,0 +1,40 @@
...
+      )}`
+    );
+  }
+  return pickWithoutKeys(disallowedKeys, customRequestConfigs) as Record<
+    string,
+    string
+  >;
+}
+
+const defaultDisallowedKeysOfCustomRequestConfigs = [
```
marikaner, line 34: Minor: This is a very long name, maybe just `defaultDisallowedKeys`?

packages/util/src/object.ts
```diff
@@ -98,6 +98,25 @@ export const pick = <T>(keys: string[], obj: T): Partial<T> => {
...
 
+/**
+ * Create a shallow copy of the given object, that does not contain the given keys.
+ * Non existing keys in the source object are ignored.
+ *
+ * @param keys - properties to be selected
+ * @param obj - object from which the values are taken
+ * @returns an object with the selected keys and corresponding values.
+ */
+export const pickWithoutKeys = <T>(keys: string[], obj: T): Partial<T> => {
```
marikaner, line 109: Consider `exclude` as a name. "Pick without keys" sounds like there might be no keys in the result (so an array?) ;)
This also correlates with the naming of the according TypeScript types.

packages/core/src/http-client/http-request-config.ts
```diff
@@ -0,0 +1,40 @@
...
+  messageContext: 'http-request-config'
+});
+
+/**
+ * Filter disallowed keys from a given custom request configs object.
+ * @param customRequestConfigs a given custom request configs object to be filtered
+ * @param disallowedKeys A list of keys that are not allowed to be customized.
+ * @returns Filtered custom request configs object.
+ */
+export function filterCustomRequestConfigs(
```
marikaner, line 14: I think this is one configuration with multiple configuration options, not multiple configurations, therefore I would like to refrain from using the word "configs".
Please cross check everywhere (also docs).

packages/core/src/odata-common/request-builder/request-builder-base.ts
```diff
@@ -119,6 +119,17 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
+  /**
+   * Add custom request configs to the request.
+   *
+   * @param requestConfigs - Key-value pairs denoting additional custom request configs to be set in the request.
+   * @returns The request builder itself, to facilitate method chaining.
+   */
+  addCustomRequestConfigs(requestConfigs: Record<string, string>): this {
```
marikaner, line 128: To add to my other comment on "configs":

I have no strong opinion here, but we might want to expose the full word: `addCustomRequestConfiguration`.
Although, this is quite long... but consider that we also expose `addCustomQueryParameters` (and not params)

packages/core/src/openapi/openapi-request-builder.ts
```diff
@@ -38,6 +44,19 @@ export class OpenApiRequestBuilder<ResponseT = any> {
     return this;
   }
 
+  /**
+   * Add custom request configs to the request.
```
marikaner, line 48: This also applies to OData: As a user I would like to have a little bit more context of what this is going to be used for. Will it just be added to the request as is?

packages/util/src/object.spec.ts
```diff
@@ -146,6 +147,23 @@ describe('pickValueIgnoreCase', () => {
   });
 });
 
+describe('pickWithoutKeys', () => {
```
marikaner, line 150: If you rename, please don't forget to rename here, too.

marikaner (changes requested):
Code-wise this looks very good to me. I only left comments regarding naming.

.github/workflows/build.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
+        os: [macos-latest, ubuntu-latest]
```
marikaner, line 17: Is it not working for windows anymore?

packages/core/src/odata-common/request-builder/request-builder-base.ts
```diff
@@ -119,6 +119,20 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
+  /**
+   * Add custom request configuration to the request. Typically, this is used when specifying response type for downloading files.
+   * If the custom request configuration contains keys in this list [[defaultDisallowedKeys]], they will be removed.
+   *
```
marikaner, line 125:
```suggestion
   * Add a custom request configuration to the request. Typically, this is used when specifying a response type for downloading files.
   * If the custom request configuration contains [[defaultDisallowedKeys | disallowed keys]], those will be ignored.
```

packages/core/src/openapi/openapi-request-builder.ts
```diff
@@ -38,6 +44,22 @@ export class OpenApiRequestBuilder<ResponseT = any> {
     return this;
   }
 
+  /**
+   * Add custom request configuration to the request. Typically, this is used when specifying response type for downloading files.
+   * If the custom request configuration contains keys in this list [[defaultDisallowedKeys]], they will be removed.
+   *
+   * @param requestConfiguration - Key-value pairs denoting additional custom request configuration to be set in the request.
```
marikaner, line 51:
```suggestion
   * @param requestConfiguration - Key-value pairs denoting additional custom request configuration options to be set in the request.
```

marikaner:
```suggestion
   * @param requestConfiguration - Key-value pairs denoting additional custom request configuration options to be set in the request.
```

packages/core/src/odata-common/request-builder/request-builder-base.ts
```diff
@@ -119,6 +119,19 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
+  /**
+   * Add a custom request configuration to the request. Typically, this is used when specifying a response type for downloading files.
+   * If the custom request configuration contains [[defaultDisallowedKeys | disallowed keys]], those will be ignored.
+   * @param requestConfiguration - Key-value pairs denoting additional custom request configuration to be set in the request.
```
marikaner, line 125:
```suggestion
   * @param requestConfiguration - Key-value pairs denoting additional custom request configuration options to be set in the request.
```

.github/workflows/build.yml
jjtang1985: The [error](https://github.com/SAP/cloud-sdk-js/runs/2298458310?check_suite_focus=true) itself only happens with windows and says
```
Error: src/http-client/http-request-config.ts(1,24): error TS2305: Module '"../../../../node_modules/@sap-cloud-sdk/util/dist"' has no exported member 'exclude'.
```
The `exclude` was added later after the initial execution and was expected to be in the `node_modules`.
However, the cache was hit and the old `node_modules` was used, which did not know `exclude`.

Therefore, it's a combination of our cache and windows (that has different behaviour compared to other environment).

jjtang1985: By the way, the main build also failed only with windows: https://github.com/SAP/cloud-sdk-js/runs/2303633553?check_suite_focus=true. Retry did not work and it failed in the [PR](https://github.com/SAP/cloud-sdk-js/pull/1184) already before merging.
I guess we are not ready for introducing windows now, and it should be investigated in follow ups.

marikaner: That's ok for me, rerun did solve it on main now, but this should not be necessary (multiple times).

marikaner (approved):
LGTM

PR #1184, opened 2021-04-08
fix: Add imports for parameters in api files

marikaner:
Fix api file generation for files containing reference schemas in parameters


packages/openapi-generator/src/file-serializer/api-file.spec.ts
```diff
@@ -84,19 +106,25 @@ describe('apiFile', () => {
     };
     expect(apiFile(api, 'MyserviceName')).toMatchInlineSnapshot(`
       "import { OpenApiRequestBuilder } from '@sap-cloud-sdk/core';
-      import type { RefType, ResponseType } from './schema';
+      import type { PathParameterType, QueryParameterType, RefType, ResponseType } from './schema';
```
jjtang1985, line 109: unused import `PathParameterType`?

marikaner: Good catch

jjtang1985 (approved):

Issue #1185, opened 2021-04-08
Creating SalesQuoteItem for SalesQuote on C4C fails for technical user where no x-csrf-token is needed and C4C response with 500


nazarkulyk:
Hi,
we generated standard c4c odata client for our C4C Cloud System and i can create some entities, but not all. As example we get error with creating SalesQuoteItem because cloud-sdk tries to get CSRF Token and C4C responds with 500.

Here our code:

```typescript

   const SalesQuoteCollection = SalesQuote.requestBuilder();
   const SalesQuoteItemBuilder = SalesQuoteItem.builder();
   const SalesQuoteItemCollection = SalesQuoteItem.requestBuilder();

    const existingSalesQuote = await SalesQuoteCollection.getByKey(parentObjectId)
      .select(SalesQuote.OBJECT_ID)
      .execute(destination);

   console.log('existingSalesQuote:', existingSalesQuote);

    const newSalesQuoteItem = SalesQuoteItemBuilder.productInternalId(C4C_SALES_QUOTE_ITEM_PRODUCT_INTERNAL_ID)
      .processingTypeCode(C4C_SALES_QUOTE_ITEM_PROCESSING_TYPE_CODE)
      .zSalesQuoteFormattedTextKut('')
      .build();

    const query = SalesQuoteItemCollection.create(newSalesQuoteItem).asChildOf(existingSalesQuote, SalesQuote.SALES_QUOTE_ITEM);

    const result = await query.execute(destination).catch((err) => {
      console.log('Error:', err.message);
      console.log('Cause:', err.cause?.message);
      console.log('Root cause:', err.rootCause?.message);
      throw new Error(err);
    });
```

As result, we can get SalesQuote (check log below with ID:00163EAB5C8B1EEBA4F1F487709ACDDF), but creating SalesQuoteItem crash and event don't do any https request.
I tried to use Proxy to verify it:

```
[2021-04-08T13:52:07.987Z] WARN     (http-client): The custom headers are provided with the keys: authorization,content-type,accept. These keys will overwrite the headers created by the SDK.
[2021-04-08T13:52:07.987Z] WARN     (authorization-header): Found custom authorization headers. The given destination also provides authorization headers. This might be unintended. The custom headers from the request config will be used.
[2021-04-08T13:52:08.834Z] WARN     (response-data-accessor): The given reponse data has the format for collections instead of the standard OData v2 format for single results.
[...]

Issue #606, opened 2020-10-20
Disable csrf fetch request


meakgoz2:
**Is your feature request related to a problem? Please describe.**
I am using the sdk as an OData client for my CAP project. I do not have CSRF protection in my backend as it is not meant for users to do browsing. I assume there is no need for csrf protection when two apps talks to each other?

So it would be very nice if there was a configuration to disable the token fetching request that is sent before each request. When there are so many requests, this is even creating problems with destination service. Issuing lots of request creates a connection problem.

I checked the documentation and tutorials, unfortunately could not find any clue to do this.
Hope this description is good enoug.

artemkovalyov:
Hi @meakgoz2,

Thanks for reporting your issue. Yes, we normally do a pre-flight to fetch CSRF, most of the services implement this. We'll take a look and come back with ideas about suppressing it or other possible solutions.

Best,
Artem

artemkovalyov:
Hi @meakgoz2 ,

There's no workaround out of the box. We added this to the backlog to provide a flexibility of suppression of CSRF token retrieval. 

Can you please let us know how urgent or blocking this fix is for you? What's the state of your service? Having many competing items we have to prioritize this.

We'll keep you posted on the progress.

Thanks,
Artem

meakgoz2:
Hi Artem,

Although it would be nice to have feature, this can be low prio as it does not block us (rather slowing our service a bit).
Thanks for keeping me up to date.

Regards,
Emin.

marikaner:
@meakgoz2, please note that this repository has been moved and the link to this issue has changed.

jjtang1985:
Hi all,
now you are able to skip fetching the csrf token by using our new feature.
Here is an example:
```ts
 BusinessPartner.requestBuilder()
   .update(businessPartner)
   .skipCsrfTokenFetching();
 ```
Please note, the feature is currently available in the `canary` version and will be released in the next minor version.

I'll close this ticket.
Please feel free to re-open, if you have questions.

jjtang1985:
Hi all,
Please use the new version [1.42.0](https://github.com/SAP/cloud-sdk-js/releases), which contains this feature.

PR #1189, opened 2021-04-09
feat: support disabling csrf token request

jjtang1985:
Related to: https://github.com/SAP/cloud-sdk-js/issues/1185 and https://github.com/SAP/cloud-sdk-js/issues/606

- [x] Users can disable the csrf token request in all request builders, which fits some systems that do not need the token.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,8 @@
 
 ## New Functionality
 
-- Support custom axios options for all request builders.
-- Support disabling csrf token request as an option for all request builders.
+- [core] Support custom axios options for all request builders.
+- [core] Support disabling csrf token request as an option for all request builders.
 
 ## Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 ## New Functionality
 
 - Support custom axios options for all request builders.
+- Support disabling csrf token request as an option for all request builders.
 
 ## Improvements
 
--- a/packages/core/src/odata-common/request-builder/request-builder-base.ts
+++ b/packages/core/src/odata-common/request-builder/request-builder-base.ts
@@ -132,6 +132,16 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
+  /**
+   * Replace the default value of fetching csrf token configuration with a given value.
+   * @param fetchCsrfToken The new value of the configuration.
+   * @returns The request builder itself, to facilitate method chaining.
+   */
+  fetchCsrfToken(fetchCsrfToken: boolean): this {
+    this.requestConfig.fetchCsrfToken = fetchCsrfToken;
+    return this;
+  }
+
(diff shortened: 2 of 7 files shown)


packages/core/src/odata-common/request-builder/request-builder-base.ts
```diff
@@ -132,6 +132,16 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
+  /**
+   * Replace the default value of fetching csrf token configuration with a given value.
```
FrankEssenberger, line 136: Perhaps we change the API to `disableCsrfTokenFechting()` without the argument. Most likely a `enavleCsrfTokenFetching` method is not needed because why would you disable it in the frist place. But I am also fine with this API, just mention the default value in the docs.

FrankEssenberger: @marikaner what is your opinion on this?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 ## New Functionality
 
 - Support custom axios options for all request builders.
+- Support disabling csrf token request as an option for all request builders.
```
FrankEssenberger, line 22: I found this [line](https://github.com/SAP/cloud-sdk-js/blob/f12daa9abfd9373240d7251609f9d41e8ff6eb90/packages/core/src/http-client/http-client.ts#L237) here. Is the change from false to true as the default value potentially a change in behavior that should be mentioned in the competability notes?

FrankEssenberger (approved):
LGTM I would most likely do a disable API instead of passing the boolean but this is a matter of taste. Perhaps you can also get merikas opinion.

packages/core/src/odata-common/request-builder/request-builder-base.ts
jjtang1985: Yes, you are right. For vdm request builder level, the default behaviour should always be "enable the csrf token fetching". Therefore, having one API called disable*** is fine. I'll adjust it.

CHANGELOG.md
marikaner, line 22: BTW: here and above the "tag is missing" [odata-common] mostly, e. g.

packages/core/src/odata-common/request-builder/request-builder-base.ts
marikaner: +1

CHANGELOG.md
jjtang1985: I did not put any tags because these are something like [common] for both odata and openapi. Maybe I put [core] for both?

jjtang1985:
@FrankEssenberger @marikaner , please check the new api name.

FrankEssenberger (approved):
like it better with the skip.

marikaner (approved):
AWESOME.

PR #1203, opened 2021-04-14
chore: add generation steps

FrankEssenberger:
Added a bit more detailed generation steps and the field `serviceStatus` in the metadata.

### DIFF
--- a/packages/generator/src/sdk-metadata/generation-and-usage.ts
+++ b/packages/generator/src/sdk-metadata/generation-and-usage.ts
@@ -7,7 +7,6 @@ export async function getGenerationAndUsage(
   service: VdmServiceMetadata
 ): Promise<GenerationAndUsage> {
   return {
-    successfulGenerationVerified: true,
     genericUsage: await getGenericUsage(),
     apiSpecificUsage: await getApiSpecificUsage(service),
     links: getLinks(),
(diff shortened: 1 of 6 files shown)


packages/generator/src/sdk-metadata/sdk-metadata-types.ts
```diff
@@ -17,6 +17,14 @@ export interface SdkMetadataHeader {
...
 export interface Client {
   language: 'java' | 'javascript';
+  /**
+   * Status of the service.
+   * `certified` means we have a pregeneratedLibrary.
+   * `verified` means we have generated and transpiled the client
+   * `unknown` means we have no information.
+   * @memberof Client
+   */
+  serviceStatus: 'certified' | 'verified' | 'unknown';
```
jjtang1985, line 27: so unknown is the default status and might also applies to some services with known issues with generation.

packages/generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -42,5 +41,10 @@ export const linkGenerationDocumentaion =
   'https://sap.github.io/cloud-sdk/docs/js/features/odata/generate-odata-client';
 
 export function getGenerationDocumentation(): string {
-  return `Please follow the documentation ${linkGenerationDocumentaion}`;
+  return `<ul>
```
jjtang1985, line 44: I assume you tested the html tags.

jjtang1985 (approved):
LGTM

packages/generator/src/sdk-metadata/sdk-metadata-types.ts
FrankEssenberger: yes unknown will be the value for the generic response when we have no data. Potentially we add even a `error` status in case we have a problem. Currently the exception are not handled to make this possible but we will do this later.

packages/generator/src/sdk-metadata/generation-and-usage.ts
FrankEssenberger: Yes I copied it to the w3school online editor. Seemed to work.

PR #1208, opened 2021-04-15
fix: use head method for csrf requests

jjtang1985:
I switch to `head` method for Csrf token request instead of `get` method for performance improvement.


FrankEssenberger (approved):
LGTM - I have to remeber this head method when I want to check the existence of content.

PR #1219, opened 2021-04-19
Metadata adjustments

FrankEssenberger:
Adjust the metadata after the UI discussion.

### DIFF
--- a/packages/generator/internal.d.ts
+++ b/packages/generator/internal.d.ts
@@ -1,3 +1,3 @@
 // eslint-disable-next-line import/no-internal-modules
 export * from './dist/sdk-metadata';
-// # sourceMappingURL=v4.d.ts.map
+// # sourceMappingURL=internal.d.ts.map
--- a/packages/generator/internal.d.ts
+++ b/packages/generator/internal.d.ts
@@ -1,0 +1,3 @@
+// eslint-disable-next-line import/no-internal-modules
+export * from './dist/sdk-metadata';
+// # sourceMappingURL=v4.d.ts.map
--- a/packages/generator/internal.js
+++ b/packages/generator/internal.js
@@ -1,0 +1,15 @@
+'use strict';
+function __export(m) {
+  for (const p in m) {
+    if (!exports.hasOwnProperty(p)) {
+      exports[p] = m[p];
+    }
+  }
+}
+Object.defineProperty(exports, '__esModule', { value: true });
+/**
+ * @packageDocumentation
+ * @experimental The v4 module is related to OData v4 and is still experimental. Use with caution.
+ */
+__export(require('./dist/sdk-metadata'));
+// # sourceMappingURL=v4.js.map
--- a/packages/generator/internal.js
+++ b/packages/generator/internal.js
@@ -9,7 +9,7 @@ function __export(m) {
 Object.defineProperty(exports, '__esModule', { value: true });
 /**
  * @packageDocumentation
- * @experimental The v4 module is related to OData v4 and is still experimental. Use with caution.
+ * @experimental The internal module is related to sdk-metadata types which are used only internally.
  */
 __export(require('./dist/sdk-metadata'));
-// # sourceMappingURL=v4.js.map
+// # sourceMappingURL=internal.js.map
--- a/packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
+++ b/packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
@@ -16,6 +16,6 @@ Object {
 
 const resultPromise = BusinessPartner.requestBuilder().getAll().top(5).execute({ destinationName:'myDestinationName' });
 ",
-  "text": "Find below a generic example on execute request with the SDK.",
+  "text": "Find a generic example on how to execute requests with the SAP Cloud SDK below.",
 }
 `;
--- a/packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
+++ b/packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
@@ -1,15 +1,21 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`generation-and-usage creates api specific usage for entity 1`] = `
-"import { DummyClass } from '@sap/dummy-package';
+Object {
+  "instructions": "import { DummyClass } from '@sap/dummy-package';
 
 const resultPromise = DummyClass.requestBuilder().getAll().top(5).execute({ destinationName:'myDestinationName' });
-"
+",
+  "text": "Find below a usage example for this API.",
+}
 `;
 
 exports[`generation-and-usage creates generic usage example 1`] = `
-"import { BusinessPartner } from '@sap/cloud-sdk-vdm-business-partner-service';
+Object {
+  "instructions": "import { BusinessPartner } from '@sap/cloud-sdk-vdm-business-partner-service';
 
 const resultPromise = BusinessPartner.requestBuilder().getAll().top(5).execute({ destinationName:'myDestinationName' });
-"
+",
+  "text": "Find below a generic example on execute request with the SDK.",
+}
 `;
--- a/packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
+++ b/packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
@@ -6,7 +6,7 @@ Object {
 
 const resultPromise = DummyClass.requestBuilder().getAll().top(5).execute({ destinationName:'myDestinationName' });
 ",
-  "text": "Find below a usage example for this API.",
+  "text": "Find a usage example for this API below.",
 }
 `;
 
--- a/packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
+++ b/packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
@@ -6,7 +6,7 @@ Object {
 
 const resultPromise = DummyClass.requestBuilder().getAll().top(5).execute({ destinationName:'myDestinationName' });
 ",
-  "text": "Find a usage example for this API below.",
+  "text": "Find below a usage example for this API.",
 }
 `;
 
@@ -16,6 +16,6 @@ Object {
 
 const resultPromise = BusinessPartner.requestBuilder().getAll().top(5).execute({ destinationName:'myDestinationName' });
 ",
-  "text": "Find a generic example on how to execute requests with the SAP Cloud SDK below.",
+  "text": "Find below a generic example on execute request with the SDK.",
 }
 `;
--- a/packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
+++ b/packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
@@ -19,7 +19,7 @@ const resultPromise = TestEntity.requestBuilder().getAll().top(5).execute({ dest
 <li>Install the generator <code>npm install -g @sap-cloud-sdk/generator</code></li>
 <li>Execute the generator <code>generate-odata-client --inputDir path/to/service-spec --outputDir path/to/</code></li>
 </ul>
-The steps above will generate a TypeScript client. For additional options like transpiling to JavaScript etc. visit our <a href=\\"https://sap.github.io/cloud-sdk/docs/js/features/odata/generate-odata-client\\">documentation.</a>",
+The steps above will generate a basic TypeScript client. For additional options like transpiling to JavaScript etc. visit our <a href=\\"https://sap.github.io/cloud-sdk/docs/js/features/odata/generate-odata-client\\">documentation.</a>",
       "text": "Follow the following generation steps to generate the client.",
     },
     "generatorRepositoryLink": "https://www.npmjs.com/package/@sap-cloud-sdk/generator",
@@ -71,7 +71,7 @@ const resultPromise = BusinessPartner.requestBuilder().getAll().top(5).execute({
   },
   "serviceStatus": Object {
     "status": "certified",
-    "statusText": "A pre-generated API client exists.",
+    "statusText": "The SDK team has generated a API client and published it under npm.",
   },
 }
 `;
@@ -147,7 +147,7 @@ Object {
 
 exports[`sdk-metadata generates the header content 1`] = `
 Object {
-  "introText": "The SAP Cloud SDK is a versatile set of libraries and tools for developers to build applications in a cloud-native way and host them on SAP Business Technology Platform or other runtimes.",
+  "introText": "The SAP Cloud SDK is a versatile set of libraries and tools for developers to build applications in a cloud-native way and host them on the SAP Business Technology Platform or other runtimes.",
   "name": "API_TEST_SRV",
   "type": "odata",
   "version": "1.0.0",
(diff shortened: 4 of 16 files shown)


packages/generator/internal.d.ts
```diff
@@ -0,0 +1,3 @@
+// eslint-disable-next-line import/no-internal-modules
```
marikaner, line 1: Is this file committed on purpose?

packages/generator/internal.d.ts
```diff
@@ -0,0 +1,3 @@
...
+export * from './dist/sdk-metadata';
+// # sourceMappingURL=v4.d.ts.map
```
marikaner, line 3: Ah, probably yes, but please adjust the v4 hint here ;)

packages/generator/internal.js
```diff
@@ -0,0 +1,15 @@
...
+    }
+  }
+}
+Object.defineProperty(exports, '__esModule', { value: true });
+/**
+ * @packageDocumentation
+ * @experimental The v4 module is related to OData v4 and is still experimental. Use with caution.
+ */
+__export(require('./dist/sdk-metadata'));
+// # sourceMappingURL=v4.js.map
```
marikaner, line 15: Also, please remove reference to odata v4

packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
```diff
@@ -1,15 +1,21 @@
...
 
 exports[`generation-and-usage creates generic usage example 1`] = `
-"import { BusinessPartner } from '@sap/cloud-sdk-vdm-business-partner-service';
+Object {
+  "instructions": "import { BusinessPartner } from '@sap/cloud-sdk-vdm-business-partner-service';
 
 const resultPromise = BusinessPartner.requestBuilder().getAll().top(5).execute({ destinationName:'myDestinationName' });
-"
+",
+  "text": "Find below a generic example on execute request with the SDK.",
```
marikaner, line 19:
```suggestion
  "text": "Find a generic example on how to execute requests with the SAP Cloud SDK below.",
```

packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
```diff
@@ -1,15 +1,21 @@
...
 
 exports[`generation-and-usage creates api specific usage for entity 1`] = `
-"import { DummyClass } from '@sap/dummy-package';
+Object {
+  "instructions": "import { DummyClass } from '@sap/dummy-package';
 
 const resultPromise = DummyClass.requestBuilder().getAll().top(5).execute({ destinationName:'myDestinationName' });
-"
+",
+  "text": "Find below a usage example for this API.",
```
marikaner, line 9:
```suggestion
  "text": "Find a usage example for this API below.",
```

packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
```diff
@@ -4,72 +4,137 @@ exports[`pregenerated-lib returns description of the service 1`] = `"SAP Cloud S
 
 exports[`sdk-metadata [E2E] generates the JS metadata content for services with pregenerated lib 1`] = `
 Object {
+  "emergencyObject": undefined,
```
marikaner, line 7: ALARM! 🚨 ?

packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
```diff
@@ -4,72 +4,137 @@ exports[`pregenerated-lib returns description of the service 1`] = `"SAP Cloud S
...
+    },
+    "generationSteps": Object {
+      "instructions": "<ul>
 <li>Download the API specification to your local computer.</li>
 <li>Install the generator <code>npm install -g @sap-cloud-sdk/generator</code></li>
 <li>Execute the generator <code>generate-odata-client --inputDir path/to/service-spec --outputDir path/to/</code></li>
 </ul>
-For additional information please consider our <a href=\\"https://sap.github.io/cloud-sdk/docs/js/features/odata/generate-odata-client\\">documentation</a>",
-    "genericUsage": "import { BusinessPartner } from '@sap/cloud-sdk-vdm-business-partner-service';
+The steps above will generate a basic TypeScript client. For additional options like transpiling to JavaScript etc. visit our <a href=\\"https://sap.github.io/cloud-sdk/docs/js/features/odata/generate-odata-client\\">documentation.</a>",
```
marikaner, line 22:
```suggestion
The steps above will generate a TypeScript client. For additional options like transpiling to JavaScript etc. visit our <a href=\\"https://sap.github.io/cloud-sdk/docs/js/features/odata/generate-odata-client\\">documentation.</a>",
```

packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
```diff
@@ -4,72 +4,137 @@ exports[`pregenerated-lib returns description of the service 1`] = `"SAP Cloud S
...
+      "text": "Execute the following npm command to install the pregenerated client.",
+    },
     "repository": "npm",
     "repositoryLink": "https://www.npmjs.com/package/@sap/cloud-sdk-vdm-business-partner-service",
     "version": "1.0.0",
   },
-  "serviceStatus": "certified",
+  "serviceStatus": Object {
+    "status": "certified",
+    "statusText": "The SDK team has generated a API client and published it under npm.",
```
marikaner, line 74: I find the notion of mentioning the team here a little awkward.
```suggestion
    "statusText": "A pre-generated API client exists.",
```

packages/generator/src/sdk-metadata/generation-and-usage.spec.ts
```diff
@@ -2,34 +2,43 @@ import { resolve } from 'path';
...
-        npmPackageName: '@sap/dummy-package',
-        entities: [{ className: 'DummyClass' }]
-      } as any)
-    ).resolves.toMatchSnapshot();
+    await expect(getApiSpecificUsage(service)).resolves.toMatchSnapshot();
+  });
+
+  it('[E2E] gives a working generator repository link', async () => {
+    const generationAndUsage = await getGenerationAndUsage(service);
+    checkUrlExists(generationAndUsage.generatorRepositoryLink);
```
marikaner, line 30: Do we really need to check that the URL exists? This also applies to the already existing tests...

packages/generator/src/sdk-metadata/sdk-metadata-types.ts
```diff
@@ -19,12 +31,19 @@ export interface Client {
...
-   * `unknown` means we have no information.
    * @memberof Client
    */
-  serviceStatus: 'certified' | 'verified' | 'unknown';
+  serviceStatus: ServiceStatus;
+
+  /**
+   * This object will be set on our metadata server in case we need to show some emergency information.
+   */
+  emergencyObject: EmergencyObject | undefined;
```
marikaner, line 41: Would that suffice?
```suggestion
  emergencyObject?: EmergencyObject;
```

packages/generator/src/sdk-metadata/sdk-metadata.ts
```diff
@@ -34,6 +45,8 @@ export function getSdkMetadataFileNames(
   };
 }
 
+export const introText =
+  'The SAP Cloud SDK is a versatile set of libraries and tools for developers to build applications in a cloud-native way and host them on SAP Business Technology Platform or other runtimes.';
```
marikaner, line 49: Not sure:
```suggestion
  'The SAP Cloud SDK is a versatile set of libraries and tools for developers to build applications in a cloud-native way and host them on the SAP Business Technology Platform or other runtimes.';
```

packages/generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -10,41 +17,61 @@ export async function getGenerationAndUsage(
...
+    ),
+    text: genericUsageText
+  };
 }
 
+export const apiSpecificUsageText = 'Find below a usage example for this API.';
 export async function getApiSpecificUsage(
   service: VdmServiceMetadata
-): Promise<string> {
+): Promise<InstructionWithText<typeof apiSpecificUsageText>> {
```
marikaner, line 42: I am not sure whether it is a good idea to have the usage string as literal type here... Why not just string?

packages/generator/internal.d.ts
FrankEssenberger: yes nicer imports in the api mirror is the reason behind this.

packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
FrankEssenberger: the idea behind this object is in case we realize some emegency in a sense of broken functionality for an client we will fill this object with some information on the server immediately and provide some fix and release a new client in the mean time.

packages/generator/src/sdk-metadata/generation-and-usage.spec.ts
FrankEssenberger: I think we should do that. Imagine we change some path in the docs, who remembers to update the sdk-metatdata?

packages/generator/src/sdk-metadata/generation-and-usage.ts
FrankEssenberger: hm I wanted to be super rigorous, but I think you are right. I make it just string.

packages/generator/src/sdk-metadata/pregenerated-lib.ts
```diff
@@ -7,10 +7,19 @@ import {
...
+  PregeneratedLibrary
+} from './sdk-metadata-types';
 
-export function getInstallationSnippet(service: VdmServiceMetadata): string {
-  return `npm i ${service.npmPackageName}:latest`;
+export function getInstallationSnippet(
+  service: VdmServiceMetadata
+): InstructionWithText {
+  return {
+    instructions: `npm i ${service.npmPackageName}:latest`,
```
jjtang1985, line 19: question: do we only consider npm for the first version?

packages/generator/src/sdk-metadata/sdk-metadata.spec.ts
```diff
@@ -29,7 +29,7 @@ describe('sdk-metadata', () => {
       createOptions({ versionInPackageJson: '1.0.0' })
     );
     expect(metaData).toMatchSnapshot();
-    expect(metaData.serviceStatus).toBe('certified');
+    expect(metaData.serviceStatus.status).toBe('certified');
```
jjtang1985, line 32: Maybe you can add a test for the `statusText`.

jjtang1985 (changes requested):
Looks good, and please consider my comments.

packages/generator/src/sdk-metadata/sdk-metadata.spec.ts
FrankEssenberger: done

packages/generator/src/sdk-metadata/pregenerated-lib.ts
FrankEssenberger: hm you mean yarn? I guess npm is more standard and all yarn users know what to do.

packages/generator/src/sdk-metadata/index.ts
```diff
@@ -0,0 +1 @@
+export * from './sdk-metadata-types';
```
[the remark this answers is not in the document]
jjtang1985: Is it intended to only export the types?
I guess metadata related should be imported via `internal` now, which is safe to export all?

jjtang1985 (approved):
LGTM, I left a comment regarding the index.ts.

packages/generator/src/sdk-metadata/index.ts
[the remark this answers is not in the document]
FrankEssenberger: I wanted to export the minimum even in the submodule `generator/internal`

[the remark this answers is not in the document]
FrankEssenberger: If I once need more I will change it to the simple all approach.

packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
marikaner: Same.

packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
marikaner: makes sense 😊

packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
marikaner: Same.

packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
marikaner: Same.

packages/generator/src/sdk-metadata/generation-and-usage.spec.ts
marikaner: Hmm, true.
Somehow I am not a fan of putting this together with the unit tests. I would rather expect this to be a separate .spec in the integration tests.
If no code is changed, I would expect no (unit) tests to fail, which could happen here. It also takes a little longer than other tests I assume.

(I noticed that we have the [E2E] prefix here, which I am also not a fan of. I think if it is an E2E test, it should be even more explicit than just the description).

If you agree, maybe move it, but I won't insist, especially because we already had some tests in this direction. As long as we don't have issues, we can look at this as a matter of taste ;)

packages/generator/src/sdk-metadata/pregenerated-lib.ts
marikaner: From my side only npm is fine. But why do we use `...:latest`?

packages/generator/src/sdk-metadata/sdk-metadata-types.ts
```diff
@@ -85,3 +93,34 @@ export type DateTimeString = string;
...
+
+export interface EmergencyObject {
+  status: string;
+  description: string;
+}
+
+export interface InstructionWithText {
+  instructions: MultiLineText;
+  text: string;
+}
```
marikaner, line 126: I think those could benefit from API docs. I don't fully get what InstructionsWithText (so what is `text`? Maybe `title`?).

packages/test-util/src/url.spec.ts
```diff
@@ -8,4 +8,8 @@ describe('url', () => {
       )
     ).rejects.toThrowErrorMatchingSnapshot();
   });
+
+  it('[E2E] does not fail for existing link', async () => {
+    await expect(checkUrlExists('https://sap.github.io/cloud-sdk/')).resolves;
```
marikaner, line 13: I think this will always be green even if the link does not exist, please check that.

packages/generator/src/sdk-metadata/generation-and-usage.spec.ts
marikaner: Maybe one additional point: I would also expect the unit tests to be able to run offline without external factors.

marikaner (changes requested):
Sorry, I left some messy comments apart from the review. Please check the snapshots for comments as well.

packages/generator/src/sdk-metadata/generation-and-usage.spec.ts
FrankEssenberger: moved it to E2E test see next PR.

PR #1220, opened 2021-04-20
Refactor references and schema mapping

marikaner:
Refactoring of the document referenced and schemarefmapping into one object.

### DIFF
--- a/packages/openapi-generator/src/parser/refs.ts
+++ b/packages/openapi-generator/src/parser/refs.ts
@@ -1,34 +1,88 @@
 import { OpenAPIV3 } from 'openapi-types';
 import { $Refs, resolve } from '@apidevtools/swagger-parser';
-import { kebabCase, pascalCase } from '@sap-cloud-sdk/util';
+import { pascalCase, kebabCase } from '@sap-cloud-sdk/util';
 import { isReferenceObject } from '../schema-util';
 import { SchemaNaming } from '../openapi-types';
 import { SchemaRefMapping } from './parsing-info';
 import { ensureUniqueNames } from './unique-naming';
 
 /**
- * Check whether the given object is a reference object and resolve if necessary.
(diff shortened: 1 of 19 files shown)


packages/openapi-generator/src/parser/refs.ts
```diff
@@ -15,3 +19,81 @@ export function resolveObject<T>(
 ): T {
   return isReferenceObject(obj) ? refs.get(obj.$ref) : obj;
 }
+
+export async function createRefs(
```
FrankEssenberger, line 23: some docs for the methods

FrankEssenberger: I would move this as a static method of the class -> builder

packages/openapi-generator/src/parser/refs.ts
```diff
@@ -15,3 +19,81 @@ export function resolveObject<T>(
...
+    if (!schemaNaming) {
+      throw new Error(
+        `Could not find schema naming for reference path '${ref}'. Schema does not exist.`
+      );
+    }
+    return schemaNaming;
+  }
+}
+
+function parseSchemaRefMapping(document: OpenAPIV3.Document): SchemaRefMapping {
```
FrankEssenberger, line 70: Also matter of taste, I would put this as a private method in the class and call this in the constructor.

FrankEssenberger (approved):
LGTM - I like your class. Consider my comments on the methods and where to put them.

packages/openapi-generator/src/parser/refs.ts
marikaner: Done, but I kept an additional function for convenience as this much shorter and "functional" to write.

packages/openapi-generator/src/parser/refs.ts
marikaner: done

PR #1221, opened 2021-04-21
chore: Refactor unique naming functions

marikaner:
Simplify unique naming functions by a lot. Left-over suggestions from @FrankEssenberger.

### DIFF
--- a/packages/openapi-generator/src/parser/operation.ts
+++ b/packages/openapi-generator/src/parser/operation.ts
@@ -115,7 +115,7 @@ export function parsePathParameters(
 ): OpenApiParameter[] {
   const sortedPathParameters = sortPathParameters(pathParameters, pathPattern);
   const parsedParameters = parseParameters(sortedPathParameters, refs);
-  const names = ensureUniqueNames(
+  const uniqueNames = ensureUniqueNames(
     parsedParameters.map(({ originalName }) => originalName),
     {
       reservedWords: ['body', 'queryParameters', ...reservedJsKeywords]
@@ -123,7 +123,7 @@ export function parsePathParameters(
   );
 
   return parsedParameters.map((param, i) => {
-    param.name = names[i];
+    param.name = uniqueNames[i];
     return param;
   });
 }
--- a/packages/openapi-generator/src/parser/operation.ts
+++ b/packages/openapi-generator/src/parser/operation.ts
@@ -1,10 +1,5 @@
 import { OpenAPIV3 } from 'openapi-types';
-import {
-  camelCase,
-  filterDuplicatesRight,
-  partition,
-  UniqueNameGenerator
-} from '@sap-cloud-sdk/util';
+import { filterDuplicatesRight, partition } from '@sap-cloud-sdk/util';
 import { OpenApiOperation, OpenApiParameter } from '../openapi-types';
 import { parseRequestBody } from './request-body';
 import { OpenApiDocumentRefs } from './refs';
@@ -12,6 +7,7 @@ import { parseSchema } from './schema';
 import { parseResponses } from './responses';
 import { OperationInfo } from './parsing-info';
 import { reservedJsKeywords } from './reserved-words';
+import { ensureUniqueNames } from './unique-naming';
 
 /**
  * Parse an operation info into a serialization-ready object.
@@ -118,16 +114,18 @@ export function parsePathParameters(
   refs: OpenApiDocumentRefs
 ): OpenApiParameter[] {
   const sortedPathParameters = sortPathParameters(pathParameters, pathPattern);
-  const nameGenerator = new UniqueNameGenerator('', [
-    'body',
-    'queryParameters',
-    ...reservedJsKeywords
-  ]);
+  const parsedParameters = parseParameters(sortedPathParameters, refs);
+  const names = ensureUniqueNames(
+    parsedParameters.map(({ originalName }) => originalName),
+    {
+      reservedWords: ['body', 'queryParameters', ...reservedJsKeywords]
+    }
+  );
 
-  return parseParameters(sortedPathParameters, refs).map(param => ({
-    ...param,
-    name: nameGenerator.generateAndSaveUniqueName(camelCase(param.originalName))
-  }));
+  return parsedParameters.map((param, i) => {
+    param.name = names[i];
+    return param;
+  });
 }
 
 export function parseParameters(
--- a/packages/openapi-generator/src/parser/unique-naming.spec.ts
+++ b/packages/openapi-generator/src/parser/unique-naming.spec.ts
@@ -1,31 +1,34 @@
 import { pascalCase } from '@sap-cloud-sdk/util';
 import { ensureUniqueNames } from './unique-naming';
 
-it('ensureUniqueNames replaces duplicate names using defaults', () => {
-  expect(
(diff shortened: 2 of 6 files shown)


packages/openapi-generator/src/parser/operation.ts
```diff
@@ -118,16 +114,18 @@ export function parsePathParameters(
   refs: OpenApiDocumentRefs
 ): OpenApiParameter[] {
   const sortedPathParameters = sortPathParameters(pathParameters, pathPattern);
-  const nameGenerator = new UniqueNameGenerator('', [
-    'body',
-    'queryParameters',
-    ...reservedJsKeywords
-  ]);
+  const parsedParameters = parseParameters(sortedPathParameters, refs);
+  const names = ensureUniqueNames(
```
FrankEssenberger, line 118: perhaps you can call this uniqueNames

packages/openapi-generator/src/parser/unique-naming.spec.ts
```diff
@@ -1,69 +1,31 @@
...
-    { name: 'someDuplicateName2', id: '3' }
-  ]);
+it('ensureUniqueNames replaces duplicate names using defaults', () => {
+  expect(
+    ensureUniqueNames([
+      'someDuplicateName1',
+      'someDuplicateName',
+      'someDuplicateName'
+    ])
+  ).toEqual(['someDuplicateName1', 'someDuplicateName', 'someDuplicateName2']);
```
FrankEssenberger, line 11: This test is special in a sense that it also checks that the obvious replacement `someDuplicateName1` is already in so it uses a 2. Perhaps you can add more simple test to cover the `[a,a,b]` case and adjust the text here.

packages/openapi-generator/src/parser/unique-naming.ts
```diff
@@ -2,107 +2,70 @@ import { UniqueNameGenerator, camelCase } from '@sap-cloud-sdk/util';
...
  * Ensure uniqueness of names.
- * Takes a list of items, identifies duplicate names and renames the duplicate names.
- * @param items List of items to rename.
- * @param nameHandler Object containing a name getter and/or setter and/or formatter.
- * @param nameHandler.getName Function to get the name of an item. Retrieves the `name` property by default.
- * @param nameHandler.transformItem Function to transform the given item with the new name. Sets the `name` property by default.
- * @param nameHandler.formatName Function to transform the name when finding a unique name. Defaults to camel case.
- * @param nameHandler.reservedWords Reserved words that should be handled as duplicates.
- * @returns The given items with unique names.
+ * Takes a list of names and renames those that are duplicate.
```
FrankEssenberger, line 5: Mention that the oder is contained this is imported for later usage I think.

packages/openapi-generator/src/parser/unique-naming.ts
```diff
@@ -2,107 +2,70 @@ import { UniqueNameGenerator, camelCase } from '@sap-cloud-sdk/util';
...
+ * @param options.reservedWords Reserved words that should be handled as duplicates.
+ * @returns Unique names in the given order.
  */
-export function ensureUniqueNames<ItemT, UniqueItemT>(
-  items: ItemT[],
-  nameHandler: {
-    getName?: (item: ItemT) => string;
-    transformItem?: (item: ItemT, name: string) => UniqueItemT;
-    formatName?: (name: string) => string;
+export function ensureUniqueNames(
```
FrankEssenberger, line 12: perhaps the name of the method is not optimal anymore. The methods gives you uniqueNames it does not changes the objects anymore.

packages/openapi-generator/src/parser/unique-naming.ts
```diff
@@ -2,107 +2,70 @@ import { UniqueNameGenerator, camelCase } from '@sap-cloud-sdk/util';
...
+ * @param format Function to format the name.
  * @param reservedWords Reserved words that should be handled as duplicates.
- * @returns An object containing the unique operations, denoted by `unique` and operations with (potentially) duplicate names, denoted by `duplicate`.
+ * @returns A list of names that do not need to be renamed.
  */
-function getCorrectlyNamedItems<ItemT>(
-  items: ItemT[],
-  getName: (item: ItemT) => string,
-  formatName: (name: string) => string,
+function getCorrectNames(
```
FrankEssenberger, line 45: like this helper methods with the other helper methods inside. Also wonder about the name but have no good alternative. `getNonConflictingNames()`

FrankEssenberger: but not sure

FrankEssenberger: I know that this method and the other `isReserve,isDuplicate,isFormatted` are only used internaly but are great candidates for nice small unit tests. Should we make them visible for tests?

FrankEssenberger (approved):
Have a look at my comments, but overall I like the code now much better with this simple make it unique method and the extraction and transformation moved out.

packages/openapi-generator/src/parser/unique-naming.ts
FrankEssenberger: Also general: Should this go to the util like the unique-name-generator?

packages/openapi-generator/src/parser/operation.ts
marikaner: Done.

packages/openapi-generator/src/parser/unique-naming.spec.ts
marikaner: Ok.

packages/openapi-generator/src/parser/unique-naming.ts
marikaner: Done.

packages/openapi-generator/src/parser/unique-naming.ts
marikaner: For now this is not necessary, but once we want to use it somewhere else we should move it out.

marikaner: > perhaps the name of the method is not optimal anymore. The methods gives you uniqueNames it does not changes the objects anymore.

I renamed it when working on it, but `uniqueNames` is a bit annoying because sometimes that is the name of the variable ;)
I was considering `transformUniqueNames` or `formatUniqueNames` but this implies that the names are unique already. I also considered `formatToUniqueNames` but this is rather long, so I stayed with `ensureUniqueNames` for now.
That said, I don't have a strong opinion, let me know what you think, I am already working on the next PR, I would change the name there, then.

packages/openapi-generator/src/parser/unique-naming.ts
marikaner: Hmm, I only moved isReserved, isDuplicate, isFormatted out, so that they are not executed if one of the others already returns true (isDuplicate and isFormatted go over "all" names every time they are called).
I think those functions are trivial and because ensureUniqueNames is tested it shouldn't be necessary to test them in my opinion.

marikaner: > like this helper methods with the other helper methods inside. Also wonder about the name but have no good alternative. `getNonConflictingNames()`

I don't like the negation in the name, but I think it is more on point than "correct". Will use it.

PR #1222, opened 2021-04-21
more comments metadata adjustment

FrankEssenberger:
I missed a few comments by @marikaner  on my PR: https://github.com/SAP/cloud-sdk-js/pull/1219
I adjsuted them here:
- Mocked the request so the unit test are offline
- Moved the link test to the E2E test folder
- Added some docs 
- Adjusted some texts


packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
```diff
@@ -2,16 +2,23 @@
...
+}
+`;
+
+exports[`sdk-metadata generates the JS metadata content for services with pregenerated lib 1`] = `
 Object {
   "emergencyObject": undefined,
   "generationAndUsage": Object {
     "apiSpecificUsage": Object {
-      "instructions": "import { TestEntity } from '@sap/cloud-sdk-vdm-business-partner-service';
+      "instructions": "import { TestEntity } from 'non-existing-package';
```
marikaner, line 17: Is this intended? Seems odd.

marikaner (approved):
I left one question because the import in the snapshot looks wrong. Other than that, LGTM, thanks for going this extra step!

packages/generator/src/sdk-metadata/__snapshots__/sdk-metadata.spec.ts.snap
FrankEssenberger: I did not really pay attention. Since it does not matter what is used there I put the test service.

PR #1223, opened 2021-04-21
feat: Fail on duplicates by default

marikaner:
Introduce strict naming option that is true by default. Fails generation for duplicate names.


packages/openapi-generator/src/cli/index.ts
```diff
@@ -40,6 +40,14 @@ class GenerateOpenApiClient extends Command {
       default: false,
       required: false
     }),
+    strictNaming: flags.boolean({
+      name: 'strictNaming',
+      description:
+        'By default, the generation fails, when there are duplicate names for operations and/or path parameters after transforming them to camel case. Set this to true to enable unique name generation. The names will then be generated by appending numbers.',
+      default: true,
+      required: false,
+      allowNo: true
```
jjtang1985, line 49: I guess oclif will add additional doc when `allowNo` is true? If not, we might want to mention it.

packages/openapi-generator/src/parser/unique-naming.spec.ts
```diff
@@ -2,32 +2,104 @@ import { pascalCase } from '@sap-cloud-sdk/util';
...
 
 describe('ensureUniqueNames', () => {
-  it('replaces duplicate names using defaults', () => {
-    expect(ensureUniqueNames(['duplicate', 'duplicate'])).toEqual([
-      'duplicate',
-      'duplicate1'
-    ]);
-  });
+  describe('with strictNaming disabled', () => {
+    const options = { strictNaming: true };
```
jjtang1985, line 6: I know the meaning of the option, but `strictNaming disabled` => `{ strictNaming: true }` seems a bit confusing.

packages/openapi-generator/src/parser/unique-naming.spec.ts
```diff
@@ -2,32 +2,104 @@ import { pascalCase } from '@sap-cloud-sdk/util';
...
+        	Formatted name: 'duplicate2', original names: 'duplicate2', 'Duplicate2'."
+      `);
+    });
   });
 
-  it('replaces duplicate names if they occur in the reserved words', () => {
-    const uniqueItems = ensureUniqueNames(['reserved', 'reserved1'], {
-      reservedWords: ['reserved']
+  describe('with strictNaming enabled', () => {
+    const options = { strictNaming: false };
```
jjtang1985, line 70: Also this combination.

packages/openapi-generator/src/parser/unique-naming.spec.ts
jjtang1985: A simple way might be `with strictNaming true`.

packages/openapi-generator/src/parser/unique-naming.ts
```diff
@@ -1,24 +1,75 @@
...
+ * @param namingOptions Object containing options to configure the transformation.
  * @param options.format Function to format the name. Defaults to camel case.
  * @param options.reservedWords Reserved words that should be handled as duplicates.
  * @returns Unique names in the given order.
  */
 export function ensureUniqueNames(
   names: string[],
-  options: {
+  options: ParserOptions,
+  namingOptions: {
```
jjtang1985, line 19: Since you created `ParserOptions`, should we also have a `NamingOptions`?

packages/openapi-generator/src/parser/unique-naming.ts
```diff
@@ -1,24 +1,75 @@
...
     format?: (name: string) => string;
     reservedWords?: string[];
   } = {}
 ): string[] {
-  const { format: format = camelCase, reservedWords = [] } = options;
+  const { format = camelCase, reservedWords = [] } = namingOptions;
```
jjtang1985, line 24: As default value, why not moving it to line 22, where `{}` is used as default `namingOptions`.

packages/openapi-generator/src/parser/unique-naming.ts
```diff
@@ -74,3 +125,81 @@ function isDuplicate(name: string, correctNames: string[]): boolean {
...
+function getDuplicateErrorMessage(
+  names: string[],
+  formattedNames: string[],
+  reservedWords: string[]
+): string {
+  const uniqueNames = unique(formattedNames);
+
+  let errorMessage = '';
+
+  if (hasDuplicates(formattedNames)) {
```
jjtang1985, line 147: It seems, when reaching this line, it is already checked in the validation step before. Not sure, whether for a large spec, whether this will be some scalability issue.
Same for the `hasReservedWords` below.
If we want to improve, we can return some information about e.g., "duplicate names" during the validation, which should be a relatively small data set.

packages/openapi-generator/src/cli/index.ts
marikaner: Good point. I just checked and this is what it shows on `--help`:

```sh
--[no-]strictNaming                          By default, the generation fails, when there are duplicate names for operations and/or path parameters after transforming them to camel case. Set this to true to enable unique name generation. The names will 
                                               then be generated by appending numbers.
```

Good enough, right?

packages/openapi-generator/src/parser/unique-naming.ts
marikaner: Yes, I thought about this, but it is only used in two places... what do you think?

packages/openapi-generator/src/parser/unique-naming.ts
marikaner: Because I don't know the syntax for it if even possible.
If I move it up and do this: `{ format: camelCase, reservedWords: [] }` it will only be used if no `namingOptions` parameter was provided.
Moving it up with `{ format = camelCase, reservedWords = [] }` doesn't work.

I think it doesn't work, unless you know something maybe?

packages/openapi-generator/src/parser/unique-naming.ts
marikaner: Yes, I also didn't like this too much. I did it similarly to the way you suggested initially, but I changed it, because I think it is better to separate the validation and error handling. 
I will try to find a way to implement a nicer "result" for the validation...

packages/openapi-generator/src/parser/unique-naming.ts
jjtang1985: This is a minor comment, you decide;)

packages/openapi-generator/src/parser/unique-naming.ts
marikaner: Argh, I just tried something and I struggle to find a way that is still nicely readable while not getting the unique values multiple times.

I also did the reserved words reduction before, but there I have to go over all reserved words once anyways to find those that are reserved, so I wouldn't really gain much from this.

I think it is just not worth to sacrifice the readability here, especially because this will only be executed in the error case. Of course, feel free to correct me and provide an example, maybe you have a better way in mind than me.

packages/openapi-generator/src/parser/unique-naming.ts
jjtang1985: I think you are right. If one parameter (`namingOptions`) is used, the default value is only applicable for this parameter and cannot be flexible enough for setting two properties (format/reservedWords) of the parameter.

packages/openapi-generator/src/parser/unique-naming.ts
```diff
@@ -1,24 +1,75 @@
...
  */
 export function ensureUniqueNames(
   names: string[],
-  options: {
+  options: ParserOptions,
+  namingOptions?: {
     format?: (name: string) => string;
     reservedWords?: string[];
-  } = {}
+  }
```
jjtang1985, line 22: Nice to remove the "default value".

packages/openapi-generator/src/parser/unique-naming.spec.ts
marikaner: Oh, the text is is just wrong...

packages/openapi-generator/src/parser/unique-naming.spec.ts
```diff
@@ -2,32 +2,104 @@ import { pascalCase } from '@sap-cloud-sdk/util';
 import { ensureUniqueNames } from './unique-naming';
 
 describe('ensureUniqueNames', () => {
-  it('replaces duplicate names using defaults', () => {
-    expect(ensureUniqueNames(['duplicate', 'duplicate'])).toEqual([
-      'duplicate',
-      'duplicate1'
-    ]);
-  });
+  describe('with strictNaming disabled', () => {
```
marikaner, line 5:
```suggestion
  describe('with strictNaming enabled', () => {
```

packages/openapi-generator/src/parser/unique-naming.spec.ts
```diff
@@ -2,32 +2,104 @@ import { pascalCase } from '@sap-cloud-sdk/util';
...
+        	Formatted name: 'duplicate1', original names: 'duplicate1', 'Duplicate1'.
+        	Formatted name: 'duplicate2', original names: 'duplicate2', 'Duplicate2'."
+      `);
+    });
   });
 
-  it('replaces duplicate names if they occur in the reserved words', () => {
-    const uniqueItems = ensureUniqueNames(['reserved', 'reserved1'], {
-      reservedWords: ['reserved']
+  describe('with strictNaming enabled', () => {
```
marikaner, line 69:
```suggestion
  describe('with strictNaming disabled', () => {
```

packages/openapi-generator/src/parser/unique-naming.spec.ts
marikaner: fixed.

packages/openapi-generator/src/parser/unique-naming.ts
marikaner: Then I would leave it unless it becomes problematic or @FrankEssenberger also wants this ;)

packages/openapi-generator/src/parser/unique-naming.ts
marikaner: Schade.

packages/openapi-generator/src/parser/unique-naming.ts
marikaner: Yes, your comment inspired this ;)

marikaner:
@jjtang1985 @FrankEssenberger 
I have one open question on the behavior, that we didn't discuss.
Currently the generation will fail when the spec does not yield unique names. I think this is the expected behavior when there is one spec. For "batch mode" I think we should still continue the generation (currently not the case), but throw an error in the end stating that services, x, y z, could not be generated because they contained conflicting names.
Do you think that is fine?

FrankEssenberger:
> @jjtang1985 @FrankEssenberger
> I have one open question on the behavior, that we didn't discuss.
> Currently the generation will fail when the spec does not yield unique names. I think this is the expected behavior when there is one spec. For "batch mode" I think we should still continue the generation (currently not the case), but throw an error in the end stating that services, x, y z, could not be generated because they contained conflicting names.
> Do you think that is fine?

As discussed in standup this is a reasonable behavior.

marikaner:
> > @jjtang1985 @FrankEssenberger
> > I have one open question on the behavior, that we didn't discuss.
> > Currently the generation will fail when the spec does not yield unique names. I think this is the expected behavior when there is one spec. For "batch mode" I think we should still continue the generation (currently not the case), but throw an error in the end stating that services, x, y z, could not be generated because they contained conflicting names.
> > Do you think that is fine?
> 
> As discussed in standup this is a reasonable behavior.

Implemented on a separate PR #1227.

jjtang1985 (approved):

PR #1227, opened 2021-04-22
chore: Add error handling for skipped services

marikaner:
Generate all services and list failures in the end.


packages/openapi-generator/src/generator.ts
```diff
@@ -60,7 +63,18 @@ export async function generate(options: GeneratorOptions): Promise<void> {
...
+
+  try {
+    await finishAll(promises);
+  } catch (err) {
+    if (promises.length > 1) {
+      cli.error(
+        new ErrorWithCause('Some clients could not be generated.', err)
+      );
+    } else {
+      new ErrorWithCause('Could not generate client.', err);
```
jjtang1985, line 75: question: why not using `cli.error` for alignment?

jjtang1985 (approved):
Looks good, I left some minor comments.

packages/util/src/promise.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Throws an error consisting of a list of reasons.
+ * @param promises Promises to settle.
+ */
+export async function finishAll(promises: Promise<any>[]): Promise<void> {
+  const settledPromises = await Promise.allSettled(promises);
+  const rejectedPromises = settledPromises.filter(
+    promise => promise.status === 'rejected'
+  ) as PromiseRejectedResult[];
+  if (rejectedPromises.length) {
+    const reasons = rejectedPromises.map(promise => promise.reason).join(', ');
```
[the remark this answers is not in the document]
marikaner: Totally, I also added a tab before each reason + `Errors:` in the begininning. Looks much better.

packages/openapi-generator/src/generator.ts
marikaner: I just forgot ;) But I now removed it in general.

packages/openapi-generator/src/cli/index.ts
```diff
@@ -101,13 +100,8 @@ class GenerateOpenApiClient extends Command {
...
   async run(): Promise<void> {
-    try {
-      const parsed = this.parse(GenerateOpenApiClient);
-      await generate(parsed.flags);
-    } catch (e) {
-      logger.error(e.message);
-      return cli.exit(1);
-    }
+    const parsed = this.parse(GenerateOpenApiClient);
+    await generate(parsed.flags);
```
jjtang1985, line 104: Please correct me. 
In the `generate()`, potentially an error will be thrown, which is not handled in this line.
This should result in `UnhandledPromiseRejectionWarning` instead of "exit 1" as the old behaviour.
Is this intended.

marikaner: Actually no. In the CLI this will be all handled through the run script which already handles errors. However, I reverted this back, because our logger basically makes the error red on the command line and this currently goes nicer together with our other logs. Let's reconsider this for the verbose PR.

PR #1230, opened 2021-04-23
more adjustments on sdk metadata

FrankEssenberger:
With Junjies PR and the common package the changes were redundant, only the export and some small clean up is left.


packages/generator/src/sdk-metadata/sdk-metadata.ts
```diff
@@ -33,6 +33,17 @@ export async function sdkMetaDataJS(
   };
 }
 
+export async function sdkMetaDataJSFallback(
```
jjtang1985, line 36: question: Is this function used?
What would be the use case?

FrankEssenberger: not in the generator but it will be used in the metadata server app if we are asked for a service where we have no information. The reason I wanted to put it in the repo is a better code reuse.

packages/core/package.json
```diff
@@ -16,8 +16,6 @@
     "access": "public"
   },
   "files": [
-    "v4.d.ts",
```
marikaner, line 19: Is this intended?

marikaner, line 19: Is this intended?

packages/generator/src/sdk-metadata/sdk-metadata.ts
marikaner, line 36: According to our styleguide it should be:
```suggestion
export async function sdkMetaDataJsFallback(
```

marikaner: Hmm, but it really doesn't belong in the generator, right?
Would it be much less convenient if we put this into the metadata service?
What is the reuse potential?

marikaner (commented):
Mostly questions.

packages/core/package.json
jjtang1985: I deleted the real files in a previous PR without touching the `package.json`. So yes.

packages/generator/src/sdk-metadata/sdk-metadata.ts
FrankEssenberger: okay I also had not a good feeling about this. So I adjusted the method signature to take strings so that we can easily use them in the metadata server as well where we do not have generator options of service objects.

jjtang1985 (approved):

PR #1231, opened 2021-04-23
chore: Update Generator ADR

marikaner:
There are two more Options to be considered for this.
Also, I realised, that for npm and ts `files` has a slightly different notion and would like to propose `include` instead.


knowledge-base/adr/0020-generator-options.md
```diff
@@ -51,8 +51,11 @@ We should discuss the details of those defaults and make a concious decision on
...
 | `--generateTypedocJson`                                  | `--typedocJson`                  | Writes a default typedoc.json, default true.                                                      | Remove/deprecate, use `files` instead.                                                                                                                       |    -     |
 | **Currently hidden options**                             |
-| `--additionalFiles`                                      | `--files`                        | Copy additional files, identified by glob. Hidden.                                                | _same_ Expose it.                                                                                                                                            |  false   |
+| `--additionalFiles`                                      | `--include`                      | Copy additional files, identified by glob. Hidden.                                                | _same_ Expose it.                                                                                                                                            |  false   |
```
marikaner, line 54: We have discussed `files` for this. I would like to propose `include`. Let me know what you think.

marikaner: @FrankEssenberger @jjtang1985

FrankEssenberger: For me `--include` is fine since it is followed by a path than so`--include select/something/via/Glob` makes sense.

FrankEssenberger (approved):

jjtang1985 (approved):

PR #1232, opened 2021-04-23
chore: Update Service Mapping Option ADR

marikaner:
Basis for discussion.


jjtang1985:
For the service mapping topic, here is my opinion.

- It makes sense to generate a default file (e.g., in the default path like odata), when there is no user input. The default service mapping can then be used as initial state and the user can adjust it.
- I prefer having one service mapping file like the odata instead of N files for every service, because for example when using Concur, I as a user would like to make one config file instead of maintaining 10 files in different folders.
- The missing part is the generation of the default service mapping then, I would also argue this can be improved after GA.
- The naming of the service mapping is not perfect, but still ok. Some other alternatives can be `service-config.json` (`--serviceConfig` option).

marikaner:
@jjtang1985 We have now decided to *NOT* generate a service mapping by default. In know you were leaning in the other direction about this, therefore please review the outcome of the discussion. I added some more context on the service mapping specifically.
This is also discussed with @artemkovalyov.

jjtang1985:
As discussed, it's nice to have the alignment between odata and openapi. Removing the default file makes sense typically for openapi, which does not really need this mapping. For some special odata services, which needs this mapping, we'll improve it as a 2.0 topic.

knowledge-base/adr/0020-generator-options.md
```diff
@@ -37,26 +37,35 @@ We should discuss the details of those defaults and make a concious decision on
...
+| `--generatePackageJson`                                  | `--[no]-packageJson`             | Writes a default package.json, default true.                                                      | Writes a default package.json, default true (needs core dependency). Optionally in the future: Writes a custom package.json if passed. Keep boolean for now.                                                   |
+| `--serviceMapping`                                       | `--perServiceConfig`             | Considers a service mapping file. Shows a warning if none is given (differs from OData behavior). | Only generates a per service configuration, if a file path is provided. No warning if it is not provided. If provided, `packageName` and `directoryName` will be generated, but are not required to be set. \* |
```
FrankEssenberger, line 47:
```suggestion
| `--serviceMapping`                                       | `--perServiceConfig`             | Considers a per service configuration file. Shows a warning if none is given (differs from OData behavior). | Only generates a per service configuration, if a file path is provided. No warning if it is not provided. If provided, `packageName` and `directoryName` will be generated, but are not required to be set. \* |
```

knowledge-base/adr/0020-generator-options.md
```diff
@@ -37,26 +37,35 @@ We should discuss the details of those defaults and make a concious decision on
...
+| -                                                        | `--verbose`                      | Logs everything that happens during generation by deafult.                                        | Log only success / error per service by default. Enable verbosity through this flag.                                                                                                                           |
+| -                                                        | `--version`, `-v`                | -                                                                                                 | Prints the version of the generator.                                                                                                                                                                           |
+
+\* Service Mapping - In OData we currently generate a `service-mapping.json` by default.
+Some services require a configuration to exist, in order to specify a service path, that could not be parsed from the specification by default.
+In those cases we plan to fail the generation when no service mapping was provided, similar to the behavior of `strictNaming`.
```
FrankEssenberger, line 63: DO we want to fail. I think warnig is finde with a remark to the `withCustomServicePath`  or config file option. At runtime if no path is given we have to fail also giving the options to include the file.

knowledge-base/adr/0020-generator-options.md
```diff
@@ -37,26 +37,35 @@ We should discuss the details of those defaults and make a concious decision on
...
 | **Currently hidden options**                             |
-| `--additionalFiles`                                      | `--include`                      | Copy additional files, identified by glob. Hidden.                                                | _same_ Expose it.                                                                                                                                            |
-| `--writeReadme`                                          | `--readme`                       | Writes a default README.md, default false.                                                        | _same_                                                                                                                                                       |
+| `--additionalFiles`                                      | `--include`                      | Copy additional files, identified by glob. Hidden.                                                | _same_ Expose it.                                                                                                                                                                                              |
```
FrankEssenberger, line 54: was the name not supposed to change to `includes` or `files` in the future?

knowledge-base/adr/0020-generator-options.md
jjtang1985: What about `--serviceConfig` as a shorter option?

knowledge-base/adr/0020-generator-options.md
marikaner: Yes, it is `include` (you have to look in the second column for the new names).

knowledge-base/adr/0020-generator-options.md
marikaner, line 63:
```suggestion
In those cases we plan to fail the generation or log a warning when no service mapping was provided, similar to the behavior of `strictNaming`.
```

marikaner: Adjusted it. I would suggest, that the final behavior in that situation should be discussed as well and noted down.

knowledge-base/adr/0020-generator-options.md
marikaner: @FrankEssenberger What do you say?

knowledge-base/adr/0020-generator-options.md
```diff
@@ -37,26 +37,35 @@ We should discuss the details of those defaults and make a concious decision on
...
+| `--generatePackageJson`                                  | `--[no]-packageJson`             | Writes a default package.json, default true.                                                                | Writes a default package.json, default true (needs core dependency). Optionally in the future: Writes a custom package.json if passed. Keep boolean for now.                                                   |
+| `--serviceMapping`                                       | `--perServiceConfig`             | Considers a per service configuration file. Shows a warning if none is given (differs from OData behavior). | Only generates a per service configuration, if a file path is provided. No warning if it is not provided. If provided, `packageName` and `directoryName` will be generated, but are not required to be set. \* |
```
marikaner, line 47:
```suggestion
| `--serviceMapping`                                       | `--optionsPerService`             | Considers a per service configuration file. Shows a warning if none is given (differs from OData behavior). | Only generates a per service configuration, if a file path is provided. No warning if it is not provided. If provided, `packageName` and `directoryName` will be generated, but are not required to be set. \* |
```

marikaner:
> For the service mapping topic, here is my opinion.
> 
> * It makes sense to generate a default file (e.g., in the default path like odata), when there is no user input. The default service mapping can then be used as initial state and the user can adjust it.
> * I prefer having one service mapping file like the odata instead of N files for every service, because for example when using Concur, I as a user would like to make one config file instead of maintaining 10 files in different folders.
> * The missing part is the generation of the default service mapping then, I would also argue this can be improved after GA.
> * The naming of the service mapping is not perfect, but still ok. Some other alternatives can be `service-config.json` (`--serviceConfig` option).

@jjtang1985 We have discussed options around the naming here and came up with `--optionsPerService`. Is that name ok for you?

jjtang1985:
> @jjtang1985 We have discussed options around the naming here and came up with --optionsPerService. Is that name ok for you?
Not bad, `option` is even better than `config`.

jjtang1985 (approved):

PR #1233, opened 2021-04-23
chore: Rename generator options

marikaner:
Rename options as defined in the according ADR: https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/adr/0020-generator-options.md


packages/openapi-generator/README.md
```diff
@@ -20,38 +20,60 @@ $ npm install @sap-cloud-sdk/openapi-generator
...
-  --tsConfig=tsConfig                          tsconfig.json file to overwrite the default "tsconfig.json".
+  --include=<glob/to/include>                             Include files matching the given glob into the root of each
+                                                          generated client directory.
 
-  --versionInPackageJson=versionInPackageJson  By default, when generating package.json file, the generator will set a
-                                               version by using the generator version. It can also be set to a specific
-                                               version.
+  --[no-]packageJson                                      When set to false, no `package.json` is generated. By default,
+                                                          a `package.json` that specifies dependencies and scripts for
+                                                          transpilation and documentation generation.
```
FrankEssenberger, line 59: verb is missing in `By default, a `package.json` that specifies dependencies and scripts for                                                          transpilation and documentation generation.`

packages/openapi-generator/README.md
```diff
@@ -20,38 +20,60 @@ $ npm install @sap-cloud-sdk/openapi-generator
...
+                                                          and JSON files. Throws an error if the path does not exist.
 
-  --generateJs                                 By default, the generator will also generate transpiled .js, .js.map,
-                                               .d.ts and .d.ts.map files. When set to false, the generator will only
-                                               generate .ts files.
+  -o, --outputDir=<path/to/output>                        (required) Specify the path to the directory to generate the
+                                                          client(s) in. Each client is generated into a subdirectory
+                                                          within the given output directory. Creates the directory if it
+                                                          does not exist. Customize subdirectory naming through
+                                                          `--serviceMapping`.
```
FrankEssenberger, line 43: This could be subject to change right? See https://github.com/SAP/cloud-sdk-js/pull/1232

packages/openapi-generator/README.md
```diff
@@ -20,38 +20,60 @@ $ npm install @sap-cloud-sdk/openapi-generator
...
 
-  --generatePackageJson                        By default, the generator will generate a package.json file, specifying
-                                               dependencies and scripts for compiling and generating documentation. When
-                                               set to false, the generator will skip the generation of the package.json.
+  -t, --transpile                                         Transpile the generated TypeScript code. When enabled a
+                                                          default `tsconfig.json` will be generated. It emits `.js`,
```
FrankEssenberger, line 46:
```suggestion
                                                          default `tsconfig.json` will be generated and used. It emits `.js`,
```

packages/openapi-generator/README.md
```diff
@@ -20,38 +20,60 @@ $ npm install @sap-cloud-sdk/openapi-generator
...
+                                                          generated client directory.
 
-  --versionInPackageJson=versionInPackageJson  By default, when generating package.json file, the generator will set a
-                                               version by using the generator version. It can also be set to a specific
-                                               version.
+  --[no-]packageJson                                      When set to false, no `package.json` is generated. By default,
+                                                          a `package.json` that specifies dependencies and scripts for
+                                                          transpilation and documentation generation.
+
+  --serviceMapping=<path/to/custom-service-mapping.json>  Set the path to the service mapping file. By default, a
```
FrankEssenberger, line 61: see above. I think the name and default behavior will change. Perhaps we will make this hidden in first GA, right?

FrankEssenberger (approved):
Only the serviceMapping is a bit moving target. Also I made the suggestions in the wrong place. Of course move them to the options and not to the auto generated code :-)

packages/openapi-generator/README.md
marikaner: Yes, will do this on a separate PR.

packages/openapi-generator/README.md
marikaner: Good point, we can make it hidden in the first version(s) now that it is not going to be the default behavior.

PR #1234, opened 2021-04-23
chore: Update packageJson option

marikaner:
When discussing the generator options with @artemkovalyov, he pointed out, that not generating a package.json might be a more commonly necessary default. Instead, we could add instructions on how to install the core in the success message.


FrankEssenberger (approved):
Fair

knowledge-base/adr/0020-generator-options.md
```diff
@@ -37,26 +37,26 @@ We should discuss the details of those defaults and make a concious decision on
...
+| `--generateJs`                                           | `--transpile`, `-t`              | Transpiles, default true                                                                          | Transpiles, default false. If set if tsconfig is enabled with default (unless configured). This should be explicitly stated in the documentation.                                                                               |
+| `--generatePackageJson`                                  | `--packageJson`                  | Writes a default package.json, default true.                                                      | Writes a default package.json, default false. Add a log statment on successful generation, that the core package needs to be installed. Optionally in the future: Writes a custom package.json if passed. Keep boolean for now. |
```
jjtang1985, line 46: I see this as a performance improvement, if the core dependency is the only reason for generating the package json. By the way, the optional custom package.json can also be covered by the new option `--include`.

jjtang1985 (approved):

PR #1240, opened 2021-04-26
escape ' for enum strings

FrankEssenberger:
Escape `'` in enum strings. Since the outer `'` are already added in the parsing, I added the escaping also directly there.

### DIFF
--- a/packages/openapi-generator/src/parser/schema.ts
+++ b/packages/openapi-generator/src/parser/schema.ts
@@ -170,9 +170,13 @@ function parseEnumSchema(
   return {
     type,
     enum: (schema.enum || []).map(entry =>
-      type === 'string' ? `'${entry}'` : entry
+      type === 'string' ? toEnumString(entry) : entry
     )
   };
+}
+
+function toEnumString(input: string): string {
+  return `'${input.replace(/'/g, "\\'")}'`;
 }
 
 /**
--- a/packages/openapi-generator/src/parser/schema.ts
+++ b/packages/openapi-generator/src/parser/schema.ts
@@ -170,12 +170,12 @@ function parseEnumSchema(
   return {
     type,
     enum: (schema.enum || []).map(entry =>
-      type === 'string' ? toEnumString(entry) : entry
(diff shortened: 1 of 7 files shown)


packages/openapi-generator/src/parser/schema.ts
```diff
@@ -170,11 +170,15 @@ function parseEnumSchema(
   return {
     type,
     enum: (schema.enum || []).map(entry =>
-      type === 'string' ? `'${entry}'` : entry
+      type === 'string' ? toEnumString(entry) : entry
     )
   };
 }
 
+function toEnumString(input: string): string {
```
marikaner, line 178: Consider a function name with a verb, e. g. `getStringEnumValue` or similar. This is also in our [style guide](https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-verbs-in-function-names).

marikaner (commented):
I left a minor comment. Should we add some kind of integration / e2e test to see that invocation in fact would work with `"valueWith'Quote"`?

packages/openapi-generator/src/parser/schema.ts
FrankEssenberger: change. Regarding this in the style guide. We use often noun like funciton like `tsConfig()` when we create objects in the generator. Should we add this to the style guide?

test-packages/e2e-tests/test/openapi.spec.ts
```diff
@@ -32,6 +32,13 @@ describe('openapi request builder', () => {
     const postCreateCount = await countEntities();
     expect(postCreateCount).toEqual(preCreateCount + 1);
   });
+
+  it('executes request with quote in property', async () => {
+    const response = await EntityApi.getAllEntities({
+      enumStringParameter: 'value1'
```
marikaner, line 38:
```suggestion
      enumStringParameter: "valueWith'Quote"
```

FrankEssenberger: ups, this was a test I commited by accident.

marikaner (approved):
LGTM

PR #1241, opened 2021-04-26
fix: Fix schema starting or ending with numbers

FrankEssenberger:
add support for schemas starting with integers.


packages/core/test/test-util/test-services/openapi/test-service/schema/index.ts
```diff
@@ -3,6 +3,7 @@
  *
  * This is a generated file powered by the SAP Cloud SDK for JavaScript.
  */
+export * from './schema-123456';
```
jjtang1985, line 6: this file is missing.

packages/openapi-generator/src/parser/refs.ts
```diff
@@ -72,6 +104,15 @@ export class OpenApiDocumentRefs {
     );
   }
 
+  /**
+   * If names end with integers it is better to use a separator to easy the mapping between the original and new object.
+   * @param names List of names investigated to find the best separator.
+   * @returns Either '' or '_' depending on the names.
+   */
+  private static getSeperator(names: string[]) {
+    return names.find(name => name.match(/\d+$/)) ? '_' : '';
```
jjtang1985, line 113: Nice to have `_` for schema names ending with numbers. How about using this rule for all cases instead of numbers?

jjtang1985 (dismissed):
Looks good, I left a suggestion and there is a missing schema in the core.

packages/core/test/test-util/test-services/openapi/test-service/schema/index.ts
jjtang1985: Interesting, the `check:test-service` in the pipeline passed.

packages/openapi-generator/src/parser/refs.ts
FrankEssenberger: I think @marikaner would prefer just a `Class1` instead of a `Class_1` since it is more JS style?

jjtang1985: @marikaner , here is a question about how should we handle the index pattern when facing name conflicts.
Previous, `Class1` is good enough as JS like style.
However, `Schema400` will become `Schema4001` with the current rule, so we might want to use `Schema400_1` for the schema names ending with number.
Now the question is whether we should align with the same separator `_` or only apply it for number.
I don't have strong opinion, but personally would prefer the alignment a little bit.

packages/core/test/test-util/test-services/openapi/test-service/schema/index.ts
jjtang1985: As discussed, the `check:test-service` should fail as well as the compilation, while the pipeline was green.
Please create a follow-up ticket for the false-green pipeline.

marikaner (changes requested):
I left one comment and basically a +1 on Junjie's comment.

packages/openapi-generator/src/parser/refs.ts
FrankEssenberger: I use now `_` always but kept it as an optional option.

packages/core/test/test-util/test-services/openapi/test-service/schema/index.ts
FrankEssenberger: added a follow up ticket.

packages/openapi-generator/src/parser/refs.spec.ts
```diff
@@ -78,34 +101,30 @@ describe('OpenApiDocumentRefs', () => {
           schemaName: 'Schema123456_1'
         }
       );
+
+      // The to camalCase removes the _ in the origirnal name.
+      expect(
+        refs.getSchemaNaming('#/components/schemas/schema12345_6')
```
jjtang1985, line 107: This means the following are treated the same, which is not optimal but since these are corner cases I would not consider for now.
- 123456
- 123_456
- 12_34_56

packages/openapi-generator/src/parser/schema-naming.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  it('throws if the strict naming is on', () => {
+    expect(() =>
+      ensureUValidSchemaNames(['1234ABC', '#som%eth.ing'], {
+        strictNaming: true
+      })
+    ).toThrowErrorMatchingInlineSnapshot(`
+      "Your OpenApi definition contains the invalid schema names. The SDK generator can adjust such names automatically if you disable 'strictNaming' or you adjust the schema. The errors are:
+      The schema 1234ABC starts with an integer.
+      The schema #som%eth.ing contains an illegal character."
```
jjtang1985, line 24: 👍 for all findings

jjtang1985 (approved):
LGTM

packages/openapi-generator/src/parser/refs.spec.ts
FrankEssenberger: yes, the `pascalCase` or `camelCase` methods will remove all this kind of formatting also small/capital letters will be removed. But I also think these are corner cases where we should not do a premature effort.

packages/openapi-generator/src/parser/unique-naming.ts
```diff
@@ -19,9 +20,11 @@ export function ensureUniqueNames(
   namingOptions?: {
     format?: (name: string) => string;
     reservedWords?: string[];
+    separator?: string;
```
marikaner, line 23: Would it make sense to always have the same separator? Now we use it for schema names only, but file names and operation names would still suffer from the same issues.

marikaner (commented):
I left one question

packages/openapi-generator/src/parser/unique-naming.ts
FrankEssenberger: I added the seperator also for files and operation names.

marikaner (approved):
LGTM

PR #1247, opened 2021-04-28
chore: generate metadata for OpenApi services

jjtang1985:
- [x] metadata files are generated with the same generatoroption as odata
- [x] a package `generator-common` is created for shared functions, which is not protocol specific.
- [x] ESLint config: `no-duplicate-imports` is disabled and use `import/no-duplicates` for fixing false positive findings. [See ESLint issue](https://github.com/eslint/eslint/issues/13669)


.github/workflows/build.yml
```diff
@@ -30,9 +30,6 @@ jobs:
             */*/node_modules
           key: ${{ matrix.os }}-${{ hashFiles('**/yarn.lock') }}
       - run: yarn install --frozen-lockfile
-        if: steps.cache.outputs.cache-hit != 'true'
-      - run: yarn lerna run compile
-        if: steps.cache.outputs.cache-hit == 'true'
```
jjtang1985, line 35: I have to disable this optimisation, as the compilation fails.

packages/generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -1,43 +1,50 @@
...
-import { getGeneratorVersion } from './pregenerated-lib';
 
 export async function getGenerationAndUsage(
   service: VdmServiceMetadata
 ): Promise<GenerationAndUsage> {
   return {
-    genericUsage: await getGenericUsage(),
-    apiSpecificUsage: await getApiSpecificUsage(service),
+    genericUsage: getGenericUsage(),
+    apiSpecificUsage: getApiSpecificUsage(service),
```
FrankEssenberger, line 20: I think the links also contain a generation manual which depends if it is OData or OpenApi

test-packages/e2e-tests/test/sdk-metadata.spec.ts
```diff
@@ -19,10 +18,14 @@ describe('sdk-metadata', () => {
   });
 
   it('gives instruction with working link', async () => {
-    expect(getGenerationSteps().instructions).toContain(
-      linkGenerationDocumentaion
-    );
-    checkUrlExists(linkGenerationDocumentaion);
+    expect(
```
FrankEssenberger, line 21: you could add similar tests for the openapi metadata that the generator link is ok and all other specific links work.

packages/openapi-generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -0,0 +1,61 @@
...
+      apiWithOperations.name,
+      operation.operationId,
+      openApiDocument.npmPackageName
+    );
+    return { instructions, text: apiSpecificUsageText };
+  }
+  return { instructions: '', text: apiSpecificUsageText };
+}
+
+const installCommand = '';
```
FrankEssenberger, line 58: Empty seem strange.

packages/openapi-generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -0,0 +1,61 @@
...
+const generateCommand = '';
```
FrankEssenberger, line 59: Empty seem strange.

packages/openapi-generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export async function getGenerationAndUsage(
+  openApiDocument: OpenApiDocument
+): Promise<GenerationAndUsage> {
+  return {
+    genericUsage: getGenericUsage(),
+    apiSpecificUsage: getApiSpecificUsage(openApiDocument),
+    links: getLinks(),
+    generationSteps: getGenerationSteps(
+      installCommand,
```
FrankEssenberger, line 22: if this is not used anywhere else just write it inline instead of a const?

FrankEssenberger: same would aplly to the OData version, but it is also a matter of taste if you use a constant or put it to inline.

packages/openapi-generator/src/sdk-metadata/sdk-metadata.ts
```diff
@@ -0,0 +1,25 @@
...
+import { packageDescription } from './package-description';
+import { getGenerationAndUsage } from './generation-and-usage';
+
+export async function sdkMetaDataJS(
+  openApiDocument: OpenApiDocument,
+  options: GeneratorOptions
+): Promise<Client> {
+  const [pregeneratedLibrary, generationAndUsage] = await Promise.all([
+    getPregeneratedLibrary(
+      packageDescription(openApiDocument.npmPackageName),
```
FrankEssenberger, line 17: I think we should add a follow up ticket to simply the description. For OData we use in one if the s4HanaCloud option is there some specific test. However, if we start creating and publishing also other OData services we need to adjust this anyway.

FrankEssenberger (approved):
LGTM - I think two strings need to be set but with the generator-common package we start to get some reuse. We can now also tackle our duplicated code issues.

packages/openapi-generator/src/sdk-metadata/generation-and-usage.ts
jjtang1985: good catch

PR #1249, opened 2021-04-29
feat: Update service config behavior

marikaner:
This PR is mainly adds the behavior we specified for the `per-service-config`. However, I also refactored some parts along this:
* Added per service configuration
  * Only read if path is provided
  * Always written if path is provided
* Reduced some of the information stored in the openapi document
* Reduced duplicate defaults (ParsedGeneratorOptions)
  * The defaults specified in the cli/index.ts where specified in many other places as well, there should be no duplication anymore.
* Moved tsConfig to be read only once in the beginning, not for every service separately
* Updated package.json behavior as well.
* Removed sdk prefix from metadata
* Fixed occurrences of metaData => metadata


packages/openapi-generator/src/options/generator-options.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+        input: 'inputDir',
+        outputDir: 'outputDir',
+        perServiceConfig: 'someDir/config.json'
+      })
+    ).toMatchObject({
+      perServiceConfig: `${process.cwd()}/someDir/config.json`
+    });
+  });
+
+  it('parses per service config file for directory path', () => {
```
jjtang1985, line 37: Should we also have similar tests (fill path/dir path) but without `perServiceConfig`?

packages/openapi-generator/src/cli/index.ts
```diff
@@ -62,27 +61,29 @@ $ generate-openapi-client --input ./my-spec.yaml --outputDir ./client --transpil
...
     }),
-    serviceMapping: flags.string({
+    perServiceConfig: flags.string({
       description:
-        'Set the path to the service mapping file. By default, a `service-mapping.json` is generated in the input directory. The service mapping ensures consistent names between multiple generation runs with updated service definitions.',
-      helpValue: '<path/to/custom-service-mapping.json>'
+        'Set the path to a file containing the configuration per service. The configuration allows to set a `directoryName` and `packageName` for every service. It also makes sure that names do not change between generator runs. If a directory is passed, a `per-service-config.json` file is read/created in this directory.',
+      helpValue: '<path/to/per-service-config.json>',
+      parse: input =>
+        extname(input)
```
jjtang1985, line 84: checking extension name might not be enough, because you might pass `config` without extension and is used as folder here.

packages/openapi-generator/README.md
```diff
@@ -31,47 +31,45 @@ USAGE
   $ generate-openapi-client --input <input> --outputDir <outputDirectory>
 
 OPTIONS
-  -i, --input=<path/to/input>                             (required) Specify the path to the directory or file
```
jjtang1985, line 34: I assume this is generated without manual adjustment.

packages/openapi-generator/src/openapi-types.ts
```diff
@@ -14,24 +15,9 @@ export interface OpenApiDocument {
...
-  /**
-   * Name of the original specification file.
-   */
-  originalFileName: string;
-
-  /**
-   * Path of the original specification file.
-   */
-  filePath: string;
+  serviceConfig: ServiceConfig;
```
jjtang1985, line 21: Question: 4 properties are deleted here, but the `ServiceConfig` only contains `packageName` and `directoryName`. Are the missing two properties not useful?
I know we can get e.g., `filePath`, but if we dont need to reuse it, I'm happy to delete.

FrankEssenberger (commented):
why the test-services in core are removed?

jjtang1985:
It would be nicer to split the PR to small pieces.

packages/openapi-generator/README.md
```diff
@@ -31,47 +31,45 @@ USAGE
...
-                                                          generated, when transpilation is enabled (`--transpile`).
+  -i, --input=<path/to/input>
+      (required) Specify the path to the directory or file containing the OpenAPI service definition(s) to generate 
+      clients for. Accepts Swagger and OpenAPI definitions as YAML and JSON files. Throws an error if the path does not 
+      exist.
+
+  -o, --outputDir=<path/to/output>
+      (required) Specify the path to the directory to generate the client(s) in. Each client is generated into a 
+      subdirectory within the given output directory. Creates the directory if it does not exist. Customize subdirectory 
+      naming through `--serviceMapping`.
```
FrankEssenberger, line 42: perServiceConfig?

package.json
```diff
@@ -58,6 +58,7 @@
     "analytics": "yarn workspace @sap-cloud-sdk/analytics",
     "core": "yarn workspace @sap-cloud-sdk/core",
     "generator": "yarn workspace @sap-cloud-sdk/generator",
+    "generator-common": "yarn workspace @sap-cloud-sdk/generator-common",
```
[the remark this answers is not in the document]
marikaner: to run tests, e.g. `yarn generator-common test` nicer than `yarn workspace @sap-cloud-sdk/generator-common test`

packages/openapi-generator/README.md
```diff
@@ -31,47 +31,45 @@ USAGE
...
+      generation is generated.
+
+  --perServiceConfig=<path/to/per-service-config.json>
+      Set the path to a file containing the configuration per service. The configuration allows to set a `directoryName` 
+      and `packageName` for every service. It also makes sure that names do not change between generator runs. If a 
+      directory is passed, a `per-service-config.json` file is read/created in this directory.
+
+  --[no-]strictNaming
+      By default, the generation fails, when there are duplicate names for operations and/or path parameters after 
+      transforming them to camel case. Set this to true to enable unique name generation. The names are then be generated
```
FrankEssenberger, line 66: I thought you have set this to `false` so that we get `enable unique name generation.`

packages/generator/src/metadata/metadata.spec.ts
```diff
@@ -1,16 +1,16 @@
 import nock = require('nock');
 import { createOptions } from '../../test/test-util/create-generator-options';
 import { getTestService } from './pregenerated-lib.spec';
-import { sdkMetaDataJS } from './sdk-metadata';
+import { metadata } from './metadata';
```
FrankEssenberger, line 4: I think this is a fair refactoring, but I would like to see it in a seperate PR the next time so that we have smaller number of files to review.

packages/openapi-generator/src/generator.ts
```diff
@@ -44,17 +49,26 @@ const logger = createLogger('openapi-generator');
  * @param options Options to configure generation.
  */
 export async function generate(options: GeneratorOptions): Promise<void> {
-  options.serviceMapping =
-    options.serviceMapping ||
-    resolve(options.input.toString(), 'service-mapping.json');
+  return generateWithParsedOptions(parseGeneratorOptions(options));
```
FrankEssenberger, line 52: I do not like that in the generate method there is something done parsing the options. This should take place in my opinion in the command (cli/index.ts). I found it also hard to understand what is the difference between the options and parsed options from the `ParsedGeneratorOptions` type.

packages/openapi-generator/src/generator.ts
```diff
@@ -74,94 +93,55 @@ export async function generate(options: GeneratorOptions): Promise<void> {
       ? 'Some clients could not be generated.'
       : 'Could not generate client.';
   await finishAll(promises, errorMessage);
+
+  if (options.perServiceConfig) {
+    writeFile(
```
FrankEssenberger, line 98: I am confused. The getPerServiceConfig just reads it and than we write it here. Where is the point where it is initialized with default values if it is not there?

packages/openapi-generator/src/generator.ts
```diff
@@ -64,7 +78,12 @@ export async function generate(options: GeneratorOptions): Promise<void> {
     return generateService(
       inputFilePath,
       options,
-      vdmMapping,
+      getOrCreateServiceConfig(
```
FrankEssenberger, line 81: ah here it is. For me it would be cleaner to use a readOrGetDefault in line 70 and not not change the value behind the const `perServiceConfig`.

packages/openapi-generator/src/generator.ts
FrankEssenberger: Ah, part of the confusion is related to the fact that the `getPerServiceConfig` return a mapping for all services. I did not get this at first. Perhaps you can adjust the name to be plural?

packages/openapi-generator/src/options/per-service-config.ts
```diff
@@ -0,0 +1,66 @@
+import { existsSync, promises } from 'fs';
+import { relative } from 'path';
+
+const { readFile } = promises;
+
+/**
+ * Represents the service configuration for all services.
+ */
+export interface PerServiceConfig {
```
FrankEssenberger, line 9: From the name I thought it is a the config for one service, object below. I would perhaps just use a list of ServiceConfig[]. I guess filename is the name of the service specification file?

packages/openapi-generator/src/generator.ts
FrankEssenberger: or use a list of service configs and use a find to get the right one in the getOrCreateServiceConfig()

packages/openapi-generator/src/generator.ts
```diff
@@ -74,94 +93,55 @@ export async function generate(options: GeneratorOptions): Promise<void> {
       ? 'Some clients could not be generated.'
       : 'Could not generate client.';
   await finishAll(promises, errorMessage);
+
+  if (options.perServiceConfig) {
+    writeFile(
+      options.perServiceConfig,
```
FrankEssenberger, line 99: This can be the path to the file or a folder. In case it is a folder a name should be set right?

FrankEssenberger: Ah I found that in the cli parse step this is already done.

packages/openapi-generator/src/options/per-service-config.ts
```diff
@@ -0,0 +1,66 @@
...
+      : {};
+  }
+}
+
+export function getOrCreateServiceConfig(
+  perServiceConfig: PerServiceConfig | undefined,
+  inputFilePath: string,
+  serviceName: string
+): ServiceConfig {
+  const relativeFilePath = relative(process.cwd(), inputFilePath);
```
FrankEssenberger, line 51: Should we consider doing a spread here i.e. adding all non existing keys?

packages/openapi-generator/src/options/generator-options.ts
```diff
@@ -0,0 +1,40 @@
...
+  metadata?: boolean;
+}
+
+export type ParsedGeneratorOptions = typeof GenerateOpenApiClient extends Parser.Input<
+  infer F
+>
+  ? F
+  : never;
+
+export function parseGeneratorOptions(
```
FrankEssenberger, line 25: a doc would help. Is this needed? Should oclif not set the default?

FrankEssenberger (changes requested):
I have two main concerns:
- The PerSerivceConfig part was hard to understand for me since I did not get which object contains all configs and which is a single one. Some minor small things
- The parsing seems ugly and scattered. I would prefer to have it done in one central point not in the generator but the CLI. But I guess there are reasons so explanation would be helpful.

General: Try to split in three PRs the next time: PerServiceConfig, metadata renaming and generateDoc/Tsconfig adjustments.

packages/generator-common/src/metadata/metadata-types.ts
```diff
@@ -1,11 +1,9 @@
 /**
  * Represents the language independent header data of the sdk metadata
  */
-export interface SdkMetadataHeader {
+export interface MetadataHeader {
```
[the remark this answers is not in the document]
marikaner: Yes, I went a little crazy on the renaming here. I shouldn't have done this on this PR.

Just to clarify, I changed everything in the code to be without sdk prefix, but the generated directory is still `sdk-metadata`.
Would that resolve your concern? If not let's have a quick discussion and make a decision.

test-packages/test-services/openapi/no-schema-service/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "es5",
     "module": "commonjs",
-    "lib": ["esnext", "dom"],
+    "lib": ["esnext"],
```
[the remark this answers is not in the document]
marikaner: Yes, this was the old generator, this is not necessary anymore. I also removed the according test.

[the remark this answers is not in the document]
marikaner: (Reason was URLSearchParams were used for quer parameters)

test-packages/test-services/openapi/test-service/entity-api.d.ts
```diff
@@ -22,8 +22,8 @@ export declare const EntityApi: {
             | 'value2'
             | "valueWith'Quote"
             | undefined;
-          enumInt32Parameter?: 1 | 2 | undefined;
-          enumDoubleParameter?: 1 | 2 | undefined;
+          enumInt32Parameter?: 2 | 1 | undefined;
```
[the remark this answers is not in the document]
marikaner: I think it is not random and I believe the reason is that 'dom' is not there anymore (I have no other explanation) ;)
But, let's keep an eye on this.

packages/generator/src/metadata/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './generation-and-usage';
```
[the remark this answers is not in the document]
marikaner: But this is not exported to the root right? @FrankEssenberger

packages/generator-common/src/metadata/metadata-types.ts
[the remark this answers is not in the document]
jjtang1985: For the generated directory name, perhaps there is a contract between java team, which @FrankEssenberger can comment.
My concern is about the typescript type name, `MetadataHeader` does not say anything and from the import statement, I only see "metadata" without knowing which metadata.
```
import {MetadataHeader} from '@sap-cloud-sdk/generator-common'
```

test-packages/test-services/openapi/test-service/entity-api.d.ts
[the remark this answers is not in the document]
jjtang1985: Ok, let's see.

packages/generator/src/metadata/index.ts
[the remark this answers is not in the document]
jjtang1985: No, we agreed to treat it like the OData V4 beta case like the example below.
```
import {Metadata} from `@sap-cloud-sdk/openapi-generator/internal`
```

packages/generator-common/src/metadata/metadata-types.ts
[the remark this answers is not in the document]
marikaner: But `SdkMetadataHeader` doesn't tell me more either, right? Maybe it should be somethink like `ApiHubMetadata`?

packages/core/test/test-util/test-services/openapi/no-schema-service/README.md
```diff
@@ -1,16 +0,0 @@
-# no-schema-service
```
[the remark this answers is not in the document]
marikaner: They are still generated for e2e tests, I just removed them from the test-util in core (I guess there is more potential here).

packages/openapi-generator/src/options/generator-options.spec.ts
marikaner: Not sure, I understand this: You mean to check the default? Or other options?

packages/openapi-generator/src/cli/index.ts
marikaner: Good point, will fix.

marikaner: Actually, should we then always treat this as a file and refrain from adding a default name?

packages/openapi-generator/README.md
marikaner: Correct.

packages/openapi-generator/src/openapi-types.ts
marikaner: It is now passed as a separate parameters. I was considering this for the serviceConfig as well, but then decided against it to be pragmatic.

packages/generator-common/src/metadata/metadata-types.ts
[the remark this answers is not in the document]
jjtang1985: Good point, I was also thinking about the same name, but the metadata is generated for the api hub but reflects the type-safe client that we generate. So I guess `Sdk*` or `Client*` might be better than `ApiHub*`?

packages/openapi-generator/src/options/generator-options.spec.ts
jjtang1985: Yes, the purpose is to check the default behaviour with the two cases: file path and dir path.

packages/openapi-generator/src/generator.ts
marikaner: Actually this was done before already, only that it wasn't explicit.
I was also struggling with this, because I don't see a good way to unify the the commandline and programmatic use.
The cli parses the arguments (`this.parse`) and if the generation is invoked programatically, it should behave the same way.

Before, the default values where assigned/assumed everywhere in the code.

`GeneratorOptions` are the options that we allow to set when using the generator programatically. `ParsedGeneratorOptions` are the `GeneratorOptions`, only that default values and default parsing (e.g. `input => resolve(input)`) is already handled. You don't have to take care of optional values anymore.

packages/openapi-generator/src/generator.ts
marikaner: Correct 😊

packages/openapi-generator/src/cli/index.ts
marikaner: Forget this question. Done.

packages/openapi-generator/README.md
marikaner: Fixed

packages/generator/src/metadata/metadata.spec.ts
marikaner: I agree, sorry for that. I will be more careful in the future.

packages/openapi-generator/src/generator.ts
marikaner: A list is problematic, because we never need it in this format. It is currently a map from file name to `ServiceConfig`. As the parameter is called `perServiceConfig` it only makes sense to give the structure a name that relates to that.
I understand the confusion, but the plural of "configuration" is "configuration", so I am not sure how to improve that.

What I can do though is set the config explicitly. It will make the core more complex, but more explicit. I'll check what this means in practice and either just do it, or come back with questions.

packages/openapi-generator/src/generator.ts
marikaner: Let's keep this one for discussion: https://github.com/SAP/cloud-sdk-js/pull/1249#discussion_r622987222

packages/openapi-generator/src/options/per-service-config.ts
marikaner: https://github.com/SAP/cloud-sdk-js/pull/1249#discussion_r622987222

packages/generator/src/metadata/index.ts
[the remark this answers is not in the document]
FrankEssenberger: yes so export everythig under `internal`.

packages/openapi-generator/src/generator.ts
FrankEssenberger: > What I can do though is set the config explicitly. It will make the core more complex, but more explicit. I'll check what this means in practice and either just do it, or come back with questions.

I do not really understand what you mean with that but let's see.

marikaner:
> why the test-services in core are removed?

I changed the configuration a bit, so that in the unit test there are no readme's or additional files. We don't need them there. The test services are still there, though.

marikaner:
> It would be nicer to split the PR to small pieces.

image

packages/openapi-generator/src/options/generator-options.ts
marikaner: See my other comments. I added documentation as well 😊

packages/openapi-generator/src/options/per-service-config.ts
marikaner: https://github.com/SAP/cloud-sdk-js/pull/1249#discussion_r622987222

packages/openapi-generator/src/options/generator-options.spec.ts
marikaner: Added a few tests for extended behavior of the perserviceconfig.

packages/generator-common/src/metadata/metadata-types.ts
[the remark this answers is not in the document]
marikaner: Decision: `SdkMetadata`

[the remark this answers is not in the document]
marikaner: Also do this for the rest of the metadata functions, properties, etc.

packages/generator/src/metadata/index.ts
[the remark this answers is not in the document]
marikaner: Because interal, this can stay as is.

packages/openapi-generator/src/generator.ts
marikaner: Leave as is, but document more thoroughly.

packages/openapi-generator/src/generator.ts
marikaner: Always create the object in the code, always get the service config and in the end only write if the parameter was set.

marikaner: Rename: 
`openapi-generator --optionsPerService`
`OptionsPerService`
`ServiceOptions`

+ update ADR

packages/openapi-generator/src/options/per-service-config.ts
marikaner: Will consider this or create a follow up.

marikaner:
File is not generated.

FrankEssenberger (approved):
As discussed approved  under the condition that the open points are addressed so that the demo tomorrow can run on canary. I will do a rereview on main tomorrow morning.

packages/generator-common/src/metadata/metadata-types.ts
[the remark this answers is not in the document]
marikaner: check.

packages/openapi-generator/src/generator.ts
marikaner: check

packages/openapi-generator/src/generator.ts
marikaner: check

packages/openapi-generator/src/options/per-service-config.ts
marikaner: check

marikaner:
> As discussed approved under the condition that the open points are addressed so that the demo tomorrow can run on canary. I will do a rereview on main tomorrow morning.

Thanks, I adjusted everything let me know if you have additional findings (you can also comment on the closed pr)

PR #1250, opened 2021-04-29
chore: Rename generator command

marikaner:
Rename the generator command to `openapi-generator`


FrankEssenberger (approved):
LGTM - please add a competability note even if openapi generator was still experimental. Do we have a task to collect the things we have to do for OData part when we do sdk 2.0?

marikaner:
> LGTM - please add a competability note even if openapi generator was still experimental. Do we have a task to collect the things we have to do for OData part when we do sdk 2.0?

Updated the according ticket for this.

PR #1251, opened 2021-04-29
feat: Introduce verbose logging

marikaner:
Log only success an error messages per service by default, but allow setting verbosity for more information.


FrankEssenberger (approved):
LGTM - add a entry in the changelog and should we perhaps add the same for the OData generator?

PR #1252, opened 2021-04-29
feat: Add overwrite option

marikaner:
This PR adds an `overwrite` option, so that by default files will not be overwritten.


jjtang1985 (approved):
Add a compatibility notes, as overwrite needs to be enabled for some users?

PR #1255, opened 2021-04-30
chore: disable node 10

jjtang1985:
Today is the last day of the node 10 lifecycle.
https://nodejs.org/en/about/releases/


FrankEssenberger (approved):
LGTM

FrankEssenberger:
@jjtang1985 when should we add node 14?

jjtang1985:
> when should we add node 14?
There is a backlog item. I guess we need to prioritise then.

PR #1256, opened 2021-04-30
chore: Negate strictNaming option (skipValidation)

marikaner:
Rename the `strictNaming` option and change the default to `false`.


jjtang1985 (approved):
LGTM

jjtang1985:
it might be better to use the new name of the option as part of the PR name.

FrankEssenberger (approved):
LGTM

PR #1257, opened 2021-04-30
followup openapi generator merge

FrankEssenberger:
I actually had only small suggestions when I looked at the merge from yesterday, but than it turned out to be a bit more. Please have a look @marikaner.


packages/openapi-generator/src/generator.spec.ts
```diff
@@ -16,7 +16,7 @@ describe('generator', () => {
     '../../../test-packages/test-services/openapi/test-service'
   );
 
-  it('getSdkVersion returns a valid stable version', async () => {
+  xit('getSdkVersion returns a valid stable version', async () => {
```
marikaner, line 19: Why?

packages/openapi-generator/src/generator.spec.ts
```diff
@@ -181,7 +185,7 @@ describe('generator', () => {
             `);
     });
 
-    it.only('does not fail when overwrite is enabled', async () => {
+    it('does not fail when overwrite is enabled', async () => {
```
marikaner, line 188: good catch O.o

packages/openapi-generator/src/options/options-per-service.ts
```diff
@@ -29,12 +42,73 @@ export interface ServiceOptions {
...
+  options: ParsedGeneratorOptions
+): Promise<OptionsPerService> {
+  const originalOptionsPerService = await getOriginalOptionsPerService(
+    options.optionsPerService
+  );
+
+  const uniqueNameGenerator = new UniqueNameGenerator('-');
+  const nonUniqueFiles: string[] = [];
+
+  const optionsPerSerivice: OptionsPerService = inputPaths.reduce(
```
marikaner, line 70:
```suggestion
  const optionsPerService: OptionsPerService = inputPaths.reduce(
```

packages/openapi-generator/src/options/options-per-service.ts
```diff
@@ -29,12 +42,73 @@ export interface ServiceOptions {
...
+      return optionsPerService;
+    },
+    {}
+  );
+
+  if (nonUniqueFiles.length > 0 && options.strictNaming) {
+    throw new Error(
+      `The following service specs lead to non unique file names: ${nonUniqueFiles.join(
+        ','
+      )}. You can either introduce/adjust a operions-per-service.json or disable the strictNaming flag.`
```
marikaner, line 96: * Consider the new name: skipValidation
* I would suggest to decide for one verb. We can throw different errors depending on whether an original file was given or not.
* The file does not have to be named options-per-service.json, therefore we should avoid mentioning it here.

packages/openapi-generator/src/options/options-per-service.ts
```diff
@@ -29,12 +42,73 @@ export interface ServiceOptions {
...
+        uniqueServiceName
+      );
+      return optionsPerService;
+    },
+    {}
+  );
+
+  if (nonUniqueFiles.length > 0 && options.strictNaming) {
+    throw new Error(
+      `The following service specs lead to non unique file names: ${nonUniqueFiles.join(
```
marikaner, line 94: This is going to be hard to read, add `\n` before the list and between the file paths.

packages/openapi-generator/src/options/options-per-service.ts
```diff
@@ -1,11 +1,20 @@
...
+import { AbsoluteAndRelativePath } from '../generator';
+import { ParsedGeneratorOptions } from './generator-options';
 
 const { readFile } = promises;
 
 /**
  * Represents the service options for all services, mapped from the input file path to the service configuration.
  */
-export type OptionsPerService = Record<string, Partial<ServiceOptions>>;
+export type OptionsPerService = Record<string, ServiceOptions>;
```
marikaner, line 12: I understand that this is the actual return type of `getOptionsPerService`, but is this ever relevant? After it was created we never really use it and it might be confusing to have two types here.

packages/openapi-generator/src/options/options-per-service.ts
```diff
@@ -20,6 +29,10 @@ export interface ServiceOptions {
    * Name of the package to reference in the package.json.
    */
   packageName: string;
+  /**
+   * Name of the service
+   */
+  serviceName: string;
```
marikaner, line 35: 👍 Great idea.

packages/openapi-generator/src/options/options-per-service.ts
```diff
@@ -29,12 +42,73 @@ export interface ServiceOptions {
...
+export async function getOptionsPerService(
+  inputPaths: AbsoluteAndRelativePath[],
+  options: ParsedGeneratorOptions
+): Promise<OptionsPerService> {
+  const originalOptionsPerService = await getOriginalOptionsPerService(
+    options.optionsPerService
+  );
+
+  const uniqueNameGenerator = new UniqueNameGenerator('-');
+  const nonUniqueFiles: string[] = [];
```
marikaner, line 68: Consider removing the negation from this name. `pathsForDuplicateServiceNames` (or something shorter ;) )

packages/openapi-generator/src/options/options-per-service.ts
```diff
@@ -29,12 +42,73 @@ export interface ServiceOptions {
...
+/**
+ * Get the options per service for the list of services.
+ * If optionsPerServicePath is not given default values are used for the services.
+ * If optionsPerServicePath is given exisiting values for the services are.
+ * @param inputPaths Paths for the service spec files
+ * @param options Options of the generator
+ * @returns The parsed configuration for all services in relativeServicePaths.
+ */
+export async function getOptionsPerService(
+  inputPaths: AbsoluteAndRelativePath[],
```
marikaner, line 60: pass the relative path directly to avoid `.relativePath` everywhere.

packages/openapi-generator/src/parser/document.ts
```diff
@@ -28,7 +26,7 @@ export async function parseOpenApiDocument(
 
   return {
     apis: parseApis(document, refs, options),
-    serviceName: pascalCase(serviceName),
+    serviceName: pascalCase(serviceOptions.serviceName),
```
marikaner, line 29: I think the service name in the service options should already be pascalCase, don't you think? If users use a non-pascal case name here I would say that's ok.

packages/openapi-generator/src/options/options-per-service.spec.ts
```diff
@@ -75,7 +80,139 @@ describe('getServiceOptions', () => {
...
+              "path/service": Object {
+                "directoryName": "service",
+                "packageName": "service",
+                "serviceName": "service",
+              },
+            }
+          `);
+  });
+
+  it('builds PerService config with options per service.', async () => {
```
marikaner, line 107: Probably a left over from the previous renaming, but from the name I really don't know what this test does. 
```suggestion
  it('builds options per service with existing config file', async () => {
```

Consider removing the notion of `PerService config` from the other tests as well, it implies that we have a class that is named like this.

marikaner: After looking at this for a while I think the reason is, that for me "options per service" is what I want to get out of this function.

packages/openapi-generator/src/options/options-per-service.spec.ts
```diff
@@ -75,7 +80,139 @@ describe('getServiceOptions', () => {
...
+            absolutePath: '/user/path1/service',
+            relativePath: 'path1/service'
+          },
+          { absolutePath: '/user/path2/service', relativePath: 'path2/service' }
+        ],
+        {
+          strictNaming: false
+        } as ParsedGeneratorOptions
+      )
+    ).resolves.toMatchInlineSnapshot(`
```
marikaner, line 203: Consider using `.toEqual({...})` here, because this is easier to maintain. Also applies to the other tests.

packages/openapi-generator/src/generator.ts
```diff
@@ -275,26 +245,44 @@ async function generateService(
...
     ],
-    Promise.resolve([])
+    Promise.resolve([] as AbsoluteAndRelativePath[])
   );
 }
 
+export interface AbsoluteAndRelativePath {
+  relativePath: string;
+  absolutePath: string;
+}
```
marikaner, line 284: I am not a fan of this. Seems more complex than before. Why did you decide to go this way?

marikaner (changes requested):
I like the additional error cases and the fact that the service name is part of the service mapping.
I am not a big fan of the path handling, though.

packages/openapi-generator/src/generator.spec.ts
FrankEssenberger: I think this reads a file?

packages/openapi-generator/src/generator.ts
FrankEssenberger: hm I did not want to calculate the relative path multiple times.

packages/openapi-generator/src/options/options-per-service.ts
FrankEssenberger: put it to a variable so it is shorter. If you include the relative path calc in the method it will work with absolute and relative as input which is a bit more resilient I think.

packages/openapi-generator/src/generator.spec.ts
FrankEssenberger: locally if i run jest it works. There was a //TODO that all the tests rely on state of files beeing created which is not a unit test. Hence I disabled the tests where this is the case.

marikaner: Hmm, but shouldn't we then rewrite the tests to not access any files? I have recently used mock-fs a bit more extensively and I think this should actually work for most of the tests.

marikaner: As discussed, consider reenabling the tests and "fixing" them on a separate PR.

FrankEssenberger: ok I make one more follow up task :-) in a separate PR.

packages/openapi-generator/src/options/options-per-service.ts
```diff
@@ -29,12 +43,77 @@ export interface ServiceOptions {
...
+        originalOptionsPerService,
+        relativePath,
+        uniqueServiceName
+      );
+      return previousOptions;
+    },
+    {}
+  );
+
+  if (duplicateServicePaths.length > 0 && options.strictNaming) {
```
marikaner, line 95:
```suggestion
  if (!options.skipValidation && duplicateServicePaths.length) {
```

packages/openapi-generator/src/options/options-per-service.ts
```diff
@@ -1,11 +1,21 @@
...
 /**
  * Represents the service options for all services, mapped from the input file path to the service configuration.
  */
-export type OptionsPerService = Record<string, Partial<ServiceOptions>>;
+export type OptionsPerService = Record<string, ServiceOptions>;
+
+/**
+ * Partial OptionsPerService
+ */
+type PartialOptionsPerService = Record<string, Partial<ServiceOptions>>;
```
marikaner, line 18: I think this comment got lost and is on the wrong line:https://github.com/SAP/cloud-sdk-js/pull/1257/files#r623923247

packages/openapi-generator/src/options/options-per-service.ts
FrankEssenberger: I removed the export for the partial config since it is only used internally.

marikaner: That doesn't really address my concern, but it is minor anyway. Let's leave it at that 😊

marikaner (approved):
LGTM

PR #1258, opened 2021-04-30
try something

marikaner:

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -6,6 +6,7 @@
 name: 'code-ql'
 
 on:
+  pull_request: ~
   schedule:
     - cron: '0 20 * * *'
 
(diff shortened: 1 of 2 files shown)


packages/generator/src/name-formatting-strategies.ts
```diff
@@ -52,6 +52,8 @@ function nextSuffix(name: string, previouslyGeneratedNames: string[]): number {
   );
   const lastElem = last(sortedList);
   const match = lastElem ? lastElem.match(/(\d+)$/) : null;
+  const test1 = lastElem ? lastElem.match(/(\d+)$/) : null;
```
marikaner, line 55:
```suggestion
  const test1 = lastElem ? lastElem.match(/(?<!\d)(\d+)$/) : null;
```

.github/workflows/codeql-analysis.yml
```diff
@@ -6,6 +6,7 @@
 name: 'code-ql'
 
 on:
+  pull_request: ~
   schedule:
     - cron: '0 20 * * *'
```
marikaner, line 11:
```suggestion
```

PR #1260, opened 2021-05-03
chore: update typedoc in the client

jjtang1985:
The `typedoc` version of the generated client is up-to-date.

- [ ] adjust compatibility notes? I did not do it, since updating a generated client with new version of generator is not common? Typically, the version update is in the `package.json`, I would not overwrite my `package.json` for a working project so my `scripts` and other `dependencies` are gone. Feel free to challenge here.


FrankEssenberger (approved):
LGTM

PR #1261, opened 2021-05-03
chore: Fix metadata test to be sdk version independent

marikaner:
This test fails with other versions than 1.41.1


FrankEssenberger (approved):
LGTM

PR #1265, opened 2021-05-04
chore: Fix potential polynomial redos

marikaner:
Fix this: warning https://github.com/SAP/cloud-sdk-js/security/code-scanning/152?query=ref%3Arefs%2Fheads%2Fmain

### DIFF
--- a/packages/generator/src/name-formatting-strategies.ts
+++ b/packages/generator/src/name-formatting-strategies.ts
@@ -51,15 +51,15 @@ function nextSuffix(name: string, previouslyGeneratedNames: string[]): number {
     previouslyGeneratedNames.filter(n => n.startsWith(name))
   );
   const lastElem = last(sortedList);
-  const match = lastElem ? lastElem.match(/(\d+)$/) : null;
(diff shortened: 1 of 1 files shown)


FrankEssenberger (approved):

packages/generator/src/name-formatting-strategies.ts
```diff
@@ -51,15 +51,15 @@ function nextSuffix(name: string, previouslyGeneratedNames: string[]): number {
     previouslyGeneratedNames.filter(n => n.startsWith(name))
   );
   const lastElem = last(sortedList);
-  const match = lastElem ? lastElem.match(/(\d+)$/) : null;
+  const match = lastElem ? lastElem.match(/(?<!\d)(\d+)$/) : null;
```
FrankEssenberger, line 54: Just out of curiosity. The old regex said end with number the new said start with non numer (lookahead) and than end with number. Is it obvious why the evaluation is faster with the new one?

jjtang1985: question: I saw the reason why the old is slow from the link, but could you please answer my stupid question why the new one works?
My online friend (https://regexr.com) does not understand the new one;(

PR #1268, opened 2021-05-04
use compiler API

FrankEssenberger:
Use the compiler API. This could become handy when we want to mock all file system interaction in the generation process.


packages/generator-common/src/compiler.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+    await transpileDirectory('test-src', compilerConfig);
+    const files = await promises.readdir('test-dist');
+    expect(files).toIncludeAnyMembers([
+      'file.js',
+      'file.d.ts',
+      'file.d.ts.map'
+    ]);
+    const filesSubfolder = await promises.readdir('test-dist/subfolder');
+    expect(filesSubfolder).toIncludeAnyMembers([
+      'file.js',
```
jjtang1985, line 114: minor: maybe rename the test files with different names to make sure test works.

packages/generator-common/src/compiler.ts
```diff
@@ -0,0 +1,139 @@
...
+      `Compilation Errors:${EOL}${getErrorList(allDiagnostics).join(EOL)}`
+    );
+  }
+  logger.verbose(`Transpiling files in directory: ${path} finished.`);
+}
+
+function getErrorList(diagnostis: Diagnostic[]): string[] {
+  return diagnostis.map(
+    diagnostic =>
+      `${diagnostic.file?.fileName}:${diagnostic.start}:${diagnostic.length} - error TS${diagnostic.code}: ${diagnostic.messageText}`
```
jjtang1985, line 48: question:
depending on the value of `diagnostic.file?.fileName`, we use `diagnostic.start` or `diagnostic.length`.
However, there is only one test for the error message.
Could you please share more about `diagnostic.start` and `diagnostic.length`, because I see the type of these two are both number, while in the test, it's like a file path.
Maybe even add a test for the missing one?

packages/generator-common/src/compiler.ts
```diff
@@ -0,0 +1,139 @@
...
+    throw new Error(`No tsconfig found under path ${fullPath}`);
+  }
+  const options: CompilerOptions =
+    JSON.parse(
+      await promises.readFile(fullPath, {
+        encoding: 'utf8'
+      })
+    )['compilerOptions'] || {};
+
+  // TODO map other values to enum later as well
```
jjtang1985: I see in line 96, this is done?

packages/generator-common/src/compiler.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      'file.d.ts',
+      'file.d.ts.map'
+    ]);
+  });
+
+  it('throws on broken source file', async () => {
+    await expect(transpileDirectory('broken-src', compilerConfig)).rejects
+      .toMatchInlineSnapshot(`
+            [Error: Compilation Errors:
+            /Users/d068544/WebstormProjects/cloud-sdk-second-checkout/packages/generator-common/broken-src/file.ts:17:3 - error TS2588: Cannot assign to 'foo' because it is a constant.]
```
jjtang1985, line 124: Local path seems to be wrong.

jjtang1985 (changes requested):

packages/generator-common/src/compiler.spec.ts
FrankEssenberger: the file name is the same but they are in different folders so I think this should be safe.

packages/generator-common/src/compiler.spec.ts
```diff
@@ -117,11 +117,22 @@ describe('compilation', () => {
...
-    await expect(transpileDirectory('broken-src', compilerConfig)).rejects
-      .toMatchInlineSnapshot(`
-            [Error: Compilation Errors:
-            /Users/d068544/WebstormProjects/cloud-sdk-second-checkout/packages/generator-common/broken-src/file.ts:17:3 - error TS2588: Cannot assign to 'foo' because it is a constant.]
-          `);
+  it('throws error with file information on broken source file', async () => {
+    await expect(
+      transpileDirectory('broken-src', compilerConfig)
+    ).rejects.toThrowError(
+      "broken-src/file.ts:17:3 - error TS2588: Cannot assign to 'foo' because it is a constant"
```
jjtang1985, line 124: As discussed, the `17:3` seems to be wrong.

jjtang1985 (approved):
LGTM

PR #1275, opened 2021-05-06
linter and tests

FrankEssenberger:


FrankEssenberger:
> LGTM.
> From the AC, I see one missing unit test?

When I wrote the AC I was not aware that on the getDestination level the no_proxy was already correctly there. Only for the two service calls it was not.

PR #1276, opened 2021-05-06
feat: support appending path to generated url for all request builders

jjtang1985:
The `appendPath` is implemented as a workaround for almost all unsupported odata functionalities.
Typically, it can help queryi navigation properties (`GET` `/TestEntity(key)/to_MultiLink`).
Unfortunately, however, this function is too powerful, so the deserialisation cannot work as the return type is not clear.
I disabled the `execute` function when `appendPath` is called, and the users should use `executeRaw` instead.

### DIFF
--- a/packages/core/src/odata-common/request-builder/request-builder-base.ts
+++ b/packages/core/src/odata-common/request-builder/request-builder-base.ts
@@ -132,6 +132,12 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
+  /**
+   * Append additional path to the generated path. Typically, this can be used for querying navigation properties of an entity.
+   * When this function is used, the "execute" function of the request builder is disabled and use "executeRaw" instead.
+   * @param path Path to be appended.
+   * @returns The request builder itself without "execute" function, to facilitate method chaining.
+   */
   appendPath(path: string): Omit<this, 'execute'> {
     this.requestConfig.appendPath(path);
     return this;
--- a/packages/core/src/odata-common/request-builder/request-builder-base.ts
+++ b/packages/core/src/odata-common/request-builder/request-builder-base.ts
@@ -132,9 +132,9 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
-  /**
-   * Append additional path to the generated path. Typically, this can be used for querying navigation properties of an entity.
-   * When this function is used, the "execute" function of the request builder is disabled and use "executeRaw" instead.
+   * Append the given path to the URL.
+   * This can be used for querying navigation properties of an entity.
+   * To execute a request with an appended path use `executeRaw` to avoid errors during deserialization. When using this, the `execute` method is omitted from the return type.
    * @param path Path to be appended.
    * @returns The request builder itself without "execute" function, to facilitate method chaining.
    */
--- a/packages/core/src/odata-common/request-builder/request-builder-base.ts
+++ b/packages/core/src/odata-common/request-builder/request-builder-base.ts
@@ -132,6 +132,11 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
+  appendPath(path: string): Omit<this, 'execute'> {
+    this.requestConfig.appendPath(path);
+    return this;
+  }
+
   /**
    * Skip fetching csrf token for this request, which is typically useful when the csrf token is not required.
    * @returns The request builder itself, to facilitate method chaining.
--- a/packages/core/src/odata-common/request-builder/request-builder-base.ts
+++ b/packages/core/src/odata-common/request-builder/request-builder-base.ts
@@ -132,14 +132,15 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
+  /**
    * Append the given path to the URL.
    * This can be used for querying navigation properties of an entity.
    * To execute a request with an appended path use `executeRaw` to avoid errors during deserialization. When using this, the `execute` method is omitted from the return type.
    * @param path Path to be appended.
    * @returns The request builder itself without "execute" function, to facilitate method chaining.
    */
-  appendPath(path: string): Omit<this, 'execute'> {
-    this.requestConfig.appendPath(path);
+  appendPath(...path: string[]): Omit<this, 'execute'> {
+    this.requestConfig.appendPath(...path);
     return this;
   }
 
--- a/packages/core/src/odata-common/request/odata-request-config.ts
+++ b/packages/core/src/odata-common/request/odata-request-config.ts
@@ -25,7 +25,7 @@ export abstract class ODataRequestConfig {
   private _customHeaders: Record<string, string> = {};
   private _customQueryParameters: Record<string, string> = {};
   private _customRequestConfiguration: Record<string, string> = {};
-  private _appendPaths: string[] = [];
+  private _appendedPaths: string[] = [];
   private _fetchCsrfToken = true;
 
   /**
@@ -100,8 +100,8 @@ export abstract class ODataRequestConfig {
     return this._customRequestConfiguration;
   }
 
-  get appendPaths(): string[] {
-    return this._appendPaths;
+  get appendedPaths(): string[] {
+    return this._appendedPaths;
   }
 
   set fetchCsrfToken(fetchCsrfToken: boolean) {
@@ -148,8 +148,8 @@ export abstract class ODataRequestConfig {
     });
   }
 
-  appendPath(path: string): void {
-    this.appendPaths.push(path);
+  appendPath(...path: string[]): void {
+    this.appendedPaths.push(...path);
   }
 
   protected prependDollarToQueryParameters(
--- a/packages/core/src/odata-common/request/odata-request-config.ts
+++ b/packages/core/src/odata-common/request/odata-request-config.ts
@@ -25,6 +25,7 @@ export abstract class ODataRequestConfig {
   private _customHeaders: Record<string, string> = {};
   private _customQueryParameters: Record<string, string> = {};
   private _customRequestConfiguration: Record<string, string> = {};
+  private _appendPaths: string[] = [];
   private _fetchCsrfToken = true;
 
   /**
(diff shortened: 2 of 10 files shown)


packages/core/src/odata-common/request-builder/request-builder-base.ts
```diff
@@ -132,6 +132,17 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
+  /**
+   * Append additional path to the generated path. Typically, this can be used for querying navigation properties of an entity.
+   * When this function is used, the "execute" function of the request builder is disabled and use "executeRaw" instead.
+   * @param path Path to be appended.
+   * @returns The request builder itself without "execute" function, to facilitate method chaining.
+   */
```
marikaner, line 140: "generated" is not the right word in my opinion. The third sentence was also a bit bumpy, consider this:
```suggestion
   * Append the given path to the URL.
   * This can be used for querying navigation properties of an entity.
   * To execute a request with an appended path use `executeRaw` to avoid errors during deserialization. When using this, the `execute` method is omitted from the return type.
   * @param path Path to be appended.
   * @returns The request builder itself without "execute" function, to facilitate method chaining.
   */
```

packages/core/src/odata-common/request/odata-request-config.ts
```diff
@@ -25,6 +25,7 @@ export abstract class ODataRequestConfig {
   private _customHeaders: Record<string, string> = {};
   private _customQueryParameters: Record<string, string> = {};
   private _customRequestConfiguration: Record<string, string> = {};
+  private _appendPaths: string[] = [];
```
marikaner, line 28: The naming here sounds like a boolean  i.e. "should I append paths?"
Consider `_appendedPaths`, `_customPaths` or something in this direction.

packages/core/src/odata-common/request/odata-request.ts
```diff
@@ -55,7 +55,9 @@ export class ODataRequest<RequestConfigT extends ODataRequestConfig> {
    * @returns The absolute URL for the request
    */
   url(): string {
-    return `${removeTrailingSlashes(this.resourceUrl())}${this.query()}`;
+    return `${removeTrailingSlashes(this.resourceUrl())}${removeTrailingSlashes(
```
marikaner, line 58: Don't you need a slash in between the resource url and the appended path?

marikaner: Please also add a test for this.

packages/core/src/odata-v2/request-builder/get-by-key-request-builder.spec.ts
```diff
@@ -119,6 +140,34 @@ describe('GetByKeyRequestBuilder', () => {
       expect(actual.data.d).toEqual(entityData);
       expect(actual.request.method).toBe('GET');
     });
+
+    it('returns navigation properties', async () => {
```
marikaner, line 144: I think this test is unnecessary. In the other test we are checking that the url is correct. In this one we are mocking data for this url and checking that this data comes back when invoking the url. As we also have an E2E test I don't see a good reason for this test.
Is there something tested that I don't see?

test-packages/e2e-tests/test/request-builder.spec.ts
```diff
@@ -64,6 +64,25 @@ describe('Request builder', () => {
...
+  it('should return one to many navigation property of an entity', async () => {
+    const multiLinks = (
+      await TestEntity.requestBuilder()
+        .getByKey(101)
+        .appendPath('/ToMultiLink')
+        .executeRaw(destination)
+    ).data.value as any[];
+    const actual = multiLinks.map(
+      multiLink =>
+        deserializeEntityV4(multiLink, TestEntityLink) as TestEntityLink
```
marikaner, line 76: Why is `as TestEntityLink` necessary?

marikaner (changes requested):
Mostly wording and naming, but I also think we should consider usage without a leading slash. (see comments)

packages/core/src/odata-common/request/odata-request-config.ts
```diff
@@ -99,6 +100,10 @@ export abstract class ODataRequestConfig {
     return this._customRequestConfiguration;
   }
 
+  get appendPaths(): string[] {
```
FrankEssenberger, line 103: see comment from Marika above

packages/core/src/odata-common/request/odata-request.ts
jjtang1985: Here I have a [test](https://github.com/SAP/cloud-sdk-js/pull/1276/files#diff-6cf9bb4dc4da9e788a878f6ee227258a0b6b3d5e44e2057fd85f0a8bdf32f950R22).

jjtang1985: > Don't you need a slash in between the resource url and the appended path?

This is about whether the user should provide `/` as part of the parameter, currently here is one example: 
`.appendPath('/to_MultiLink')`

packages/core/src/odata-common/request/odata-request-config.ts
```diff
@@ -143,6 +148,10 @@ export abstract class ODataRequestConfig {
     });
   }
 
+  appendPath(path: string): void {
```
FrankEssenberger, line 151: Could we use a `...string` for this method so that you can add one path or multiple in one call.

packages/core/src/odata-common/request/odata-request.ts
```diff
@@ -55,7 +55,9 @@ export class ODataRequest<RequestConfigT extends ODataRequestConfig> {
...
+      this.config.appendPaths.join('/')
```
FrankEssenberger, line 59: is ensured that the path which are joined here do not contain trailing slashes? Test would be good.

packages/core/src/odata-v2/request-builder/get-by-key-request-builder.spec.ts
jjtang1985: As you said, the  most unit tests in this file test the `url`/`execute` methods for verifying the final URL of the request.
As `appendPath` can only use `executeRaw`, this test is the replacement of the one for `execute`.

packages/core/src/odata-common/request-builder/request-builder-base.ts
```diff
@@ -132,6 +132,17 @@ export abstract class MethodRequestBuilder<
...
+  appendPath(path: string): Omit<this, 'execute'> {
```
FrankEssenberger, line 141: We changed the naming in the builder to a `addXYZ` and `setXYZ` which adds or sets something. This one is currently a `add` so perhaps follow this pattern @marikaner?

packages/core/src/odata-common/request-builder/request-builder-base.ts
FrankEssenberger: I would be perhaps a bit more general here: 
Typically, this can be used for building requests which are not yet supported by the request-builder e.g. querying navigation properties of an entity.

packages/core/src/odata-common/request-builder/request-builder-base.ts
jjtang1985: I personally would use `append` as a verb, because it clearly means you "add A to the **end** of B"

packages/core/src/odata-common/request/odata-request.ts
FrankEssenberger: what happens if user includes query parameters in path?

packages/core/src/odata-common/request/odata-request.ts
jjtang1985: I would then make this test as simple as possible and only check the URL from `url`/`executeRaw`.

packages/core/src/odata-v2/request-builder/get-by-key-request-builder.spec.ts
jjtang1985: I would then make this test as simple as possible and only check the URL from `url`/`executeRaw`.

packages/core/src/odata-common/request/odata-request.ts
jjtang1985: As discussed with @FrankEssenberger , based on the reasons below, the user needs to provide all the appended path:
- if the SDK adds the missing `/` in the beginning, it cannot handle use case when not `/` is needed
- as a workaround API, we aim to provide a powerful API instead of an intelligent solution
- also having everything from the parameter provided by the user is explicit.

marikaner: Almost convinced, still:
* Do you have an example of a use case, where the leading `/` is not needed?
* Isn't this inconsistent with the fact that we are removing trailing slashes? Should we consider to not remove trailing slashes? (See next question)
* What happens if someone uses `.appendPath('/')`? Currently it will basically be ignored, but should it?

packages/core/src/odata-v2/request-builder/get-by-key-request-builder.spec.ts
marikaner: But what is this test checking in addition to the one above? I don't see any difference except, that we check, that `executeRaw` is behaving the way we expect (which should already be tested in other tests as this is not new functionality).
Maybe it would be worth testing `relativeUrl` instead.

packages/core/src/odata-common/request/odata-request.ts
jjtang1985: As discussed, we remove the `removeTrailingSlashes` for appended path so users have the full power to control the url.

packages/core/src/odata-v2/request-builder/get-by-key-request-builder.spec.ts
jjtang1985: As discussed, this test will be kept for testing the `appendPath` API in the request builder. The tests for `releativeUrl` is added for lighter tests.

packages/core/src/odata-common/request-builder/request-builder-base.ts
```diff
@@ -132,6 +132,18 @@ export abstract class MethodRequestBuilder<
     return this;
   }
 
+  /**
+   * Append the given path to the URL.
```
FrankEssenberger, line 136: Explain in the docs that all strings are simply joined and appended to the path. No logic like adding slashes etc. is performend.

FrankEssenberger (approved):
LGTM only adjust the docs

PR #1282, opened 2021-05-10
chore: update typescript version in all clients

jjtang1985:
Update `typescript` version to `~4.1.2` in the `package.json` of the client


FrankEssenberger (approved):
As discussed make sense.

marikaner (approved):
Looks good

PR #1292, opened 2021-05-12
feature: config.json for openapi-generator parameters

tomfrenken:
Adds a feature to use a config file to pass all command line options for the openapi-generator.
- If a directory is used as the path, a `config.json` file is used by default.
- If other options than `-c/--config` are used, they overwrite the options that are set in the config file.


packages/openapi-generator/src/generator.ts
```diff
@@ -60,6 +60,9 @@ export async function generate(options: GeneratorOptions): Promise<void> {
 export async function generateWithParsedOptions(
   options: ParsedGeneratorOptions
 ): Promise<void> {
+  if (options.input === '' || options.outputDir === '') {
```
jjtang1985, line 63: Since these are not required, please add a test for the error message.

packages/openapi-generator/src/cli/index.ts
```diff
@@ -110,13 +113,24 @@ $ openapi-generator --input ./my-spec.yaml --outputDir ./client --transpile`
...
 
   async run(): Promise<void> {
     try {
       const parsed = this.parse(OpenApiGenerator);
-      await generateWithParsedOptions(parsed.flags);
+      if (parsed.flags.config) {
+        await generate(parseOptionsFromConfig(parsed.flags.config));
+      } else {
+        await generateWithParsedOptions(parsed.flags);
+      }
```
jjtang1985, line 133: This means we cannot use both config file and command line options, which is fine.
However, this is possible with the OData generator (not sure, who wins when reaching conflicts though).
We should make a decision for the alignment and make it explicit in the docs.

Personally, I'm not a big fan of using the `parse` (line 128) provided by the lib and using `parseOptionsFromConfig` (line 130) made by us together, which might be error prone.

packages/openapi-generator/src/generator-utils.ts
```diff
@@ -0,0 +1,6 @@
+import { readFileSync } from 'fs';
+import { GeneratorOptions } from './options';
+
+export function parseOptionsFromConfig(path: string): GeneratorOptions {
+  return JSON.parse(readFileSync(path, 'utf8'));
```
jjtang1985, line 5: As currently, we are not able to use config and command line options, I guess we should move the validation `if (options.input === '' || options.outputDir === '')` to here, as it returns `GeneratorOptions` that says `input` is not nullable.

jjtang1985 (commented):
I guess before moving forward we should align about the oclif and yargs usage.

However, for yargs, we also have our own parse function for config file [here](https://github.com/SAP/cloud-sdk-js/blob/main/packages/generator/src/generator-cli.ts#L39).

packages/openapi-generator/src/generator-utils.ts
tomfrenken: I think it's better if the check stays in `generateWithParsedOptions`, since `input` and `outputDir` are no longer required, it is possible to use no config but also not use `input  & `outputDir`, with this guard, both possiblities are checked. 🤔

jjtang1985: Good point.
Then this will be a point that `yargs` does better.
- When using config, we cannot mark `input` as required with `oclif` for the current implementation.
- We can still mark `inputDir` as required with `yargs` even using config.

tomfrenken: Agreed!

marikaner:
> here

marikaner:
> I guess before moving forward we should align about the oclif and yargs usage.
> 
> However, for yargs, we also have our own parse function for config file [here](https://github.com/SAP/cloud-sdk-js/blob/main/packages/generator/src/generator-cli.ts#L39).

Agreed, they both have pros and cons. As we have a ticket to align both generators to one tool, I think we can postpone the decision to there. What do you think?

jjtang1985:
> greed, they both have pros and cons. As we have a ticket to align both generators to one tool, I think we can postpone the decision to there. What do you think?

I'm fine to work on the follow-ups. Could you please point out the ticket so we can add more findings from this PR?

marikaner (changes requested):
I think this is a sensible workaround to handle oclif. However, I left you some comments to consider 😊

packages/openapi-generator/src/options/flags.ts
```diff
@@ -0,0 +1,100 @@
...
+  }),
+  metadata: flags.boolean({
+    description: 'When enabled, metadata for the API hub is generated.',
+    default: false,
+    hidden: true
+  }),
+  config: flags.string({
+    char: 'c',
+    parse: input => resolve(input),
+    description: 'Set the path to the config.json file for generation.',
```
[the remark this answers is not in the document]
tomfrenken: Agreed, I changed the function to work like the `tsconfig` function and added some context.

packages/openapi-generator/src/cli/index.ts
tomfrenken: Agreed, I've implemented it (with some additions) and now additional flags are overwriting the options that are set in the config file.

CHANGELOG.md
```diff
@@ -18,7 +18,7 @@
 
 ## New Functionality
 
--
+- [openapi-generator] Added a feature to use a config.json with `-c/--config` to use the openapi-generator. The config.json can be used for all generator options that would otherwise be used with command line options. If command line options are used, they overwrite the options set in the config file.
```
marikaner, line 21: Sorry, I think I wasn't clear enough.
I think in the release notes you don't need too much detail, try to be concise but explicit. I think "config.json" is not 100% correct.
The note should always be imperative ("add" instead of "added").
```suggestion
- [openapi-generator] Add the `-c/--config` command line option to specify options through a configuration file instead of on the command line.
```

packages/openapi-generator/src/options/flags.ts
```diff
@@ -0,0 +1,101 @@
...
+  metadata: flags.boolean({
+    description: 'When enabled, metadata for the API hub is generated.',
+    default: false,
+    hidden: true
+  }),
+  config: flags.string({
+    char: 'c',
+    parse: input => resolve(input),
+    description:
+      'Set the path to the file containing the options for generation. If other flags are used, they overwrite the options set in the config. If a directory is passed, a `config.json` file is read from this directory.',
```
marikaner, line 97:
```suggestion
      'Set the path to a file containing the options for generation instead of setting the options on the command line. When combining the `config` option with other options on the command line, the command line options take precedence. If a directory is passed, a `config.json` file is read from this directory.',
```

packages/openapi-generator/src/generator-utils.ts
```diff
@@ -0,0 +1,50 @@
...
+  generatorFlags
+} from './options';
+const { readFile, lstat } = promises;
+
+/**
+ * Parses a given path to a config file or directory and returns its content
+ * @param configPath path to a config file or a directory containing a config.json
+ * @returns Options to configure generation
+ */
+export async function parseOptionsFromConfig(
```
marikaner, line 16: Actually, why isn't this in the `generator-options.ts`? I think it would fit there better ;)

packages/openapi-generator/src/generator-utils.ts
```diff
@@ -0,0 +1,50 @@
...
+  configPath: string
+): Promise<GeneratorOptions> {
+  try {
+    if ((await lstat(configPath)).isDirectory()) {
+      configPath = resolve(configPath, 'config.json');
+    }
+    return JSON.parse(await readFile(configPath, 'utf8'));
+  } catch (err) {
+    throw new ErrorWithCause(
+      `Could not read config.json at ${configPath}.`,
```
marikaner, line 26: It isn't always config.json ;)
```suggestion
      `Could not read configuration file at ${configPath}.`,
```

packages/openapi-generator/src/generator-utils.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Removes default values from a given set of parsed generator options
+ * @param options parsed generator options
+ * @returns generator options with removed default values
+ */
+export function removeDefaultValues(
+  options: ParsedGeneratorOptions
+): GeneratorOptions {
+  return Object.entries(generatorFlags).reduce(
```
marikaner, line 40: I am not a fan of removing the values based based on the comparison of values.
If someone sets the default value on purpose it will be ignored.
Also, it is a bit counterintuitive and harder to read than the other approach with filtering the values based on the parsed 'raw' values.

marikaner (changes requested):
I think the remove approach does not cover all corner cases, consider using the `raw` property of the parsed result.

packages/openapi-generator/src/generator-utils.ts
tomfrenken: Default values are added after the `generate` function is executed regardless, but I agree that filtering with the raw values is the better approach.

packages/openapi-generator/src/generator-utils.ts
tomfrenken: Agreed, moved it over to generator-options 👍

marikaner (changes requested):
Damn, I forgot to submit the review. Sorry for making you wait...
This is mostly about test cases, I think we should add some tests, while I believe some are unnecessary.

packages/openapi-generator/src/cli/index.ts
```diff
@@ -25,98 +28,19 @@ $ openapi-generator --input ./my-spec.yaml --outputDir ./client --transpile`
...
+  static flags = generatorFlags;
 
   async run(): Promise<void> {
     try {
       const parsed = this.parse(OpenApiGenerator);
-      await generateWithParsedOptions(parsed.flags);
+      if (parsed.flags.config) {
+        await generate({
+          ...(await parseOptionsFromConfig(parsed.flags.config)),
+          ...removeDefaultValues(parsed.flags, parsed.raw)
```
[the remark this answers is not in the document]
tomfrenken: I believe it's still necessary, the issue with the raw values is that all paths values (e.g. input, outputDir, ...) aren't resolved (`resolve` function from `path`), if we would use that raw values, we would need to execute `resolve` on all path values again, which would make the `parseFromConfig` function a lot bigger I believe (and also more code duplication since we would resolve twice but only use it once)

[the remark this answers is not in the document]
marikaner: Got it. Then I have some additional comments:
1. Consider removing the notion of "default values" in the name of this function, consider something like `getSpecifiedFlags` (or maybe you come up with something more fitting).
2. I think you can express `Object.entries(parseRawInput(rawInput))` + the following usage as this: `parsed.raw.map(({flag}) => flag)`. If you pass the result of this to the (currently) `removeDefaultValues` function, you can remove the import of `ParsingTokens` and instead pass a list of keys (`keys: string[]`).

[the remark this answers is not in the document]
tomfrenken: Yes just passing a string array is a good idea, but instead of importing the ParsingToken I now need to import FlagToken, because ParsingTokens can theoretically also be ArgTokens and ArgTokens don't have the property "flag" which would lead to a type error if I don't typecast it (AFAIK we don't have ArgTokens so typecasting should be fine)

packages/openapi-generator/src/options/generator-options.spec.ts
```diff
@@ -97,4 +107,81 @@ describe('parseGeneratorOptions', () => {
...
+      config: '/path/config.json'
+    };
+    mock({
+      '/path/': {
+        'config.json': JSON.stringify(config)
+      }
+    });
+    await expect(parseOptionsFromConfig(parameters.config)).resolves.toEqual({
+      input: 'some-repository'
+    });
```
marikaner, line 164:
```suggestion
    await expect(parseOptionsFromConfig(parameters.config)).resolves.toEqual(config);
```

marikaner (approved):
LGTM

PR #1299, opened 2021-05-17
Fix: Re-enable circuit breaker

marikaner:
The circuit breaker was basically unused. This PR fixes that and add tests to ensure this does not happen again.

### DIFF
--- a/packages/core/src/connectivity/scp-cf/destination/circuit-breaker.spec.ts
+++ b/packages/core/src/connectivity/scp-cf/destination/circuit-breaker.spec.ts
@@ -1,0 +1,86 @@
+import nock from 'nock';
+import * as jwt123 from 'jsonwebtoken';
+import {
+  destinationServiceUri,
+  providerXsuaaClientCredentials,
+  providerXsuaaUrl
+} from '../../../../test/test-util/environment-mocks';
+import { privateKey } from '../../../../test/test-util/keys';
+import { circuitBreakerDefaultOptions } from '../resilience-options';
+import { clientCredentialsGrant } from '../xsuaa-service';
+import { fetchDestination } from './destination-service';
+
+const jwt = jwt123.sign(
+  JSON.stringify({ user_id: 'user', zid: 'tenant' }),
+  privateKey(),
+  {
+    algorithm: 'RS512'
+  }
+);
+
+const attempts = circuitBreakerDefaultOptions.volumeThreshold!;
+
+describe('circuit breaker', () => {
+  it('opens after 50% failed request attempts (with at least 10 recorded requests) for destintation service', async () => {
+    const request = () =>
+      fetchDestination(destinationServiceUri, jwt, 'FINAL-DESTINATION');
+
+    nock(destinationServiceUri)
+      .get(/.*/)
+      .times(1)
+      .reply(200, JSON.stringify({ URL: 'test' }));
+    nock(destinationServiceUri).get(/.*/).times(attempts).reply(400);
+
+    // First attempt should succeed
+    expect(await request()).toBeDefined();
+
+    // Following attempts should fail
+    for (let i = 0; i < attempts - 1; i++) {
+      await expectToThrowWithRootCause(
+        request,
+        'Request failed with status code 400'
+      );
+    }
+
+    // Last attempt should open the circuit breaker
+    await expectToThrowWithRootCause(request, 'Breaker is open');
+  });
+
+  it('opens after 50% failed request attempts (with at least 10 recorded requests) for xsuaa service', async () => {
+    const request = () =>
+      clientCredentialsGrant(providerXsuaaClientCredentials, {
+        username: 'user',
+        password: 'pass'
+      });
+
+    nock(providerXsuaaUrl).post('/oauth/token').times(1).reply(200);
+
+    nock(providerXsuaaUrl).post('/oauth/token').times(attempts).reply(400);
+
+    // First attempt should succeed
+    expect(await request()).toBeDefined();
+
+    // Following attempts should fail
+    for (let i = 0; i < attempts - 1; i++) {
+      await expectToThrowWithRootCause(
+        request,
+        'Request failed with status code 400'
+      );
+    }
+
+    // Last attempt should open the circuit breaker
+    await expectToThrowWithRootCause(request, 'Breaker is open');
+  });
+});
+
+async function expectToThrowWithRootCause(
+  fn: () => any,
+  rootCause: string
+): Promise<void> {
+  try {
+    await fn();
+    fail('Expected to throw.');
+  } catch (err) {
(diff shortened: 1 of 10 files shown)


packages/core/src/connectivity/scp-cf/resilience-options.ts
```diff
@@ -8,5 +8,6 @@ export interface ResilienceOptions {
 export const circuitBreakerDefaultOptions = {
   timeout: 10000,
   errorThresholdPercentage: 50,
+  volumeThreshold: 10,
```
jjtang1985, line 11: We now have more hard coded options, should we make a follow up so users can customize the options?

packages/core/src/connectivity/scp-cf/destination/destination-service.ts
```diff
@@ -234,18 +248,19 @@ function callDestinationService(
...
 
   return executeHttpRequest(destination, config);
 }
 
-function getInstanceCircuitBreaker(breaker?: any): any {
-  return typeof breaker === 'undefined'
-    ? new CircuitBreaker(executeHttpRequest, circuitBreakerDefaultOptions)
-    : breaker;
+function getCircuitBreaker(): DestinationCircuitBreaker {
+  if (typeof circuitBreaker === 'undefined') {
```
jjtang1985, line 259: why not `if (circuitBreaker === 'undefined') {`

packages/core/src/connectivity/scp-cf/destination/destination-service.ts
```diff
@@ -22,17 +24,29 @@ import {
...
-// For some reason, the equivalent import statement does not work
-/* eslint-disable-next-line @typescript-eslint/no-var-requires */
-const CircuitBreaker = require('opossum');
-
 const logger = createLogger({
   package: 'core',
   messageContext: 'destination-service'
 });
+
+type DestinationCircuitBreaker = CircuitBreaker<
```
jjtang1985, line 39: +1 for have separate CB for destination and xsuaa.

packages/core/src/connectivity/scp-cf/xsuaa-service.ts
```diff
@@ -343,10 +343,14 @@ function accessTokenError(error: Error, grant: string): Error {
   );
 }
 
-function getInstanceCircuitBreaker(breaker?: any | undefined): any {
-  return typeof breaker === 'undefined'
-    ? new CircuitBreaker(executeHttpRequest, circuitBreakerDefaultOptions)
-    : breaker;
+function getCircuitBreaker(): XsuaaCircuitBreaker {
+  if (typeof circuitBreaker === 'undefined') {
```
jjtang1985, line 347: same question like destination

packages/core/src/connectivity/scp-cf/destination/destination-service.spec.ts
```diff
@@ -486,16 +470,32 @@ describe('destination service', () => {
...
         })
-        .catch(error => {
-          expect(error.message).toBeDefined();
-          done();
-        });
+      ).rejects.toThrowError();
+    });
+
+    it('circuit breaker opens after 10 failed request attempts', async () => {
+      const attempts = 10;
```
jjtang1985, line 481: This magic number comes from the hard coded option `volumeThreshold`, maybe reuse that value.
Also, I know the other 3 options were there for long time, could we also add some tests as you are enabling the circuit breaker.

packages/core/src/connectivity/scp-cf/xsuaa-service.spec.ts
```diff
@@ -182,6 +182,27 @@ describe('xsuaa', () => {
         expect(error.stack).toContain('500');
       }
     });
+
+    it('circuit breaker opens after 10 failed request attempts', async () => {
+      const attempts = 10;
```
jjtang1985, line 187: Here is also a magic number.

jjtang1985 (changes requested):
Nice fix for the singleton pattern.
I would like to have more tests if possible, as the circuit breaker is enable by this PR, which is more like a new feature.

packages/core/src/connectivity/scp-cf/destination/destination-service.ts
marikaner: Generally this is the reason: https://stackoverflow.com/questions/2703102/typeof-undefined-vs-null.
But I agree we don't need this "security" here (nor in most other places).

packages/core/src/connectivity/scp-cf/destination/destination-service.spec.ts
marikaner: Good point, I will adjust to volumeThreshold.

I tried to make a test for the reset and thresholdpercentage, but due to an [issue](https://github.com/facebook/jest/issues/10221) with jest fake timers and nock, I was not able to do this. (Took me a long time already unfortunately).
Anyways, I think testing that the circuit breaker is in fact used is important, while checking that the circuit breaker behaves as we expect should not necessarily be our business. I think, here it would make sense, but the benefit does not justify the cost.

packages/core/src/connectivity/scp-cf/resilience-options.ts
marikaner: I think users should not be able to customize the options on which we break for the destination and xsuaa service as this is meant to protect the BTP services. At least that is how we designed it initially and how I discussed it with @FrankEssenberger recently. That said, the timeout will be removed from those options and will be made configurable.

packages/core/src/connectivity/scp-cf/destination/destination-service.ts
marikaner: Done

packages/core/src/connectivity/scp-cf/destination/destination-service.spec.ts
marikaner: As discussed, partially this was tested manually, but due to the jest / nock combination issue it is currently not easily possible to automatically test this.

packages/core/src/connectivity/scp-cf/destination/circuit-breaker.spec.ts
```diff
@@ -0,0 +1,86 @@
+import nock from 'nock';
```
jjtang1985, line 1: It makes sense to have a centralised location for testing circuit breaker instead of moving to multiple locations.
However, this also violates our spec file naming rule, because there is no `circuit-breaker.ts` ;)
I'm fine with it as the exception.

jjtang1985 (approved):
Very good improvement.

packages/core/src/connectivity/scp-cf/destination/circuit-breaker.spec.ts
marikaner: Agreed, we also have this in some other places, maybe we should write it down somewhere.

Issue #1320, opened 2021-05-27
`content-length` header should not be passed to CSRF fetch


marikaner:
CAP adds `content-length` headers to all requests. For write operations, e.g. `patch`, custom headers are passed to the CSRF fetch request. Those requests fail during validation, because the fetch request is a HEAD request without content. We should not pass on the `content-length`.

jjtang1985:
I'll create this ticket, please reopen if it does not work.

PR #1327, opened 2021-05-31
fix: Fix CSRF fetch content length header

marikaner:
Always overwrite the content length header to be 0, which is the correct length of a HEAD request.

Related to #1320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,9 @@
 
 ## Fixed Issues
 
-- [core] Re-enable the circuit breakers for the destination and xsuaa services.
+- [core] Re-enable the circuit breakers for the destination and XSUAA services.
 - [odata] Fix encoding of filter strings.
+- [core] Add or overwrite the content length header for CSRF fetch requests. As those are always `HEAD` requests, the content length is always set to 0.
 
 
 # 1.44.0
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -26,8 +26,9 @@
 
 ## Fixed Issues
 
-- [core] Re-enable the circuit breakers for the destination and xsuaa services.
+- [core] Re-enable the circuit breakers for the destination and XSUAA services.
 - [odata] Fix encoding of filter strings.
+- [core] Add or overwrite the content length header for CSRF fetch requests. As those are always `HEAD` requests, the content length is always set to 0.
```
jjtang1985, line 31: Alternative would be to delete the `content-length` key. I guess setting to 0 might be a bit easier.

jjtang1985, line 31:
```suggestion
- [core] Add or overwrite the `Content-Length` header for CSRF fetch requests. As those are always `HEAD` requests, the content length is always set to 0.
```

jjtang1985 (approved):
LGTM and I left a minor suggestion.

CHANGELOG.md
marikaner: The CAP colleagues mentioned that some servers need the content length header. That is why they add it.

jjtang1985: I guess they need it for the real request.
I mean only remove the key for `HEAD` request, which should make sense?

marikaner: But the HEAD request also goes against the same system.

marikaner: I wanted to avoid writing out the case, because it is case independent.

marikaner:
I will go ahead and merge, but let me know, if you would like to add changes.

PR #1328, opened 2021-05-31
chore: Reduce test scripts

marikaner:
This simplifies the test scripts to be just "test" and removes debug test scripts as they are not necessary anymore


jjtang1985 (approved):
Without coverage, the pipeline gets faster.

PR #1334, opened 2021-06-02
fix: enumeration exceptions

tomfrenken:
The bug was caused because we couldn't handle two types of enumerations.

The first error was caused when the type was declared as `string` but the actual values were not in string format, this is most likely caused by yaml conversion, because in yaml this format is correct.

Example:

```
"some-param": {
    "name": "category",
    "in": "query",
    "type": "string",
    "required": true,
    "enum": [
        0,
        1,
        2
    ],
    "description": "Some Description"
}
```

The second error was caused because the values can actually be `null` when the field `nullable` is declared as `true`

Example:
```
 "some-param": {
    "type": "string",
    "maxLength": 1,
    "title": "Some Title",
    "description": "Some Description",
    "nullable": true,
    "enum": [
        null,
        "X"
    ]
  }
```

To fix 
1) 
- Whenever the type is declared as `string`, all values in the enum are typecasted to string (except if their value is `null`, then either the value null is used or if `nullable` wasn't set to `true`, an error is thrown.

2) 
- When a value of the enum is `null` and `nullable` is declared as `true`, the value is used,
- When the value is `null` but `nullable` is not declared as true, an error is thrown


packages/openapi-generator/src/parser/schema.ts
```diff
@@ -169,9 +169,17 @@ function parseEnumSchema(
   const type = schema.type ? getType(schema.type) : 'string';
   return {
     type,
-    enum: (schema.enum || []).map(entry =>
-      type === 'string' ? getEnumStringValue(entry) : entry
-    )
+    enum: (schema.enum || []).map(entry => {
+      if (type === 'string' && entry !== null) {
+        return getEnumStringValue(String(entry));
```
FrankEssenberger, line 174: A general question. In general we use `undefined` over `null` because the optional set it to undefined. Should we perhaps also use `undefined` instead of `null`. However `null` is closer to the enum definiotn. @marikaner  @jjtang1985 what do you think?

FrankEssenberger (changes requested):
LGTM - I would like to discuss if the value `null` is good or if `undefined` is better.

packages/openapi-generator/src/parser/schema.ts
jjtang1985: In general, if the schema defines `null` as a valid enum value, I think we should just use it without any changes. The spec definition should be treated over any programming language (if possible) to avoid confusing.

In this case, `nullable` only allows `null` as a valid enum value.
So the type should look like:
```ts
param: 'v1' | 'v2' | null
```

On the other hand, there is another keyword called `required`. When setting to false, it's optional, so the type should look like below, where `undefined` is assigned if no value is provided:
```ts
param?: 'v1' | 'v2'
```

FrankEssenberger: That is a good point so we do:
- nullable = true add the null value to enum list. (as it is)
-  required = false make the enum parameter optional

In the schema.ts the `required`property is considered, but I am not sure if there is a test. So would be good to add a test in the schema.spec.ts to see if serialization and `required:false` leads to `?` schema file.

marikaner: Also fits our [styleguide](https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#null-and-undefined).

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -169,9 +169,17 @@ function parseEnumSchema(
...
+      }
+      if (entry === null && !schema.nullable) {
+        throw new Error(
+          "'Null' was used as a parameter, but nullable wasn't declared"
```
marikaner, line 178: Should we be this restrictive? Shouldn't we rather log a warning, but still add null as a possible value in the enum? @FrankEssenberger @jjtang1985 what are your opinions?

marikaner, line 178: Either way I would suggest the following improvements to the message:
* Use null instead of 'Null', even in the beginning of a sentence (lower case + no quotes).
* Use "was not" instead of "wasn't".
* Add a fullstop at the end of the sentence.
* Add the notion of an enum somehow.
```suggestion
          "null was used as a parameter in an enum, although the schema was not declared as nullable"
```

tomfrenken: Agreed! I wasn't satisfied with the error message as well but wasn't sure what the best one would be, your approach looks clean.

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -169,9 +169,17 @@ function parseEnumSchema(
...
-    enum: (schema.enum || []).map(entry =>
-      type === 'string' ? getEnumStringValue(entry) : entry
-    )
+    enum: (schema.enum || []).map(entry => {
+      if (type === 'string' && entry !== null) {
+        return getEnumStringValue(String(entry));
+      }
+      if (entry === null && !schema.nullable) {
+        throw new Error(
+          'null was used as a parameter in an enum, although the schema was not declared as nullable'
```
marikaner, line 178: I'll just repeat my suggestion ;)

Should we be this restrictive? Shouldn't we rather log a warning, but still add null as a possible value in the enum? @FrankEssenberger @jjtang1985 what are your opinions?

jjtang1985: I'm fine to use warning instead of error for enabling the user. The worst case would be, the user receive an error from server side and stop using `null`. For that we can argue, the spec is anyway confusing and needs some actions from the service owner.

packages/openapi-generator/src/parser/schema.ts
FrankEssenberger: I would propose the following:
- If skipValidation is set to `false` throw an execption
- If skipeValidation is set to `true` log a warning and except the not perfect spec files.

packages/openapi-generator/src/parser/schema.ts
FrankEssenberger: I added a comment above

marikaner (approved):
LGTM. Consider Franks comment about using `skipValidation` for the warning / error decision.

tomfrenken:
@FrankEssenberger I've added an additional test and changed the function to use the `--skipValidation` flag, however, this introduces a lot of changes to many functions, as I need to pass this argument down the function chain and therefore needed to change many tests, etc.

FrankEssenberger (approved):

PR #1353, opened 2021-06-11
chore: disable audit-ci

jjtang1985:
As discussed with @marikaner and @FrankEssenberger , having two vulnerability checks does not bring us much value. 
Dependabot alerts will be used and audit-ci will be disabled.
I updated the doc for release process.


FrankEssenberger (approved):
LGTM

Issue #1338, opened 2021-06-06
OR-Filter on Property & NavigationProperty


HeneryHawk:
Hi @jjtang1985

Creating a filter query consisting of two filters, one on an entity property and one filter on a NavigationProperty, which are associated with an OR expression, results in an error.

I want to create the following oData filter:
`$filter=(Id eq <id> or (ExternalIds/any(a0:(a0/ExternalId eq '<id>'))))`

This should be done by the following code:
```typescript
.filter(or(Entity.ID.equals(id), Entity.EXTERNAL_IDS.filter(any(ExternalIds.EXTERNAL_ID.equals(id)))))
```

But unfortunately the TypeScript compiler returns the following error message:
```
Argument of type 'OneToManyLink<Entity, ExternalIds>' is not assignable to parameter of type 'Filterable<Entity>'.
  Type 'OneToManyLink<Entity, ExternalIds>' is missing the following properties from type 'Filter<Entity, FieldType | FieldType[]>': field, operator, valuets(2345)
```

However, linking both filters with an AND expression works fine
```typescript
.filter(Entity.ID.equals(id), Entity.EXTERNAL_IDS.filter(any(ExternalIds.EXTERNAL_ID.equals(id))))
```
`$filter=(Id eq <id> and (ExternalIds/any(a0:(a0/ExternalId eq '<id>'))))`

See also the [question](https://answers.sap.com/questions/13405271/cloud-sdk-or-filter-on-property-and-navigationprop.html) on the SAP Community.

**Used Versions:**
- SAP Cloud SDK: 1.44.0

Regards
Simon

jjtang1985:
Hi @HeneryHawk ,

thanks for approaching us in this repo.

I could reproduce this issue so I created a backlog item.

Could you please share your time line and your project status?
Are you working on a PoC project or live system?

Best regards,
Junjie

HeneryHawk:
Hi @jjtang1985,
thank you for your efforts.

The feature that includes this functionality of the Cloud SDK should be delivered to our customers shortly.
The feature will work in its entirety, but a part of it is affected by this Cloud SDK bug. So this non working part will have to be delivered afterwards, as soon as a bugfix for the Cloud SDK is available.

**In a nutshell:**
The sooner a bugfix for the Cloud SDK is available, the better it is for us, as we can then complete the feature.

Best
Simon

jjtang1985:
Hi @HeneryHawk ,

thanks for sharing the status of your current project.

[...]

PR #1356, opened 2021-06-11
fix: Fix filtering on one-to-many navigation properties

marikaner:
This adds one-to-many navigation property filters to `Filterable`. This should allow using those filters in all places, where filtering is possible.

Related to https://github.com/SAP/cloud-sdk-js/issues/1338.


FrankEssenberger (approved):
LGTM

CHANGELOG.md
```diff
@@ -26,7 +26,7 @@
 
 ## Fixed Issues
 
--
+- [core] Fix type error to allow filtering on one-to-many navigation properties.
```
jjtang1985, line 29: Minor: It would be great if you can mention lambda expression (is typed as one to many navigation properties in core), as the author of the issue failed to use lambda expression.

marikaner, line 29:
```suggestion
- [core] Fix type error to allow filtering on one-to-many navigation properties in lambda expressions.
```

PR #1359, opened 2021-06-15
fix: error messages of invalid path patterns

jjtang1985:
There are two failing generation due to the special paths:
- /somePath/{param1}:{param2}, where `:` is used as delimiter between multiple parameters
- /somePath/value?param={date}-{time}, where `?` and query parameters are shown in the path

I added a validation step and improved the error message.

### DIFF
--- a/packages/openapi-generator/src/parser/operation.ts
+++ b/packages/openapi-generator/src/parser/operation.ts
@@ -76,8 +76,14 @@ function validatePlaceholder(pathPart: string): boolean {
   return /^\{[^{}]+\}$/.test(pathPart);
 }
 
-function validatePathPattern(pathPattern: string, placeholders: string[]): boolean{
-  return pathPattern.includes('?') || placeholders.some(placeholder => !validatePlaceholder(placeholder))
+function validatePathPattern(
+  pathPattern: string,
+  placeholders: string[]
+): boolean {
+  return (
+    pathPattern.includes('?') ||
+    placeholders.some(placeholder => !validatePlaceholder(placeholder))
+  );
 }
 
 function sortPathParameters(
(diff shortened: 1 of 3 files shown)


packages/openapi-generator/src/parser/operation.ts
```diff
@@ -72,13 +72,33 @@ function isPlaceholder(pathPart: string): boolean {
   return /^\{.+\}$/.test(pathPart);
 }
 
+function validatePlaceholder(pathPart: string): boolean {
```
FrankEssenberger, line 75: Could you add some small statement what this method checks with an example. The regex is so hard to read.

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -72,13 +72,33 @@ function isPlaceholder(pathPart: string): boolean {
...
+  return /^\{[^{}]+\}$/.test(pathPart);
+}
+
+function validatePathPattern(
```
FrankEssenberger, line 79: General: The methods are called validateXYZ and give back a boolean. For boolean I like the `isSomeThing` form. For validate I expect that the method throws an exeception if not valid.

packages/openapi-generator/src/parser/operation.ts
FrankEssenberger: The argument is called pathPart but it is called with placeholder. Rename?

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -72,13 +72,33 @@ function isPlaceholder(pathPart: string): boolean {
...
+  pathPattern: string,
+  placeholders: string[]
+): boolean {
+  return (
+    pathPattern.includes('?') ||
+    placeholders.some(placeholder => !validatePlaceholder(placeholder))
```
FrankEssenberger, line 85: Is place holder a two word thing so `placeHolder` I am not sure, perhaps you can find out.

packages/openapi-generator/src/parser/operation.ts
FrankEssenberger: Also here add some small comment what the method does.

FrankEssenberger (approved):
LGTM some minor comments on names. Please have a look.

packages/openapi-generator/src/parser/operation.ts
jjtang1985: [placeholder](https://dictionary.cambridge.org/dictionary/english/placeholder) is a word.

PR #1361, opened 2021-06-15
fix: Fix uniqueness check on directory names

marikaner:
This PR includes some minor refactoring of descriptions and error messages.
The main purpose is to fix the uniqueness check for directory names. Before, we checked the `serviceName` in `optionsPerService` for uniqueness. The purpose of this field is solely to have a human readable text in the readme and API docs. It is ok if this is duplicate. It is not for directory names, though.

### DIFF
--- a/packages/openapi-generator/src/options/options-per-service.ts
+++ b/packages/openapi-generator/src/options/options-per-service.ts
@@ -1,7 +1,6 @@
 import { existsSync, promises } from 'fs';
 import { parse } from 'path';
-import { EOL } from 'os';
-import { UniqueNameGenerator } from '@sap-cloud-sdk/util';
+import { unique, UniqueNameGenerator } from '@sap-cloud-sdk/util';
 import { getRelativePath } from '../generator';
 import { ParsedGeneratorOptions } from './generator-options';
 
@@ -31,7 +30,7 @@ export interface ServiceOptions {
    */
   packageName: string;
   /**
-   * Name of the service
+   * Human readable name of the service. Used in API documentation and readme.
    */
   serviceName: string;
 }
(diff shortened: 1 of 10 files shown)


packages/openapi-generator/src/options/options-per-service.ts
```diff
@@ -66,76 +65,120 @@ export async function getOptionsPerService(
...
+ * @param serviceOptions The original options for this service as specified in the per service options.
+ * @param directoryName The directory name of the according service.
  * @returns Service options.
  */
 export function getServiceOptions(
-  optionsPerService: PartialOptionsPerService,
-  relativeInputFilePath: string,
-  serviceName: string
+  serviceOptions: Partial<ServiceOptions> | undefined,
+  directoryName: string
```
florian-richter, line 171:
```suggestion
  directoryName: string,
  serviceOptions?: Partial<ServiceOptions>
```

Would changing the order to allow `serviceOptions` to be optional instead of `... | undefined` be worth it?

packages/openapi-generator/src/parser/schema-naming.ts
```diff
@@ -36,39 +29,52 @@ function makeSchemaNameValid(name: string): string {
...
-function getNonValidNameErrorMessages(nonValidSchemaNames: string[]) {
-  const header =
-    "Your OpenApi definition contains the invalid schema names. The SDK generator can adjust such names automatically if you disable 'strictNaming' or you adjust the schema. The errors are:";
-  const errors = nonValidSchemaNames.map(nonValidSchemaName =>
-    getErrorMessage(nonValidSchemaName)
+function validateSchemaNames(names: string[]): void {
+  const namesStartingWithNumber = names.filter(name =>
+    name.match(startsWithNumberRegex)
+  );
+  const namesContainingIllegalChars = names.filter(name =>
```
florian-richter, line 47:
```suggestion
  const namesContainingIllegalChars = names.some(name =>
```

I think the `length` is not relevant so `.some` would be the better fit instead of `.filter`

florian-richter: Nevermind, it's needed for the errors...

florian-richter (approved):
LGTM

packages/openapi-generator/src/options/options-per-service.ts
marikaner: Not necessarily, but it looks slightly nicer 😊

PR #1363, opened 2021-06-17
fix: error messages for invalid swagger 2.0 specs

jjtang1985:
When using `ErrorWithCause`, the root cause is saved as part of the `stack` attribute. As a ts object, the `stack` information can be read. However, when using generator as a CLI, the `stack` info is omitted from the stderr so you have no ideas what's wrong with the spec.


packages/openapi-generator/src/document-converter.ts
```diff
@@ -61,7 +61,7 @@ export async function convertDocToOpenApiV3(
     return (await convert(openApiDocument, {})).openapi;
   } catch (err) {
     throw new ErrorWithCause(
-      'Could not convert OpenAPI specification to OpenAPI version 3.',
+      `Could not convert OpenAPI specification to OpenAPI version 3. ${err.message}`,
```
marikaner, line 64: Seems a bit unclean to me, because when using this programatically, you would get both errors. Is there a way to configure how we log the errors in the CLI?

FrankEssenberger (approved):

packages/openapi-generator/src/document-converter.ts
jjtang1985: Some bullet points after the discussion:
- Just like the upcoming discussion with CAP, we need to consider different audience.
    - CLI only uses the `message`, while programming call  uses the whole object `ErrorWithCause`
- For different audiences, we might build error to different format
    - A single error message
    - Like the stack object structure with hierarchy
    - Nicer and shorter error message that contains all useful information from stack.

Before the CAP meeting, I would keep a smaller change, as this brings the value (real reason) to the CLI user.

jjtang1985: @marikaner , please approve.

PR #1366, opened 2021-06-18
fix: Allow using custom header for principal propagation

marikaner:
Related to this issue: https://answers.sap.com/questions/13416513/sap-cloud-sdk-core-giving-issues-with-principal-pr.html.

The user tried to set the pp header per custom headers. This did not work, because it isn't treated as auth header. Although this header was set, the error message stated, that the header was missing. This PR improves this behavior.


packages/core/src/connectivity/scp-cf/authorization-header.spec.ts
```diff
@@ -59,28 +81,21 @@ describe('buildAuthorizationHeaders', () => {
...
+      const headers = await buildAuthorizationHeaders(
+        principalPropagationDestination
+      );
       checkHeaders(headers);
+    });
 
-      delete destination!.proxyConfiguration!.headers![
-        'SAP-Connectivity-Authentication'
-      ];
+    it('does not throw on Principal Propagation', async () => {
```
FrankEssenberger, line 91: it says does not throw but in the test it is expected to throw -> confused.

packages/core/src/connectivity/scp-cf/authorization-header.ts
```diff
@@ -61,11 +61,21 @@ function hasAuthHeaders(destination: Destination): boolean {
...
+}
+
 export async function getAuthHeaders(
   destination: Destination,
   customHeaders?: Record<string, any>
 ): Promise<Record<string, string>> {
-  const customAuthHeaders = pickIgnoreCase(customHeaders, 'authorization');
+  const customAuthHeaders = getCustomAuthHeaders(destination, customHeaders);
 
   if (Object.keys(customAuthHeaders).length && hasAuthHeaders(destination)) {
```
FrankEssenberger, line 80: the `hasAuthHeaders` is not triggered in the `SAP-Connectivity-Authentication` case right? I would also like to have a warning is this is done.

FrankEssenberger (changes requested):
LGTM but there are to git checks and I have two minor comments.

packages/core/src/connectivity/scp-cf/authorization-header.spec.ts
marikaner: Fixed.

packages/core/src/connectivity/scp-cf/authorization-header.ts
marikaner: It does. It just checks for authentication type.

FrankEssenberger: And we do not want to add a log statement if the `SAP-Connectivity-Authentication` is set via the custom header and is overwriting the automatic one?

marikaner: I meant, that the warning is logged in case there is a custom header.

marikaner: But I'll add a test for this.

FrankEssenberger (approved):
LGTM - as discussed small renaming of `hasAuthHeader`

PR #1367, opened 2021-06-18
chore: export sdk-metatdata in generator common

FrankEssenberger:
For the oData and openApi generators we were already exporting the metadata under the non root to be used by us internally. We did not do this for the generator-common part which was introduced later.


marikaner (approved):
LGTM. Please don't forget to adhere to conventional commits (here, I guess `chore:`) will do.

PR #1379, opened 2021-06-22
Fix: Allow setting null in filters for nullable properties

marikaner:
I tried to use mixins for this, but failed miserably because of the generics. I am not sure that this can be done when using generics. Anyways, I found a few articles with tricks, that might help this.
Instead I added another class to reduce some boiler plate code for fields that can be compared by `greater` and `less`.
* Add `GreaterOrLessEdmTypeField` class
* Add new classes for nullable field types
* Use new field types in generator


packages/generator/src/edmx-parser/legacy/common/edmx-to-vdm.ts
```diff
@@ -390,7 +390,7 @@ export function transformComplexTypes(
             ? 'CollectionField'
             : isComplex
             ? formattedTypes[parsedType] + 'Field'
-            : edmToComplexPropertyType(type),
+            : edmToComplexPropertyType(type, true),
```
FrankEssenberger, line 393: This seems wrong. Why it is always nullable?

packages/generator/src/edmx-parser/legacy/common/edmx-to-vdm.ts
```diff
@@ -480,7 +480,7 @@ export function transformFunctionImportBase(
       ),
       edmType: parseType(p.Type),
       jsType: edmToTsType(p.Type)!,
-      fieldType: edmToFieldType(p.Type),
+      fieldType: edmToFieldType(p.Type, true),
```
FrankEssenberger, line 483: This seems wrong. Why it is always nullable?

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -135,7 +135,7 @@ function getEdmReturnType(
   extractResponse: ExtractResponse,
   oDataVersion: ODataVersion
 ): VdmFunctionImportReturnType {
-  const typeMapping = getTypeMappingActionFunction(edmType);
+  const typeMapping = getTypeMappingActionFunction(edmType, true);
```
FrankEssenberger, line 138: see above

packages/generator/src/edmx-parser/legacy/common/edmx-to-vdm.ts
marikaner: Before, everything was non-nullable, because the nullable fields didn't exist. This is the legacy code, so I decided not to touch it. Do you think this is ok?

packages/generator/src/edmx-parser/legacy/common/edmx-to-vdm.ts
marikaner: See above.

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
marikaner: Here, I think we don't have the information whether this is nullable or not...

packages/core/src/odata-common/selectable/nullable-string-field.ts
```diff
@@ -0,0 +1,71 @@
...
+import { EdmTypeField, SelectableEdmTypeField } from './edm-type-field';
+
+/**
+ * Represents a property with a string value, that is nullable.
+ *
+ * @typeparam EntityT - Type of the entity the field belongs to
+ */
+class NullableStringFieldBase<EntityT extends Entity> extends EdmTypeField<
+  EntityT,
+  string | null
```
FrankEssenberger, line 16: General: I know you tried a lot but it breaks my heart a bit to see so many extra classes. If I got it right the only relvant difference  is here this value for the field type which is `string` and `string|null` in the two cases right?

marikaner: Right

marikaner:
Closing due to new strategy on this

PR #1385, opened 2021-06-25
fix: Proxy config not considered for destination service calls

### LINKED ISSUE #1384: HTTP_PROXY/HTTPS_PROXY setting not respected when accessing destination service
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@sap-cloud-sdk/core@1.45.0` for the project I'm working on.

I observed that XSUAA and target system requests went through the specified proxy, but access to the destination service did not.
While the `destinationRequestConfig` contains the correct values from the environment, it gets overwritten by the default settings from `requestConfig`. Switching the merge operands seems to resolve it.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@sap-cloud-sdk/core/dist/http-client/http-client.js b/node_modules/@sap-cloud-sdk/core/dist/http-client/http-client.js
index b4c3b6a..b3d5d9a 100644
--- a/node_modules/@sap-cloud-sdk/core/dist/http-client/http-client.js
+++ b/node_modules/@sap-cloud-sdk/core/dist/http-client/http-client.js
@@ -147,7 +147,7 @@ function execute(executeFn) {
                     case 0: return [4 /*yield*/, buildHttpRequest(destination, requestConfig.headers)];
                     case 1:
                         destinationRequestConfig = _b.sent();
-                        request = merge(destinationRequestConfig, requestConfig);
+                        request = merge(requestConfig, destinationRequestConfig);
                         _a = request;
                         return [4 /*yield*/, addCsrfTokenToHeader(destination, request, httpRequestOptions)];
                     case 2:
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

FrankEssenberger:
closes #1384 

@johenning found an error in the proxy handling. We included the default agents in the passed request config. This config beats the one from the destination, which is intended to have a way to overwrite things.

However, if the destination had a valid internet proxy config this would would overwrite the config from the destination and leads the error @johenning observed.

Note that es always add the [axios default config](https://github.com/SAP/cloud-sdk-js/blob/40c006573602c6813d51714073f6ee6055006b4c/packages/core/src/http-client/http-client.ts#L205) with the agents which gets overwritten by everything else. So there is no need to add this in the config at all.

### DIFF
--- a/packages/core/src/connectivity/scp-cf/destination/destination-service.ts
+++ b/packages/core/src/connectivity/scp-cf/destination/destination-service.ts
@@ -8,7 +8,6 @@ import CircuitBreaker from 'opossum';
 import { decodeJwt, wrapJwtInHeader } from '../jwt';
 import {
   executeHttpRequest,
-  getAxiosConfigWithDefaults,
   HttpRequestConfig,
   HttpRequestOptions,
   HttpResponse
@@ -239,9 +238,7 @@ function callDestinationService(
   options: ResilienceOptions = { enableCircuitBreaker: true }
 ): Promise<HttpResponse> {
   const config: HttpRequestConfig = {
-    ...getAxiosConfigWithDefaults(),
-    url: uri,
-    method: 'GET',
+    method: 'get',
     headers
   };
 
--- a/packages/openapi-generator/src/parser/schema-naming.ts
+++ b/packages/openapi-generator/src/parser/schema-naming.ts
@@ -4,12 +4,6 @@ import { ParserOptions } from './options';
 const illegalCharacterRegex = /[.#@/"'*%]+/g;
 const startsWithNumberRegex = /^\d+/g;
 
-function isValidSchemaName(name: string): boolean {
-  return (
-    !name.match(startsWithNumberRegex) && !name.match(illegalCharacterRegex)
-  );
-}
-
 function makeSchemaNameValid(name: string): string {
   while (name.match(illegalCharacterRegex)) {
     name = name.replace(illegalCharacterRegex, '');
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 
 ## Fixed Issues
 
--
+- [proxy] Fix destination service calls using web proxies.
 
 
 # 1.46.0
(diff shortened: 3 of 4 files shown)


packages/openapi-generator/src/parser/schema-naming.ts
```diff
@@ -4,12 +4,6 @@ import { ParserOptions } from './options';
 const illegalCharacterRegex = /[.#@/"'*%]+/g;
 const startsWithNumberRegex = /^\d+/g;
 
-function isValidSchemaName(name: string): boolean {
```
marikaner, line 7: Finally! 😊

packages/core/src/connectivity/scp-cf/destination/destination-service.ts
```diff
@@ -239,9 +238,7 @@ function callDestinationService(
   options: ResilienceOptions = { enableCircuitBreaker: true }
 ): Promise<HttpResponse> {
   const config: HttpRequestConfig = {
-    ...getAxiosConfigWithDefaults(),
-    url: uri,
-    method: 'GET',
+    method: 'get',
```
marikaner, line 241: The uri parameter is unused now, isn't it?

marikaner (approved):
LGTM, but please remove the unused parameter.

packages/core/src/connectivity/scp-cf/destination/destination-service.ts
FrankEssenberger: yes, but it is part of the HttpRequestConfig.

marikaner: I mean the parameter in the function, `uri` - it is not used within the function, so you can remove it from the signature.

FrankEssenberger: it is used in line 245: ```  let destination: Destination = { url: uri, proxyType: 'Internet' };```

marikaner: Ah ok, then 👍

PR #1408, opened 2021-07-06
Fix: Allow fields to be null

marikaner:
This PR introduces a field builder and more generic fields, that can be nullable (or not).

TODOs:
- [x] Fix tests
- [x] Add tests
- [x] Fix/add documentation

Notes for the reviewer(s):
- I am not 100% happy with the naming: I deprecated `EdmTypeField` and added a new `EdmField`, because it didn't manage to make my changes non-breaking. The according builder function is called `buildEdmTypeField`. I am looking for your suggestions here 😊
- For entity generation I am using the builder, for complex type field generation not. The reason is, that in the complex type I would have to do this in the constructor (which is not a big deal).


packages/core/src/odata-common/selectable/edm-field.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export type FieldTypeByEdmType<
+  EdmT extends EdmTypeShared<'any'>,
+  NullableT extends boolean
+> = NullableFieldType<NonNullableFieldTypeByEdmType<EdmT>, NullableT>;
+
+/**
+ * Represents a property of an OData entity with an Edm type.
+ *
+ * `EdmTypeField`s are used as static properties of entities and are generated from the metadata, i.e. for each property of
```
FrankEssenberger, line 107: perhaps avoid the term metadata -> service definitions, edmx files just to avoid confusion with the the sdk metadata WRT APIhub

packages/core/src/odata-common/selectable/selectable.ts
```diff
@@ -29,3 +23,19 @@ export type Selectable<EntityT extends Entity> =
         | CollectionField<EntityT>
         | AllFields<EntityT>
     : never;
+
+/**
+ * Interface to denote a selectable [[Field]].
+ */
+export interface SelectableField {
```
FrankEssenberger, line 30: Question: The property selectable feels like a boolean like the `Nullable` which is a property of the field. Could we include this also in the generic signature of the field?

packages/core/src/odata-common/selectable/orderable-edm-field.ts
```diff
@@ -0,0 +1,58 @@
+import { Filter } from '../filter';
+import { Entity } from '../entity';
+import { EdmTypeShared } from '../edm-types';
+import { EdmField, FieldTypeByEdmType } from './edm-field';
+
+export class OrderableEdmField<
```
FrankEssenberger, line 6: Also ordreable is a boolean like property. You could either pass the `NonOderType` or the `OrderType` instead of true/false where the `OrderType` class implements the methods. Doing so you could even break the inheritance chain in the fields.

FrankEssenberger: I have thought about this 3mins so this is just a basis for discussion.

packages/core/src/odata-common/selectable/selectable.ts
FrankEssenberger: I have thought about this 3mins so this is just a basis for discussion.

packages/core/src/odata-common/selectable/collection-field.ts
```diff
@@ -1,24 +1,18 @@
...
- *
- * Represents a static field of an entity or complex type.
- *
- * @typeparam EntityT - Type of the entity the field belongs to
- * @typeparam FieldT - Type of the entries of the collection in the field
- */
 export class CollectionField<
     EntityT extends Entity,
-    FieldT extends EdmTypeShared<'any'> | Record<string, any> = any
+    CollectionFieldT extends EdmTypeShared<'any'> | Record<string, any> = any,
```
FrankEssenberger, line 11: Also this type feels like a property encoding like `isNullable` so perhaps this should also go to a generic property?

FrankEssenberger (commented):
So I see the following properties on a field:
- isNullable
- isCollection
- isSelectable
- isOrderable
- isComplex // isEdmx 

Would it be possible to model them all in the same way in other words not be inheritance as in the past but with a generic parameter. I see that only `isOrderable` would need some implementation right?

packages/core/src/odata-common/selectable/orderable-edm-field.ts
marikaner: I was also considering `Sortable` as a keyword or `GreaterOrLess`. I like none of them. `Orderable` could easily be confused with the actual `Orderable` interface (asc, desc), but this keyword is already used in another place.

packages/core/src/odata-common/selectable/collection-field.ts
```diff
@@ -1,40 +1,57 @@
...
+   * @param _fieldType - Edm type of the field according to the metadata description.
+   *  @param fieldOptions - Optional settings for this field.
    */
   constructor(
     fieldName: string,
-    fieldOf: ConstructorOrField<EntityT>,
-    readonly _fieldType: FieldT | ComplexTypeNamespace<FieldT>
+    readonly _fieldOf: ConstructorOrField<EntityT>,
+    readonly _fieldType: CollectionFieldType<CollectionFieldT>,
+    fieldOptions?: Partial<FieldOptions<NullableT, SelectableT>>
```
FrankEssenberger, line 34: I see very often the `Partial<FieldOptions...` in the constructor arguments. Why not make the properties optional in the `FiledOptions` type?

packages/core/src/odata-common/selectable/field.ts
```diff
@@ -23,6 +23,29 @@ export type FieldType =
  */
 export type DeepFieldType = FieldType | { [keys: string]: DeepFieldType };
 
+export interface FieldOptions<
```
FrankEssenberger, line 26: why is it necessary to have the generic here? Simple booleans enough?

packages/core/src/odata-common/selectable/field-builder.ts
```diff
@@ -0,0 +1,139 @@
...
+export type IsSelectableField<T> = T extends Constructable<any> ? true : false;
+export type IsOrderableField<T> = T extends OrderableEdmType ? true : false;
+
+export class FieldBuilder<
+  EntityT extends Entity,
+  FieldOfT extends ConstructorOrField<EntityT>
+> {
+  constructor(public fieldOf: FieldOfT) {}
+
+  buildEdmTypeField<EdmT extends OrderableEdmType, NullableT extends boolean>(
```
FrankEssenberger, line 41: I think this is a matter of taste. You could have three different builders with a defining name but than you need in the generator invoke the fitting one. So I am fine with the overload. But could you add some small docs like: creates wither Edm or OrderableEdm depending on inptu.....

packages/core/src/odata-common/selectable/selectable.ts
```diff
@@ -6,26 +6,31 @@ import type { CustomField } from './custom-field';
...
-        | ComplexTypeField<EntityT>
-        | CustomField<EntityT>
-        | CollectionField<EntityT>
+        | ComplexTypeField<EntityT, any, boolean, boolean>
+        | CustomField<EntityT, boolean>
+        | CollectionField<EntityT, any, boolean, boolean>
         | AllFields<EntityT>
     : never;
+
+/**
```
FrankEssenberger, line 27: should we move this to legacy folder?

FrankEssenberger (changes requested):
Only small question and remarks. When the tests are there I can approve.

FrankEssenberger:
> I am not 100% happy with the naming: I deprecated EdmTypeField and added a new EdmField, because it didn't manage to make my changes non-breaking. The according builder function is called buildEdmTypeField. I am looking for your suggestions here

Is this still relevant, I did not see the `EdmField` class anymore?

packages/core/src/odata-common/selectable/selectable.ts
marikaner: As discussed.

packages/core/src/odata-common/selectable/orderable-edm-field.ts
marikaner: Won't fix, as discussed.

packages/core/src/odata-common/selectable/collection-field.ts
marikaner: Done.

packages/core/src/odata-common/selectable/edm-field.ts
marikaner: Added OData notion

packages/core/src/odata-common/selectable/field-builder.ts
marikaner: But that would also mean, that you have to know which function to invoke when you use `Edm.String` vs. `Edm.Int16`.
I forgot the docs before and added them now.

packages/core/src/odata-common/selectable/field.ts
marikaner: No, because both those information are relevant to types down the line. E.g., when looking at the filter, I want to have something like `equals(string | null)`. A boolean does not allow me to do any type magic, because its type is only determined at runtime. It is `true` **and** `false` at design time.

packages/core/src/odata-common/selectable/collection-field.ts
marikaner: In the end I need to have all types somewhere. They cannot be optional in the type in the end, because then their type is not `true` or `false` as I need it to be, but `true | undefined` or `false | undefined`.
I am pretty sure I could achieve this by turning it around and using `Required<...>` in the end. I initially decided against that, because I think that semantically the field options are required by default and only the constructors have some sensible defaults.

packages/core/src/odata-common/selectable/selectable.ts
marikaner: Done.

marikaner:
> > I am not 100% happy with the naming: I deprecated EdmTypeField and added a new EdmField, because it didn't manage to make my changes non-breaking. The according builder function is called buildEdmTypeField. I am looking for your suggestions here
> 
> Is this still relevant, I did not see the `EdmField` class anymore?

It's not, thanks.

FrankEssenberger (approved):
LGTM test are now there, after you checked the Partial<> versus Required<> please go ahead and merge.

PR #1429, opened 2021-07-12
feat: Support OAuth2JWTBearer

jjtang1985:
- [x] add `OAuth2JWTBearer` to the support list
- [x] test the auth type with workflow in ` auth-flow.spec.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 
 ## New Functionality
 
--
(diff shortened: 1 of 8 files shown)


FrankEssenberger (approved):
LGTM - so you could use the same `addAuthSamlAuth` method because the destination service takes care of the actual token etching. Consider perhaps renaming the method. I also wonder if the system user property check would make sense in all cases for the method but I guess it does not hurt.

PR #1430, opened 2021-07-12
Fix: lambda expression with or operand

jjtang1985:
Last time, the type error (compile time) was fixed, but the runtime error still exists: `Could not construct query parameters from filter. Filter is not valid`, as mentioned in the [issue](https://github.com/SAP/cloud-sdk-js/issues/1338).

This PR will fix the runtime error.


FrankEssenberger (approved):
LGTM

PR #1449, opened 2021-07-20
feat: Additional properties

### LINKED ISSUE #1362: Consider header & query destination property
A colleague reported this to me. and I did not find that we handle this in the SDK.

There is a convention to define headers to the target system as [destination properties](https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/4e1d742a3d45472d83b411e141729795.html?q=header) as `URL.headers.<key>`  and `URL.queries.<key>`with the value. The doc states:

>This is a naming convention. As the call to the target endpoint is performed on the client side, the service only provides the configured properties. The expectation for the client-side processing logic is to parse and use them. If you are using higher-level libraries and tools, please check if they support this convention.

I think the SDK should consider these properties if they are on a destination.

marikaner:
Closes #1362, Closes SAP/cloud-sdk-backlog#111

This PR introduces additional query parameters and headers that are set on the destination. They take precedence over any SDK built properties. Custom properties take precedence over destination properties.

I also did some refactoring along the way. 🙄


packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -497,56 +504,38 @@ class DestinationFromServiceRetriever {
...
-    if (
-      this.options.selectionStrategy === subscriberFirst &&
-      resultFromSubscriber
-    ) {
-      return false;
-    }
-
-    return true;
+    return (
+      this.options.selectionStrategy !== alwaysSubscriber &&
```
FrankEssenberger, line 510: For me this here is not an improvement. I find it hard to process.

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -497,56 +504,38 @@ class DestinationFromServiceRetriever {
...
-    }
-
-    if (this.isProviderAndSubscriberSameTenant()) {
-      return false;
-    }
-
-    return true;
+    return (
+      !!this.decodedUserJwt &&
+      this.options.selectionStrategy !== alwaysProvider &&
```
FrankEssenberger, line 519: same here

packages/core/src/connectivity/scp-cf/destination/destination.ts
```diff
@@ -55,7 +55,40 @@ export function parseDestination(
...
+      );
+      if (mapping) {
+        const [configProperty, destinationProperty] = mapping;
+        if (!additionalHeaders[destinationProperty]) {
+          additionalHeaders[destinationProperty] = {};
+        }
+        const headerKey = originalKey.replace(configProperty, '');
+        additionalHeaders[destinationProperty][headerKey] = value;
+      }
+      return additionalHeaders;
```
FrankEssenberger, line 88: Are the query parameters also handled here? In general I do not like this method because it is hard to read. My suggestion:
- One method: getDestinationPropertiesMathcingPattern(): {key:value}.
- One method: adjustKeys() replacing the keys

FrankEssenberger (commented):
General: Perhaps consider two PRs for refactoring and code.

packages/core/src/http-client/http-client.ts
```diff
@@ -43,15 +43,19 @@ export async function buildHttpRequest(
 ): Promise<DestinationHttpRequestConfig> {
   if (customHeaders) {
     logger.warn(
-      `The custom headers are provided with the keys: ${Object.keys(
+      `The custom headers are provided with the keys: [ ${Object.keys(
         customHeaders
-      )}. These keys will overwrite the headers created by the SDK.`
+      ).join(
+        ', '
+      )} ]. These keys will overwrite the headers created by the SDK.`
```
florian-richter, line 50: I'd prefer something like
```
`The custom headers are provided with the keys: 
  - ${Object.keys(
  customHeaders
).join(
  '\n  - '
)}
These keys will overwrite the headers created by the SDK.`
```
As it probably reads easier. But I can appreciate the easier copy paste with the `[]` but in that case it should also be double quoted like JSON...

florian-richter (commented):
Some minor improvements

packages/core/src/connectivity/scp-cf/destination/destination.ts
```diff
@@ -55,7 +55,40 @@ export function parseDestination(
...
+  const additionalHeadersAndQueryParameters =
+    getAdditionalHeadersAndQueryParameters(destinationConfig);
+
+  return sanitizeDestination({
+    ...destination,
+    ...additionalHeadersAndQueryParameters
+  });
+}
+
+function getAdditionalHeadersAndQueryParameters(
```
FrankEssenberger, line 67: I do not get where the query parameters are handled.

FrankEssenberger (changes requested):
Main point is the `getAdditionalHeadersAndQueryParameters` method where I do not see the query parameter being handled.

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -497,56 +504,38 @@ class DestinationFromServiceRetriever {
   }
 
   private isProviderNeeded(
-    resultFromSubscriber: DestinationSearchResult
+    resultFromSubscriber: DestinationSearchResult | undefined
```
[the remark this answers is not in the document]
marikaner: Not 100%. See [styleguide](https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-undefined-for-explicit-absence-of-values).

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
marikaner: I kind of agree and kind of disagree. Will revert ;)

packages/core/src/http-client/http-client.ts
marikaner: Done

packages/core/src/connectivity/scp-cf/destination/destination.ts
marikaner: I only handle them for the OData case, totally forgot about the http-client. I will try to get this in here, might be more complex than expected, because we do funny stuff with query parameters for OData...

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@
 
 ## New Functionality
 
+- [core] Support `OAuth2JWTBearer` authentication type.
```
```suggestion
```

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -156,30 +159,33 @@ class DestinationFromServiceRetriever {
...
-      options.isolationStrategy || IsolationStrategy.Tenant;
-    this.options.selectionStrategy =
-      options.selectionStrategy || subscriberFirst;
-    this.options.useCache = options.useCache || false;
+    this.options = {
+      isolationStrategy: IsolationStrategy.Tenant,
+      selectionStrategy: subscriberFirst,
+      useCache: false,
+      ...options
+    };
```
florian-richter, line 182: 👍  I like this, but it is a bit implicit that `...options` may overrides those default values. Maybe it would be better like this?
```suggestion
    const defaultOptions =  {
      isolationStrategy: IsolationStrategy.Tenant,
      selectionStrategy: subscriberFirst,
      useCache: false,
      ...options
    };
    this.options = { ...defaultOptions, ...options };
```

packages/core/src/http-client/http-client.ts
```diff
@@ -207,6 +206,8 @@ function mergeRequestWithAxiosDefaults(request: HttpRequest): HttpRequest {
 }
 
 function executeWithAxios(request: HttpRequest): Promise<HttpResponse> {
+  // https://example.com/sap/opu/odata/sap/API_TEST_SRV/TestActionImportNoParameterNoReturnType?$format=json
+  // https://example.com/sap/opu/odata/sap/API_TEST_SRV/TestActionImportNoParameterNoReturnType?$format=json
```
florian-richter, line 210: Is this left in here intentionally? Why the same URL twice?

florian-richter (approved):
LGTM but I had one more minor improvement idea:

packages/core/src/http-client/http-client.ts
marikaner: No, I also noticed that, it's gone.

packages/core/src/connectivity/scp-cf/destination/destination.ts
```diff
@@ -55,7 +55,54 @@ export function parseDestination(
...
+    (additionalProperties: Record<string, any>, [originalKey, value]) => {
+      const headerKey = originalKey.replace(originalKeyPrefix, '');
+      additionalProperties[headerKey] = value;
+      return additionalProperties;
+    },
+    {}
+  );
+}
+
+function getAdditionalHeadersAndQueryParameters(
```
FrankEssenberger, line 85: I find the code much clearer now. Anyway could you perhaps add a little doc about the situation in the kind of "There is a convention of putting parameters with XYZ on destination, client should handle bla bla. Link to the destination service docs" so that it is 100% clear what this is about.

packages/core/src/connectivity/scp-cf/destination/destination.ts
```diff
@@ -55,7 +55,54 @@ export function parseDestination(
...
+  destinationConfig: DestinationConfiguration
+): Record<string, any> {
```
FrankEssenberger, line 87: Could you perhaps use a strong type like an interface with the keys headers and query parameters?

packages/core/src/connectivity/scp-cf/destination/destination.ts
marikaner: Makes sense!

FrankEssenberger (approved):
LGTM only minor questions comments left

packages/core/src/connectivity/scp-cf/destination/destination.ts
marikaner: No, because I  don't know the keys. Those will either be the headers or the query parameters. The keys `headers` and `queryParameter` are not returned here, but wrapped around outside.

marikaner: Done.

PR #1450, opened 2021-07-21
Adjust onpremise_proxy_port environment variable

tomfrenken:
With this PR we are now able to use `onpremise_proxy_http_port` if present in the connectivity service binding, otherwise we still use the deprecated `onpremise_proxy_port`.

### DIFF
--- a/packages/core/src/connectivity/scp-cf/connectivity-service.ts
+++ b/packages/core/src/connectivity/scp-cf/connectivity-service.ts
@@ -49,9 +49,16 @@ interface HostAndPort {
   protocol: Protocol.HTTP;
 }
 
-function proxyHostAndPort(): HostAndPort {
+export function proxyHostAndPort(): HostAndPort {
   const service = readConnectivityServiceBinding();
 
+  if(typeof service.credentials.onpremise_proxy_http_port !== 'undefined'){
+    return {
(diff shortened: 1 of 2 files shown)


packages/core/src/connectivity/scp-cf/connectivity-service.ts
```diff
@@ -49,9 +49,16 @@ interface HostAndPort {
   protocol: Protocol.HTTP;
 }
 
-function proxyHostAndPort(): HostAndPort {
+export function proxyHostAndPort(): HostAndPort {
   const service = readConnectivityServiceBinding();
 
+  if (typeof service.credentials.onpremise_proxy_http_port !== 'undefined') {
```
FrankEssenberger, line 55: cloudn't we just use something like:
```
port: service.credentials.onpremise_proxy_http_port || service.credentials.onpremise_proxy_port
```

FrankEssenberger (approved):
LGTM only minor comment on if statement.

packages/core/src/connectivity/scp-cf/connectivity-service.ts
tomfrenken: Ahh I didn't know I can use logical operators like that, that's good to know!

PR #1453, opened 2021-07-22
chore: Fix incorrect deprecation notes

marikaner:
This PR fixes the incorrect deprecation notes introduced by #1408.


packages/core/src/odata-common/selectable/legacy/enum-field.ts
```diff
@@ -6,7 +6,6 @@ import { ConstructorOrField } from '../constructor-or-field';
 import { EdmTypeField } from '../edm-type-field';
 
 /**
- * @deprecated Since v1.27.0. Use [[XY]] instead.
```
jjtang1985, line 9: 👍

jjtang1985 (approved):
LGTM

PR #1454, opened 2021-07-22
chore: Improve metadata

FrankEssenberger:
Improve our metadata. Things done:
- Use the package version if a pregenerated lib exists
- Header added to the usage examples
- Sentence before usage example now API specific.
- `repository` property now also added for generic part.


packages/openapi-generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -41,8 +40,7 @@ export async function getGenericGenerationAndUsage(): Promise<GenerationAndUsage
 
 function getGenericUsage(): InstructionWithTextAndHeader {
   return {
-    instructions: genericCodeSample(),
-    text: genericUsageText,
```
jjtang1985, line 45: question: for generic usage, I guess we decided to show static text.
The requirement has been changed, right?

packages/openapi-generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -62,14 +60,14 @@ function getApiSpecificUsage(
...
-      text: apiSpecificUsageText,
+      ...instructions,
+
       header: usageHeaderText
     };
   }
   return {
     instructions: '',
-    text: apiSpecificUsageText,
+    text: '',
```
jjtang1985, line 70: minor: I see many cases that we set empty string to `text`, which means this can be optional. The same to other properties like `instructions`.

packages/generator-common/src/sdk-metadata/pregenerated-lib.spec.ts
```diff
@@ -39,6 +40,13 @@ describe('pregenerated-lib', () => {
     );
   }, 30000);
 
+  it('gives version for exisitng client', async () => {
+    nock('http://registry.npmjs.org/')
+      .get(/@sap-cloud-sdk\/core\/latest/)
+      .reply(200, { version: '1.2.3' });
+    await getLatestVersionOfNpmPackage('@sap-cloud-sdk/core');
```
jjtang1985, line 47: I know it does not fail, but having `expect` would be better.

jjtang1985 (approved):
LGTM. I left some minor comments.

packages/openapi-generator/src/sdk-metadata/generation-and-usage.ts
FrankEssenberger: we have an issue for handling also the cases where there is no text yet.

packages/generator-common/src/sdk-metadata/pregenerated-lib.spec.ts
FrankEssenberger: no this is a mistake. I forgot the assertion. I will change that.

PR #1448, opened 2021-07-20
Fix: Enum as key of entities.

jjtang1985:
This PR is related to the following tickets:
- [Fix: OData generation failed with enum key.](https://github.com/SAP/cloud-sdk-backlog/issues/213)
- [Fix Enum Types](https://github.com/SAP/cloud-sdk-backlog/issues/43)

- [x] ADR for enum handling
- [x] update all test services
- [x] changelog

PR #1456, opened 2021-07-22
fix: Allow only valid enum values when building filters

jjtang1985:
This PR contains further improvement based on https://github.com/SAP/cloud-sdk-js/pull/1448.
When building a filter on enum fields, it will not allow a random string to be used as a parameter for `equals()`.
e.g., `TestEntity.ENUM_PROPERTY.equals('random string');` will show type error.

- [ ] ~~compatibility notes~~
- [x] changelog


packages/core/src/odata-common/entity-serializer.ts
```diff
@@ -77,6 +78,9 @@ export function entitySerializer(tsToEdm: TsToEdmType): EntitySerializer {
     if (fieldValue === null || fieldValue === undefined) {
       return null;
     }
+    if (field instanceof EnumField) {
```
marikaner, line 81: Again, very minor, maybe move this below `EdmTypeField`.

packages/core/src/odata-common/selectable/field-builder.ts
```diff
@@ -194,4 +195,22 @@ export class FieldBuilder<FieldOfT extends ConstructorOrField<any>> {
       isSelectable
     });
   }
+
+  buildEnumField<EnumT extends string, NullableT extends boolean>(
+    fieldName: string,
+    enumType: Record<string, EnumT>,
+    isNullable: NullableT
```
marikaner, line 202: Should we have a default value here?

packages/core/src/odata-v4/payload-value-converter.ts
```diff
@@ -144,17 +143,15 @@ const deserializers: EdmTypeMapping = {
   'Edm.Date': edmDateToMoment,
   'Edm.DateTimeOffset': edmDateTimeOffsetToMoment,
   'Edm.Duration': edmDurationToMoment,
-  'Edm.TimeOfDay': edmTimeOfDayToTime,
-  'Edm.Enum': identity
+  'Edm.TimeOfDay': edmTimeOfDayToTime
```
marikaner, line 146: 👍 Do we need a compatibility note? And while I think it is good to get rid of this, isn't this a breaking change?

packages/core/test/test-util/filter-factory.ts
```diff
@@ -126,6 +126,6 @@ export const testFilterLambdaExpressionFilterFunctionOnLink = {
 };
 
 export const testFilterEnum = {
-  filter: TestEntityV4.ENUM_PROPERTY.equals('test'),
-  odataStr: "EnumProperty eq 'test'"
+  filter: TestEntityV4.ENUM_PROPERTY.equals('Member1'),
```
marikaner, line 129:
```suggestion
  filter: TestEntityV4.ENUM_PROPERTY.equals(TestEnum.Member1),
```

packages/generator/src/complex-type/field-type-class.ts
```diff
@@ -96,5 +96,9 @@ export function createPropertyFieldInitializer(prop: VdmProperty): string {
     return `this._fieldBuilder.buildComplexTypeField('${prop.originalName}', ${prop.fieldType}, ${prop.nullable})`;
   }
 
+  if (prop.isEnum) {
```
marikaner, line 99: Do you need another branch for enum collection?

packages/generator/src/edmx-to-vdm/common/complex-type.ts
```diff
@@ -118,9 +118,9 @@ export function getTypeMappingComplexProperties(
 
   if (isEnum) {
     return {
-      edmType: 'Edm.Enum',
+      edmType: 'Edm.String',
```
marikaner, line 121: Is this correct?

packages/generator/src/edmx-to-vdm/common/enum-type.ts
```diff
@@ -50,6 +50,7 @@ function parseMember(edmxEnumType: EdmxEnumType): VdmEnumMemberType[] {
 
 function validateUniqueness(edmxEnumType: EdmxEnumType) {
   if (unique(edmxEnumType.Member).length !== edmxEnumType.Member.length) {
+    // TODO: when validation option is added, throw an error based on the option
```
marikaner, line 53: Maybe mention the name of the option, because this could be easily missed.

packages/generator/src/generator-utils.ts
```diff
@@ -48,7 +48,7 @@ export type EdmTypeMapping = {
 };
 
 type EdmTypeMappingWithoutEnum = {
-  [key in Exclude<EdmTypeShared<'any'>, 'Edm.Enum'>]: string | undefined;
+  [key in EdmTypeShared<'any'>]: string | undefined;
```
marikaner, line 51: Is the name still appropriate or can we remove `WithoutEnum`?

marikaner (commented):
Looks quite good to me, I just left a few comments and questions.

packages/generator/src/edmx-to-vdm/common/complex-type.ts
jjtang1985: The goal is to get rid of the `Edm.Enum`. Actually, the `edmType` of the enum situation is not useful.
If we don't want to make it optional, I guess `Edm.String` is the type fits well.

packages/core/test/test-util/test-services/v4/multiple-schemas-service/TestEntity3.ts
```diff
@@ -13,6 +13,7 @@ import {
   EdmTypeField,
   EntityBuilderType,
   EntityV4,
+  EnumFieldV2,
```
FrankEssenberger, line 16: As I mentioned the name with `V2` strongly triggers oData `V2` like here EntityV4 is for oData version. So I would prefer to have a different extension.

test-packages/type-tests/test/v4/filters.spec.ts
```diff
@@ -10,3 +11,12 @@ and(
     any(TestEntityMultiLink.STRING_PROPERTY.equals('test'))
   )
 );
+
+// $ExpectType Filter<TestEntity, string>
+TestEntity.ENUM_PROPERTY.equals('Member1');
```
FrankEssenberger, line 16: Could you also add a test for integer based enum fixing what we expect. I know there were some limitation but a test is always good to fix our expected behaviour.

FrankEssenberger: or is this the test from above?

FrankEssenberger (approved):
LGTM only minor comments

packages/generator/src/complex-type/field-type-class.ts
jjtang1985: ok, this was missing from the beginning and I fixed it.
It's like you have you have the following boolean, where every boolean might be true or false.
- isEnum
- isCollection
- isComplex

test-packages/type-tests/test/v4/filters.spec.ts
jjtang1985: Discussed, the number based enum is about the Option A and Option B in the ADR, which is not implemented.
There are tests to make sure numbers are not allowed to be used as parameter.

marikaner (dismissed):
I created another [PR](#1465) where I tried to show that two fields are not necessary. Please check whether we can apply this.

packages/core/src/odata-common/selectable/enum-field.ts
```diff
@@ -0,0 +1,66 @@
...
+   * Creates an instance of Filter for this field and the given value using the operator 'ne', i.e. `!=`.
+   * @param value - Value to be used in the filter
+   * @returns The resulting filter
+   */
+  notEquals(value: EnumType<EnumT>): Filter<EntityT, string> {
+    return new Filter(this.fieldPath(), 'eq', value, this.edmType);
+  }
+}
+
+export type EnumType<T extends string> = `${T}`;
```
marikaner, line 66: Please add some API documentation

packages/core/src/odata-common/selectable/enum-field.ts
```diff
@@ -0,0 +1,66 @@
+import { Entity } from '../entity';
+import { EdmTypeShared } from '../edm-types';
+import { Filter } from '../filter';
+import { Field, FieldOptions } from './field';
+import { ConstructorOrField } from './constructor-or-field';
+import { ComplexTypeField, getEntityConstructor } from './complex-type-field';
+
+export class EnumField<
```
marikaner, line 8: Please add some API documentation

packages/core/src/odata-common/selectable/field-builder.ts
```diff
@@ -194,4 +195,22 @@ export class FieldBuilder<FieldOfT extends ConstructorOrField<any>> {
       isSelectable
     });
   }
+
+  buildEnumField<EnumT extends string, NullableT extends boolean>(
```
marikaner, line 199: Same here, please add docs.

packages/generator/src/complex-type/field-type-class.ts
```diff
@@ -85,7 +85,7 @@ function property(
 
 export function createPropertyFieldInitializer(prop: VdmProperty): string {
   if (prop.isCollection) {
-    if (prop.isComplex) {
+    if (prop.isComplex || prop.isEnum) {
```
marikaner, line 88: 👍 I hope we have some kind of test for this? (I think I remember we had something)

packages/generator/src/edmx-to-vdm/common/complex-type.ts
marikaner: Generally our API is ugly because we have edmType for all properties, even if they are not edm properties, but well...
If I am not mistaken this type is not used anymore, is it? So should we use anything here? Maybe assign undefined or null or something like that? I still find it confusing to have "Edm.String" here.

packages/generator/src/edmx-to-vdm/edmx-to-vdm-util.ts
```diff
@@ -135,7 +135,7 @@ export function typesForCollection(
   if (isComplexTypeOrEnumType(typeInsideCollection)) {
     if (isEnumType(typeInsideCollection, enumTypes)) {
       return {
-        edmType: typeInsideCollection,
+        edmType: 'Edm.String',
```
marikaner, line 138: I guess it is the same here, this is not used anywhere, right?

packages/generator/src/generator-utils.ts
```diff
@@ -190,7 +190,11 @@ export function getGenericParameters(
 ): string {
   const params = [entityClassName];
   if (prop.isEnum) {
-    params.push("'Edm.Enum'");
+    if (prop.isCollection) {
+      params.push(`typeof ${prop.jsType}`);
+    } else {
+      params.push(`${prop.jsType}`);
+    }
```
marikaner, line 197: This is hard to grasp and is is easy to make errors here, maybe we should rather write this down explicitly as we have done it in some other cases. It is a bit more verbose, but so much easier to read.

```ts
if (isCollection) {
  if(isEnum) {
    ...
  } else if (isComplex) {
    ...
  } else {
    ...
 }
} else {
    if(isEnum) {
    ...
  } else if (isComplex) {
    ...
  } else {
    ...
 }
}
```

packages/generator/src/generator.ts
```diff
@@ -235,7 +235,7 @@ export async function generateSourcesForService(
 
   service.enumTypes.forEach(enumType => {
     logger.info(
-      `[${service.originalFileName}] Generating enum type '${enumType.originalName}' ...`
+      `[${service.originalFileName}] Generating enum type ${enumType.originalName} ...`
```
marikaner, line 238: I liked the version with `'` better.

marikaner (commented):
Looks good, I am still unsure about the usage of `Edm.String`

packages/generator/src/complex-type/field-type-class.ts
jjtang1985: It's about a collection of enum type in a complex type, which was missing from the test spec file.
I added this case and the generation went well.

packages/core/src/odata-common/selectable/enum-field.ts
```diff
@@ -0,0 +1,80 @@
+import { Entity } from '../entity';
+import { EdmTypeShared } from '../edm-types';
+import { Filter } from '../filter';
+import { Field, FieldOptions } from './field';
+import { ConstructorOrField } from './constructor-or-field';
+import { ComplexTypeField, getEntityConstructor } from './complex-type-field';
+
+/**
+ * Represents a property with a enum value.
+ *
```
marikaner, line 10:
```suggestion
 * Represents a property with an enum value.
```

packages/core/src/odata-common/selectable/enum-field.ts
```diff
@@ -0,0 +1,80 @@
...
+ * @typeparam EntityT - Type of the entity the field belongs to
+ * @typeparam EnumT - Enum type that contains all valid enum entries for this field
```
marikaner, line 12:
```suggestion
 * @typeparam EntityT - Type of the entity the field belongs to.
 * @typeparam EnumT - Enum type that contains all valid enum entries for this field.
```

packages/core/src/odata-common/selectable/enum-field.ts
```diff
@@ -0,0 +1,80 @@
...
+> extends Field<EntityT, NullableT, SelectableT> {
+  /**
+   * @deprecated Since v1.48.0. This property is not used anymore.
+   */
+  readonly edmType: EdmTypeShared<any> = 'Edm.Enum';
+
+  /**
+   *
+   * Creates an instance of EnumField.
+   *
```
marikaner, line 30:
```suggestion
   * Creates an instance of EnumField.
```

packages/generator/src/generator-utils.ts
```diff
@@ -189,12 +189,22 @@ export function getGenericParameters(
...
+    } else if (prop.isComplex) {
+      params.push(prop.jsType);
+    } else {
+      params.push(`'${prop.edmType}'`);
+    }
+  } else {
+    if (prop.isEnum) {
+      params.push(`${prop.jsType}`);
+    } else if (prop.isComplex) {
+      // prettier-ignore
```
marikaner, line 204:
```suggestion
      continue;
```

marikaner (approved):
LGTM, I made minor adjustments in the comments.

PR #1463, opened 2021-07-28
chore: fix flaky tests

marikaner:
The test-util tests were writing a lot of files with `writeFileSync` causing the tests to fail 90% of the time for me. I replaced this with `mock-fs`.


FrankEssenberger (approved):
LGTM

PR #1465, opened 2021-07-29
chore: Enum combined

marikaner:
I think we can easily combine the two enum fields instead of having a separate one, unless I am missing anything.

@jjtang1985, @FrankEssenberger I would be interested in your opinions on this.


jjtang1985 (approved):
Thanks for testing the combined version.
LGTM.

PR #1471, opened 2021-07-30
fix: Fix parsing of proxy env variable with `-` in host name

### LINKED ISSUE #1464: Dash in HTTP_PROXY host breaks proxy configuration
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@sap-cloud-sdk/core@1.47.0` for the project I'm working on.

When setting `HTTP_PROXY` or `HTTPS_PROXY` to e.g. `http://service-name:8080`, the dash is not considered part of the host, but instead lands inside the `port`. Effectively deactivating the proxy configuration entirely.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@sap-cloud-sdk/core/dist/connectivity/scp-cf/proxy-util.js b/node_modules/@sap-cloud-sdk/core/dist/connectivity/scp-cf/proxy-util.js
index 67a6f98..d01d8d7 100644
--- a/node_modules/@sap-cloud-sdk/core/dist/connectivity/scp-cf/proxy-util.js
+++ b/node_modules/@sap-cloud-sdk/core/dist/connectivity/scp-cf/proxy-util.js
@@ -151,7 +151,7 @@ var addHost = function (groups) { return function (proxyConfiguration) {
  * @returns Configuration with default values or undefined if the parsing failed.
  */
 function parseProxyEnv(proxyEnvValue) {
-    var regex = /(?<protocolWithDelimiter>(?<protocol>^.+):\/\/)?(?<userPwdWithDelimeter>(?<user>.+):(?<pwd>.+)@)?(?<hostAndPort>(?<host>[\w.]+):?(?<port>.+)?)/;
+    var regex = /(?<protocolWithDelimiter>(?<protocol>^.+):\/\/)?(?<userPwdWithDelimeter>(?<user>.+):(?<pwd>.+)@)?(?<hostAndPort>(?<host>[\w.-]+):?(?<port>.+)?)/;
     var parsed = regex.exec(proxyEnvValue);
     if (parsed === null || parsed === void 0 ? void 0 : parsed.groups) {
         var groups = parsed.groups;
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

marikaner:
Proxy parsing only allowed alphanumeric characters, `_` and `.` to be present in the host name. `-` should also be allowed. Same applies to the username and password.

Closes #1464.
Closes SAP/cloud-sdk-backlog#240.


packages/core/src/connectivity/scp-cf/proxy-util.spec.ts
```diff
@@ -232,6 +232,11 @@ describe('proxy-util', () => {
       );
       expect(parseProxyEnv('user:password@some.host')?.host).toBe('some.host');
 
+      expect(parseProxyEnv('some-host')?.host).toBe('some-host');
+      expect(parseProxyEnv('us-er:pass-word@some-host')?.host).toBe(
+        'some-host'
+      );
```
FrankEssenberger, line 238: Could you also make a test with `us-er:pass-word@some-host:1234` where also the port is correctly asserted.

FrankEssenberger (approved):
Only one minor comment.

packages/core/src/connectivity/scp-cf/proxy-util.ts
```diff
@@ -200,13 +200,13 @@ export function parseProxyEnv(
   proxyEnvValue: string
 ): ProxyConfiguration | undefined {
   const regex =
-    /(?<protocolWithDelimiter>(?<protocol>^.+):\/\/)?(?<userPwdWithDelimeter>(?<user>.+):(?<pwd>.+)@)?(?<hostAndPort>(?<host>[\w.]+):?(?<port>.+)?)/;
+    /(?<protocolWithDelimiter>(?<protocol>^\w+):\/\/)?(?<userPwdWithDelimiter>(?<user>.+):(?<pwd>.+)@)?(?<hostAndPort>(?<host>[\w.-]+):?(?<port>.+)?)/;
```
jjtang1985, line 203: The regex is too complicated so I can only rely on the tests.
I wonder, why don't we use the node native [URL class](https://nodejs.org/api/url.html#url_url). Maybe I miss something?
Here is the playground [tests](https://www.typescriptlang.org/play?ts=4.1.5#code/GYVwdgxgLglg9mABABwIYCcDOBTAqugGwAo11UBbALkUynRjAHMBKa3AJQBkBvAWAChEiCAlooMOACaIAvIjDYA7og6cSGCswDcAoSLCY4BbADoCcRuqzZJJ5OjhQ4Igtt3DRR0+cukpJgAs4WjdBDwMvMwsrf2Q4dChQvU9jKN8JGxMQHHQwCmwk8MNUnxjMtExMRXjJQvRsKBBc8WtJHX4AXwEBPzxCIgByAKgoZEoAenHsgFpsdEoKzGnq9EkAAUNybGmg2koADgAGI4HQ3vxiIZGxyahsPZOzjIvBu9oTEXJT9u7+IA), where shows you might have to handle the special case by our own:
- port number is 80, see [doc](https://nodejs.org/api/url.html#url_url_port)
- protocol is missing

I know these two special case also cost effort from our side, but getting rid of complicated regex might be better in my point of view.

jjtang1985 (changes requested):
I think it would be better to delegate the parsing to other libs like the node native URL.

packages/core/src/connectivity/scp-cf/proxy-util.ts
FrankEssenberger: I was not aware, that the URL class also handles the user and password. I guess this should be fine. The only thing we would lose is the log statements in case the password and or user contain unencoded characters.

FrankEssenberger: besides that I would also say the library is better than the ugly regex

packages/core/src/connectivity/scp-cf/proxy-util.spec.ts
marikaner: Done.

packages/core/src/connectivity/scp-cf/proxy-util.ts
marikaner: Done.

jjtang1985 (approved):

Issue #1398, opened 2021-07-02
Breaking change in URL encoding of OData key properties


lennart-m:
Hi, I found that the cloud SDK does handle URL encoding for key properties differently since 1.45.
I would like to know if this was intended as I did not find anything about this breaking change in the release notes.

The change is this: up to 1.44 the key properties were encoded when performing an OData GET.
Since 1.45 this does not happen anymore.

### Example:

I created an application for reading and writing BSP applications to ABAP repositories to use in our CI/CD pipelines, consuming the SAP standard service `/UI5/ABAP_REPOSITORY_SRV`. The entity `Repository` has a key consisting of only one property, `Name`.

In order to read information about a repository I need to perform this request:
```http
GET /sap/opu/odata/UI5/ABAP_REPOSITORY_SRV/Repositories(Name='%2FUI2%2FUSHELL')?$format=json
```
When using version 1.44 of the cloud SDK exactly this URL is generated.

But since 1.45 the request looks like this:
```http
GET /sap/opu/odata/UI5/ABAP_REPOSITORY_SRV/Repositories(Name='/UI2/USHELL')?$format=json
```
This is a valid URL consisting of the segments 
- `Repositories(Name='`
- `UI2`
- `USHELL')?$format=json`

where there should only be one segment
- `Repositories(Name='%2FUI2%2FUSHELL')?$format=json`

So the system responds with this message:
_Invalid URI segment 'Repositories(Name=''_

Is this behavior intended? In this case the calling application would have to do the encoding itself; but in this case the cloud SDK must not do additional encoding in order to prevent duplicate encoding of strings.

Thanks a lot,

Lennart

jjtang1985:
Hi @lennart-m ,

Thanks for approaching us.
I assume you use the code snippets like 
```ts
YourEntity
.requestBuilder()
.getByKey(`/UI2/USHELL`)
```
and I can reproduce the issue, where the `/` is not encoded.

I'll create a backlog for it.

May I ask your timeline and do you have a workaround for it?

Best regards,
Junjie

lennart-m:
Hi @jjtang1985,

Thanks for your reply. You're right this is the URL I use. I haven't tested using `update` or `delete` but I guess URL generation will show the same behavior for those as well.

We are using the cloud SDK for internal purposes only (deployment in CI/CD) so there is no urgency. The workaround is not to upgrade the dependencies to 1.45 but stick to 1.44 or lower and this works fine.

Thanks a lot,

[...]

PR #1481, opened 2021-08-03
fix: encoding OData get by key

jjtang1985:
Related to https://github.com/SAP/cloud-sdk-js/issues/1398.

Closes SAP/cloud-sdk-backlog#200.

### DIFF
--- a/knowledge-base/adr/0024-url-encoding.md
+++ b/knowledge-base/adr/0024-url-encoding.md
@@ -9,6 +9,7 @@ Accepted.
 OData and OpenAPI requests built by the SDK needs to be encoded.
 We should decide, which parts of the URL need to be encoded and which not.
 Let's start from the structure of a URL.
+Emojis :thumbsup: and :thumbsdown: are used in this doc for specifying whether encoding will be applied by the SDK.
 
 ### URL
 
@@ -127,7 +128,7 @@ Option B is the winner, as Option A does not fit our requirement.
 ##### Pros
 
 - We just need this method ONCE for the whole URL
-- The special characters like `/` (as path delimiter), `&` (as query delimiter) etc. are not encoded. (See Appendix for more details)
+- The special characters like `/` (as path delimiter), `&` (as query delimiter) etc. are not encoded. (See [Appendix](#appendix))
 
 ##### Cons
 
@@ -154,7 +155,7 @@ Not accepted, as it does not work for edge cases.
 ##### Pros
 
 - We can choose specific components for encoding, which is flexible.
-- It solves the problem of Option A, that `/` can be encoded. (See Appendix for more details)
+- It solves the problem of Option A, that `/` can be encoded. (See [Appendix](#appendix))
 - For custom values, the encoding can be skipped.
 
 ##### Cons
@@ -171,7 +172,8 @@ Chosen, as this is a valid solution.
 - For the chosen components, we are able to apply encoding properly.
 - There might be still some other parts like "host", "service path" and "entity name" etc., which need coding. Potential solutions are mentioned in this doc.
 
-## Appendix: The difference between `encodeURI` and `encodeURIComponent`.
+## Appendix
+The table below shows the difference between `encodeURI` and `encodeURIComponent`.
 
 | Character | encodeURI | encodeURIComponent |
 | --------- | --------- | ------------------ |
--- a/knowledge-base/adr/0024-url-encoding.md
+++ b/knowledge-base/adr/0024-url-encoding.md
@@ -158,7 +158,7 @@ Not accepted, as it does not work for edge cases.
 
 ##### Cons
 
-- We have to call this method multiple times for all the parts that need encoding.
+- We have to call this function multiple times for all the parts that need encoding.
 - We have to be very careful to call `encodeURIComponent('a/b/c')` instead of `encodeURIComponent('/Entity('a/b/c')')`, as the latter one is wrong.
 
 #### Decision
--- a/knowledge-base/adr/0024-url-encoding.md
+++ b/knowledge-base/adr/0024-url-encoding.md
@@ -173,6 +173,7 @@ Chosen, as this is a valid solution.
 - There might be still some other parts like "host", "service path" and "entity name" etc., which need coding. Potential solutions are mentioned in this doc.
 
 ## Appendix
+
 The table below shows the difference between `encodeURI` and `encodeURIComponent`.
 
 | Character | encodeURI | encodeURIComponent |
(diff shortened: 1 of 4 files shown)


knowledge-base/adr/0024-url-encoding.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Host :thumbsdown:
+
+There are some cases where the host contains special characters like umlaut.
+As not all the browsers/http clients can handle such special characters, Internationalized Domain Name (IDN) plays a big role here.
+One can use an IDN converter so `www.häuser.de` can be converted to `www.xn--huser-gra.de`.
+As the host is usually configured in the destination, the SDK will not make any adjustment.
+
+### Path
+
```
FrankEssenberger, line 37: something missing?

knowledge-base/adr/0024-url-encoding.md
```diff
@@ -0,0 +1,191 @@
...
+- encode `a/b/c` first then apply Option A
+- ask the user to provide encoded value
+
+Both of the 2 proposals will lead to double encoding.
+
+#### Decision
+
+Not accepted, as it does not work for edge cases.
+
+####
```
FrankEssenberger, line 149: remove?

knowledge-base/adr/0024-url-encoding.md
```diff
@@ -0,0 +1,191 @@
...
+### Which parts are encoded by the SDK?
+
+To wrap up, the SDK will only encode the parts below:
+
+- The values of path parameters (the get by key structure of OData is called path parameter in OpenAPI)
+- The values of query parameters
+
+## Decision
+
+Option B is the winner, as Option A does not fit our requirement.
```
FrankEssenberger, line 120: Could you perhaps change the order:
- Alternatives
- Decision->encode parts not full URL
- Explain which parts are encoded by SDK

FrankEssenberger (dismissed):
Code wise fine. The ADR should be a bit polished and rearanged.

jjtang1985:
@FrankEssenberger , I rearranged the order of some sections and kept the long details into background.

knowledge-base/adr/0024-url-encoding.md
```diff
@@ -0,0 +1,193 @@
...
+
+Below is one example about how the path should look like in terms of OData
+
+```
+/servicePath/Entity('key')/to_MultiLink('key')/to_SingleLink
+```
+
+- service path: `/servicePath` :thumbsdown:
+- entity path: `/Entity(KeyName='value')` :thumbsup:
+- navigation property path: `/to_MultiLink(KeyName='value')/to_SingleLink` :thumbsup:
```
marikaner, line 57: See above, I find this information misleading.

knowledge-base/adr/0024-url-encoding.md
```diff
@@ -0,0 +1,193 @@
...
+
+OData allows you to use the following query options:
+
+- `$format=json` :thumbsdown:
+- `$top=1` :thumbsdown:
+- `$skip=1` :thumbsdown:
+- `$orderby=StringProperty` :thumbsdown:
+- `$select=StringProperty` :thumbsdown:
+- `$expand=to_MultiLink` :thumbsup:
+- `$filter=StringProperty eq 'string'` :thumbsup:
```
marikaner, line 110: Again, I find this misleading. I am not even sure whether we would encode $filter=**StringProperty eq 'string'** or $filter=StringProperty eq **'string'**, but this information here rather tells me that we would encode the whole thing.

knowledge-base/adr/0024-url-encoding.md
```diff
@@ -0,0 +1,193 @@
...
+
+##### Pros
+
+- We can choose specific components for encoding, which is flexible.
+- It solves the problem of Option A, that `/` can be encoded. (See [Appendix](#appendix))
+- For custom values, the encoding can be skipped.
+
+##### Cons
+
+- We have to call this method multiple times for all the parts that need encoding.
```
marikaner, line 161: It's not a method ;)

```suggestion
- We have to call this function multiple times for all the parts that need encoding.
```

knowledge-base/adr/0024-url-encoding.md
```diff
@@ -0,0 +1,193 @@
...
+- We have to be very careful to call `encodeURIComponent('a/b/c')` instead of `encodeURIComponent('/Entity('a/b/c')')`, as the latter one is wrong.
```
marikaner, line 162: Just my opinion: Both "cons" do not sound like real "cons".
- We have to use it as many times as we need it
- We have to use it only where it is correct
;)

knowledge-base/adr/0024-url-encoding.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### Decision
+
+Chosen, as this is a valid solution.
+
+## Decision
+
+Option B is the winner, as Option A does not fit our requirement.
```
marikaner, line 170: I think that option A is not an option as it does not solve any of our issues and just doesn't work. I would therefore prefer to follow the standard ADR approach and just describe what we are doing (decision) and give a background + consequences.
If we just have one option, there is less to read and the whole section on option A could just be mentioned in one sentence in the background.

marikaner (commented):
Implementation looks good 👍 

I am not a fan of the ADR to be very honest. I won't insist in changing it too much, however I would prefer it to not mention any options and just straight forward explain what we do.
In my opinion the important information are difficult to find because there is a lot less relevant information hiding it (especially option A - see my comments).

knowledge-base/adr/0024-url-encoding.md
```diff
@@ -0,0 +1,193 @@
...
+## Status
+
+Accepted.
+
+## Requirement
+
+The SDK will only encode the parts below:
+
+- The values of path parameters (the get by key structure of OData is called path parameter in OpenAPI)
+- The values of query parameters
```
FrankEssenberger, line 12: perhaps give an example - filter are query parameters?

knowledge-base/adr/0024-url-encoding.md
FrankEssenberger: I think we encode `eq 'string'` so the empty spaces are encoded and the value put in.

knowledge-base/adr/0024-url-encoding.md
FrankEssenberger: I had the same remark. Keep it much shorter:
- Say what we do: encode each user input part
- Say why we do it: encode the full case `host/something/?filter=Foo eq 'abc\def'` can not be solved because the encode method will not encode inside argument where it should but keep the path untouched.

marikaner (approved):

jjtang1985:
@FrankEssenberger , I'm going to merge this afternoon, and you still have some time to say no;)

FrankEssenberger (approved):
LGTM

PR #1488, opened 2021-08-05
chore: Fix dependencies

marikaner:
Fixes some references to vulnerable dependencies:
- color-string (although this was ok already)
- path-parse
- tar
- postcss

Closes SAP/cloud-sdk-backlog#217.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,10 +94,14 @@
   },
   "dependencies": {},
   "resolutions": {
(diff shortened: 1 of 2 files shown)


FrankEssenberger (approved):
LGTM - good we have yarn with the resolution option. BTW does yarn check automatically if the resolution is still necessary?

Issue #1262, opened 2021-05-03
Failed to fetch instance destinations. (Timed out after 10000ms)


LaxOn:
### Description of erroneous behaviour 

We have a Node.js CAP application [edom-retailer-core](https://github.wdf.sap.corp/c4u/edom-retailer-core/tree/feat/25-mdiclient-outbound). In [MDIClient.js](https://github.wdf.sap.corp/c4u/edom-retailer-core/blob/4983f3f46e72e5f12f24d6b144e4387872101724/srv/alpha/mdiclient/MDIClient.js), we are making a request via destination connecting to `MdiDestination`.

```
// package.json
"MdiDestination": {
   "kind": "rest",
   "credentials": {
      "destination": "mdi-dest",
      "requestTimeout": 1000000
```

The API in the destination retrieves 2000 items at a time and if it has more data, it will provide a `deltaToken` that we use to make the same call but with the `deltaToken` to retrieve the next 2000 items. We repeat this till all the items are retrieved. Currently our API has 19000+ items.

```javascript
// MDIClient.js
const MDIDest = await cds.connect.to('MdiDestination');
const logRequest = new cds.Request();
const baseQuery = `GET /log/sap.odm.businesspartner.BusinessPartner`;

let moreData = true;
while (moreData) {
   if (deltaToken !== false) {
     logRequest.query = `${baseQuery}?since=${deltaToken}`;
   }
   const response = await MDIDest.tx(logRequest).run(logRequest.query);

   ...

   if (!response.hasMoreEvents) {
      moreData = false;
   }
}
```

The request is successful for the first 6-9 iterations but throws an error afterwards:
```
Error: Timed out after 10000ms
    at buildError (/mnt/c/Users/C5315270/Desktop/projects/edom-retailer-core/node_modules/opossum/lib/circuit.js:710:17)
    at Timeout.<anonymous> (/mnt/c/Users/C5315270/Desktop/projects/edom-retailer-core/node_modules/opossum/lib/circuit.js:511:29)
    at listOnTimeout (internal/timers.js:554:17)
    at processTimers (internal/timers.js:497:7) {
  cause: ErrorWithCause: Failed to fetch instance destinations.
      at new ErrorWithCause (/mnt/c/Users/C5315270/Desktop/projects/edom-retailer-core/node_modules/@sap-cloud-sdk/util/dist/error-with-cause.js:31:16)
[...]

PR #1490, opened 2021-08-05
fix: add destination option

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#229

Added the option to pass a `{iss:someURL}` as a user jwt to the execute mehtods. This is usefull if you do not have a user JWT but want the execute for a different account than the provider. You need to know what you are doing, but this feature was requested in the past and we have it on the bare execute methods so this PR was about forwarding it to the openapi request builder.

I added test on the:
- http-client execute-http-request
- odata-request-builder
- openapi-request-builder

to see that on all entry levels the obption is considered. Put for discussion I see is the many argument. Perhaps it makes sense to make a combined config for request and destination fetching? Up for discussion.


packages/core/test/test-util/mocked-access-tokens.ts
```diff
@@ -25,6 +26,16 @@ const subscriberServiceTokenPayload = {
 
 export const subscriberServiceToken = signedJwt(subscriberServiceTokenPayload);
 
+// This tokens are used to test cases when as a user Jwt only {iss:someXSUAAurl} is provided.
```
marikaner, line 29:
```suggestion
// These tokens are used to test cases when the provided user JWT only contains `{ iss: someXSUAAurl }`.
```

packages/core/test/test-util/mocked-access-tokens.ts
```diff
@@ -25,6 +26,16 @@ const subscriberServiceTokenPayload = {
...
+// See docs on [[DestinationAccessorOptions]] for mor details
```
marikaner, line 30:
```suggestion
// See docs on [[DestinationAccessorOptions]] for more details.
```

marikaner: This doc comment is intended for internal use only?

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -380,8 +380,8 @@ class DestinationFromServiceRetriever {
 
   private async addClientCertAuth(): Promise<Destination> {
     const accessToken = await serviceToken('destination', {
-      userJwt: this.decodedUserJwt,
-      ...this.options
+      ...this.options,
+      userJwt: this.decodedUserJwt
```
marikaner, line 384: Does this mean, that if I set `iss` in my options and this is also set in the JWT, it will be ignored?
Is this what we want? I see advantages/disadvantages on both sides.

If we leave it as is, we should add a compatibility note.

packages/core/src/http-client/http-client.spec.ts
```diff
@@ -222,6 +234,40 @@ describe('generic http client', () => {
...
+          .reply(200, 'iss token used on the way')
+      ];
+
+      const response = await executeHttpRequest(
+        { destinationName: 'ERNIE-UND-CERT' },
+        { method: 'get' },
+        {},
+        { iss: onlyIssuerXsuaaUrl }
+      );
+      allMocksUsed(nocks);
```
marikaner, line 267: Consider renaming this function to `expectAllMocksUsed`. It isn't entirely clear that this contains assertions.

packages/core/src/http-client/http-client.ts
```diff
@@ -139,14 +147,21 @@ export async function buildAxiosRequestConfig<T extends HttpRequestConfig>(
...
  * @param options - An [[HttpRequestOptions]] of the http request for configuring e.g., csrf token delegation. By default, the SDK will not fetch the csrf token.
+ * @param destinationOptions - Options to employ when fetching destinations.
  * @returns A promise resolving to an [[HttpResponse]].
  */
 export function executeHttpRequest<T extends HttpRequestConfig>(
   destination: Destination | DestinationNameAndJwt,
   requestConfig: T,
-  options?: HttpRequestOptions
+  options?: HttpRequestOptions,
+  destinationOptions?: DestinationOptions
```
marikaner, line 157: Two optional parameters can be inconvenient, if you need the second one.
We actually already talked about this in the past and decided not to add this.
We might want to consider writing it down somewhere (ADR?)
Please check this [comment](https://github.com/SAP/cloud-sdk-js/issues/1262#issuecomment-848918155).

The same of course applies to the other functions here.

packages/core/src/openapi/openapi-request-builder.spec.ts
```diff
@@ -68,16 +91,67 @@ describe('openapi-request-builder', () => {
           limit: 100
         }
       },
-      { fetchCsrfToken: true }
+      { fetchCsrfToken: true },
+      undefined
+    );
+  });
+
+  it('executes a request using the (iss) token instead of the whole jwt', async () => {
```
marikaner, line 99:
```suggestion
  it('executes a request using the (iss) token instead of the whole JWT', async () => {
```

marikaner: I am missing the fact that `iss` is passed through options. To me this sounds like we are using only `iss` from the JWT, not the whole token.
Also, isn't `iss` a URL, not a token?

marikaner (changes requested):
As usual, mostly minor stuff. However, I think we should not pass the destination options in the generic http client, as discussed but not documented properly in the past.

Also, please add to the changelog.

packages/core/test/test-util/mocked-access-tokens.ts
FrankEssenberger: yes

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
FrankEssenberger: no in the initialitaion a JWT is build from the iss if present and then used. It is not validated like a JWT>

packages/core/src/http-client/http-client.ts
FrankEssenberger: I also had the feeling that this is not a nice API. But I was not aware we made it a clear `no go`, I adjusted the API.

packages/core/src/openapi/openapi-request-builder.spec.ts
FrankEssenberger: On line 126 it is passed. The reason for the difference of passing JWT or the iss (and creating a JWT from it) is, that we do a validation if we get the `userJwt` property and not validation if we have the `iss`.

packages/core/src/http-client/http-client.ts
```diff
@@ -134,13 +137,14 @@ export async function buildAxiosRequestConfig<T extends HttpRequestConfig>(
  * and executes it (using Axios).
  * @param destination - A destination or a destination name and a JWT.
  * @param requestConfig - Any object representing an HTTP request.
- * @param options - An [[HttpRequestOptions]] of the http request for configuring e.g., csrf token delegation. By default, the SDK will not fetch the csrf token.
+ * @param options - An [[HttpRequestOptions]]  of the http request for configuring e.g., csrf token delegation and [[DestinationOptions]] to configure destination retrieval options. By default, the SDK will not fetch the csrf token.
```
marikaner, line 140:
```suggestion
 * @param options - An [[HttpRequestOptions]] of the http request for configuring e.g., csrf token delegation and [[DestinationOptions]] to configure destination retrieval options. By default, the SDK will not fetch the csrf token.
```

packages/core/src/openapi/openapi-request-builder.spec.ts
FrankEssenberger: I adjusted also the test name

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -559,3 +594,12 @@ class DestinationFromServiceRetriever {
...
 }
+
+interface Jwt {
+  decoded: JwtPayload;
+  encoded: string;
+}
+
+interface IssuerJwt {
+  decoded: { iss: string };
+}
```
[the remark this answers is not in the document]
marikaner: Please also note that `Jwt` might be misleading as the `jsonwebtoken` lib already specifies that type. I don't have a good name... maybe something with "pair" or "meta"
¯\\_(ツ)_/¯

marikaner (changes requested):
Let's talk about some of this, as suggested.

packages/core/src/connectivity/scp-cf/destination/destination-accessor.ts
```diff
@@ -29,10 +29,10 @@ export async function useOrFetchDestination(
   options: DestinationOptions = {}
 ): Promise<Destination | null> {
   return isDestinationNameAndJwt(destination)
-    ? getDestination(destination.destinationName, {
-        userJwt: destination.jwt,
-        ...options
-      })
+    ? getDestination(
+        destination.destinationName,
+        destination.jwt ? { userJwt: destination.jwt, ...options } : options
```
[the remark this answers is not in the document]
FrankEssenberger: The code before added a {userJwt:undefined} to the options. If you did than a spread not in the right order it could overwrite your existing value.

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -165,22 +180,17 @@ class DestinationFromServiceRetriever {
...
+    Omit<DestinationOptions, 'userJwt' | 'iss'>,
     'isolationStrategy' | 'selectionStrategy' | 'useCache'
   >;
 
   private constructor(
     readonly name: string,
     options: DestinationOptions,
-    readonly decodedUserJwt: JwtPayload | undefined,
-    readonly providerClientCredentialsToken: string
+    readonly userOrIssToken: Jwt | IssuerJwt | undefined,
```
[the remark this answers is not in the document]
FrankEssenberger: because you can also give a user JWT for the provider account.

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -269,10 +279,16 @@ class DestinationFromServiceRetriever {
     );
   }
 
+  // The user JWT can be a full JWT containing user information but also a reduced one setting only the iss value
+  // This method divides the two cases.
```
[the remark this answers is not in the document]
FrankEssenberger: done

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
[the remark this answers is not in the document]
FrankEssenberger: I went in circles over hours. This is what I came up with.

packages/core/src/connectivity/scp-cf/jwt.ts
```diff
@@ -409,3 +409,27 @@ export function checkMandatoryValue<InterfaceT, JwtKeysT>(
     );
   }
 }
+
+export interface JwtPair {
```
marikaner, line 413: Don't forget the API Docs!

marikaner (approved):
LGTM 👍

PR #1498, opened 2021-08-10
chore: X509 research

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#232.

Investigation on the options to support x509 authentication for token retrieval.


knowledge-base/adr/0024-x509-service-authentication.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Comparison A versus B
+
+We all agree to go for library.  
+With option B we are future-proof for other things to come.
+Also, from a security point of view we should use the the `xssec` and not a homemade implementation.
+We make the assumptions:
+
+- non-breaking change
+- licence of xssec compatible with SDK licence
```
marikaner, line 157: Checked again, we're good. Maybe remove this from the ADR then?

marikaner (approved):
I just overflew this and read only the most relevant sections, as we had the meeting on it already. Hope that is enough.

knowledge-base/adr/0024-x509-service-authentication.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+In this case the `key` property is not part of the VCAP service variables, and you need to bring it from the outside.
+
+## Decision
+
+For the self-managed certificates we have to find a secure way for the customer to pass the private part of the certificate to the application.
+In the first version this is not considered.
+Implementation is easy it is more a security question.
+
```
jjtang1985, line 126: minor: maybe mention B is the winner explicitly in the decision.

jjtang1985 (approved):
The decision is clear.
I hope we have the timeline somewhere, so we can prepare this timely, since it takes time for switching to a new lib.

PR #1502, opened 2021-08-12
feat: add more information in openapi schema comments

deekshas8:
Closes SAP/cloud-sdk-backlog#42

Extended OpenApiSchema with new properties to include additional type-specific information in the generated comments for schemas. Example, tags like deprecated, example, etc or information such as maxLength, minimum, etc.  
Updated the tests to include these properties in the schema.


FrankEssenberger (changes requested):
Could you include the properties in one of our test spec files  - some example like `minimum`, `maximum` or `maxLenght` and see that they are ending up in our generated clients (regeneration is needed)

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -160,7 +174,37 @@ function parseObjectSchemaProperties(
           ? undefined
           : propSchema.description,
         name: propName,
-        required: schema.required?.includes(propName) || false
+        required: schema.required?.includes(propName) || false,
+        typeProperties: {
+          format: isReferenceObject(propSchema) ? undefined : propSchema.format,
```
FrankEssenberger, line 179: I wonder if this can be done in a more generic way so that we do not have to do all the properties @marikaner

packages/openapi-generator/src/file-serializer/schema.spec.ts
```diff
@@ -18,7 +18,8 @@ describe('serializeSchema for object schemas', () => {
           {
             name: 'simpleProperty',
             required: true,
-            schema: { type: 'integer' }
+            schema: { type: 'integer' },
```
FrankEssenberger, line 21: can you perhaps add a small unit test here that a schema with this properties is correctly parsed.

packages/openapi-generator/src/file-serializer/schema-file.spec.ts
```diff
@@ -139,14 +143,16 @@ describe('schemaFile', () => {
...
+              typeProperties: {}
             },
             {
               name: 'string-property-no-description',
               required: true,
               schema: {
                 type: 'string'
-              }
+              },
+              typeProperties: {}
```
FrankEssenberger, line 155: also here add a test that the generated codeblock contains the new properties like @deprecated or @example and some other docs.

packages/openapi-generator/src/file-serializer/schema.ts
```diff
@@ -110,5 +115,25 @@ function serializePropertyWithDocumentation(
...
+  if (schema.description) {
+    signature.push(schema.description);
+  }
+
+  if (schema.schema?.deprecated) {
+    signature.push('@deprecated');
+  }
+
+  if (schema.schema?.example) {
+    signature.push(`@example ${JSON.stringify(schema.schema?.example)}`);
```
marikaner, line 128: Cool idea, I would not have come up with this for values that are not JSON. But I checked, it works.

This will make it look even nicer:
```suggestion
    signature.push(`@example${unixEOL}${JSON.stringify(schema.schema?.example, null, 2)}`);
```

marikaner (changes requested):
This looks quite good already. I have many minor remarks and I think we should add this feature for parameters and persisted schemas as well.

packages/openapi-generator/src/openapi-types.ts
```diff
@@ -153,7 +153,9 @@ export type OpenApiSchema =
   | OpenApiOneOfSchema
   | OpenApiAllOfSchema
   | OpenApiAnyOfSchema
-  | OpenApiNotSchema;
+  | OpenApiNotSchema
+) &
+  OpenApiAdditionalSchemaProperty;
```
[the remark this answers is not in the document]
deekshas8: I have combined properties from interface `OpenApiAdditionalSchemaProperty `to the `OpenApiTypeSpecificProperty `(now renamed to `OpenApiSchemaProperties `). Since `deprecated` and `example` can be at both persisted schema level and object property level, and we are adding `typeProperties ` to both too, it made more sense to me to combine the interfaces into one.

packages/openapi-generator/src/file-serializer/schema.ts
deekshas8: Adding EOL formats it like 
/**
  * @example
  * 1.2
  **/
as opposed to in the same line which looks more consistent with how other comments are formatted.
/**
  * @example 1.2
  **/

Do you think it makes more sense like this or I should move it to the next line?

packages/openapi-generator/src/openapi-types.ts
```diff
@@ -297,6 +299,11 @@ export interface OpenApiObjectSchemaProperty extends OpenApiNamedSchema {
    * Denotes whether the parameter is required for the according object.
    */
   required: boolean;
+
+  /**
+   * Denotes the type specific keywords for the according object's properties.
+   */
+  typeProperties?: OpenApiTypeSpecificProperty;
```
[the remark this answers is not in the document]
deekshas8: Moved the properties to `OpenApiNamedSchema` and `OpenApiPersistedSchema`.

packages/openapi-generator/src/parser/schema.spec.ts
```diff
@@ -90,7 +90,12 @@ describe('parseSchema', () => {
     const schema: OpenAPIV3.SchemaObject = {
       required: ['simpleProperty'],
       properties: {
-        simpleProperty: { type: 'string' },
+        simpleProperty: {
```
marikaner, line 93: Please add a unit test (or more) for `parseSchemaProperties` in this file.

marikaner (changes requested):
I think it would be great to add a few more tests and improve naming in a few places. Other than that we're almost good to go.

FrankEssenberger (approved):
With the additional test, good for me.

packages/openapi-generator/src/file-serializer/schema.ts
marikaner: Both are fine 😊, ideally I would put short examples on the same line (e. g. less than 20 characters or so) and long examples below. But I also think this doesn't matter.
https://tsdoc.org/pages/tags/example/

packages/openapi-generator/src/parser/schema.spec.ts
```diff
@@ -177,6 +225,23 @@ describe('parseSchema', () => {
...
+      default: 10
+    };
+    expect(parseSchemaProperties(schemaProperties)).toEqual({
+      deprecated: true,
+      example: 100,
+      minimum: 10,
+      maximum: 1000,
+      default: 10
+    });
+  });
```
marikaner, line 243: This test only covers the happy path, but I think it would be good to see that unknown properties are not added for example.

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -174,6 +174,7 @@ export function parseParameters(
   return pathParameters.map(param => ({
     ...param,
     originalName: param.name,
-    schema: parseSchema(param.schema, refs, options)
+    schema: parseSchema(param.schema, refs, options),
+    schemaProperties: {}
```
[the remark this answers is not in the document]
marikaner: A simple test would probably be a good idea.

marikaner (changes requested):
We're almost there, but I think we are missing something in the parameters here ;)

marikaner (approved):

PR #1503, opened 2021-08-12
chore: Update license information

marikaner:
This PR adds a note to the license section in the readme.
It makes it more transparent, that we have some dependencies, that are free, but not Open Source.


jjtang1985 (approved):
👍

PR #1521, opened 2021-08-18
chore: Fix loads of typos and increase the level of the typedoc rule

marikaner:
I fixed many typos manually and increased the level of the typedoc rule.


yarn.lock
```diff
@@ -74,7 +74,8 @@
   integrity sha512-DSA7ruZrY4WI8VxuS1jWSRezFnghEoYEFrZcw9BizQRmOZiUsiHl59+qEARGPqPikwA/GPTyRCi7isuCK/oyqg==
   dependencies:
     "@babel/types" "^7.12.17"
-    jsesc "^2.5.1"
+    
+     "^2.5.1"
```
marikaner, line 78:
```suggestion
    jsesc "^2.5.1"
```

jjtang1985 (approved):
LGTM.
Thanks for taking care.
I think we should either have a pipeline check or split the PR to smaller to avoid potential overlooking.

PR #1537, opened 2021-08-23
Test errors windows

deekshas8:
Fixed unit test errors that fail when run on windows. Used path.js lib to resolve or normalize paths to handle file path delimiters. 

Closes SAP/cloud-sdk-backlog#270

### DIFF
--- a/packages/openapi-generator/src/generator.spec.ts
+++ b/packages/openapi-generator/src/generator.spec.ts
@@ -11,6 +11,7 @@ jest.mock('../../generator-common', () => {
 });
 
 const { readFile } = promises;
+const path = require('path');
 
 describe('generator', () => {
   afterAll(() => {
@@ -30,9 +31,9 @@ describe('generator', () => {
     });
 
     expect(await getInputFilePaths('/path/to/test/dir')).toEqual([
-      '/path/to/test/dir/sub-dir/sub-directory-service.txt',
-      '/path/to/test/dir/sub-dir/test-service.txt',
-      '/path/to/test/dir/test-service.txt'
+      path.resolve('/path/to/test/dir/sub-dir/sub-directory-service.txt'),
+      path.resolve('/path/to/test/dir/sub-dir/test-service.txt'),
+      path.resolve('/path/to/test/dir/test-service.txt')
     ]);
 
     mock.restore();
@@ -154,7 +155,7 @@ describe('generator', () => {
           })
         },
         existingConfig:
-          '{ "inputDir/spec.json": {"directoryName": "customName" } }'
+          JSON.stringify({ [path.normalize("inputDir/spec.json")]: {"directoryName": "customName" } })
       });
     });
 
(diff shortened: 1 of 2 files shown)


packages/openapi-generator/src/generator.spec.ts
```diff
@@ -192,16 +193,13 @@ describe('generator', () => {
...
-              "{
-                \\"inputDir/spec.json\\": {
-                  \\"packageName\\": \\"customName\\",
-                  \\"directoryName\\": \\"customName\\",
-                  \\"serviceName\\": \\"customName\\"
-                }
-              }
-              "
-            `);
+      expect(JSON.parse(await actual)).toEqual({
```
marikaner, line 196: I know we have talked about this, but I did not look at the bigger picture back then.
This test actually should remain as it was and the generator should be adjusted, so that the paths look the same for all OS.

Reason:
If this file is created when generating with certain options on Mac it should result in the same output as on Windows, so that users don't play ping pong between PRs.

This applies to all changed tests in this file.

marikaner (changes requested):
I think we should rather adjust how the generator creates options per service than the test, se reasoning in the comments.

Also, please don't forget to adhere to the conventional commit guidelines and use the imperative to describe what your PR does.

packages/openapi-generator/src/generator.spec.ts
deekshas8: I updated the `getRelativePath(absolutePath: string)` method. It now returns the relative path with posix delimiters. This path gets saved in the options-per-service json, so the output of the file will now be same on both OS. 

Therefore, I also did not change `getOriginalOptionsPerService` method (where we read existing options) assuming that the file will have `/` separated paths, since we save them that way now. 

Is this assumption correct? Or should I still add a check to make sure if we read an existing file with `\\` , it is processed correctly?

packages/openapi-generator/src/generator.ts
```diff
@@ -255,7 +264,7 @@ async function generateService(
  * @hidden
  */
 export function getRelativePath(absolutePath: string): string {
-  return relative(process.cwd(), absolutePath);
+  return relative(process.cwd(), absolutePath).split(sep).join(posix.sep);
```
marikaner, line 267: As discussed, please adjust the name and docs of this function.

marikaner (approved):
LGTM, please don't forget to update the name and docs + add to the changelog.

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@
 
 ## Compatibility Notes
 
--
+- [openapi-generator] Paths referenced in options per service configuration  file use `/` separator. Existing config files generated on Windows using `\\` in paths should be updated manually to `/`.
```
marikaner, line 17:
```suggestion
- [openapi-generator] Paths referenced in the options per service configuration now use POSIX-style separators (`/`) independent of the operating system. Existing configuration files generated on Windows contain Windows-style separators (`\\`) in the paths. Those paths will not be read correctly and should be updated manually to `/`.
```

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@
 
 ## Compatibility Notes
```
marikaner, line 15: I changed it a bit to make it even more explicit. Could you also add an entry to fixed issues?

PR #1553, opened 2021-08-30
chore: A small refactoring 

FrankEssenberger:
Clean up some code I came across. The multiple `then()` and other chained methods are not good to read.

### DIFF
--- a/packages/analytics/src/usage-analytics.spec.ts
+++ b/packages/analytics/src/usage-analytics.spec.ts
@@ -1,5 +1,6 @@
 import os from 'os';
 import nock from 'nock';
+import { expectAllMocksUsed } from '@sap-cloud-sdk/core/test/test-util';
 import { UsageAnalyticsOptions } from './analytics-types';
 import { sendAnalyticsData } from './usage-analytics';
 import { getAnalyticsData } from './analytics-data';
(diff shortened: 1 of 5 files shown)


packages/core/src/connectivity/scp-cf/jwt.ts
```diff
@@ -146,23 +146,24 @@ export async function verifyJwt(
...
+    if (options?.cacheVerificationKeys) {
+      verificationKeyCache.set(verificationKeyURL, key);
+    }
+    return verifyJwtWithKey(token, key.value);
+  } catch (error) {
+    throw new ErrorWithCause(
+      'Failed to verify JWT. Could not retrieve verification key.',
+      error
+    );
+  }
```
marikaner, line 166: This is different behavior than before. The error message does not fit the error cause.

```suggestion
    const key = await getVerificationKey(creds, verificationKeyURL).catch(error => {
      throw new ErrorWithCause(
        'Failed to verify JWT. Could not retrieve verification key.',
        error
      );
    });
    if (options?.cacheVerificationKeys) {
      verificationKeyCache.set(verificationKeyURL, key);
    }
    return verifyJwtWithKey(token, key.value);
  }
```

marikaner (changes requested):
Looks good, except the one error message that in my eyes is incorrect now. I think the behavior before made more sense.

packages/analytics/src/usage-analytics.spec.ts
```diff
@@ -1,5 +1,6 @@
 import os from 'os';
 import nock from 'nock';
+import { expectAllMocksUsed } from '@sap-cloud-sdk/core/test/test-util';
```
marikaner, line 3: Meh, do you think it makes sense to put this somewhere else outside of core, e.g. `repo/test-util`?

marikaner (approved):
LGTM

PR #1554, opened 2021-08-30
feat: Support client certificates

marikaner:
Add xssec library for client credentials grant. I simplified/generalized some test functions for easier testing with certificates.
The most relevant changes can be found in `token-accessor.ts` and `environment-accessor-types.ts` + tests.

Relates to SAP/cloud-sdk-backlog#264.

Things to consider:
- I did not find a way to decode a JWT without verifying it, when using xssec. Otherwise I would have used xssec for decoding in the `serviceToken` function. We haven't verified the JWT before either. I guess that should be ok?
- The client credentials grant cache created a cache key using clientid + clientsecret. I found this approach unexpected and I don't see a good reason to add the secret to the key. Instead, I changed it to use the clientid only, which can be used for certificate authorization too.


jjtang1985 (commented):
LGTM, I left some minor comments.
Have you tested it with e2e scenario?

marikaner:
> LGTM, I left some minor comments.
> Have you tested it with e2e scenario?

No not yet, I thought it would be easier with canary in place.

packages/core/test/test-util/environment-mocks.ts
```diff
@@ -43,7 +45,23 @@ export const mockDestinationServiceBinding: Service = {
   credentials: {
     clientid: 'destinationClient',
     clientsecret: 'destinationSecret',
-    uri: destinationServiceUri
+    uri: destinationServiceUri,
+    url: providerXsuaaUrl
+  }
+};
+
+export const mockDestinationServiceCertBinding: Service = {
```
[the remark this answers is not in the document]
marikaner: Good idea, although very long ;)

packages/core/test/test-util/environment-mocks.ts
```diff
@@ -74,10 +90,14 @@ interface MockServiceBindings {
   connectivity: Service[];
 }
 
-export function mockServiceBindings(): MockServiceBindings {
+export function mockServiceBindings(mockWithCert = false): MockServiceBindings {
```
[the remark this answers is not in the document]
marikaner: Hmm, I think this might be overkill. But I agree that this is not very intuitive. I tried to make it a bit nicer. What do you think?

jjtang1985:
> Yes, I did a manual test and it seems to work 😊

Usually, if possible, you can add some manual tests in the `auth-flow.spec.ts`.
Not sure whether it's applicable in your case?

FrankEssenberger:
> I did not find a way to decode a JWT without verifying it, when using xssec. Otherwise I would have used xssec for decoding in the serviceToken function. We haven't verified the JWT before either. I guess that should be ok?

A JWT has three parts separated by dots. Each data containing part is just base64 endcoded so decoding should be easy. Easy here for details: https://jwt.io/.

FrankEssenberger:
> > LGTM, I left some minor comments.
> > Have you tested it with e2e scenario?
> 
> Yes, I did a manual test and it seems to work 😊

I guess we have a follow up ticket to include a test in our nightly E2E tests?

packages/core/src/connectivity/scp-cf/destination/destination-accessor-proxy-configuration.spec.ts
```diff
@@ -1,6 +1,6 @@
 import nock from 'nock';
 import {
-  mockedConnectivityServiceProxyConfig,
+  connectivityProxyConfigMock,
```
FrankEssenberger, line 3: General feedback: The renaming makes it hard to find the actual core of the PR. Consider doing such refactoring in a separate PR.

marikaner:
> > I did not find a way to decode a JWT without verifying it, when using xssec. Otherwise I would have used xssec for decoding in the serviceToken function. We haven't verified the JWT before either. I guess that should be ok?
> 
> A JWT has three parts separated by dots. Each data containing part is just base64 endcoded so decoding should be easy. Easy here for details: https://jwt.io/. But I guess the thing with xssec is that you need the security context to get new tokes right? And when you get the context they verify the token.

Yes, I am aware of this and I am just using the old decoder from the `jsonwebtoken` package (with types). Still, the API from `xssec` has a very different style and theoretically they can parse the subdomain for us, which is part of the security context in `xssec`.

marikaner:
> > > LGTM, I left some minor comments.
> > > Have you tested it with e2e scenario?
> > 
> > 
> > Yes, I did a manual test and it seems to work 😊
> 
> I guess we have a follow up ticket to include a test in our nightly E2E tests?

No it is part of this ticket and the PR is already open ;) Just couldn't confirm it is working for sure yet.

FrankEssenberger:
> The client credentials grant cache created a cache key using clientid + clientsecret. I found this approach unexpected and I don't see a good reason to add the secret to the key. Instead, I changed it to use the clientid only, which can be used for certificate authorization too.

I think this is fine, but only because an update of the VCAP env variables requires a application restart which will empty the cache. Note that the clienID stays the same for a service if you have the same number of services. So you could bind undbind a service and have the same ID. However, the secret changes. But This could only become a problem if you could change this at runtime without restart.

packages/core/src/connectivity/scp-cf/client-credentials-token-cache.ts
```diff
@@ -38,9 +37,13 @@ const ClientCredentialsTokenCache = (
 
 export function getGrantTokenCacheKey(
   url: string,
-  credentials: ClientCredentials
```
FrankEssenberger, line 41: Could you add a `TODO` to remove the ClientCredentials in version 2.0 and also remove all internal usages with this API.

packages/core/src/connectivity/scp-cf/token-accessor.ts
```diff
@@ -40,40 +69,48 @@ export async function serviceToken(
...
       xsuaa.url,
-      serviceCreds
+      serviceCredentials.clientid
     );
     if (cachedToken) {
       return cachedToken.access_token;
     }
   }
+
+  // TODO: handle Payload as well
```
FrankEssenberger, line 92: what about this TODO?

packages/core/src/connectivity/scp-cf/token-accessor.ts
```diff
@@ -40,40 +69,48 @@ export async function serviceToken(
...
         xsuaa.url,
-        serviceCreds,
-        resp
+        serviceCredentials.clientid,
+        token
       );
     }
-    return resp.access_token;
-  } catch (error) {
+    // TODO: Token responses can have id_token instead of access_token (xssec does that correctly)
```
FrankEssenberger, line 108: also here?

packages/core/src/connectivity/scp-cf/client-credentials-token-cache.ts
marikaner: Done.

packages/core/test/test-util/xsuaa-service-mocks.ts
```diff
@@ -1,20 +1,46 @@
 import nock from 'nock';
-import { basicHeader } from '../../src/connectivity/scp-cf';
+import { basicHeader, ServiceCredentials } from '../../src/connectivity/scp-cf';
```
FrankEssenberger, line 2: I put this comment here, but it is about the `xsuaa-service.ts` should we deprecate the `clientCredentialsGrant` method or also replace the implementation using the xssec?

packages/core/src/connectivity/scp-cf/client-credentials-token-cache.ts
marikaner: I added a TODO, but this type is used in other places still and often part of public API. I would not deprecate it yet, because I didn't think the replacement through yet... (e. g. `jwtBearerTokenGrant`).

packages/core/test/test-util/xsuaa-service-mocks.ts
FrankEssenberger: Currently this method will fail it the credential type is `x509`

packages/core/src/connectivity/scp-cf/token-accessor.ts
marikaner: Totally missed that, fixed.

packages/core/src/connectivity/scp-cf/token-accessor.ts
```diff
@@ -40,40 +69,48 @@ export async function serviceToken(
...
+    return token.access_token;
+  } catch (err) {
     throw new ErrorWithCause(
-      `Fetching an access token for service "${resolvedService.label}" failed.`,
-      error
+      `Could not fetch client credentials token for service of type "${service.label}".`,
+      err
```
FrankEssenberger, line 113: I put the comment here but it is about the `jwtBearerToken` in this file. From the docs of xssec it is stated that also grant type `urn:ietf:params:oauth:grant-type:saml2-bearer` is supported so we could also replace the method call in there or deprecate the method and ask users to use the xssec lib. I guess the validation could be a potential issue, but for this grant I think it is fair to validate the JWT before even if it is a breaking change in behavior.

packages/core/test/test-util/xsuaa-service-mocks.ts
marikaner: Probably yes, but I would do this on a separate PR. I would like to see E2E tests work an so on and then slowly (but soon) replace the rest. I think this can still be done as part of this ticket.

packages/core/src/connectivity/scp-cf/token-accessor.ts
marikaner: Same as above. Agreed, but I would do it on a separate PR.

packages/core/src/connectivity/scp-cf/token-accessor.ts
marikaner: I am aware of this one, but this is the old behavior therefore I wouldn't touch it on this PR either. I'll add it to the ACs of the ticket to consider/discuss this.

packages/core/src/connectivity/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -9,7 +9,7 @@ import {
   subscriberUserJwt
 } from '../../../../test/test-util/mocked-access-tokens';
 import {
-  mockedConnectivityServiceProxyConfig,
+  connectivityProxyConfigMock,
```
FrankEssenberger, line 12: This is a test comment using the new webStorm PR review tool.

packages/core/src/connectivity/scp-cf/token-accessor.ts
FrankEssenberger: ok

packages/core/src/connectivity/scp-cf/client-credentials-token-cache.ts
FrankEssenberger: ok please create a follow up ticket containing the points we identified in this review which will be tackeled afterwards. Perhaps by a different colleague to spread the knowledge.

packages/core/src/connectivity/scp-cf/token-accessor.ts
FrankEssenberger: Perhaps you can explain what is the difference between the two and will there be a separate property? In the type I only see `access_token`

packages/core/src/connectivity/scp-cf/destination/destination-cache.spec.ts
marikaner: Worx ;)

FrankEssenberger (approved):
LGTM - left overs are done in separate PR

PR #1555, opened 2021-08-30
chore: Fix ReDos vulnerability

marikaner:
This most likely is not an actual problem for the SDK, but I think the new regex reads nicer anyways.


FrankEssenberger (approved):
LGTM can you perhaps also fix the two static codecheck findings.

jjtang1985 (approved):
I think it will fix the issue and also has better readability.
Hopefully, we have good test coverage.

marikaner:
> LGTM can you perhaps also fix the two static codecheck findings.

Done.

PR #1562, opened 2021-08-31
doc: verify IAS token via http requests

jjtang1985:
The doc shows how to call the token endpoint for the following grant types: 
- client credential grant
- password grant

### DIFF
--- a/knowledge-base/how-tos/0006-set-up-identity-service.md
+++ b/knowledge-base/how-tos/0006-set-up-identity-service.md
@@ -149,14 +149,14 @@ The payload of the decoded IAS (JWT) looks like below, where `user_uuid` (IdP `u
 
 #### Verify IAS token via HTTP request
 
-You can use the client id/secret for getting the IAS token via the token endpoint, after creating the service key in step 2.
-
+You can use the client id/secret for getting the IAS token via the token endpoint (`/oauth2/token`), after creating the service key in step 2.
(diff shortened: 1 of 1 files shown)


knowledge-base/how-tos/0006-set-up-identity-service.md
```diff
@@ -146,3 +146,34 @@ The payload of the decoded IAS (JWT) looks like below, where `user_uuid` (IdP `u
   "email": "xxx"
 }
 ```
+
+#### Verify IAS token via HTTP request
+
+You can use the client id/secret for getting the IAS token via the token endpoint (`/oauth2/token`), after creating the service key in step 2.
+We assume the domain of your default IdP looks like `https://xx.xx.ondemand.com/`.
```
FrankEssenberger, line 153: Is this really the IdP domain or the XSUAA? For all other services like destination you call the XSUAA.

FrankEssenberger (approved):

FrankEssenberger:
Only one minor comment/question

knowledge-base/how-tos/0006-set-up-identity-service.md
jjtang1985: This the default IdP, as so far I only used the `identity service` without any XSUAA service binding. The default IdP is bound to the `identity service` if you create an `identity service key`.

PR #1565, opened 2021-08-31
chore: Verify JWT using xssec lib

marikaner:
When running the E2E tests I realized that just getting the token using the certificate is not enough. Online JWT verification needs this too (if the xsuaa is configured as x509).

This is a proposal for a quick fix without breaking changes.


packages/core/src/connectivity/scp-cf/jwt.ts
```diff
@@ -115,38 +134,14 @@
...
   options = { ...defaultVerifyJwtOptions, ...options };
+  await verifyJwtXssec(token, options).catch(err => {
+    throw new ErrorWithCause('Could not verify JWT.', err);
+  });
 
-  const creds = getXsuaaServiceCredentials(token);
-  const verificationKeyURL = getVerificationKeyURL(token);
-  if (verificationKeyURL) {
-    checkDomainVerificationKeyURL(verificationKeyURL, creds.uaadomain);
-  }
```
FrankEssenberger, line 123: Does this method not use the json webtoken build verification. How is this different to the xssec validation?

packages/core/src/connectivity/scp-cf/token-accessor.ts
```diff
@@ -24,19 +24,20 @@ import {
...
   let subdomain: string;
   let zoneId: string;
 
   if (userJwt) {
-    const decodedJwt = decodeJwt(userJwt);
-    if (decodedJwt.iss) {
-      subdomain = parseSubdomain(decodedJwt.iss);
+    const jwtPayload =
+      typeof userJwt === 'string' ? decodeJwt(userJwt) : userJwt;
+    if (jwtPayload.iss) {
```
FrankEssenberger, line 37: Is this working for IAS token where the iss subdomain is not the one from the subaccoutn?

FrankEssenberger (commented):
I do not really get the changes. Lets dicuss in person.

packages/core/src/connectivity/scp-cf/token-accessor.ts
jjtang1985: From the IAS token scenario, the `iss` key has the value, which is the default IdP demain: `https://XX.accountsXXX.ondemand.com`

FrankEssenberger:
So after the discussion I understood the problem. When we try to get the verification key from URL it fails already because this endpoint needs also the credentials or certificate. So the lib is the save thing to use.

FrankEssenberger (approved):
LGTM - Please discuss with Junjie the iss since for IAS toknen the value is the IdP URL.

packages/core/src/connectivity/scp-cf/token-accessor.ts
```diff
@@ -24,19 +24,20 @@ import {
...
+      subdomain = parseSubdomain(jwtPayload.iss);
     }
-    zoneId = decodedJwt.zid;
+    zoneId = jwtPayload.zid;
```
jjtang1985, line 40: I have a question that is not directly related to this PR.
We can only get potential `subdomain` and `zoneId`, if `userJwt` is provided.
Otherwise, you'll pass `undefined` in line 44.
It seems these two parameters can be `null` (`subdomain`) and optional (`zoneId`) from the [doc](https://github.wdf.sap.corp/CPSecurity/node-xs2sec/blob/master/doc/TokenFlows.md#client-credentials-flow).
Do we have tests that cover the case when no JWT is provided?

jjtang1985 (approved):
Looks good, let's merge this before my xssec migration PR.

packages/core/src/connectivity/scp-cf/jwt.ts
```diff
@@ -117,52 +152,46 @@ export async function verifyJwt(
...
+        );
+        logger.warn(`Original error: ${error.message}`);
+
+        return fetchAndCacheKeyAndVerify(creds, header, token, options);
+      });
+    }
   }
 
-  return fetchAndCacheKeyAndVerify(creds, verificationKeyURL, token, options); // Verify only here
+  return fetchAndCacheKeyAndVerify(creds, header, token, options); // Verify only here
```
FrankEssenberger, line 174: Good that all these things will be done in the lib soon.

packages/core/src/connectivity/scp-cf/jwt.ts
```diff
@@ -178,30 +207,41 @@ const defaultVerifyJwtOptions: VerifyJwtOptions = {
...
-  key: TokenKey,
-  options: VerifyJwtOptions
-): TokenKey {
-  if (options.cacheVerificationKeys) {
-    verificationKeyCache.set(verificationKeyURL, key);
+function buildCacheKey(
+  jku: string | undefined,
+  kid: string | undefined
+): string {
+  if (!jku || !kid) {
```
FrankEssenberger, line 239: In the  validateJwtHeaderForVerification it is already checked if the `jku` and `kid` are there. Is this second check neede?

packages/core/src/connectivity/scp-cf/jwt.ts
```diff
@@ -315,11 +355,11 @@ export function readPropertyWithWarn(
  * @param token - Encoded JWT as a string.
  * @returns The value of the `jku` property of the JWT header.
  */
-function getVerificationKeyURL(token: string): string {
+function getVerificationKeyUrl(token: string): string {
   const decodedJwt = decodeJwtComplete(token);
-  if (!decodedJwt.header.jku) {
+  if (!decodedJwt.header.jku || !decodedJwt.header.kid) {
```
FrankEssenberger, line 360: here once more the same test

packages/core/src/connectivity/scp-cf/token-accessor.ts
marikaner: As discussed with Junjie, this will become relevant once he works on IAS.

packages/core/src/connectivity/scp-cf/jwt.ts
marikaner: Yes, because the types we pass here are optional properties. This error message will likely never be reached, but I thought just to be safe it is better to check it here.

packages/core/src/connectivity/scp-cf/xsuaa-service.ts
```diff
@@ -204,23 +204,37 @@ export function fetchVerificationKeys(
       logger.warn(
         'JKU field from the JWT not provided. Use xsuaaClient.url/token_keys as fallback. ' +
           'This will not work for subscriber accounts created after 14th of April 2020.' +
-          'Please provide the right URL given by the field JKU present in the JWT header.'
+          'Please provide the right URL given by the field `jku` in the JWT header.'
+      );
+      return executeFetchVerificationKeys(
+        `${xsuaaUriOrCredentials.url}/token_keys`,
+        xsuaaUriOrCredentials.clientid,
+        xsuaaUriOrCredentials.clientsecret
```
FrankEssenberger, line 212: I thought we discussed that the credentials are not needed to fethc the verification key.

packages/core/src/connectivity/scp-cf/jwt.ts
marikaner: I mean an alternative could be to add another type that has only jku und kid, but it seems a bit like overkill to me. But only a bit, I could also do this. What do you think?

packages/core/src/connectivity/scp-cf/jwt.ts
```diff
@@ -178,30 +207,41 @@ const defaultVerifyJwtOptions: VerifyJwtOptions = {
...
-        'No verification keys have been returned by the XSUAA service.'
+  return fetchVerificationKeys(xsuaaCredentials, header.jku).then(
+    verificationKeys => {
+      if (!verificationKeys.length) {
+        throw Error(
+          'No verification keys have been returned by the XSUAA service.'
+        );
+      }
+      const verificationKey = verificationKeys.find(
+        key => key.keyId === header.kid
```
FrankEssenberger, line 220: +1

packages/core/src/connectivity/scp-cf/xsuaa-service.ts
marikaner: Ah, this is the same URL, right? Or can it be something different?

packages/core/src/connectivity/scp-cf/jwt.ts
FrankEssenberger: yes keep it - will be removed soon when the lib supports cache invalidation.

packages/core/src/connectivity/scp-cf/xsuaa-service.ts
FrankEssenberger: I do not understand the question.

marikaner: Clarified in person, behavior should be the same for old xsuaa as well.

PR #1566, opened 2021-08-31
feat: odata code snippet generation

deekshas8:
Closes SAP/cloud-sdk-backlog#70

Updated `getApiSpecificUsage` method to generate usage code snippet for entity/function/action based on the closest match found using the levenshtein algo against a threshold.

### DIFF
--- a/packages/generator/src/sdk-metadata/code-sample-util.spec.ts
+++ b/packages/generator/src/sdk-metadata/code-sample-util.spec.ts
@@ -2,8 +2,7 @@ import {
   VdmActionImport,
   VdmEntity,
   VdmFunctionImport,
-  VdmParameter,
-  VdmServiceMetadata
+  VdmParameter
 } from '../vdm-types';
 import {
   getActionFunctionImport,
@@ -16,25 +15,25 @@ import {
 
 describe('code-sample-utils entity', () => {
   it('gets entity based on levenshtein algorithm', () => {
-     const entities = [
-       { className: 'DummyCollection' },
-       { className: 'DummyClass' }
-      ] as VdmEntity[];
-
-    expect(getODataEntity('DummyClass', entities)).toEqual(
+    const entities = [
+      { className: 'DummyCollection' },
       { className: 'DummyClass' }
-    );
+    ] as VdmEntity[];
+
+    expect(getODataEntity('DummyClass', entities)).toEqual({
+      className: 'DummyClass'
+    });
   });
 
   it('gets entity with shortest name when closest match is not found', () => {
     const entities = [
       { className: 'TestEntity' },
       { className: 'TestCollection' }
-     ] as VdmEntity[];
-     
-    expect(getODataEntity('DummyClass', entities)).toEqual(
-      { className: 'TestEntity' }
-    );
+    ] as VdmEntity[];
+
+    expect(getODataEntity('DummyClass', entities)).toEqual({
+      className: 'TestEntity'
+    });
   });
 });
 describe('code-sample-utils imports', () => {
@@ -52,7 +51,13 @@ describe('code-sample-utils imports', () => {
   const functionImportsZeroParams = [
     { name: 'serviceFuncA', parameters: [{ parameterName: 'dummyParam' }] },
     { name: 'serviceFuncB', parameters: [] },
-    { name: 'serviceFuncC', parameters: [{ parameterName: 'dummyParam' }, { parameterName: 'dummyParam' }] },
+    {
+      name: 'serviceFuncC',
+      parameters: [
+        { parameterName: 'dummyParam' },
+        { parameterName: 'dummyParam' }
+      ]
+    }
   ] as VdmFunctionImport[];
 
   it('gets undefined when func import based on levenshtein algorithm is not found', () => {
@@ -62,9 +67,10 @@ describe('code-sample-utils imports', () => {
   });
 
   it('gets func import without parameters when no closest match found', () => {
-    expect(getFunctionWithoutParameters(functionImportsZeroParams)).toEqual(
-      { name: 'serviceFuncB', parameters: [] }
-    );
+    expect(getFunctionWithoutParameters(functionImportsZeroParams)).toEqual({
+      name: 'serviceFuncB',
+      parameters: []
+    });
   });
 
   const functionImportsMinParams = [
@@ -81,12 +87,10 @@ describe('code-sample-utils imports', () => {
     }
   ] as VdmFunctionImport[];
   it('gets func import with min parameters when methods without params not found ', () => {
-    expect(getFunctionWithMinParameters(functionImportsMinParams)).toEqual(
-      {
-        name: 'serviceFunctionA',
-        parameters: [{ parameterName: 'dummyParam1' }]
-      }
-    );
+    expect(getFunctionWithMinParameters(functionImportsMinParams)).toEqual({
+      name: 'serviceFunctionA',
+      parameters: [{ parameterName: 'dummyParam1' }]
+    });
   });
 
   const actionsImports = [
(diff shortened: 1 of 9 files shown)


packages/generator/src/sdk-metadata/code-samples.ts
```diff
@@ -4,6 +4,7 @@ import {
   InstructionWithTextAndHeader,
   usageHeaderText
 } from '@sap-cloud-sdk/generator-common';
+import { VdmActionImport, VdmFunctionImport, VdmParameter } from '../vdm-types';
 
 export function codeSamples(
```
FrankEssenberger, line 9: Perhaps name this entityCodeSample

packages/generator/src/sdk-metadata/code-samples.ts
```diff
@@ -29,3 +30,47 @@ export function genericCodeSample(): InstructionWithTextAndHeader {
     header: usageHeaderText
   };
 }
+
+export function importsCodeSample(
```
FrankEssenberger, line 34: Could you export two methods functionImportCodeSample and actionImportCodeSample which delegate to this private method.

packages/generator/src/sdk-metadata/code-samples.ts
```diff
@@ -29,3 +30,47 @@ export function genericCodeSample(): InstructionWithTextAndHeader {
...
+  actionFunctionImport: VdmFunctionImport | VdmActionImport,
+  packageName: string
+): InstructionWithText {
+  return {
+    text: 'To consume the service via the pregenerated typed client library run the code snippet below. For more details about OData client libraries chose "OData Consumption Manual" from the "Helpful Links" menu.',
```
FrankEssenberger, line 39: make a constant for this text since it is used in multple places.

packages/generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -76,3 +105,106 @@ export function getODataLinks(): Links {
...
+    }
+  });
+
+  return closestFunc;
+}
+
+function getFunctionWithoutParameters(
+  actionFunctionImports: VdmFunctionImport[] | VdmActionImport[]
+): VdmFunctionImport | VdmActionImport | undefined {
+  const functionsWithoutParams = actionFunctionImports.filter(
```
FrankEssenberger, line 174: You can use `find` here I guess.

packages/generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -76,3 +105,106 @@ export function getODataLinks(): Links {
...
+  if (functionsWithoutParams) {
+    return functionsWithoutParams[0];
+  }
+}
+
+// Sorts and gets a function import having minimum input parameters.
+function getFunctionWithMinParameters(
+  actionFunctionImports: VdmFunctionImport[] | VdmActionImport[]
+): VdmFunctionImport | VdmActionImport {
+  const getFunctions = actionFunctionImports.filter(
```
FrankEssenberger, line 186: The method takes minimal parameter but onlu on the getFunctions. Either reflect this in the name or remove it if not needed.

packages/generator/src/sdk-metadata/generation-and-usage.spec.ts
```diff
@@ -19,6 +70,18 @@ describe('generation-and-usage', () => {
     expect(getApiSpecificUsage(service)).toMatchSnapshot();
   });
 
+  it('create api specific usage for function import', () => {
```
FrankEssenberger, line 73: I do not see the checked in snapshots.

packages/generator/src/sdk-metadata/code-samples.ts
```diff
@@ -29,3 +30,47 @@ export function genericCodeSample(): InstructionWithTextAndHeader {
...
+  if (functionImport.parameters) {
+    return `const parameter: ${
+      functionImport.parametersTypeName
+    } = ${getActionFunctionParams(functionImport.parameters)};
+    `;
+  }
+  return '';
+}
+
+function getActionFunctionParams(parameters: VdmParameter[]): string {
```
FrankEssenberger, line 65: For this method a dedicated unit test would make sense to see the cases: below 2 above 2 parameters and that the string is correct.

FrankEssenberger (changes requested):
Add more tests for the subparts of the generation. Besides that the allgorithm looks good.

FrankEssenberger:
One more thing which came to my mind. I think it would make sense to put all the `find right entity allgorithm` things in a separate file in the `sdk-metadata`. This makes testing also easier. You can than do the same later for the OpenApi case and perhaps there is even some reuse potential.

packages/generator/src/sdk-metadata/code-sample-util.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+} from './code-sample-util';
+
+describe('code-sample-utils entity', () => {
+  it('gets entity based on levenshtein algorithm', () => {
+    const service = {
+      originalFileName: 'DummyClass',
+      entities: [{ className: 'DummyCollection' }, { className: 'DummyClass' }]
+    } as VdmServiceMetadata;
+    expect(getODataEntity(service.originalFileName, service.entities)).toEqual(
+      service.entities[1]
```
FrankEssenberger, line 23: Here I would prefer to see the actual object because it is faster to grep that index 1-> which is in the array.

packages/generator/src/sdk-metadata/code-sample-util.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    );
+  });
+
+  it('gets entity with shortest name when closest match is not found', () => {
+    const service = {
+      originalFileName: 'DummyClass',
+      entities: [{ className: 'TestEntity' }, { className: 'TestCollection' }]
+    } as VdmServiceMetadata;
+    expect(getODataEntity(service.originalFileName, service.entities)).toEqual(
```
FrankEssenberger, line 32: same here

packages/generator/src/sdk-metadata/code-sample-util.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+  });
+
+  it('gets undefined when func import based on levenshtein algorithm is not found', () => {
+    expect(
+      getLevensteinClosestFunction('DummyClass', functionImportsZeroParams)
+    ).toBeUndefined();
+  });
+
+  it('gets func import without parameters when no closest match found', () => {
+    expect(getFunctionWithoutParameters(functionImportsZeroParams)).toEqual(
```
FrankEssenberger, line 74: For both test this and the one below the "right" value is the first one. So you could swap the elements in the `functionImportsZeroParams` and assert that than the second element is returned as a second check here. Perhaps you implemente just give me the first :-).

packages/generator/src/sdk-metadata/code-sample-util.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+      functionImportsZeroParams[0]
+    );
+  });
+
+  it('gets func import with min parameters when methods without params not found ', () => {
+    expect(getFunctionWithMinParameters(functionImportsMinParams)).toEqual(
+      functionImportsMinParams[0]
```
FrankEssenberger, line 81: same here as above.

packages/generator/src/sdk-metadata/code-sample-util.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    );
+  });
+
+  it('gets action import', () => {
```
FrankEssenberger, line 85: perhaps tell here a bit more, like get base on closest name. Should there be a second test like if not name is close take least parameters?

FrankEssenberger (approved):
WIth the additional test it is already good to go. Please consider my last remarks in making the tests even a bit better and go ahead and merge.

PR #1567, opened 2021-09-02
chore: Use xssec lib for XSUAA requests

jjtang1985:
This PR migrates to `@sap/xssec` lib for xsuaa service call.

- replace `jwtBearerTokenGrant` with `xssec.requests.requestUserToken`
- create a wrapper function for the circuit breaker feature
- deprecate unused functions

> scopes: object if you need to set scopes (or null)

- send `null` instead of `undefined`, as the doc above mentions `null` and the source code has null check like `if (scopes !== null)` everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ## Compatibility Notes
 
--
+- [core] Do not use client credentials to fetch public keys for token verification.
 
 ## New Functionality
 
@@ -26,7 +26,8 @@
 
 ## Fixed Issues
 
--
+- [core] Fix verification key retrieval to match the given key id (`kid`).
+- [core] Fix the missing export for files under `http-agent` directory.
 
 
 # 1.48.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@
 
 ## Improvements
 
--
+- [core] Uses `@sap/xssec` for requesting `user token` and `client credentials token` from the XSUAA service instead of our own implementations.
 
 ## Fixed Issues
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@
 
 ## Compatibility Notes
 
-- [core] Do not use client credentials to fetch public keys for token verification.
+- [core] Stop using client credentials to fetch public keys for token verification.
+- [core] Replace own implementation for requesting user tokens and client credential tokens from the XSUAA service with the `@sap/xssec` library.
 
 ## New Functionality
 
@@ -22,13 +23,12 @@
 
 ## Improvements
 
-- [core] Uses `@sap/xssec` for requesting `user token` and `client credentials token` from the XSUAA service instead of our own implementations.
+-
 
 ## Fixed Issues
 
 - [core] Fix verification key retrieval to match the given key id (`kid`).
 - [core] Fix the missing export for files under `http-agent` directory.
-
 
 # 1.48.1
 
--- a/packages/core/src/connectivity/scp-cf/index.ts
+++ b/packages/core/src/connectivity/scp-cf/index.ts
@@ -1,6 +1,7 @@
 export * from './jwt';
 export * from './cache';
 export * from './legacy/client-credentials-token';
+export * from './legacy/xsuaa-service';
 export * from './client-credentials-token-cache';
 export * from './connectivity-service';
 export * from './connectivity-service-types';
@@ -9,7 +10,7 @@ export * from './environment-accessor';
 export * from './environment-accessor-types';
 export * from './resilience-options';
 export * from './token-accessor';
-export * from './legacy/xsuaa-service';
+export * from './verification-keys';
 export * from './xsuaa-service-types';
 export * from './protocol';
 export * from './authorization-header';
--- a/packages/core/src/connectivity/scp-cf/index.ts
+++ b/packages/core/src/connectivity/scp-cf/index.ts
@@ -1,6 +1,6 @@
 export * from './jwt';
 export * from './cache';
-export * from './client-credentials-token';
+export * from './legacy/client-credentials-token';
 export * from './client-credentials-token-cache';
 export * from './connectivity-service';
 export * from './connectivity-service-types';
@@ -9,7 +9,7 @@ export * from './environment-accessor';
 export * from './environment-accessor-types';
 export * from './resilience-options';
 export * from './token-accessor';
-export * from './xsuaa-service';
+export * from './legacy/xsuaa-service';
 export * from './xsuaa-service-types';
 export * from './protocol';
 export * from './authorization-header';
--- a/packages/core/src/connectivity/scp-cf/token-accessor.ts
+++ b/packages/core/src/connectivity/scp-cf/token-accessor.ts
@@ -11,40 +11,27 @@ import {
 } from './environment-accessor';
 import { Service } from './environment-accessor-types';
 import { ResilienceOptions } from './resilience-options';
-import { parseSubdomain, replaceSubdomain } from './subdomain-replacer';
-import {
-  jwtBearerTokenGrant,
-  refreshTokenGrant,
-  userTokenGrant
-} from './xsuaa-service';
+import { replaceSubdomain } from './subdomain-replacer';
+import { refreshTokenGrant, userTokenGrant } from './xsuaa-service';
 import {
   ClientCredentialsResponse,
   UserTokenResponse
 } from './xsuaa-service-types';
+import { getSubdomainAndZoneId, requestUserToken } from './xsuaa-service-xssec';
 
 async function getClientCredentialsToken(
   service: string | Service,
   userJwt?: string
 ): Promise<ClientCredentialsResponse> {
   const serviceCredentials = resolveService(service).credentials;
-
-  let subdomain: string;
-  let zoneId: string;
-
-  if (userJwt) {
-    const decodedJwt = decodeJwt(userJwt);
-    if (decodedJwt.iss) {
-      subdomain = parseSubdomain(decodedJwt.iss);
-    }
-    zoneId = decodedJwt.zid;
-  }
+  const subdomainAndZoneId = getSubdomainAndZoneId(userJwt);
 
   return new Promise((resolve, reject) => {
     xssec.requests.requestClientCredentialsToken(
-      subdomain,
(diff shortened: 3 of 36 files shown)


CHANGELOG.md
```diff
@@ -22,7 +22,7 @@
 
 ## Improvements
 
--
+- [core] Uses `@sap/xssec` for requesting `user token` and `client credentials token` from the XSUAA service instead of our own implementations.
```
marikaner, line 25: For me this is an implementation detail. Does this affect the user in any way? (Like - is the behavior any different maybe?) Otherwise, I think it shouldn't be in the changelog. What do you think?

packages/core/src/connectivity/scp-cf/xsuaa-service.ts
```diff
@@ -135,6 +136,7 @@ export async function refreshTokenGrant(
 }
 
 /**
+ * @deprecated Since v1.49.0 Use `@sap/xssec` lib instead.
```
marikaner, line 139: Instead of deprecating the function right away, I would consider to use the xssec implementation to replace our implementation and see that/whether our tests are passing with this. That way you still have the option to roll back if there are problems down the line.
Also you can identify differences in behavior, because of the assumptions in the tests.

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
```diff
@@ -0,0 +1,85 @@
+import * as xssec from '@sap/xssec';
```
marikaner, line 1: Why is this file named `...-xssec` - seems not very speaking to me. I would expect this to be in the `xsuaa-service.ts`.

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
```diff
@@ -0,0 +1,85 @@
...
+    return { subdomain, zoneId };
+  }
+
+  const decodedJwt = decodeJwt(jwt);
+  if (decodedJwt.iss) {
+    subdomain = parseSubdomain(decodedJwt.iss);
+  }
+  zoneId = decodedJwt.zid || zoneId;
+  return { subdomain, zoneId };
+}
```
marikaner, line 51: I checked the xssec library and it does not seem to be correct that it can only handle null. Let's discuss this, maybe I missed something.
Not using null could make this code a little more concise.

Also: we could consider changing `parseSubdomain` to allow and return `undefined` (or `null`) values to avoid the `if`s.

```suggestion
export function getSubdomainAndZoneId(jwt?: string): SubdomainAndZoneId {
  let subdomain: string | null = null;
  let zoneId: string | null = null;

  if (jwt) {
    const decodedJwt = decodeJwt(jwt);
    if (decodedJwt.iss) {
      subdomain = parseSubdomain(decodedJwt.iss);
    }
    if (decodedJwt.zid) {
      zoneId = decodedJwt.zid;
     }
  }
  return { subdomain, zoneId };
}
```

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (options.enableCircuitBreaker) {
+    return getCircuitBreaker().fire(param);
+  }
+
+  return new Promise((resolve: (token: string) => void, reject) => {
+    xssec.requests.requestUserToken(
+      param.userJwt,
+      param.serviceCredentials,
+      null,
+      null,
```
marikaner, line 79: Are scopes not relevant for us? Did we not consider them before either?

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
```diff
@@ -0,0 +1,85 @@
...
+      param.subdomainAndZoneId.subdomain,
+      param.subdomainAndZoneId.zoneId,
+      (err: Error, token: string) => (err ? reject(err) : resolve(token))
```
marikaner, line 82: To align with the old API, there is a third parameter that gives you the response.

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
```diff
@@ -0,0 +1,85 @@
...
+ * Make a user token request against the XSUAA service
+ * @param param - The parameters for make the request.
+ * @param options - Options to influence resilience behavior (see [[ResilienceOptions]]). By default, usage of a circuit breaker is enabled.
+ * @returns User token.
+ * @hidden
+ */
+export function requestUserToken(
+  param: RequestUserTokenParam,
+  options: ResilienceOptions = { enableCircuitBreaker: true }
+): Promise<string> {
```
marikaner, line 69: I think the old API was not optimal, but having to pass subdomain and zoneId + JWT seems inconvenient in my opinion.

marikaner (changes requested):
I am a little surprized that we didn't replace the old functionality, but just put this next to it. It seems to me that it would be easy to keep the old API. Please find more details in the comments.

CHANGELOG.md
jjtang1985: I mentioned this just because this would be a note for both users and us about the potential unexpected behaviour after migrating to a new lib. I feel the migration might be risky even we have e2e tests.

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
jjtang1985: Discussed, we can rename this as `xsuaa-service.ts` and rename the old `xsuaa-service.ts` to `...-legacy.ts`

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
jjtang1985: Double checked. For the time being, 
- the `subdomain` works fine with both `null` and `undefined`.
- the `zoneId` works fine with `null` and `undefined`, but it will be set to `null` (`zoneId = null`) under some conditions, which might be risky in the future.
- the `scopes` works with `null`, because the source code checks `if(scopes !== null)`.
- some other parameters have `null` check like `if(abc !== null)`.

Therefore, I still think this brings some values.

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
jjtang1985: No, `scopes` has never been considered.

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
jjtang1985: The third parameter is not used, but we can add it to the signature to make it consistent with the client credential token.

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
jjtang1985: This is not a public api, and is only used internally.

packages/core/src/connectivity/scp-cf/xsuaa-service.ts
jjtang1985: Discussed, this function was exported, but should not be used.

CHANGELOG.md
marikaner: But then maybe we mention it in the compatibility notes?

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
marikaner: Agreed.

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
marikaner: True, still I aligned the two token request functions.

packages/core/src/connectivity/scp-cf/xsuaa-service-xssec.ts
marikaner: As discussed, this might be overkill and as is it is not public, we can change it anytime.

packages/core/src/connectivity/scp-cf/index.ts
```diff
@@ -9,7 +9,7 @@ export * from './environment-accessor';
 export * from './environment-accessor-types';
 export * from './resilience-options';
 export * from './token-accessor';
-export * from './xsuaa-service';
+export * from './legacy/xsuaa-service';
```
jjtang1985, line 12: I guess we miss the new `xsuaa-service`.

packages/core/src/connectivity/scp-cf/token-accessor.ts
```diff
@@ -11,46 +10,10 @@ import {
...
-  jwtBearerTokenGrant,
-  refreshTokenGrant,
-  userTokenGrant
-} from './xsuaa-service';
-import {
-  ClientCredentialsResponse,
-  UserTokenResponse
-} from './xsuaa-service-types';
-
-async function getClientCredentialsToken(
```
jjtang1985, line 25: Moving to xsuaa-service makes lots of sense.

packages/core/src/connectivity/scp-cf/xsuaa-service.ts
```diff
@@ -1,172 +1,155 @@
...
-  clientId: string,
-  options?: ResilienceOptions
-): Promise<UserTokenResponse> {
-  const authHeader = 'Bearer ' + userJwt;
-  const body = objectToXWwwUrlEncodedBodyString({
-    client_id: clientId,
-    grant_type: GrantType.USER_TOKEN,
-    response_type: 'token'
-  });
+function wrapInCircuitBreaker<T extends (...args: any[]) => any>(
```
jjtang1985, line 52: Nice to have this wrapper function.

jjtang1985 (commented):
LGTM.
The circular dependency has to be fixed.

packages/core/src/connectivity/scp-cf/index.ts
marikaner: I thought it was internal on purpose. However, I moved the `fetchVerificationkeys` to a separate file and exportet it to avoid breaking changes.

jjtang1985:
I approved via comment and please merge once the circular dependency is fixed.

marikaner (approved):

PR #1574, opened 2021-09-06
feat: open-api code snippet changes

deekshas8:

Updated `getApiSpecificUsage `method to generate open-api usage code snippet for main API and its operationId based on the closest match found using the levenshtein algo measured against a threshold.

Closes SAP/cloud-sdk-backlog#245

### DIFF
--- a/packages/generator-common/src/sdk-metadata/sdk-metadata.ts
+++ b/packages/generator-common/src/sdk-metadata/sdk-metadata.ts
@@ -1,4 +1,5 @@
 import { removeFileExtension } from '@sap-cloud-sdk/util';
+import levenstein from 'fast-levenshtein';
 import { getSdkVersion } from './util';
 import {
   Client,
@@ -8,6 +9,7 @@ import {
   ServiceStatus
 } from './sdk-metadata-types';
 
+const distanceThreshold = 5;
 export function getSdkMetadataFileNames(originalFileName: string): {
   clientFileName: string;
   headerFileName: string;
@@ -51,6 +53,30 @@ export function getSdkMetadataClient(
     pregeneratedLibrary,
     generationAndUsage
   };
+}
+
+export function getLevenshteinClosest<T>(name: string, objectsToCheck: T[], extractorFn: (x: T) => string): T|undefined {
+  const distBelowThreshold = objectsToCheck.reduce((prev, obj) => {
+    const levenshteinDist = getLevenshteinDistance(name, extractorFn(obj));
+    if(levenshteinDist < distanceThreshold) {
+     return [ ...prev, { dist: levenshteinDist, obj } ];
+    }
+    return prev;
+  },[]);
+  if(distBelowThreshold.length > 0) {
+    return distBelowThreshold.sort((a,b) => a.dist < b.dist ? -1 : 1)[0].obj;
+  }
+}
+
+ function getLevenshteinDistance(stringA: string, stringB: string): number {
+  return levenstein.get(
+    getSanitizedString(stringA),
+    getSanitizedString(stringB)
+  );
+}
+
+function getSanitizedString(text: string): string {
+  return text.replace(/[^A-Za-z]/g, '').toLowerCase(); // new RegExp('[^A-Za-z]/g')
 }
 
 export const sdkMetadataHeaderIntroText =
--- a/packages/generator-common/src/sdk-metadata/sdk-metadata.ts
+++ b/packages/generator-common/src/sdk-metadata/sdk-metadata.ts
@@ -54,7 +54,13 @@ export function getSdkMetadataClient(
     generationAndUsage
   };
 }
-
+/**
+ * Gets the closest matching object using Levenshtein distance algorithm.
+ * @param name - Name of the service or api class.
+ * @param objectsToCheck - List of objects, e.g. VdmEntity, FunctionImports, etc.
+ * @param extractorFn - Function to get the object's property to match against name.
+ * @returns - closest matched object or undefined if not found.
+ */
 export function getLevenshteinClosest<T>(
   name: string,
   objectsToCheck: T[],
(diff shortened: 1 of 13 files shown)


packages/openapi-generator/src/sdk-metadata/code-sample-util.ts
```diff
@@ -0,0 +1,167 @@
...
+
+const distanceThreshold = 5;
+export function getMainApi(
+  serviceName: string,
+  apis: OpenApiApi[]
+): OpenApiApi {
+  let closestApi: OpenApiApi | undefined;
+  let minDistance = distanceThreshold;
+
+  apis.forEach(api => {
```
FrankEssenberger, line 13: This first part is a bit ugly and also does two things, finding the closest which is also below the threshold. Perhaps you can use reduce to buld a mapping object {distance->api} which contains only elements below the threshold. On this object you can use then Math.min on the keys to get the smallest. This does the same but uses methods which describe the flow better.

Or you wrap these lines in a helper method: findApiBelewLevensteinThreshold(apis,threshold)

packages/openapi-generator/src/sdk-metadata/code-sample-util.ts
```diff
@@ -0,0 +1,167 @@
...
+  return (
+    getLevensteinClosestOperation(api.name, api.operations) ||
+    getGetAllOperation(api.operations) ||
+    getGetOperation(api.operations) ||
+    getAnyOperationWithoutParams(api.operations) ||
+    api.operations[0]
+  );
+}
+
+export function getLevensteinClosestOperation(
```
FrankEssenberger, line 52: Perhaps you can make a reuse method so that you can use this in the case above. Potentential signaure:
getLevensteinClosestOperation<T>(name:string,objectToCheck:T,extractorFn:(T)=>string,options?:{threshold:number}):T[]|undefined

packages/openapi-generator/src/sdk-metadata/code-sample-util.ts
```diff
@@ -0,0 +1,167 @@
...
+    const distance = getLevensteinDistance(apiName, operation.operationId);
+    if (distance <= minDistance) {
+      minDistance = distance;
+      closestOperation = operation;
+    }
+  });
+  return closestOperation;
+}
+
+export function getGetAllOperation(
```
FrankEssenberger, line 69: Also this method could be reduce in the two methods. Also this check on no parameters. Perhaps split it?

packages/openapi-generator/src/sdk-metadata/code-sample-util.ts
```diff
@@ -0,0 +1,167 @@
...
+  if (
+    operation.pathParameters?.length === 0 &&
+    operation.queryParameters?.filter(param => param.required).length === 0 &&
+    operation.requestBody === undefined
+  ) {
+    return '';
+  }
+
+  const paramSignature: string[] = [];
+  if (operation.pathParameters?.length > 0) {
```
FrankEssenberger, line 122: is the check necessary? If the array is empty it will not do something or do you check for undefined? if so just do if(operation.pathParameters)

packages/openapi-generator/src/sdk-metadata/code-sample-util.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  if (operation.requestBody) {
+    let reqBody = 'undefined';
+    if (operation.requestBody.required) {
+      reqBody = 'myRequestBody';
+    }
+    paramSignature.push(reqBody);
+  }
+
+  if (operation.queryParameters?.length > 0) {
```
FrankEssenberger, line 136: same as above

packages/openapi-generator/src/sdk-metadata/code-sample-util.ts
```diff
@@ -0,0 +1,167 @@
...
+ * @hidden
+ */
+function getLevensteinDistance(stringA: string, stringB: string): number {
+  return levenstein.get(
+    getWordWithoutSpecialChars(stringA).toLowerCase(),
+    getWordWithoutSpecialChars(stringB).toLowerCase()
+  );
+}
+
+function getWordWithoutSpecialChars(text: string): string {
```
FrankEssenberger, line 165: The name does not match 100%, either use a regex with replacing all non word,number characters or call it removeUnderscore.

packages/openapi-generator/src/sdk-metadata/code-sample-util.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/**
+ * Calculate levenshtein distance of the two strings.
+ * @param stringA - The first string.
+ * @param stringB - The second string.
+ * @returns The levenshtein distance (0 and above).
+ * @hidden
+ */
+function getLevensteinDistance(stringA: string, stringB: string): number {
```
FrankEssenberger, line 158: There is a generator-common package which also contains some metadata parts. You could put this method in there. Potentially also a few of the more general method from above after you changed the signature.

FrankEssenberger (changes requested):
Looks good in general. Only a few comments to reduce code duplication on the code-sample-util.ts within this PR but also to reuse some parts for the OData case.

packages/openapi-generator/src/sdk-metadata/code-sample-util.ts
deekshas8: Moved the generic method to calculate and find the closest api/operation/entity/imports to generator-common.

packages/generator-common/src/sdk-metadata/sdk-metadata.ts
```diff
@@ -53,6 +55,34 @@ export function getSdkMetadataClient(
   };
 }
 
+export function getLevenshteinClosest<T>(
```
FrankEssenberger, line 58: Please add a JS doc what this method does.

packages/generator-common/src/sdk-metadata/sdk-metadata.ts
```diff
@@ -53,6 +55,34 @@ export function getSdkMetadataClient(
...
+
+function getLevenshteinDistance(stringA: string, stringB: string): number {
+  return levenstein.get(
+    getSanitizedString(stringA),
+    getSanitizedString(stringB)
+  );
+}
+
+function getSanitizedString(text: string): string {
+  return text.replace(/[^A-Za-z]/g, '').toLowerCase(); // new RegExp('[^A-Za-z]/g')
```
FrankEssenberger, line 83: comment can go away?

FrankEssenberger (approved):
LGTM - consider my two minor comments and go ahead and merge it.

Issue #1564, opened 2021-08-31
Unable to get local issuer certificate when fetching CSRF token


JulianGlattki:
Hello, 
I am trying to use the SAP Cloud SDK to create Purchase Requisitions. 
I am using the following code: 

```javascript
let builder = purchaseReqService
            .PurchaseRequisitionHeader
            .requestBuilder()
            .create(payload)
            .addCustomQueryParameters({'sap-client': 123})
            return await builder
            .execute({url: baseUrl});
```
When executing this code the request will fail when fetching the CSRF token: 

> Error: unable to get local issuer certificate

I can workaround this problem by using isTrustingAllCertificates. This is not what I want however. 
I had a similar issue in the past and was able to fix it by overwriting the CA certificates used by the http agent. 

I saw that you are using axios as an underlying library and allow passing of custom request configurations to it. So I tried overwriting the httpsAgent.

```javascript
let builder = purchaseReqService
            .PurchaseRequisitionHeader
            .requestBuilder()
            .create(payload)
            .addCustomQueryParameters({'sap-client': 123})
            .addCustomRequestConfiguration({httpsAgent: new https.Agent({
                ca: rootCAs,
              })})
            return await builder
            .execute({url: baseUrl});
```

This did not help however and I still receive the error above. I cannot figure out how to proceed. Do you any idea?

jjtang1985:
Hi @JulianGlattki ,

thanks for approaching us.

I see you tried to use `addCustomRequestConfiguration()` for setting a custom `httpsAgent`, which should work, as `httpsAgent` is one of the valid options of `axios`.

In order to understand better, I would like to understand how did you fix the CA certificates issue before as you mentioned?
It would be great, if you can share some code snippets.

Thank you.
BR,
Junjie

JulianGlattki:
Hi Junjie, 
thank you for the answer. 

In another part of the code we use [got](https://www.npmjs.com/package/got). 
It allows you to set this option (see on the npm page linked above): 
![image](https://user-images.githubusercontent.com/48208850/131510490-235ddee9-b412-43ac-9178-1343614a0c08.png)

Of course, it's not the same library but it seems to tackle the same problem (and fix the same error message). 

For the code in the initial post I used [this](https://stackoverflow.com/a/53585725) answer on SO.

While debugging to me it looked like the option provided in addCustomRequestConfiguration is not taken into account when fetching the CSRF token.

jjtang1985:
[...]

PR #1580, opened 2021-09-07
fix: consider http(s)Agent key when sending CSRF token requests

jjtang1985:
Related to: https://github.com/SAP/cloud-sdk-js/issues/1564

When sending CSRF token request, the following keys of the requestConfig are now considered like the real request:
- proxy
- httpAgent
- httpsAgent

### DIFF
--- a/packages/core/src/http-client/http-client.ts
+++ b/packages/core/src/http-client/http-client.ts
@@ -266,7 +266,10 @@ async function getCsrfHeaders(
     : buildCsrfHeaders(destination, {
         params: request.params,
         headers: request.headers,
-        url: request.url
+        url: request.url,
+        proxy: request.proxy,
+        httpAgent: request.httpAgent,
+        httpsAgent: request.httpsAgent
       });
 }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/http-client/http-client.ts
```diff
@@ -266,7 +266,10 @@ async function getCsrfHeaders(
     : buildCsrfHeaders(destination, {
         params: request.params,
         headers: request.headers,
-        url: request.url
+        url: request.url,
```
FrankEssenberger, line 269: why we do not pass the full request Config along? Are the parts which we do not want to use for the csrfHeader fetch?

FrankEssenberger (approved):
Only one minor question

packages/core/src/http-client/http-client.ts
jjtang1985: Here are some reasons:
- We built request with partial request config before. Using all configs might introduce issues. 
- Some keys should be replaced anyways like:
    - `method` should be `head`
    - `data` should be empty
    - `responseType` should not be some special value like `stream`/`arrayBuffer`.
    - `maxContentLength` should not be related to the `head` request (CSRF token request).

Issue #1575, opened 2021-09-06
oData v2 Client orderBy Escaping Issue


schw4be:
Hi there,

a simple oData v2 Call with orderby runs in the following error:

`TypeError [ERR_UNESCAPED_CHARACTERS]: Request path contains unescaped characters
    at new ClientRequest (_http_client.js:151:13)
    at Object.request (https.js:313:10)`

This is the sample code:

`        return await MyEntity.requestBuilder().getAll().filter(MyEntity.Field.equals(someValue)
        ).orderBy(desc(MyEntity.ID)).top(1).execute({
            url: api_url,
            username: api_user,
            password: api_pwd,
        });
    }`

The build URI looks like this:
`odata/v2/MyEntity/MyEntity?$format=json&$filter=(Field%20eq%20'VALUE')&$orderby=Id desc&$top=1
`
SDK Version used: 1.48.1

I think the unescaped "**id desc**" causes the issue.

jjtang1985:
Hi @schw4be ,

thanks for reaching out.

I can reproduce the issue and we will work on it soon.

Thank you.

jjtang1985:
Hi @schw4be ,

I fixed the encoding bug.
Please try `canary` version and let me know whether it solves your issue.

Thank you.

schw4be:
Hey @jjtang1985,
looks good! Thank you very much for the quick fix :)
Greetings

PR #1582, opened 2021-09-08
Fix: missing encoding of `orderby`

jjtang1985:
Related to: https://github.com/SAP/cloud-sdk-js/issues/1575

The space of the `orderby` query parameter, for example `$orderby=Id desc`, should be encoded to `$orderby=Id%20desc`.


FrankEssenberger (approved):
LGTM - Hopefully we have found all places of missing encoding. I also wonder why not all backend systems react the same. But it is the way it is.

PR #1585, opened 2021-09-09
chore: Lock version of xssec to 3.2.3 until we find a workaround with…

marikaner:
… browsers

As we currently cannot find a solution with react + webpack + sdk, it is a risk to use the latest version of xssec. In order to fulfill the requirements for mTLS we should finally make a release. Therefore, this is a proposal to keep the older version of xssec for the time being.

Relates to SAP/cloud-sdk-backlog#264.


jjtang1985 (approved):
LGTM.
Nice timing for introducing xssec with 3.2.3 and the next patch version 3.2.4 broke frontend.

PR #1593, opened 2021-09-13
fix: Reduce number of warnings on XSUAA selection

FrankEssenberger:

Closes SAP/cloud-sdk-backlog#314.

We called the `getXsuaaServiceCredentials()` multiple times in the destination-from-service.ts. This alone would be only inefficient. However, since in one case we made a call without a JWT there was no way to avoid the warning: `Could not match XSUAA instance`. This PR reduces the number of calls and also reduces the number of logs or the level. 

Note that the current XSUAA instance is mainly used to the get URL for caching. Once the xssec supports cache options this can all go away.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 
 ## Improvements
 
--
+- [xsuaa] Reduce the number of warn message when selecting a XSUAA instance.
 
 ## Fixed Issues
 
--- a/packages/core/src/connectivity/scp-cf/destination/destination-accessor-loading-precendence.spec.ts
+++ b/packages/core/src/connectivity/scp-cf/destination/destination-accessor-loading-precendence.spec.ts
@@ -76,7 +76,7 @@ describe('destination loading precedence', () => {
         { cacheVerificationKeys: false }
       )
     ).rejects.toThrowErrorMatchingInlineSnapshot(
-      'No binding to an XSUAA service instance found. Please make sure to bind an instance of the XSUAA service to your application.'
+      '"No binding to an XSUAA service instance found. Please make sure to bind an instance of the XSUAA service to your application."'
     );
   });
 });
--- a/packages/core/src/connectivity/scp-cf/destination/destination-accessor-loading-precendence.spec.ts
+++ b/packages/core/src/connectivity/scp-cf/destination/destination-accessor-loading-precendence.spec.ts
@@ -75,6 +75,8 @@ describe('destination loading precedence', () => {
         { destinationName: 'non-existent' },
         { cacheVerificationKeys: false }
       )
-    ).rejects.toThrowError('No binding to an XSUAA service instance found. Please make sure to bind an instance of the XSUAA service to your application.')
+    ).rejects.toThrowError(
+      'No binding to an XSUAA service instance found. Please make sure to bind an instance of the XSUAA service to your application.'
+    );
   });
 });
--- a/packages/core/src/connectivity/scp-cf/destination/destination-accessor-loading-precendence.spec.ts
+++ b/packages/core/src/connectivity/scp-cf/destination/destination-accessor-loading-precendence.spec.ts
@@ -75,8 +75,6 @@ describe('destination loading precedence', () => {
         { destinationName: 'non-existent' },
         { cacheVerificationKeys: false }
       )
-    ).rejects.toThrowErrorMatchingInlineSnapshot(
-      '"Unable to get access token for \\"destination\\" service. No service instance of type \\"destination\\" found."'
-    );
+    ).rejects.toThrowError('No binding to an XSUAA service instance found. Please make sure to bind an instance of the XSUAA service to your application.')
   });
 });
--- a/packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
+++ b/packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
@@ -10,10 +10,14 @@ import { jwtBearerToken, serviceToken } from '../token-accessor';
 import { addProxyConfigurationOnPrem } from '../connectivity-service';
 import {
   getDestinationService,
-  getDestinationServiceCredentialsList
+  getDestinationServiceCredentialsList,
+  getXsuaaServiceCredentials
 } from '../environment-accessor';
 import { isIdenticalTenant } from '../tenant';
-import { DestinationServiceCredentials } from '../environment-accessor-types';
+import {
+  DestinationServiceCredentials,
+  XsuaaServiceCredentials
+} from '../environment-accessor-types';
 import { Destination } from './destination-service-types';
 import {
   alwaysProvider,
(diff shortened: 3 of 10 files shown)


CHANGELOG.md
```diff
@@ -23,7 +23,7 @@
 
 ## Improvements
 
--
+- [xsuaa] Reduce the number of warn message when selecting a XSUAA instance.
```
marikaner, line 26:
```suggestion
- [core] Reduce the number of warnings when selecting an XSUAA instance.
```

packages/core/src/connectivity/scp-cf/destination/destination-accessor-loading-precendence.spec.ts
```diff
@@ -75,8 +75,8 @@ describe('destination loading precedence', () => {
         { destinationName: 'non-existent' },
         { cacheVerificationKeys: false }
       )
-    ).rejects.toThrowErrorMatchingInlineSnapshot(
-      '"Unable to get access token for \\"destination\\" service. No service instance of type \\"destination\\" found."'
+    ).rejects.toThrowError(
```
marikaner, line 78: Why not snapshot?
```suggestion
    ).rejects.toThrowErrorMatchingInlineSnapshot(
```

packages/core/src/connectivity/scp-cf/destination/destination-from-service.ts
```diff
@@ -80,13 +84,16 @@ class DestinationFromServiceRetriever {
   ): Promise<Destination | null> {
     const subscriberToken =
       await DestinationFromServiceRetriever.getSubscriberToken(options);
+
+    const xsuaaCredentials = getXsuaaServiceCredentials(options.userJwt);
```
marikaner, line 88: Not related to this PR, but seems important: When I added support for x509, I updated the `ServiceCredentials` but not the `XsuaaServiceCredentials`... Shouldn't this support x509 as well?!

packages/core/src/connectivity/scp-cf/token-accessor.ts
```diff
@@ -41,7 +43,7 @@ export async function serviceToken(
 
   service = resolveService(service);
   const serviceCredentials = service.credentials;
-  const xsuaa = multiTenantXsuaaCredentials(opts.userJwt);
+  const xsuaa = multiTenantXsuaaCredentials({ ...options });
```
marikaner, line 46:
```suggestion
  const xsuaa = multiTenantXsuaaCredentials(options);
```

marikaner (changes requested):
Didn't get through the whole thing yet, but maybe some of the comments can already be addressed?

packages/core/src/connectivity/scp-cf/environment-accessor.ts
```diff
@@ -245,86 +245,95 @@ export function extractClientCredentials(
 }
 
 function selectXsuaaInstance(token?: JwtPayload): XsuaaServiceCredentials {
-  const xsuaaInstances = getServiceList('xsuaa');
+  const xsuaaInstances: XsuaaService[] = getServiceList('xsuaa') as any;
 
-  if (!xsuaaInstances.length) {
+  return (
+    selectViaJwt(xsuaaInstances, token) || selectWithoutJwt(xsuaaInstances)
+  );
```
marikaner, line 252: I would expect, that I either have a JWT or not and act based on this.
```suggestion
  return token ? selectViaJwt(xsuaaInstances, token) : selectWithoutJwt(xsuaaInstances);
```

Currently, if I have a token and no xsuaa instance could be matched, we go over to selection without jwt, where the error message states that there is no jwt ("and no JWT is given to decide which one to use").

Also, if I have no jwt, in the current flow I have to call the "with jwt" function although it isn't really useful, which is a bit confusing.

Consider adding back a function like `takeFirstAndWarn` that handles the case generically that there are multiple xsuaa instances matched (independently of whether there was a jwt or not).

marikaner (changes requested):
I am wondering why we have to pass around the xsuaa credentials. Maybe we can quickly have a chat about why this is beneficial or necessary.

packages/core/src/connectivity/scp-cf/environment-accessor.ts
```diff
@@ -238,91 +237,125 @@ export function resolveService(service: string | Service): Service {
...
+): XsuaaServiceCredentials {
+  if (xsuaaCredentials.length > 1) {
     logger.warn(
-      `Multiple XSUAA instances could be matched to the given JWT! Choosing the first one (xsappname: ${
-        first(selected)!.credentials.xsappname
+      `The following XSUAA instances are bound: ${xsuaaCredentials.map(
+        x => x.xsappname
+      )} and no JWT is given to decide which one to use. Choosing the first one (xsappname: ${
+        first(xsuaaCredentials)!.xsappname
       }).`
```
marikaner, line 293: formatting
```suggestion
      `The following XSUAA instances are bound: ${xsuaaCredentials.map(
        x => `\n\t- {x.xsappname}`
      )}
      No JWT given to select XSUAA instance. Choosing the first one (xsappname: "${
        first(xsuaaCredentials)!.xsappname
      }").`
```

packages/core/src/connectivity/scp-cf/environment-accessor.ts
```diff
@@ -238,91 +237,125 @@ export function resolveService(service: string | Service): Service {
...
-  xsuaaInstances: Record<string, any>[],
-  token?: JwtPayload
-) => Record<string, any>[];
+  if (selected.length > 1) {
+    logger.warn(
+      `Multiple XSUAA instances could be matched to the given JWT: ${xsuaaCredentials.map(
+        x => x.xsappname
+      )}. Choosing the first one (xsappname: ${first(selected)!.xsappname}).`
+    );
+    return selected[0];
```
marikaner, line 321: formatting
```suggestion
    logger.warn(
      `Multiple XSUAA instances could be matched to the given JWT: ${xsuaaCredentials.map(
        x => `\n\t- ${x.xsappname}
      )}
      Choosing the first one (xsappname: "${first(selected)!.xsappname}").`
    );
    return selected[0];
```

packages/core/src/connectivity/scp-cf/environment-accessor.ts
```diff
@@ -238,91 +237,125 @@ export function resolveService(service: string | Service): Service {
...
       'No binding to an XSUAA service instance found. Please make sure to bind an instance of the XSUAA service to your application.'
     );
   }
 
-  const strategies = [matchingClientId, matchingAudience, takeFirstAndWarn];
-  const selected = applyStrategiesInOrder(strategies, xsuaaInstances, token);
+  if (token) {
+    return selectXsuaaInstanceWithJwt(xsuaaCredentials, token);
+  }
+  return selectXsuaaInstanceWithoutJwt(xsuaaCredentials);
```
marikaner, line 265:
```suggestion
return token ? selectXsuaaInstanceWithJwt(xsuaaCredentials, token) : selectXsuaaInstanceWithoutJwt(xsuaaCredentials);
```

packages/core/src/connectivity/scp-cf/environment-accessor.ts
```diff
@@ -238,91 +237,125 @@ export function resolveService(service: string | Service): Service {
...
+}
 
-  if (selected.length === 0) {
-    throw Error('No XSUAA instances are found from the given JWT.');
+function handleOneXsuuaInstance(
+  xsuaaCredentials: XsuaaServiceCredentials[]
+): XsuaaServiceCredentials {
```
marikaner, line 270: This function and the error it throws seems a bit unnecessary to me.
I assume you wanted to have this function to avoid duplicating the debug message and TS tells you that `xsuaaCredentials[0]` could be undefined?
Maybe this can be circumvented with this:
```suggestion
function handleOneXsuuaInstance(
  xsuaaCredentials: [XsuaaServiceCredentials]
): XsuaaServiceCredentials {
```

FrankEssenberger: yes this works. For more than one element you can use `[a,...a[]]` and with type guards you can select the cases.

PR #1604, opened 2021-09-17
chore: add include/exclude to transpile directory

FrankEssenberger:
I needed this options in an other project where I wanted to exclude files trom the transpilation.

### DIFF
--- a/packages/generator-common/src/compiler.ts
+++ b/packages/generator-common/src/compiler.ts
@@ -1,7 +1,7 @@
 import { parse, resolve } from 'path';
 import { existsSync, promises } from 'fs';
 import { EOL } from 'os';
-import { createLogger } from '@sap-cloud-sdk/util';
+import { createLogger, flatten } from '@sap-cloud-sdk/util';
 import {
   CompilerOptions,
   createProgram,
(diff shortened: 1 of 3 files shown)


packages/generator-common/src/compiler.ts
```diff
@@ -22,17 +22,31 @@ const logger = createLogger('compiler');
...
+    resolve(path, exclude)
+  );
+  const allFiles: string[] = flatten(
+    includes.map(
+      include =>
+        new GlobSync(resolve(path, include), {
+          ignore: excludes
+        }).found
+    )
+  );
```
deekshas8, line 47: I came across this way of searching for more than one pattern. This avoids having to do multiple searches I believe.
```
  const allFiles = new GlobSync(resolve(path, '**/*.ts'), {
    ignore: resolve(path, '{dist/**/*, **/*.d.ts, **/*.spec.ts, node_modules/**/*}')
  }).found;
```

deekshas8 (approved):
LGTM.

packages/generator-common/src/compiler.ts
FrankEssenberger: I checked and this works only for the excludes. For the includes it does not. So I would perhaps keep it the way it is, since the number of searches is not reduced.

FrankEssenberger: But joining with `|` seems to work for the include. I add a test to ensure it.

deekshas8: For include, if its a single pattern, it works without giving any `{}`, so either like
`new GlobSync(resolve(path, '**/*.ts')` or like `new GlobSync(resolve(path, '{**/*.ts,}')` (not pretty). For more that one pattern in include this syntax worked, example: `new GlobSync(resolve(path, '{**/*.ts, **/*.d.ts}')`

PR #1617, opened 2021-09-22
chore: unpin xssec

FrankEssenberger:

Relates to SAP/cloud-sdk-backlog#315.

From version 3.2.3 onwards xssec used `GOT` as http client. This causes problem in the frontend e.g. react and webpack.  We pinned the version to come up with a fix using webpack 5. However, in the meantime xssec moved back to axios so we can remove the pinned version.


jjtang1985 (approved):
🤞 to E2E tests

PR #1619, opened 2021-09-23
feature: Auth token forwarding

FrankEssenberger:

Closes SAP/cloud-sdk-backlog#286.

Enables the `forwardAuthToken` on the enviorment destinations as the [approuter](https://help.sap.com/viewer/65de2977205c403bbc107264b8eccf4b/Cloud/en-US/3cc788ebc00e40a091505c6b3fa485e7.html) also allows it.

### DIFF
--- a/packages/core/src/connectivity/scp-cf/destination/destination-from-env.ts
+++ b/packages/core/src/connectivity/scp-cf/destination/destination-from-env.ts
@@ -10,8 +10,11 @@ import {
   isDestinationConfiguration,
   parseDestination
 } from './destination';
-import { Destination, DestinationAuthToken } from './destination-service-types';
-import { DestinationOptions } from './destination-accessor';
+import type {
+  Destination,
+  DestinationAuthToken
+} from './destination-service-types';
+import type { DestinationOptions } from './destination-accessor';
 
 const logger = createLogger({
   package: 'core',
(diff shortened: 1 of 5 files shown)


packages/core/src/connectivity/scp-cf/destination/destination-from-env.ts
```diff
@@ -161,3 +167,26 @@ export function searchEnvVariablesForDestination(
...
+): [DestinationAuthToken] | undefined {
+  if (token) {
+    const decoded = decodeJwt(token);
+    logger.info(
+      "Option 'forwardAuthToken' enabled on destination. Using the initial token for the destination."
+    );
+    return [
+      {
+        value: token,
+        expiresIn: decoded.exp!.toString(),
```
jjtang1985, line 182: If the service token expires, will you get a refreshed token automatically?

FrankEssenberger: no, The token we are talking about is the initial one from the incoming request. I would assume that his token is still valid.

jjtang1985 (approved):
LGTM.
Please consider updating the documentation.

FrankEssenberger:
Here is the PR for the documentation change: https://github.com/SAP/cloud-sdk/pull/476

PR #1634, opened 2021-09-28
fix: pac-resolver vulnerability

jjtang1985:
Related to: https://github.com/SAP/cloud-sdk-backlog/issues/310

Given that:
- `pm2` bumped `5.1.2`
- `@pm2/agent` bumped to `2.01`
for fixing the `pac-resolver` vulnerability.

However, as `pm2` uses `@pm2/agent: "~2.0.0"` and we have `@pm2/agent` as transitive dependency, I would not wait for dependabot.

### DIFF
--- a/test-packages/e2e-tests/package.json
+++ b/test-packages/e2e-tests/package.json
@@ -30,7 +30,7 @@
     "json-schema-faker": "^0.5.0-rcv.30",
     "moment": "^2.29.0",
     "openapi-backend": "^4.1.0",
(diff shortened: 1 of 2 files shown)


FrankEssenberger (approved):
LGTM but I think you said to wait for dependabot, right? If you want you can also merge now.

jjtang1985:
> LGTM but I think you said to wait for dependabot, right? If you want you can also merge now.

It's a typo. But after 24 hours, there is no dependabot PR.

PR #1644, opened 2021-10-01
fix: Removed warning for custom headers

FrankEssenberger:

Closes SAP/cloud-sdk-backlog#74.

We had a warning about `custom headers` which was not correct. We handed actually all headers to the executeRaw method. Not only the custom ones. Since the destination provided headers the warning was triggered . This PR removes this.

It also tightends the types of the `authentication-header.ts`.

I also includes the `Proxy-Authentication` headers if a custom authentication header is provided. 

Debate: I added a `debug` statement which shows the final headers and target. This is meant of you want to have son insight, which header is actually sent. Since this may content auth header I am not sure about this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ## Compatibility Notes
 
--
+- [core] Tightened the return type of `getAuthHeaders()` and `buildAuthorizationHeaders()` methods.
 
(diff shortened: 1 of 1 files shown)


marikaner (changes requested):
You only adjusted the changelog ;)

PR #1655, opened 2021-10-06
api-building

FrankEssenberger:

Closes SAP/cloud-sdk-backlog#198

ADR for our export strategy on version 2.0 of the SDK.

### DIFF
--- a/knowledge-base/adr/0028-public-api-extraction.md
+++ b/knowledge-base/adr/0028-public-api-extraction.md
@@ -1,0 +1,65 @@
+# Public API Generation
+
+## Status
+
+accepted 
+
+## Context
+
+For a typescript library it is common to use [barrels](https://basarat.gitbook.io/typescript/main-1/barrel) to make imports easy for consumers.
+A barrel is simply an `index.ts` re-exporting content of other source file.
+As a consumer you can then import all parts of the public API via the root `index.ts`.
+
+In the past we exported everything via a * in the barrels:
+```ts
+export * from '.some-file.ts'
+```
+This was convenient for us, but we had no distinction between:
+- This is part of the `public` API where a consumer can rely on stability.
+- This is part of the `internal` API and not meant for direct usage of the consumer.
+The convenience had the drawback that we were often blocked to do a refactoring.
+This ADR proposes a way to improve the situation.
+
+## API Categories 
+
+Our new strategy should provide the following options:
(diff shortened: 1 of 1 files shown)


knowledge-base/adr/0028-public-api-extraction.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Context
+
+For a typescript library it is common to use [barrels](https://basarat.gitbook.io/typescript/main-1/barrel) to make imports easy for consumers.
+A barrel is simply an `index.ts` re-exporting content of other source file.
+As a consumer you can then import all parts of the public API via the root `index.ts`.
+
+In the past we exported everything via a * in the barrels:
+```ts
+export * from '.some-file.ts'
```
florian-richter, line 15:
```suggestion
export * from './some-file.ts'
```

knowledge-base/adr/0028-public-api-extraction.md
```diff
@@ -0,0 +1,65 @@
...
+- This is part of the `public` API where a consumer can rely on stability.
+- This is part of the `internal` API and not meant for direct usage of the consumer.
+The convenience had the drawback that we were often blocked to do a refactoring.
+This ADR proposes a way to improve the situation.
+
+## API Categories 
+
+Our new strategy should provide the following options:
+- `Exported Public API`: This is a cautious decision and consumer can rely on stable contract for minor versions.
+- `Exported Internal API`: Exported for technical reasons but not meant to be used by consumer so no stable contract.
```
florian-richter, line 27: Maybe this could be rephrased/renamed to make it clear if this would be exported only via the full path to the exporting file or also exported in the barrel file (I would prefer the first option to make it undiscoverable externally to make clear there is no stable contract).

florian-richter (commented):
I am a bit confused what the solution will look like. The `stripInternal` seems helpful and should be a natural first step. To support the JS use case, we should also only export the public interface in the top level barrel files (which would mean no `barrelsby` or similar). Is this intended in this proposal or not?

knowledge-base/adr/0028-public-api-extraction.md
FrankEssenberger: I think you are right and we should only have one barrel file exporting the things we want to export. I will adjust the ADR to make this more clear.

FrankEssenberger:
I tried to make the flow more clear.

PR #1662, opened 2021-10-08
Split off connectivity module

FrankEssenberger:

Closes SAP/cloud-sdk-backlog#334.

Sorry for touching so many files, but there is now way around 🤷‍♂️. The main things:
- Minimal index.ts on connectivity
- A script `check-public-api.ts` to test if the index and types are aligned 
- All paths are adjusted


FrankEssenberger:
review #1692 instead

PR #1677, opened 2021-10-12
refactor!: remove aggregator package and related function calls

tomfrenken:
This PR removes the aggregator package as requested in the ticket in SAP/cloud-sdk-backlog#329

Closes SAP/cloud-sdk-backlog#329


FrankEssenberger:
please create the initial `diff` document and list the change that we have removed the aggregator package in v2.

PR #1736, opened 2021-10-26
feat: switch to es2019

jjtang1985:
This is a small PR, that changes the target version from `es5` to `es2019`.
Please ignore all the generated test services files.


FrankEssenberger (approved):

packages/odata-v4/src/uri-conversion/get-expand.ts
```diff
@@ -58,7 +58,7 @@ function getExpandAsString<EntityT extends Entity>(
       params = {
         ...params,
         ...createGetFilter(uriConverter).getFilter(
-          and(...expand._filters?.filters),
+          and(expand._filters?.filters || []),
```
FrankEssenberger, line 61: i assume this is equivalent in the new TS version.

jjtang1985: The new TS version is stricter.
Since `expand._filters?` can be `undefined`, we might pass `undefined` to `and`, which was wrong.

PR #1753, opened 2021-11-02
fix: windows tests

tomfrenken:
The issue was caused by windows error message handling, our test tested the concrete error message that looks like this on mac:

```
broken-src/file.ts:3:4 - error TS2588: Cannot assign to 'foo' because it is a constant
```

but it looks like this on windows:

```
Compilation Errors:·\n    D:/a/cloud-sdk-js/cloud-sdk-js/packages/generator-common/broken-src/file.ts:3:5 - error TS2588: Cannot assign to 'foo' because it is a constant.
```

Closes SAP/cloud-sdk-backlog#305.


tomfrenken:
worked on wrong branch, closing this PR and opening a new one

PR #1755, opened 2021-11-02
Implement de-serializers with defaults

marikaner:
Relates to SAP/cloud-sdk-backlog#399.

This introduces de-serializers and default de-serializers as a basis for customizable de-serializers according to this [ADR](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/adr/0015-date-time-handling.md#new-api-design).


jjtang1985 (commented):
Looks good.
I left some comments e.g., about the function `wrapUriSerialization`.

packages/odata-common/src/de-serializers/de-serialization-middleware.ts
```diff
@@ -0,0 +1,84 @@
+import { identity, isNullish } from '@sap-cloud-sdk/util';
+import BigNumber from 'bignumber.js';
+
+export interface DeSerializationMiddleware<
```
FrankEssenberger, line 4: for the other `deserializer` we did not split the `de` in kebap and camel case. I would like to have in everywhere the same. I find this `de-serialzer` and DeSeri..... a bit ugly so I would undo it here.

FrankEssenberger:
Just out of curiosity. What is our startegy to make the serializer changeble at runtime by the user? Which pattern will we use?

packages/odata-common/src/de-serializers/de-serialization-middleware.ts
jjtang1985: I guess this `DeSerialize` means `(de)serialize`, which is a combination of both `serialize` and `deserialize` as one word.

jjtang1985:
> Just out of curiosity. What is our startegy to make the serializer changeble at runtime by the user? Which pattern will we use?

The [ADR](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/adr/0015-date-time-handling.md#new-api-design) shows one example, which is a builder pattern with default serializer.

packages/odata-common/src/de-serializers/default-de-serializers.ts
```diff
@@ -0,0 +1,93 @@
...
+  deserializeToNumber,
+  serializeFromNumber,
+  deserializeToBigNumber,
+  validateGuid,
+  convertToNumber
+} from './payload-value-converter';
+import { DeSerializationMiddleware } from './de-serialization-middleware';
+import { convertToUriForEdmString, isInfOrNan } from './uri-value-converter';
+
+export const defaultDeSerializersRaw: DeSerializationMiddleware = {
```
FrankEssenberger, line 13: What is the `raw` reffering to? That it is raw data? If so I would a a doc explaining what is the default and leave it out.

packages/odata-common/src/de-serializers/de-serialization-middleware.ts
FrankEssenberger: hm, yes you are right. But I did not get it from the name.

packages/odata-common/src/de-serializers/default-de-serializers.ts
jjtang1985: I guess the `raw` is used with e.g., `wrapSerialization`, where:
- `raw` is the serialiser it self
- `wrap**` handles the `null` check.

packages/util/src/pipe.ts
```diff
@@ -15,9 +15,10 @@ export const asyncPipe =
...
- * Identity function
- * @param argument - Any object
- * @returns the given argument without doing something
- *
+ * Identity function.
+ * @param value - Any value.
+ * @returns The given value.
  */
-export const identity = <T>(argument: T): T => argument;
+export function identity<T>(value: T): T {
```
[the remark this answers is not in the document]
marikaner: No, we just decided, that we wouldn't use arrow functions anymore for these cases.

packages/odata-common/src/de-serializers/de-serialization-middleware.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+    return value;
+  };
+}
+
+export function wrapDeserialization<T>(
+  deserialize: DeSerializer<T>['deserialize']
+): DeSerializer<T>['serialize'] {
+  return function (value) {
+    if (isNullish(value)) {
```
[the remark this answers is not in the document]
marikaner: I did not change the behavior, it is the same as before, see [here](https://github.com/SAP/cloud-sdk-js/blob/7b5e1d6c0c5e78b83fde4e375ec0c21f2201a3f2/packages/odata-common/src/payload-value-converter.ts#L128) and [here](https://github.com/SAP/cloud-sdk-js/blob/7b5e1d6c0c5e78b83fde4e375ec0c21f2201a3f2/packages/odata-common/src/payload-value-converter.ts#L146).

I think the reason is, that in EDM there is no "undefined",  so technically this case can never occur - it is either null or not in the response.

packages/odata-common/src/de-serializers/de-serialization-middleware.ts
```diff
@@ -0,0 +1,84 @@
...
+    deSerializer: Omit<DeSerializer<T>, 'serializeToUri'>
+  ) => string;
+}
+
+export const identityDeSerializerRaw = {
+  deserialize: identity,
+  serialize: identity
+};
+
+export function wrapUriSerialization<T>({
```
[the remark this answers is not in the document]
marikaner: This function is currently unused, I will potentially delete it.

packages/odata-common/src/de-serializers/default-de-serializers.ts
marikaner: I am not happy with the naming either. Will try to improve before the next PR.

marikaner:
> Just out of curiosity. What is our startegy to make the serializer changeble at runtime by the user? Which pattern will we use?
> 
> * polomorphism?
> * funcitonal dispatching?
> * generics?

generics, especially because it is supposed to be dynamic.

marikaner:
Closing, as replaced by #1842.

PR #1763, opened 2021-11-03
fix: batch DELETE response error 

deekshas8:

Closes SAP/cloud-sdk-backlog#413

If the response is HTTP 200 OK but the body is empty (delete request SF), all the empty lines after headers are trimmed (including body) and `getResponseBody` returns the last item in array, which happens to be the header and results in invalid json error. Added a check for HTTP 200 OK and Content-Length: 0 to also append an empty line at the end.

### DIFF
--- a/packages/core/src/odata-common/request-builder/batch/batch-response-parser.ts
+++ b/packages/core/src/odata-common/request-builder/batch/batch-response-parser.ts
@@ -126,7 +126,9 @@ export function splitResponse(response: string, boundary: string): string[] {
   const newLineSymbol = detectNewLineSymbol(response);
   const parts = response.split(`--${boundary}`).map(part => {
     const trimmedPart = part.trim();
-    return trimmedPart.includes('204 No Content')
+    return trimmedPart.includes('204 No Content') ||
+      (trimmedPart.includes('200 OK') &&
+        trimmedPart.includes('Content-Length: 0'))
       ? `${trimmedPart}${newLineSymbol}`
       : trimmedPart;
   });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 
 ## Fixed Issues
 
--
+- [core] Fix parsing of batch response with HTTP 200 status code and containing empty body.
 
 
 # 1.51.0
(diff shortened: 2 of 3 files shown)


packages/core/src/odata-common/request-builder/batch/batch-response-parser.ts
```diff
@@ -126,7 +126,9 @@ export function splitResponse(response: string, boundary: string): string[] {
   const newLineSymbol = detectNewLineSymbol(response);
   const parts = response.split(`--${boundary}`).map(part => {
     const trimmedPart = part.trim();
-    return trimmedPart.includes('204 No Content')
+    return trimmedPart.includes('204 No Content') ||
+      (trimmedPart.includes('200 OK') &&
```
FrankEssenberger, line 130: What happens in the case of `200 OK` but the there is content, is there in this case perhaps a line to much. See comment on test.

FrankEssenberger (approved):
LGTM - I am not 100% if the new line can lead to errors, but most likely the content is parsed afterwards and this uncritical. Please think about the comments I added.

packages/core/src/odata-common/request-builder/batch/batch-response-parser.ts
jjtang1985: Good point. Unfortunately, in the current SDK, we assume update and delete response contains nothing.
- for a normal delete request, the response type of `execute` is `void`
- for a normal update request, the response is ignored and we build remote state as return value.

I personally would like to touch the API to reflect the real response.

jjtang1985:
Please also merge to 2.0 branch

packages/core/src/odata-common/request-builder/batch/batch-response-parser.ts
deekshas8: @FrankEssenberger : I didn't see a comment on test. 
From my understanding, if the status code is `200 OK` and response has some payload, then `Content-Length` should not be 0. The condition I added checks both:
`(trimmedPart.includes('200 OK') && trimmedPart.includes('Content-Length: 0'))` 

Therefore, when response has payload, we won't add another line. We have another test for this scenario (Line 279).

PR #1764, opened 2021-11-03
refactor!: change execute API

tomfrenken:
This PR updates the execute and executeRaw API, along with its adjacent functions, see [ADR](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/adr/0026-execute-api.md): 

Closes SAP/cloud-sdk-backlog#325.


packages/connectivity/src/scp-cf/destination/destination-accessor-types.ts
```diff
@@ -46,3 +50,40 @@ export interface DestinationAccessorOptions {
...
+  VerifyJwtOptions;
+
+/**
+ * Declaration of a destination to be retrieved from an environment variable or from the destination service on SAP Business Technology Platform, including all DestinationOptions.
+ *
+ * Use an object of this interface to specify which destination shall be used when executing a request.
+ * The destination will be retrieved via its [[DestinationFetchOptions.destinationName]] according to the following algorithm:
+ * 1. If a destination of this [[DestinationFetchOptions.destinationName]] is defined in the environment variable `destinations` (if available), it will be converted into a [[Destination]] and used for the request.
+ * 2. Otherwise, the destination service on SAP Business Technology Platform is queried for a destination with the given [[DestinationFetchOptions.destinationName]], using the access token provided as value of property [[jwt]].
+ * Additionally, you can set [[DestinationOptions]] for objects of this interface.
```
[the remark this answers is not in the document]
tomfrenken: Yes, I agree. I also updated the comments for some functions that don't use the `DestinationOptions` anymore; I'm not sure if I used the best formulation there as well.

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -25,21 +23,18 @@ import type { DestinationAccessorOptions } from './destination-accessor-types';
...
-  destination: Destination | DestinationNameAndJwt,
-  options: DestinationOptions = {}
+  destination: Destination | DestinationFetchOptions
 ): Promise<Destination | null> {
-  return isDestinationNameAndJwt(destination)
+  return isDestinationFetchOptions(destination)
     ? getDestination(
         destination.destinationName,
-        destination.jwt ? { userJwt: destination.jwt, ...options } : options
+        destination.jwt
```
FrankEssenberger, line 31: I would also adjust the API of the `getDestination()` method. This makes only sense if you have  `DestinationFetchOptions` in the other case we just sanatize the existing destination:
- create (use the existing) typeguard for `isDestination()` which decides if something is a destination or DestinationFetchOptions
- If it is the optins use then as argument for the getDestination

FrankEssenberger:
General: Make a global search for `DestinationNameAndJwt` I found it in some docs and minor places.

FrankEssenberger (changes requested):
Overall LGTM - only small  request of method signature change.

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
tomfrenken: Okay, I've changed the API of `getDestination` to use `DestinationFetchOptions`, along with all the affected tests

packages/connectivity/src/scp-cf/destination/destination-accessor-types.ts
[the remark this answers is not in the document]
tomfrenken: I've added the link, let's see if we do a follow-up or if I should do a docs PR as part of this ticket.

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -48,18 +36,27 @@ export type DestinationOptions = DestinationAccessorOptions &
...
  * @returns A promise returning the requested destination on success.
  * @internal
  */
 export async function getDestination(
-  name: string,
-  options: DestinationOptions = {}
+  destinationFetchOptions: DestinationFetchOptions
 ): Promise<Destination | null> {
+  destinationFetchOptions = destinationFetchOptions.jwt
+    ? { userJwt: destinationFetchOptions.jwt, ...destinationFetchOptions }
```
FrankEssenberger, line 47: I think we can rename the `userJwt` to `jwt` and get rid of this spread.

FrankEssenberger (approved):
Left one last comment.

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
tomfrenken: Agreed

jjtang1985:
@tomfrenken , have we updated the changelog v2 for such a breaking change?
Sorry for being late.

tomfrenken:
@jjtang1985 good point, I'll create a follow-up PR for the change-log

PR #1766, opened 2021-11-03
chore: Make  test in common test common

FrankEssenberger:
Relates to SAP/cloud-sdk-backlog#98

A first approach to use a Entity not containing protocol specific logic. I used the name `CommonXYZ` for object having this features. Feel free to suggest an other name. With the `RequestConfig`, `RequestBuilder` and `Requests`  I am pretty sure we can migrate all the tests from common to just use this entity. 

I wanted to give a first impression with this PR. Once this is merged I will do the rest. 
I will also remove the `toJSON()`  from the generated entities because it is the same for v2 and v4. 
But for this PR I wanted to keep the number of changes small.


packages/odata-common/src/entity-base.ts
```diff
@@ -260,6 +260,14 @@ export abstract class EntityBase {
       : names.filter(key => !equal(this.remoteState[key], currentState[key]));
   }
 
+  /**
+   * Overwrites the default toJSON method so that all instance variables as well as all custom fields of the entity are returned.
```
jjtang1985, line 264: I guess this is new function to be shared.
"Overwrites" seems to be a copy paste issue.

packages/odata-common/src/entity.spec.ts
```diff
@@ -1,66 +1,44 @@
...
-          TestEntitySingleLink.builder().stringProperty('singleLink').build()
-        )
-        .toMultiLink([
-          TestEntityMultiLink.builder().stringProperty('multiLink').build()
-        ])
-        .withCustomFields({ custom: 'custom' })
-        .build()
-        .setOrInitializeRemoteState();
+      const entity = new CommonEntity();
+      entity.setOrInitializeRemoteState();
```
jjtang1985, line 7: Question: does it remove the tests for navigation properties?

packages/odata-common/test/common-entity.ts
```diff
@@ -0,0 +1,147 @@
...
+        originalName: 'BooleanProperty',
+        name: 'booleanProperty',
+        type: 'Edm.Boolean',
+        isCollection: false
+      }
+    ];
+}
+
+// eslint-disable-next-line import/export
+export class CommonEntity extends EntityBase {
```
jjtang1985, line 60: Is it intended to have a different name "CommenEntity" v.s. "TestEntity"?

packages/odata-common/test/common-entity.ts
```diff
@@ -0,0 +1,147 @@
+/* eslint-disable max-classes-per-file */
```
jjtang1985, line 1: Question: Is this file is auto generated?
Or will it be?

jjtang1985 (commented):
In general, except for the `common-entity.ts`. The rest of the `common-XYZ` looks similar to some source code, but used for tests. I'm not sure the reason of the "duplication", which is a major concern.

packages/odata-common/src/entity-base.ts
FrankEssenberger: I copied it from the concrete entity: https://github.com/SAP/cloud-sdk-js/blob/f43da96fb9efbc5297f4279fa80060d542830c97/test-packages/test-services/v2/test-service/TestEntity.ts#L172

packages/odata-common/src/entity.spec.ts
FrankEssenberger: oh yes, because the links are v2 or v4 specific. I created a `entity.spec.ts` in v2 for that test

packages/odata-common/test/common-entity.ts
FrankEssenberger: no, the idea was to implement a entity based on the base which is not a v2 or v4 entity to test the content which is shared between oData versions.

packages/odata-common/test/common-entity.ts
FrankEssenberger: I wanted to emphasise that it is not v2 or v4 entity but one for common.

FrankEssenberger:
> In general, except for the common-entity.ts. The rest of the common-XYZ looks similar to some source code, but used for tests. I'm not sure the reason of the "duplication", which is a major concern.

The idea is to have serializers, oData Uri which is based on the common functionality and is used to build the shared `RequestConfig` which is shared between v2 and v4 and can be used for most tests.

jjtang1985:
> The idea is to have serializers, oData Uri which is based on the common functionality and is used to build the shared RequestConfig which is shared between v2 and v4 and can be used for most tests. I refactored one more test for the createRequestConfig where it becomes clear that for the tested funcitonality you do not need a v2 or v4 specific enttity.

The goal is to test common code in the common package.
To achieve so, we need common serializer/request config etc.

In the PR, we put the common serializer in the `test` directory, indicating 
- this is for test purposes
- the common serializer has nothing to do with the productive code
- we have to maintain one more location for any related changes.

Is it possible to have the common seralizer as source code and build a relation between common serializer and odata v2/v4 seralizer? I feel this might be a cleaner way.

FrankEssenberger:
hm, you are right but with the functional way this seems harder than with inheritance. I will try to acheive it.

packages/odata-common/test/common-entity.ts
```diff
@@ -0,0 +1,148 @@
...
+  int16Property: number;
+}
+export interface CommonEntityType {
+  keyPropertyGuid: string;
+  keyPropertyString: string;
+  stringProperty: string;
+  int16Property: number;
+}
+// eslint-disable-next-line import/export
+export namespace CommonEntity {
```
jjtang1985, line 88: I assume this is not generated, which means we have to remember this test entity when touching the generator.
When Marika's PR gets merged, for example, we might forget this, and therefore it will test the namespace which does not exist.
I feel we might want to use real test service (v2), unless the generator is able to generate a common test entity, which is useless in productive scenario.

jjtang1985 (commented):
Looks good.
I left a question/comment regarding common entity generation.

packages/odata-common/test/common-entity.ts
FrankEssenberger: hm, I guess you are right. I will try to use the generator to get these files.

FrankEssenberger: I added generation for the entity and replaced all imports after generation with the `odata-common`

packages/odata-common/test/common-entity.ts
```diff
@@ -0,0 +1,174 @@
+/* eslint-disable */
```
jjtang1985, line 1: I guess this is not needed any more as we have the generated one?

scripts/generate-test-services.ts
```diff
@@ -97,3 +95,76 @@ async function generateAll(): Promise<void> {
 }
 
 generateAll();
+
+function removeImports(str: string): string {
```
jjtang1985, line 99: I guess this line and below are the adjustment you mentioned to avoid touching the generator.

jjtang1985 (approved):
LGTM, I left one minor comment regarding the old `common-entity.ts`.

packages/odata-common/test/common-entity.ts
FrankEssenberger: I put both classes: entity and complex type in one file also the order of the attributes is not right. We disable the lint for all generated test cases because also the double export class and namespace is not ok. I did not want to disable the lint for any other file in odata-common so I used the disable.

scripts/generate-test-services.ts
FrankEssenberger: Yes, Imports are removed and replace by the ones from common which are sufficient for a entity with links. This will fail if we change the generation for entity strongly i.e. add different fields etc. But we will see this in a failling test so I guess this is ok.

PR #1769, opened 2021-11-05
fix: wrong token  for cache

FrankEssenberger:

Closes SAP/cloud-sdk-backlog#437.

Under some condition the wrong token was used to build the cache. The removed tests were misleading because the passed token from the user can be from the subscriber or provider account.


jjtang1985 (changes requested):
LGTM
You have to update a function as I deprecated one.
Please update the change log.
As a test driven fix, maybe you can share, without the fix, which tests failed?

packages/core/src/connectivity/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -140,6 +142,24 @@ describe('caching destination integration tests', () => {
       );
     });
 
+    it('cache key contains user also for provider tokens', async () => {
```
FrankEssenberger, line 145: This is the test which failed without the test. It would contain a key: `tenant:undefined:destName`. With your fix merged it would contain no cache key at all I guess.

jjtang1985 (approved):

PR #1770, opened 2021-11-05
Fix: destination cache vulnerablity

jjtang1985:
- Switch the default isolation strategy from `IsolationStrategy.Tenant` to `IsolationStrategy.Tenant_User`, when setting `useCache` to true for destination lookup functions like `getDestination`.
-  Disable destination cache, when the JWT does not contain necessary information. For example, when using `IsolationStrategy.Tenant_User`, the JWT has to contain both tenant id and user id.

Closes SAP/cloud-sdk-backlog#438
Closes SAP/cloud-sdk-backlog#439

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ## Compatibility Notes
 
--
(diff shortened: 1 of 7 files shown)


FrankEssenberger (approved):
LGTM

PR #1771, opened 2021-11-05
fix: Overwrite default expiration times

marikaner:
This fixes the issue of shorter expiration times than 5 minutes. Before this if the expiration of a token was 1 min, you would have to wait for 4 mins to make the next call to the destination instead...
Relates to SAP/cloud-sdk-backlog#437.


FrankEssenberger (approved):
LGTM

PR #1781, opened 2021-11-08
fix: batch DELETE response error

deekshas8:

Closes SAP/cloud-sdk-backlog#413

Merge #1763 to 2.0


FrankEssenberger (approved):
Same as merge on 2.0. So ok for me.

PR #1786, opened 2021-11-09
chore: Remove deprecated `FieldType` in `EdmTypeField`

marikaner:
This was deprecated and will make DeSerializer efforts easier.

This removed the FieldType from the Generics of EdmTypeField.


jjtang1985 (approved):
LGTM

PR #1778, opened 2021-11-08
refractor!: remove core packagee

tomfrenken:
This PR moves the remaining `test-util` folder from the `core` package to `test-resources`, removes the rest of `core` package and adapts all import statements.

Closes SAP/cloud-sdk-backlog#407.

PR #1788, opened 2021-11-10
refractor: remove core refrences

tomfrenken:
This PR removes the remaining refs and some package.json to the core package.
It also updates our loggers to use the respective packages.

Follow-up PR to #1778


FrankEssenberger (approved):
LGTM - one "Fleißsternchen" for Tom. Could you also search for method names containing `core` I think. For example `corePropertyTypeImportNames`. Could you also look for these in a separate PR.

tomfrenken:
I'll make a follow-up PR for the functions containing core 👍

PR #98, opened 2020-04-26
Documentation

artemkovalyov:
## Context

Automated pipeline for documentation
- Checks if build is successful on Pull request
- Publishes update on push 

I tested this in a forked repo to make sure checks and publishing work.

I added necessary private and public keys to the GH repo

PR #1792, opened 2021-11-10
chore: increase code reuse

FrankEssenberger:
In the PR #1766  @jjtang1985 requested that some code was remade this PR increases the reuse and should minimize the reimplementations.

### DIFF
--- a/packages/odata-common/src/payload-value-converter.ts
+++ b/packages/odata-common/src/payload-value-converter.ts
@@ -2,7 +2,7 @@
 
 import BigNumber from 'bignumber.js';
 import { identity } from '@sap-cloud-sdk/util';
-import { EdmTypeSameConverters } from './edm-types';
+import { EdmTypeCommon, EdmTypeSameConverters } from './edm-types';
 
 type EdmTypeMapping = { [key in EdmTypeSameConverters]: (value: any) => any };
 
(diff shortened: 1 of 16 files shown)


packages/odata-common/src/payload-value-converter.ts
```diff
@@ -111,3 +111,35 @@ export const serializersCommon: EdmTypeMapping = {
   'Edm.String': identity,
   'Edm.Any': identity
 };
+
+/**
+ * @hidden
+ */
+export function createEdmToTs<V extends EdmTypeCommon>(deserializers: {
```
deekshas8, line 118: For my understanding, shouldn't `V` also extend from `ExclusiveEdmTypeV2 | ExclusiveEdmTypeV4` or just from `EdmTypeShared<'any'>` for that matter ? . Same in `createUriConverter` below. All tests work fine though, so I might be mistaken here.

packages/odata-common/src/request/odata-get-by-key-request-config.spec.ts
```diff
@@ -1,9 +1,8 @@
 import { v4 as uuid } from 'uuid';
 import { TestEntity } from '@sap-cloud-sdk/test-services/v2/test-service';
-import { oDataUri as oDataUriV2 } from '@sap-cloud-sdk/odata-v2';
 import { ODataGetByKeyRequestConfig } from '@sap-cloud-sdk/odata-common';
-import { testEntityResourcePath } from '../../../../test-resources/test/test-util/test-data';
-import { uriConverter } from '../../../odata-v2/src/uri-conversion/uri-value-converter';
+import { uriConverter, oDataUri as oDataUriV2 } from '@sap-cloud-sdk/odata-v2/';
```
jjtang1985, line 4: I feel this should be moved to odata-v2 package.
The same applies to other tests in this folder.
Should we create a follow up?

jjtang1985 (approved):
LGTM.
I left a question.

packages/odata-common/src/request/odata-get-by-key-request-config.spec.ts
FrankEssenberger: We have already the ticket #98 for that.

packages/odata-common/src/payload-value-converter.ts
FrankEssenberger: EdmTypeShared contains EdmTypeCommon + ExclusiveEdmTypeV2 for V2. So I think to use `EdmTypeCommon` as a base type is fine. In principle you are right and the keys should be either the all V2 or all V4 ones. Not less like it would be possible now but also not more which would be the case if you use the EdmTypeShared<'any'>. Note EdmTypeShared is currently not select the version specif values.

PR #1794, opened 2021-11-11
chore: internal imports for generators

deekshas8:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/425
Closes https://github.com/SAP/cloud-sdk-backlog/issues/426

- mark methods `@internal` in both generators and ignore jsdoc check for them
- add public api in root index.ts - `generate()` and `GeneratorOptions`
- add `internal.ts`
- resolved names in `edmx-parser` 
- run `check:public-api` for generators


packages/generator/src/action-function-import/action.ts
```diff
@@ -7,7 +7,11 @@ import { additionalDocForEntityNotDeserializable } from './function';
 import { actionImportReturnType } from './return-type';
 
 const parameterName = 'parameters';
+/* eslint-disable valid-jsdoc */
```
florian-richter, line 10:
```suggestion
/* eslint-disable-next-line valid-jsdoc */
```

packages/generator/src/action-function-import/export-statement.ts
```diff
@@ -5,7 +5,11 @@ import {
   VariableStatementStructure
 } from 'ts-morph';
 import { VdmFunctionImport, VdmActionImport } from '../vdm-types';
+/* eslint-disable valid-jsdoc */
```
florian-richter, line 8:
```suggestion
/* eslint-disable-next-line valid-jsdoc */
```

packages/generator/src/action-function-import/import.ts
```diff
@@ -127,7 +131,11 @@ export function importDeclarationsFunction(
     service.oDataVersion
   );
 }
+/* eslint-disable valid-jsdoc */
```
florian-richter, line 134: only needed once per file if you use `eslint-disable` instead of `eslint-disable-next-line`

packages/generator/src/edmx-parser/edmx-file-reader.ts
```diff
@@ -29,7 +31,11 @@ function parseMetadata(
     root
   };
 }
+/* eslint-disable valid-jsdoc */
```
florian-richter, line 34: Shouldnt this be on line 8 (if it is needed at all)

packages/generator/src/edmx-to-vdm/common/function-import.ts
```diff
@@ -4,10 +4,14 @@ import { VdmFunctionImportBase } from '../../vdm-types';
 import { SwaggerPath } from '../../swagger-parser/swagger-types';
 import { functionImportDescription } from '../description-util';
 import { EdmxParameter } from '../../edmx-parser/common';
-import { EdmxFunctionImport as EdmxFunctionImportV2 } from '../../edmx-parser/v2';
-import { EdmxFunctionImport as EdmxFunctionImportV4 } from '../../edmx-parser/v4';
+import { EdmxFunctionImportV2 as EdmxFunctionImportV2 } from '../../edmx-parser/v2';
+import { EdmxFunctionImportV4 as EdmxFunctionImportV4 } from '../../edmx-parser/v4';
```
florian-richter, line 8:
```suggestion
import { EdmxFunctionImportV2 } from '../../edmx-parser/v2';
import { EdmxFunctionImportV4 } from '../../edmx-parser/v4';
```

florian-richter (commented):
Really large PR so I only scrolled through the bottom half, but everything seemed fined. Only found very minor stuff.

packages/generator/src/edmx-parser/edmx-file-reader.ts
deekshas8: The lint error in this case was only for functions with missing docs. Hence only added here. I changed it to 
`// eslint-disable-next-line valid-jsdoc` here

packages/openapi-generator/src/internal.ts
```diff
@@ -0,0 +1,10 @@
+export * from './file-serializer';
+export * from './file-writer';
+export * from './options';
+export * from './parser';
+export * from './sdk-metadata';
+export * from './document-converter';
+export * from './extensions';
+export * from './generator';
+export * from './openapi-types';
+export * from './schema-util';
```
jjtang1985, line 10: I know the `cli.ts` might not contain any internal "APIs", but the `internal.ts` should contain "all" source code. Could we all it as well?

packages/generator/src/internal.ts
```diff
@@ -0,0 +1,28 @@
...
+export * from './input-path-provider';
+export * from './internal-prefix';
+export * from './name-formatting-strategies';
+export * from './package-description';
+export * from './reserved-words';
+export * from './service-generator';
+export * from './service-mapping';
+export * from './service-name-formatter';
+export * from './typedoc';
+export * from './vdm-types';
```
jjtang1985, line 28: Same as the openapi generator. I would also add `generator-cli.ts` to internal so it contains all ts files.

jjtang1985 (approved):
LGTM.
I left some minor suggestions.

packages/generator/src/internal.ts
deekshas8: @jjtang1985 - I had added it initially in both packages, but it breaks the e2e tests. I debugged this and found that both  [cli.ts](https://github.com/SAP/cloud-sdk-js/blob/f4ad2302c7126347bc2124a016278d84f54dbbba/packages/openapi-generator/src/cli.ts#L12) and [generator-cli.ts](https://github.com/SAP/cloud-sdk-js/blob/f4ad2302c7126347bc2124a016278d84f54dbbba/packages/generator/src/generator-cli.ts#L17) directly calls the parse command line args methods. Therefore when internal.ts file loads all the modules, when it loads these files, it ends up calling the parse methods and fails. Thats why I have left them out.

PR #1795, opened 2021-11-11
chore: Add Internal import to generator-common

deekshas8:

Closes https://github.com/SAP/cloud-sdk-backlog/issues/430

- Added `barrel.config.ts` to generate `index.ts` to all subfolders.
- Removed all public apis from root `index.ts`
- Added `internal.ts` at root level to export everything. This file was created manually because
> - creating `internal.ts` at root level using barrelsby didn't consider exports from respective subfolder's index.ts but added full paths for all files
> - Updating it doesn't work properly without providing delete option, but this also deletes the root level `index.ts`
> -  using barrelsby its not possible to create another folder named `internal` with `index.ts` file
- Included `internal.js` and `internal.d.ts` under `files` in `package.json`
- Updated imports in other packages to `@sap-cloud-sdk/generator-common/internal`
- Updates tsdoc config to include `@file` which is introduced by barrelsby

TODO : Improve checks in public-api


packages/openapi-generator/src/generator.spec.ts
```diff
@@ -17,27 +17,6 @@ describe('generator', () => {
     mock.restore();
   });
 
-  it('getInputFilePaths returns an array of all file paths, including subdirectories', async () => {
```
tomfrenken, line 20: What's the reason for removing this test? 🤔

deekshas8: This was removed by mistake, have reverted it now. Thanks for pointing this out 👍

packages/generator/src/sdk-metadata/generation-and-usage.ts
```diff
@@ -1,12 +1,14 @@
...
   getSdkVersion,
   getLinks,
   getGenerationSteps,
   Links,
   InstructionWithTextAndHeader,
   usageHeaderText
-} from '@sap-cloud-sdk/generator-common';
-import type { GenerationAndUsage } from '@sap-cloud-sdk/generator-common';
+} from '@sap-cloud-sdk/generator-common/internal';
+import type { GenerationAndUsage } from '@sap-cloud-sdk/generator-common/internal';
```
FrankEssenberger, line 10: can you combine the imports? Isn't there a lint rule for this?

FrankEssenberger:
Can you create a follow up issue, that we create a check that we export all files via the intern.ts.

scripts/check-public-api.ts
```diff
@@ -4,11 +4,12 @@ import { GlobSync } from 'glob';
 import { createLogger, flatten, unixEOL } from '@sap-cloud-sdk/util';
 import mock from 'mock-fs';
 import { CompilerOptions } from 'typescript';
+/* eslint-disable import/no-internal-modules */
```
FrankEssenberger, line 7: there is a eslint-disable-next-line - I assume this does not work here?

FrankEssenberger (changes requested):
Looks already good. Left two questions.

scripts/check-public-api.ts
deekshas8: eslint-disable-next-line doesn't work when the whole import statement is not in a single line.

FrankEssenberger (approved):

PR #1809, opened 2021-11-16
chore: internal import for connectivity package

deekshas8:

Closes https://github.com/SAP/cloud-sdk-backlog/issues/451

Adjust Public/Internal API for connectivity package.


packages/connectivity/src/scp-cf/destination/index.ts
```diff
@@ -0,0 +1,12 @@
+export * from './destination-accessor';
```
tomfrenken, line 1: I like the introduction of index.ts files in subfolders, have to do that as well in my PR 👍

tomfrenken (approved):
LGTM

PR #1812, opened 2021-11-17
chore: internal import for odata-common

tomfrenken:
Adjust Public/Internal API for odata-common package.

Closes SAP/cloud-sdk-backlog#427.


packages/odata-v2/src/request-builder/response-data-accessor.ts
```diff
@@ -1,5 +1,6 @@
 import { createLogger } from '@sap-cloud-sdk/util';
-import { ResponseDataAccessor } from '@sap-cloud-sdk/odata-common';
+// eslint-disable-next-line import/no-internal-modules
```
FrankEssenberger, line 2: Could you try to use something like: (https://eslint.org/docs/user-guide/configuring/configuration-files)

    'import/no-internal-modules': ["error", {
      "allow": ["@sap-cloud-sdk/**/*"],
    }]

instead of adding very often the ignore case

FrankEssenberger (changes requested):
Pleae try to notuse the ignore lint rule but to use a allow rule in the eslint config for our modules.

packages/odata-v2/src/request-builder/response-data-accessor.ts
tomfrenken: I agree, I didn't see such a rule in the other PRs and assumed this was the go-to approach. Good to know this can be made cleaner.

tomfrenken:
I've implemented your suggestion, now it works as you suggested 👍

packages/eslint-config/index.js
```diff
@@ -121,7 +121,12 @@ module.exports = {
     'import/namespace': 'error',
     'import/no-absolute-path': 'error',
     'import/no-dynamic-require': 'error',
-    'import/no-internal-modules': 'error',
+    'import/no-internal-modules': [
```
FrankEssenberger, line 124: This is something for discussion, since this is the config we also ship to customers. Should costomers also do not get a warning if they include internal modules and use our linter? For me this makes sense.

FrankEssenberger (approved):
LGTM

PR #1818, opened 2021-11-18
chore: Update versions for peer deps of eslint-config

florian-richter:
I noticed during the setup of the samples repo that the eslint config is missing 2 peer dependencies and that it is does not use the current eslint version. I updated the peer dependencies to properly work with the current eslint version.


FrankEssenberger (approved):
LGTM

PR #1823, opened 2021-11-19
chore: Update public api checks for internal imports

deekshas8:

Closes SAP/cloud-sdk-backlog#448.

Updated the public-api script:
- checks for a single index.ts at root level
- checks for index.ts recursively in all src/** and an internal.ts at src level
- matches exports from internal barrel files with all files/modules for each package

**cli.ts and generator-cli.ts cannot be exported (even with export { parse...} from ..) hence these files are ignored in the checks


scripts/check-public-api.ts
```diff
@@ -253,3 +251,70 @@ function captureGroupsFromGlobalRegex(regex: RegExp, str: string): string[] {
...
+}
+
+function compareBarrels(
+  dirContents: string[],
+  exportedFiles: string[],
+  barrelFilePath: string
+) {
+  let setsAreEqual = true;
+
+  dirContents.forEach(tsFiles => {
```
FrankEssenberger, line 303: This is correct but I find the forEach loops a bit long to read. Perhaps you can do a:
``let difference = arrA.filter(x => !arrB.includes(x));`` and do a forEach only for the log statement. Also you can use the difference varialbes (for the two differnces) to sett the boolean return.

scripts/check-public-api.spec.ts
```diff
@@ -22,28 +23,13 @@ describe('check-public-api', () => {
         }
       }
     });
-    expect(indexFiles('root', 'index.ts')).toEqual([
+    expect(indexFiles('root', '**/index.ts')).toEqual([
```
FrankEssenberger, line 26: can you add one more test regarding the recursive check, which should fail if a file is missing in a subfolder.

FrankEssenberger, line 26: Also a second check to ensure that the content of the folder is parsed and comared to the content would be good.

FrankEssenberger (changes requested):
Looks very nice. Please consider my comments regarding one more tests.

FrankEssenberger (approved):

PR #1827, opened 2021-11-23
chore: remove deprecated from generator

FrankEssenberger:
In the generator some deprecated methods were included. This PR removes them and updates the clients. This is related to #1766 where I would like to use a clean generated entity in common.


tomfrenken (approved):
LGTM

PR #1830, opened 2021-11-23
feat: support defining headers and query params with origin

jjtang1985:
When calling `executeHttpRequest`, the SDK treats normal (v.s. custom) options `headers` and `query parameters` as custom options. This causes problems as the priority looks like below in terms of `headers` and `query parameters`:
```
custom > destination property > destination > normal option
```

With this PR, the origin information is saved and the values with highest priority can be picked depending on the origin.
This also enable potential fix for this [issue](https://github.com/SAP/cloud-sdk-backlog/issues/414), as it introduces separation for values from different origin and the encoding should not be applied for custom query parameters.

Related ADR: https://github.com/SAP/cloud-sdk-js/blob/2.0/knowledge-base/adr/0029-customer-header-handling.md

Closes SAP/cloud-sdk-backlog#361


packages/http-client/src/http-request-config.ts
```diff
@@ -40,3 +46,90 @@ const defaultDisallowedKeys = [
...
+ * Build [[HttpRequestConfig]] from [[HttpRequestConfigWithOrigin]]
+ * @param configWithOrigin - An http request config with origin information
+ * @returns The resulting [[HttpRequestConfig]]
+ * @internal
+ */
+export function buildHttpRequestConfig(
+  configWithOrigin: HttpRequestConfigWithOrigin
+): HttpRequestConfig {
+  const requestConfig = configWithOrigin as HttpRequestConfig;
+  requestConfig.headers = getOptionWithPriority(configWithOrigin.headers);
```
FrankEssenberger, line 60: I think the requirement is missing that we should merge ignoring the casing and keep the casing of the value which one. The `mergeLeftIgnoringCase` method should do it.

packages/http-client/src/http-client-types.ts
```diff
@@ -77,3 +88,29 @@ export interface HttpRequestOptions {
...
+ * @internal
+ */
+export type ValueWithOrigin = {
+  [key in Origin]?: string;
+};
+
+/**
+ * @internal
+ */
+export interface OptionWithOrigin {
```
FrankEssenberger, line 113: I found this additional type mislieading. I know it is convinient because you can get the options for a specific origin. However, you have to cast it back and fort and the `OptionWithOrigin` looks also like headers with origin from the type of `option`.

packages/http-client/src/http-client-types.ts
```diff
@@ -77,3 +88,29 @@ export interface HttpRequestOptions {
...
+/**
+ * Origins of http request options. This indicates the priority of an http request option.
+ * Http request options with higher priorities will be used when reaching conflicts.
+ * The priority is "Custom" \> "DestinationProperty" \> "Destination" \> "RequestConfig"
+ */
+export type Origin =
+  | 'Custom'
+  | 'DestinationProperty'
+  | 'Destination'
+  | 'RequestConfig';
```
FrankEssenberger, line 101: `RequestConfig` means default SDK?

packages/http-client/src/http-request-config.ts
```diff
@@ -40,3 +46,90 @@ const defaultDisallowedKeys = [
...
+export function mergeOptionsWithOrigin(
+  ...optionsWithOrigin: OptionWithOrigin[]
+): Record<string, ValueWithOrigin> {
+  return optionsWithOrigin.reduce(
+    (prevOptionWithOrigin, currentOptionWithOrigin) => {
+      if (currentOptionWithOrigin.option) {
+        Object.entries(currentOptionWithOrigin.option).reduce(
+          (prev, [optionKey, optionValue]) => {
+            if (prev[optionKey]) {
+              prev[optionKey][currentOptionWithOrigin.origin] = optionValue;
```
FrankEssenberger, line 120: In this method also some merging takes place so the order of the arguments needs to be in line with the priority.

packages/odata-common/src/request/odata-request.ts
```diff
@@ -135,17 +143,21 @@ export class ODataRequest<RequestConfigT extends ODataRequestConfig> {
...
+    }
+    return '';
   }
 
   /**
    * Create object containing all headers, including custom headers for the given request.
    * @returns Key-value pairs where the key is the name of a header property and the value is the respective value
    */
-  async headers(): Promise<Record<string, any>> {
+  async headers(): Promise<Record<string, ValueWithOrigin>> {
```
FrankEssenberger, line 160: I would prefer to a return a list of KeyValueWithOrigin or if you like to keep the OptionsFromOrigin a list of these so that no merging takes place until the http-client does the execute.

packages/http-client/src/http-client-types.ts
FrankEssenberger, line 113: I find this a bit misleading. It is more Options from origin right? So all header or parameter from one origin.

packages/http-client/src/http-request-config.ts
jjtang1985: - {'key': 'value'} (custom)
- {'KEY': 'value2'} (sdk)
=> {'key': 'value'}

packages/http-client/src/http-request-config.ts
```diff
@@ -40,3 +46,43 @@ const defaultDisallowedKeys = [
...
+ * Merge options from a given [[OriginOptions]]. When reaching conflicts, values with higher priorities are chosen.
+ * @param headersOrParams - Given options with origin information.
+ * @returns The resulting merged options.
+ */
+export function mergeOptionsWithPriority(
+  headersOrParams?: OriginOptions
+): Record<string, string> | undefined {
+  if (headersOrParams) {
+    return getAllOriginsAsc().reduce(
+      (prev, origin) => mergeIgnoreCase(prev, headersOrParams[origin]),
```
FrankEssenberger, line 75: I like this. very compact.

packages/http-client/src/http-client-types.ts
```diff
@@ -77,3 +88,18 @@ export interface HttpRequestOptions {
    */
   fetchCsrfToken?: boolean;
 }
+
+/**
+ * Origins of http request options. This indicates the priority of an http request option.
```
FrankEssenberger, line 93: `indicates` is perhaps not optimal. Perhaps `determines`

packages/http-client/src/http-request-config.spec.ts
```diff
@@ -24,3 +33,73 @@ describe('filterCustomRequestConfig', () => {
...
+      param3: 'destParam3',
+      param4: 'reqParam4'
+    };
+    expect(mergeOptionsWithPriority(originOptions)).toStrictEqual(expected);
+  });
+
+  it('should use options with higher priority and ignore case', () => {
+    const originOptions: OriginOptions = {
+      Custom: { Authorization: 'customAuth' },
+      Destination: { authorization: 'destAuth', 'sap-client': '001' }
```
FrankEssenberger, line 97: Minor: perhaps make hte casing difference stronger so that one sees it quicker: authorization versus AuThORizAtioN  or so.

FrankEssenberger (approved):
LGTM only very minor comments left. Please have a look and merge.

PR #1836, opened 2021-11-24
chore: update compareBarrels method

deekshas8:

Follow up on #1823. Improve code in compareBarrels.


FrankEssenberger (approved):
Thanks for remember to also improve the forEach loops.

PR #1839, opened 2021-11-25
chore: adjust tests third bunch

FrankEssenberger:
relates SAP/cloud-sdk-backlog#98

same as #1853 and #1766. Move tests from common to the v2 and v4 is specific or use the common entity.
The common entity is now generated on the fly with ts-morph.


packages/odata-v4/src/request-builder/get-all-request-builder.spec.ts
```diff
@@ -155,5 +156,27 @@ describe('GetAllRequestBuilder', () => {
...
+        4711,
+        TestEntity.requestBuilder().getAll()
+      );
+      const count = await requestBuilder.count().execute(defaultDestination);
+      expect(count).toBe(4711);
+    });
+  });
+
+  describe('execute raw', () => {
+    it('parses the raw number of count response', async () => {
```
marikaner, line 172: Why should this happen in the get all spec? ;)

packages/odata-v4/src/request-builder/get-all-request-builder.spec.ts
```diff
@@ -155,5 +156,27 @@ describe('GetAllRequestBuilder', () => {
         .execute(defaultDestination);
       expect(actual).toEqual([createTestEntity(testEntity)]);
     });
+
+    it('parses the raw number of count response', async () => {
```
marikaner, line 160: Same here

scripts/generate-test-services.ts
```diff
@@ -103,68 +103,77 @@ function removeImports(str: string): string {
...
-    (colllected, current) => current(colllected),
-    complexType
-  );
-  const adjusted = [
+async function generateCommonTestEntity() {
+  const files = await readClasses();
+
+  const [entity, entityLink, complexType, nestedComplex] = files
+    .map(str => removeImports(str))
+    .map(str => removeJsDoc(str));
```
marikaner, line 152: This could be done in one line

scripts/generate-test-services.ts
```diff
@@ -103,68 +103,77 @@ function removeImports(str: string): string {
...
+
+  const [adjustedEntity, adjustedEntityLink] = [entity, entityLink]
+    .map(str => adjustRequestBuilder(str))
+    .map(str => adjustCustomField(str))
+    .map(str => addODataVersion(str));
```
marikaner, line 157: Same here (then you would just iterate once).

scripts/generate-test-services.ts
```diff
@@ -103,68 +103,77 @@ function removeImports(str: string): string {
...
+    allParts,
     'utf8'
   );
 }
 
 const disclaimer = `/* This entity was generated from the COMMON_SRV.edmx and the generate-test-service.ts script.
 The idea behind this entity is to use only odata-common imports and use it in the tests for the odata-common functionality.*/`;
 const imports =
-  "import { AllFields, Constructable,  EntityBuilderType, Field,  OrderableEdmTypeField,CustomField,ComplexTypeField, ConstructorOrField, EdmTypeField, FieldBuilder, FieldOptions, FieldType, PropertyMetadata, EntityBase as Entity } from '../src/internal';";
+  "import { AllFields, Constructable,  EntityBuilderType,OneToOneLink, Field,CollectionField,  OrderableEdmTypeField,CustomField,ComplexTypeField, ConstructorOrField, EdmTypeField, FieldBuilder, FieldOptions, PropertyMetadata, EntityBase as Entity } from '../src/internal';";
```
marikaner, line 178:
```suggestion
  "import { AllFields, Constructable, EntityBuilderType, OneToOneLink, Field,CollectionField,  OrderableEdmTypeField, CustomField, ComplexTypeField, ConstructorOrField, EdmTypeField, FieldBuilder, FieldOptions, PropertyMetadata, EntityBase as Entity } from '../src/internal';";
```

marikaner (commented):
As I have not reviewed part 1 nor part 2 I am lacking a bit of background. I don't fully understand how to common service is used and why there are two versions (v2 and v4) of it. I wonder if then it wouldn't be more "honest" to just have a v2 and v4 service.
I can see that the service is slightly simpler than the service around the `TestEntity`, however I fail to recognize the benefit of this service compared to the old one.
Differences from my perspective:
- less properties
- no second level single / multi link properties
- no function imports

For OpenAPI we introduced something called a "simple" test entity, which this seems to be.
Maybe it would be an option to introduce this for our current service as well and reuse the other already existing entities.

Also, this still forces me to transpile before testing. But then again, maybe this is not part of the scope of this PR.

packages/odata-v4/src/request-builder/get-all-request-builder.spec.ts
FrankEssenberger: because count is part of the getAll? In v2 and common we have the same tests

scripts/generate-test-services.ts
FrankEssenberger: see comment below about nesting function calls

packages/odata-common/src/properties-util.spec.ts
```diff
@@ -1,18 +1,14 @@
...
-    });
-
-    it('returns true for one to many navigation property', () => {
-      expect(isNavigationProperty('toMultiLink', TestEntity)).toBe(true);
+      expect(isNavigationProperty('toSingleLink', CommonEntity)).toBe(true);
     });
 
     it('returns false for boolean property', () => {
-      expect(isNavigationProperty('booleanProperty', TestEntity)).toBe(false);
+      expect(isNavigationProperty('stringProperty', CommonEntity)).toBe(false);
```
marikaner, line 11: Ah, here we have a negative test. I think one level of `describes` can still be removed here.

marikaner: Also, as we have moved the multilink test to v2, should there be another test for v4 as well?

packages/odata-v4/src/request-builder/get-all-request-builder.spec.ts
marikaner: Hmm, not really. These tests check that `execute` and `executeRaw` of the `CountRequestBuilder` return what we expect them to return. The `GetAllRequestBuilder` only instantiates the `CountRequestBuilder`. So for a test in get all, I think we could check that a correct request builder is returned from the count function, but the response of the count request is not related to the get all request spec.

scripts/generate-test-services.ts
```diff
@@ -103,68 +103,77 @@ function removeImports(str: string): string {
...
-    'return Entity.customFieldSelector(fieldName, CommonEntity)',
-    'return new CustomField(fieldName,CommonEntity)'
+    'return Entity.customFieldSelector',
+    'return new CustomField'
   );
 }
 
 function addODataVersion(str: string): string {
   const firstProperty = "static _entityName = 'A_CommonEntity';";
+  const nameString = str.match(/static _entityName =.*/)![0];
```
marikaner: `firstProperty` can be removed

scripts/generate-test-services.ts
marikaner: I don't know which comment you are referring to, but this was supposed to just be a trigger for thought.

marikaner:
Please adjust the title of the PR to our convention.

marikaner (approved):

packages/odata-common/src/properties-util.spec.ts
FrankEssenberger: added it to this ticket: https://github.com/SAP/cloud-sdk-backlog/issues/463 about test parity.

PR #1840, opened 2021-11-25
refractor!: Implement TODO leftovers

tomfrenken:
This PR is part of the remove deprecated functionality ticket.

Closes SAP/cloud-sdk-backlog#194.


FrankEssenberger (commented):
LGTM @marikaner the legacy code was encapsulated nicely so removal was easy. 👍 I added a minor comment on the exception sentence.

FrankEssenberger (approved):

packages/connectivity/src/scp-cf/authorization-header.ts
```diff
@@ -184,9 +161,10 @@ function getProxyRelatedAuthHeaders(
     destination.proxyType === 'OnPremise' &&
     destination.authentication === 'NoAuthentication'
   ) {
-    return legacyNoAuthOnPremiseProxy(destination) as any;
+    throw Error(
+      'NoAuthentication is no longer supported for OnPremise connections'
```
FrankEssenberger, line 165: Perhaps I would repharse this to a more active form: OnPremise connections are not possible with NoAuthentication. Please select a supported auth flow e.g. principal propagation or basic.

tomfrenken: True, I'll also add the removed test case back and instead test for the error message

tomfrenken: Or is this test case too trivial?

Issue #1810, opened 2021-11-16
tenant information is not send to destination service for authentication type 'OAuth2ClientCredentials'


dinurp:
**Describe the bug**

Cloud SDK API ```getDestination``` is not sending tenant information to destination service when the authentication type is ```OAuth2ClientCredentials```. Because of this destination service is not able to return valid ```authTokens``` when the destination is configured with ```tokenServiceURLType``` as ```Common```.  In this case destination Service returns error messages instead of valid tokens:
```
Retrieval of OAuthToken failed due to: You must send the X-tenant header when the tokenServiceURLType property is Common
```

**To Reproduce**

Same as described in https://github.com/SAP/cloud-sdk-js/issues/1806 
Inspection of ```authTokens``` in destination returned with ```userJwt``` for provider and authentication type as ```OAuth2ClientCredentials``` shows the error.

**Expected behavior**
Valid ```authTokens``` should be fetched from destination service by sending the user context, either in ```X-tenant```, in case the destination is configured with ```tokenServiceURLType``` property is ```Common```.

**Used Versions:**
v14.18.1
- npm version via `npm -v`
7.16.0
- SAP Cloud SDK version you used as dependency
-  npm ls @sap-cloud-sdk/core
flights@1.0.0 C:\Work\cap-mtx-project
└─┬ @sap/cds@5.6.2
  └── @sap-cloud-sdk/core@1.52.0

**Code Examples**

same setup as in https://github.com/SAP/cloud-sdk-js/issues/1806. Inspect the returned ```accessTokens```

```
> var out = await sdk.getDestination("CAP_Incidents",{userJwt:process.env.access_token})
> out
{
  originalProperties: [Object],
  authTokens: [Array],
  certificates: [],
  name: 'CAP_Incidents,
  type: 'HTTP',
  url: 'https://httpbin.cfapps.eu10.hana.ondemand.com/anything',
  authentication: 'OAuth2ClientCredentials',
  proxyType: 'Internet',
  clientId: 'something',
  tokenServiceUrl: 'https://authentication.eu10.hana.ondemand.com/oauth/token',
  clientSecret: 'something',
  headers: [Object],
  queryParameters: [Object],
  isTrustingAllCertificates: false
}
> out.authTokens[0]
{
  type: '',
  value: '',
  expiresIn: '0',
  error: 'Retrieval of OAuthToken failed due to: You must send the X-tenant header when the tokenServiceURLType property is Common',
  http_header: undefined
}
```

[...]

PR #1845, opened 2021-11-26
Fix missing x user token header

FrankEssenberger:

relates to SAP/cloud-sdk-backlog#456
relates to SAP/cloud-sdk-backlog#457

This issue was brought up by the issues #1810 and $1806. The problem where that the wrong tokens were selected in a provider/subscriber scenario. Now for all user depended flows the `x-user-token` is set and for the OAuth2ClientCred the `x-tenant` header as both described [here](https://api.sap.com/api/SAP_CP_CF_Connectivity_Destination/resource).

@jjtang1985 can you add a sentence how to configure the ias credentials. I executed the local test and they all worked. With one exception, where I tried to reach out to the destination service colleagues.


packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -299,146 +298,135 @@ class DestinationFromServiceRetriever {
...
+      jwt:
+        origin === 'subscriber'
+          ? this.subscriberToken!.serviceJwt.decoded
+          : this.providerServiceToken.decoded
+    });
+    return { authHeaderJwt: clientGrant, exchangeTenant };
+  }
+
+  // This covers the two technical user propagation https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/3cb7b81115c44cf594e0e3631291af94.html
+  private usesSustemUser(destination: Destination): boolean {
```
jjtang1985, line 333:
```suggestion
  private usesSystemUser(destination: Destination): boolean {
```

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -151,16 +153,22 @@ class DestinationFromServiceRetriever {
...
+        ...options
+      });
       return {
-        decoded: await verifyJwt(options.jwt, options),
-        encoded: options.jwt
+        userJwt: {
+          decoded: await verifyJwt(options.jwt, options),
+          encoded: options.jwt
+        },
+        serviceJwt: { encoded, decoded: decodeJwt(encoded) }
```
jjtang1985, line 171: 👍

jjtang1985 (changes requested):
Very nice fix.
Unfortunately, the null related type is used for xssec.
We might need a new type for us.

packages/odata-v2/src/request-builder/get-all-request-builder.spec.ts
```diff
@@ -171,7 +171,7 @@ describe('GetAllRequestBuilder', () => {
       const nocks = [
         nock(onlyIssuerXsuaaUrl)
           .post('/oauth/token')
-          .times(2)
+          .times(1)
```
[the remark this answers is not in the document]
FrankEssenberger: yes I optimized the code of the `destination-from-service.ts` so that the clientCrendentials grant to call the destination service is only called once.

packages/connectivity/src/scp-cf/xsuaa-service.ts
```diff
@@ -53,8 +53,8 @@ function wrapInCircuitBreaker<T extends (...args: any[]) => any>(
 
 // `@sap/xssec` sometimes checks `null` without considering `undefined`.
 interface SubdomainAndZoneId {
-  subdomain: string | null;
-  zoneId: string | null;
+  subdomain: string | undefined;
+  zoneId: string | undefined;
```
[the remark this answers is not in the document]
FrankEssenberger: oh, I did overlook the comment. I changed it back to null and we should consider adjusting that or asking the xssec colleagues to be more `isNullish`

jjtang1985 (approved):
LGTM.
Please consider fixing it on main as well.
How about e2e tests, should this be covered or do we rely on the additional tests in the `auth-flow.spec.ts`?

FrankEssenberger:
PR for additional E2E tests is already prepared: https://github.tools.sap/cloudsdk/sdk-js/pull/100

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -299,146 +298,135 @@ class DestinationFromServiceRetriever {
...
-    if (!isUserToken(this.subscriberToken)) {
+    const { destination, origin } = destinationResult;
+    // This covers the X-Tenant case https://api.sap.com/api/SAP_CP_CF_Connectivity_Destination/resource
+    const exchangeTenant = this.getExchangeTenant(destination);
+    const clientGrant = await serviceToken('destination', {
+      jwt:
+        origin === 'subscriber'
+          ? this.subscriberToken!.serviceJwt.decoded
+          : this.providerServiceToken.decoded
+    });
```
marikaner, line 328: This seems to be the root

PR #1861, opened 2021-12-01
chore: fix circular dependencies

FrankEssenberger:

Relates SAP/cloud-sdk-backlog#152

Removed the cyclic dependencies fround by the lerna install step by moving the test-util mock destination functionality to the `connectivty` package. Also some minor API changes to get rid of a few circular dependencies. There will be a second PR to tackle the rest.

Lerna and madge will fail on new circular dependencies.


packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -158,8 +158,9 @@ class DestinationFromServiceRetriever {
           'You have provided the `userJwt` and `iss` options to fetch the destination. This is most likely unintentional. Ignoring `iss`.'
         );
       }
+      const creds = getXsuaaServiceCredentials(options.jwt);
       return {
-        decoded: await verifyJwt(options.jwt, options),
+        decoded: await verifyJwt(options.jwt, creds.uaadomain, options),
```
marikaner, line 163: This does not seem related to this PR, did you add this change on purpose?

CHANGELOG-v2.md
```diff
@@ -377,6 +377,8 @@
 ### Signature changed
 
 - [core] `EdmTypeField` only support EDM types, no field types in generics
+- [connectivity] `verifyJwt` needs the uaaDomain as input. The domain can be obtained from the XSUAA credentials.
```
marikaner, line 380: Is this related to this PR? Was this added on purpose?

packages/connectivity/src/scp-cf/destination/proxy-util.spec.ts
```diff
@@ -1,7 +1,7 @@
-import { unmockAllTestDestinations } from '@sap-cloud-sdk/test-util';
 import { createLogger } from '@sap-cloud-sdk/util';
 import { Protocol } from '../protocol';
 import { basicHeader } from '../authorization-header';
+import { unmockAllTestDestinations } from '../../../../test-util/src';
```
marikaner, line 4: I fail to recognize where the counterpart is called (`mockAllTestDestinations`). A quick search on my local state reveals that this is only called in the integration tests and in the tests of `test-util`. I wonder why we even need to call this...

marikaner (changes requested):
I think the util package is still needed for test-util (but I wonder why the dependency check does not fail, so maybe I am seeing it wrong?).
Other than that I mostly have questions.

FrankEssenberger:
Discussion marika
 - See if mock all is used at all
 - Use old JWT api and see why madge is complaining wiht modern ES imports.

marikaner (approved):
LGTM

PR #1828, opened 2021-11-23
refractor!: remove deprecated functionality

tomfrenken:
Removes the majority of deprecated functions/properties etc.
Solves some `TO-DO` comments

Closes SAP/cloud-sdk-backlog#194.

PR #1863, opened 2021-12-02
chore!: remove leftover deprecated functionality

tomfrenken:
This PR is the last PR in the series of #1828, remove leftover functionality

Closes SAP/cloud-sdk-backlog#194.


packages/odata-common/src/de-serializers/payload-value-converter.ts
```diff
@@ -2,14 +2,6 @@
 
 import BigNumber from 'bignumber.js';
 
-/**
- * @internal
- * @deprecated I think this should be removed and replaced by the number functions below.
- */
-export function convertToNumber(value: any): number {
```
marikaner, line 9: Sorry I was not very clear with this. I was wondering whether we should use `deserializeToNumber` and `serializeFromNumber` instead of `convertToNumber`. I discussed this with @florian-richter and we didn't see an issue with this, but didn't want to put this change into our previous PR to separate concerns.

packages/odata-common/src/de-serializers/default-de-serializers.ts
```diff
@@ -48,8 +47,8 @@ export const defaultDeSerializersRaw: DefaultDeSerializers = {
     serializeToUri: (value, serialize) => String(serialize(value))
   },
   'Edm.Byte': {
-    deserialize: convertToNumber,
-    serialize: convertToNumber,
+    deserialize: Number,
```
marikaner, line 50: This will work, but this was not the reason for deprecating it previously.

marikaner (commented):
LGTM, please check the following:
1. The change regarding `convertToNumber` was intended differently.
2. Make sure the PR adheres to conventional commits. I think we should discuss and document a clearer decision in our contribution guidelines, but we initially agreed on `chore`, `fix` and `feat` (and `BREAKING CHANGE` which we never used before). I think it is fine to extend our set of keywords with `refactor` but I believe we might not need the `!` ;) Anyways - I think this goes beyond refactoring and should rather be a `BREAKING CHANGE`?

tomfrenken:
According to the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) that we should adhere to:

```
Commit message with ! to draw attention to breaking change
```

That's why I used the `!`

marikaner (approved):

PR #1867, opened 2021-12-03
fix: Missing x user token header

FrankEssenberger:
apply the changes from #1845 . I tested the auth-flow tests and they worked.


jjtang1985 (changes requested):
LGTM.
Please update changelog and use more descriptive PR name like the same as old one
`Fix missing x user token header #1845`

jjtang1985 (approved):

PR #1872, opened 2021-12-06
feat: Implement 'DeSerializers' for the Temporal lib

deekshas8:
Implemented DeSerializers for the temporal lib for odata-v2 and v4.

Closes SAP/cloud-sdk-backlog#397.


FrankEssenberger:
is this ready for review?

packages/odata-v2/src/de-serializers/temporal-de-serializers.ts
```diff
@@ -0,0 +1,135 @@
...
+      dateTimeOffsetComponents.sign + dateTimeOffsetComponents.offset
+    );
+    return dateTimeInstant.toZonedDateTimeISO(timeZone);
+  }
+  return dateTimeInstant.toZonedDateTimeISO('UTC');
+}
+
+function parseEdmDateTimeOffset(edmDateTime: string): {
+  [key: string]: string;
+} {
```
FrankEssenberger, line 62: I would prefer a stronger typed return type since you use in the `deserializeToZonedDateTime` the ticks and offset and sign property.

packages/odata-v2/src/de-serializers/temporal-de-serializers.ts
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * @internal
+ */
+export function serializeFromZonedDateTime(
+  value: Temporal.ZonedDateTime
+): string {
+  const operator = value.offset[0];
+  const offset = value.offset.replace(/[-+:]/g, '');
+  return `/Date(${value.epochMilliseconds}${operator}${leftpad(offset, 4)})/`;
```
FrankEssenberger, line 91: The format looks like `+00:00`  which is replaced to `0000` by removing the sign and :. So when is the length shorter 4 and the leftpad is needed?

packages/odata-v2/src/de-serializers/temporal-de-serializers.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/**
+ * @internal
+ */
+export function deserializeToPlainTime(value: string): Temporal.PlainTime {
+  const regexResult =
+    /PT(?<hours>\d{1,2}H)?(?<minutes>\d{1,2}M)?(?<seconds>\d{1,2}S)?/.exec(
```
FrankEssenberger, line 99: In the `converters.ts` there is a very similar regex. Perhaps you can reuse it?

packages/odata-v2/src/de-serializers/temporal-de-serializers.ts
```diff
@@ -0,0 +1,135 @@
...
+      value
+    );
+  if (!regexResult) {
+    throw new Error(`Failed to parse the value: ${value} to time.`);
+  }
+  const { hours, minutes, seconds } = regexResult?.groups || {};
```
FrankEssenberger, line 105: I know that this is very compact and handles the three groups. However, if a value is not there it is `undefined` and the parseInt(undefined) gives then a `NaN` which seems to be lead to 0 in temporal. I know there is a test, but I would prefer to initialize the values with "0" if they are not there so that the parsing and passing to the new Temporal.PlainTime constructor is not critical.

packages/odata-v2/src/de-serializers/temporal-de-serializers.ts
```diff
@@ -0,0 +1,135 @@
...
+    parseInt(minutes),
+    parseInt(seconds)
+  );
+}
+
+/**
+ * @internal
+ */
+export function serializeFromPlainTime(value: Temporal.PlainTime): string {
+  return (
```
FrankEssenberger, line 118: Just a side remark. In the definition the serialize, deserialize is not idenpotent i.e. "PT12H" will become "PT12H00M00S" when you pipe it through. I think this should not be a problem but we should keep an eye on it and if it causes a problem for some backends.

packages/odata-v2/src/de-serializers/temporal-de-serializers.ts
```diff
@@ -0,0 +1,135 @@
...
+    leftpad(value.hour, 2) +
+    'H' +
+    leftpad(value.minute, 2) +
+    'M' +
+    leftpad(value.second, 2) +
+    'S'
+  );
+}
+
+function leftpad(value: any, targetLength: number): string {
```
FrankEssenberger, line 129: could this be a replacement: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

FrankEssenberger: I think we do not support IE anyway in the frontend. But feel free to ask this in the sdk-js chat to the colleages.

packages/odata-v4/src/de-serializers/temporal-de-serializers.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+export function serializePlainTimeToTime(value: Temporal.PlainTime): string {
+  return value.toString();
+}
+
+/**
+ * @internal
+ */
+export function deserializeDateToTemporal(date: string): Temporal.PlainDate {
+  const dateFormat = /\d{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])/;
```
FrankEssenberger, line 42: could you leave a comment that this is the regex for YYYY-MM-DD. It is nice that you enforced the numbers on month either 0X or 1Y and day but you did not do this on year where you also have some conditions. So I would perhaps loosen the regex to just `\d{2}". Also I wonder if this could be delegated to the `from` method. What would happen if a wrong string is parse?

FrankEssenberger: for the `deserializeDateTimeOffsetToTemporal` you did it this way

packages/odata-v4/src/de-serializers/temporal-de-serializers.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/**
+ * @internal
+ */
+export function deserializeDurationToTemporal(
+  value: string
+): Temporal.Duration {
+  const durationPattern =
+    /([+-]{1,1})?P(\d{1,2}D)?(T(\d{1,2}H)?(\d{1,2}M)?(\d{1,2}S)?(\d{2,2}\.\d+S)?)?/;
```
FrankEssenberger, line 97: also here there is a similar pattern in the `converters.ts`

PR #1896, opened 2021-12-14
chore: implement todos

tomfrenken:
This tickets solves a few of the TODOs in SAP/cloud-sdk-backlog#469.


FrankEssenberger (changes requested):
LGTM - However I would like to harden also this cache implementation like we did for the `destination-cahce`. So in case one part of the cache key is undefined by some mistake the cache key should be undefined so that the `cache.get` does not return something. The risk here is lower but you never now. Please also add a test for that.

FrankEssenberger (approved):
LGTM - try to get a test for the undefined case as discussed.

PR #1912, opened 2021-12-21
chore: Update edm converters version explicitly

jjtang1985:
The old version of this lib caused the node 14 warning.
We fixed it implicitly in the lock file by applying the upgrade by dependabot.
This PR explicitly update the version from our package.json.


tomfrenken (approved):
LGTM

PR #1917, opened 2021-12-27
feature: add `registerDestination`function

tomfrenken:
This PR adds the `registerDestination` function which sets a given destination into the environment variables.

Closes SAP/cloud-sdk-backlog#467.

### DIFF
--- a/packages/connectivity/src/scp-cf/destination/destination-from-env.spec.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-from-env.spec.ts
@@ -8,7 +8,8 @@ import {
 import { Destination } from './destination-service-types';
 import {
   getDestinationFromEnvByName,
-  getDestinationsFromEnv
+  getDestinationsFromEnv,
+  registerDestination
 } from './destination-from-env';
 import { getDestination } from './destination-accessor';
 
@@ -203,6 +204,30 @@ describe('env-destination-accessor', () => {
       expect(() => getDestinationsFromEnv()).toThrowErrorMatchingInlineSnapshot(
         '"Error in parsing the destinations from the environment variable."'
       );
+    });
+  });
+});
+
+describe('registerDestination', () => {
+  const mockDestination: Destination = {
+    name: 'MockedDestination',
+    url: 'https://example.com'
+  };
+
+  afterEach(() => {
+    unmockDestinationsEnv();
+    jest.resetAllMocks();
+  });
+
+  it('should set the destination in the environment variables', () => {
+    mockDestinationsEnv(environmentDestination);
+
+    registerDestination(mockDestination);
+    const actual = getDestinationsFromEnv();
+
+    const expected = [destinationFromEnv, mockDestination];
+    expected.forEach((e, index) => {
+      expect(actual[index]).toMatchObject(e);
     });
   });
 });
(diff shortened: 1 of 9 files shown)


packages/connectivity/src/scp-cf/destination/destination-from-env.ts
```diff
@@ -168,3 +168,47 @@ function destinationAuthToken(
...
+
+/**
+ * Set a given destination in the `destinations` environment variable.
+ *
+ * Throws an error if a destination with the same name as the given test destination already exists.
+ * @param destination - Test destination to add to the `destinations` environment variable
+ */
+export function registerDestination(destination: Destination): void {
+  const currentDestinations = getDestinationsFromEnv();
+  const existingNames = new Set<string>(
```
FrankEssenberger, line 201: we have a `unique` method in the util. This does the same but is more speaking.

packages/connectivity/src/scp-cf/destination/destination-from-env.spec.ts
```diff
@@ -206,3 +207,27 @@ describe('env-destination-accessor', () => {
...
+    name: 'MockedDestination',
+    url: 'https://example.com'
+  };
+
+  afterEach(() => {
+    unmockDestinationsEnv();
+    jest.resetAllMocks();
+  });
+
+  it('should set the destination in the environment variables', () => {
```
FrankEssenberger, line 222: Add a test to see the errors are thrown for empty name destinations

FrankEssenberger: Also add a test for the `getDestination(name)` case to see that the destination by name is working.

FrankEssenberger: Also add a test to see the name conflict exception being raised.

packages/connectivity/src/scp-cf/destination/destination-from-env.ts
```diff
@@ -168,3 +168,47 @@ function destinationAuthToken(
...
+    currentDestinations.map(dest => {
+      if (!dest.name) {
```
FrankEssenberger, line 203: Why this is a problem? I think our set method should ensure that you can not set a destination with undefined name, but if they are in the env by some mechanism is this a problem? Also without some context you get an exception which you can not investigate. So perhaps use a warning and give some info on the destination. Be careful not to expose sensitve information in the log. So perhaps it is easier to just get rid of this exception?

packages/connectivity/src/scp-cf/destination/destination-from-env.ts
```diff
@@ -168,3 +168,47 @@ function destinationAuthToken(
...
+        throw Error('The destination name is undefined.');
+      }
+      return dest.name;
+    })
+  );
+  if (!destination.name) {
+    throw Error('The destination name is undefined.');
```
FrankEssenberger, line 210: Perhaps formulate with a clear action: `registerDestination` does require to set XYZ.

FrankEssenberger: I would use a small helper method `checkDestiantionValid` which does a check on mandatory fields like name, but also URL.

FrankEssenberger (changes requested):
Some request on more tests and exception handling.

PR #1928, opened 2022-01-03
chore: remove deprecate annotation for `setOrInitializeRemoteState`

jjtang1985:
The method `setOrInitializeRemoteState` cannot be deprecated easily as this is a method of a class.
- Marking it private is not possible as the is used by e.g., deserialiser.
- Moving it outside the class to a function is not good as it calls private methods of the class.
- Using `export` keyword for controlling the accessibility is not possible.

As the initial issue was created 2 years ago and the root cause was fixed, there is no much business value for the time being.


tomfrenken (approved):
LGTM

PR #1929, opened 2022-01-03
chore: reduce code duplication

FrankEssenberger:

The `copyAdditionalFiles` is moved to common generator. This could be a blue print how to replace and then also test the `ts-morph` in the long run.


packages/generator/src/generator.ts
```diff
@@ -299,13 +316,6 @@ export async function generateSourcesForService(
     );
   }
 
-  if (options.additionalFiles) {
```
tomfrenken, line 302: Perhaps I understand this condition wrong, but shouldn't we use your new function `generateAdditionalFiles` here? Or was it removed entirely intentionally?

FrankEssenberger: at this point we used `ts-morph` to add the file, which made reuse with openapi difficult. Since we want to replace `ts-morph` in the long run anyway, I decided to use basic `fs` also here, which lead to the removal here and inclusion in the non. ts-morph file.

tomfrenken (approved):
LGTM

PR #1933, opened 2022-01-04
chore: add action and funciton imports to service

FrankEssenberger:

Closes SAP/cloud-sdk-backlog#493.

Added instances of the actions and function imports to the service object using the deserializer from the service instance.


packages/generator/src/generator-without-ts-morph/service/class.ts
```diff
@@ -46,9 +46,37 @@ export function serviceClass(service: VdmServiceMetadata): string {
...
+    
+    ${getActionFunctionImports(service, 'actionImports')}
   }`;
 }
 
+function getActionFunctionImports(
+  service: VdmServiceMetadata,
+  type: 'functionImports' | 'actionImports'
+): string {
+  const names = service[type]
```
marikaner, line 60: check the GH Actions warning.

packages/generator/src/generator-without-ts-morph/service/class.ts
```diff
@@ -46,9 +46,37 @@ export function serviceClass(service: VdmServiceMetadata): string {
...
+  const lines = service[type]!.map(
+    f =>
+      `${f.name}:(parameter:${f.parametersTypeName}<DeSerializersT>)=>${f.name}(parameter,this.deSerializers)`
+  );
+
+  return codeBlock`
+  get ${type}(){
+  return {${lines.join(',')}}
+  }
+  `;
```
marikaner, line 77: General: I know the code goes through a linter later on, but I think it would be nice, if we would apply some formatting ourselved, like spaces between ` => ` and `, `, so that this is also better readable.

packages/generator/src/generator-without-ts-morph/service/file.ts
```diff
@@ -10,12 +10,43 @@ ${serviceBuilder(service.className, service.oDataVersion)}
 ${serviceClass(service)}`;
 }
 
+function getImports(
+  actionFunctionImports:
+    | { name: string; parametersTypeName: string }[]
+    | undefined,
+  type: 'function-imports' | 'action-imports'
```
marikaner, line 17: Minor, I find `moduleIdentifier` more speaking.

marikaner (approved):
LGTM, I just left a few minor comments

packages/generator/src/generator-without-ts-morph/service/class.ts
FrankEssenberger: k

PR #1937, opened 2022-01-05
Update PULL_REQUEST_TEMPLATE.md

marikaner:
I think it is not good practice, if most of our PRs start with "Please provide a description of what your change does and why it is needed.". When we created the template the idea was to replace this sentence by a description. Hopefully a comment makes this more obvious ;)


jjtang1985 (approved):
👍

PR #1940, opened 2022-01-05
chore: try scan setup

FrankEssenberger:
relates SAP/cloud-sdk-backlog#482 When we want an exception it is good to have the tooling checked.

This PR includes a `code-ql` check on each PR. The reason the scans took so long in the past was that they scanned the docs and other useless things. After proper inclusion the scan takes 2min an runs nicely in parallel with the other checks, so I would just include it.

I created a error leading to a failing PR build:

image

with nice annotations on the lines where the errors are:

image

So I think this is a good addition to our pipeline. I kept the other triggers like schedule and merge on main so that we have security in case we do not work actively on the code.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -9,7 +9,11 @@ on:
   schedule:
     - cron: '0 20 * * *'
   push:
-    branches: [ add-check ]
+    branches: [ main,2.0 ]
+    tags: [ 'v*' ]
+    paths-ignore:
+      - 'docs/**'
+  pull_request: ~
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -6,8 +6,6 @@
 name: 'code-ql'
 
 on:
-  schedule:
-    - cron: '0 20 * * *'
   push:
     branches: [main, 2.0]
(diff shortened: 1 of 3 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -8,6 +8,13 @@ name: 'code-ql'
 on:
   schedule:
     - cron: '0 20 * * *'
+  push:
```
marikaner, line 11: I guess we don't need the scheduled one then, right?

.github/workflows/codeql-config.yml
```diff
@@ -0,0 +1,6 @@
+paths:
+  - packages
+paths-ignore:
+  - packages/**/node_modules
+  - packages/**/dist
+  - '**/*.spec.ts'
```
marikaner, line 6: When discussing with the GHAS colleagues it seemed very obvious for me, that static security checks should be running on tests as well. Currently I can't make up good examples why that was. Anyways, how much longer does it take with test files?

marikaner (approved):
LGTM, but we should make a conscious decision on the tests.

.github/workflows/codeql-analysis.yml
FrankEssenberger: I thought it is good if we do not work on the repo, but I guess you are right and the state should be save when we do it on each PR.

PR #1942, opened 2022-01-06
chore: Split v2 and v4 filter factory

marikaner:
I carved this out of my current work, for better review. In this PR, I split the v2 and v4 filter factory and moved the resources to the respective packages.


packages/odata-v2/test/test-util/test-data.ts
```diff
@@ -16,15 +16,24 @@ export const {
...
-    .keyPropertyGuid(originalData.KeyPropertyGuid)
-    .keyPropertyString(originalData.KeyPropertyString)
-    .stringProperty(originalData.StringProperty)
-    .booleanProperty(originalData.BooleanProperty)
-    .int16Property(originalData.Int16Property)
-    .build()
-    .setOrInitializeRemoteState();
+  const entityBuilder = testEntityApi.entityBuilder();
+
+  if ('KeyPropertyGuid' in originalData) {
```
tomfrenken, line 21: What was the motivation to create these `if (XY in YZ)` guards? Other than that lgtm 👍

tomfrenken (approved):
LGTM just one question in see comment

packages/odata-v2/test/test-util/test-data.ts
marikaner: If one of the proprties is not defined in the data it will be set like : `{ property: undefined }` which is not what we want ;)

tomfrenken: Ah, I see, thanks for the explanation 👍 , so you basically polished the previous implementation while you were moving the function, makes sense.

PR #1943, opened 2022-01-06
chore: remove reference to internal packages from (pre-)generated packages

tomfrenken:
This PR removes all references to internal packages from (pre-)generated packages

Closes SAP/cloud-sdk-backlog#492.


packages/odata-v4/src/de-serializers/de-serializers.ts
```diff
@@ -1,6 +1,7 @@
 import {
   DeSerializer,
-  DeSerializers as DeSerializersCommon
+  DeSerializers as DeSerializersCommon,
```
jjtang1985, line 3: Maybe we don't need this change and it can be moved to the common.ts?

packages/odata-v4/src/entity-base.ts
```diff
@@ -0,0 +1,5 @@
+export {
```
jjtang1985, line 1: Do we need this?

packages/odata-v4/src/internal.ts
```diff
@@ -10,3 +10,5 @@ export * from './entity';
 export * from './extract-odata-etag';
 export * from './filter-function';
 export * from './filter-functions';
+export * from './common';
+export { DeserializedType } from './de-serializers';
```
jjtang1985, line 14: This line seems to be a left over?

jjtang1985 (commented):
Looks good.
I left some comments.

packages/odata-v4/src/internal.ts
tomfrenken: I needed to re-export it because both `de-serliazers` and `common` contained `DeserializedType`

packages/odata-v4/src/entity-base.ts
tomfrenken: Good catch, I forgot to delete it after our new approach

packages/generator/src/batch/imports.ts
```diff
@@ -18,7 +18,8 @@ export function importBatchDeclarations(
         'GetAllRequestBuilder',
         'GetByKeyRequestBuilder',
         'ODataBatchRequestBuilder',
-        'UpdateRequestBuilder'
```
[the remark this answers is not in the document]
tomfrenken: Should they be broken? The `UpdateRequestBuilder`is still there, see the line below

packages/odata-v4/src/de-serializers/de-serializers.ts
tomfrenken: Good catch again, this is also a leftover before we switched the approach, it is already inside the common.ts, I'll remove it 👍

packages/generator/src/batch/imports.ts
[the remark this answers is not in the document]
jjtang1985: I see that the you touched multiple `import.ts` in different directories, and you adjusted most of the tests, as the implementation changed.
However, this `imports.ts` of `batch` directory is one exception, where no tests are touched. Maybe we can add a test , if it's missing?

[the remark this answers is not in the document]
tomfrenken: Ah now I understand it, you're right. I'll add similar tests to this `import.ts` like the others have.

jjtang1985 (approved):
Looks good to me and please consider adding a test.

packages/odata-v2/src/common.ts
```diff
@@ -0,0 +1,22 @@
+export {
```
marikaner, line 1: I like this approach 😊 The question is, if this should be "internal" in `odata-common`

packages/generator/src/entity/imports.ts
```diff
@@ -18,13 +17,12 @@ export function entityImportDeclarations(
...
     odataImportDeclaration(
-      ['Entity', 'DefaultDeSerializers', 'DeSerializers'],
+      ['Entity', 'DefaultDeSerializers', 'DeSerializers', 'DeserializedType'],
       oDataVersion
     ),
-    odataCommonImportDeclaration(['DeserializedType']),
     ...complexTypeImportDeclarations(entity.properties),
     ...enumTypeImportDeclarations(entity.properties)
-  ];
+  ].sort();
```
marikaner, line 25: Does this do, what you expect it to do? I am pretty sure `sort` needs a comparator to be able to handle objects, i.e. `ImportDeclarationStructure`s.

packages/generator/src/generator-without-ts-morph/service/file.ts
```diff
@@ -34,9 +30,10 @@ export function imports(service: VdmServiceMetadata): Import[] {
...
         'DeSerializers',
         'DefaultDeSerializers',
-        'mergeDefaultDeSerializersWith'
+        'mergeDefaultDeSerializersWith',
+        'Time'
       ],
       moduleIdentifier: `@sap-cloud-sdk/odata-${service.oDataVersion}`
     }
-  ];
+  ].sort();
```
marikaner, line 38: Same here.

packages/generator/src/request-builder/imports.ts
```diff
@@ -18,15 +17,18 @@ export function requestBuilderImportDeclarations(
...
+        'DeserializedType',
+        'RequestBuilder',
+        ...propertyTypeImportNames(entity.keys)
+      ],
+      oDataVersion
+    ),
     entityImportDeclaration(entity),
     ...entityKeyImportDeclaration(entity.keys)
-  ];
+  ].sort();
```
marikaner, line 31: And here ;)

marikaner (commented):
Concept-wise: LGTM.
But I agree with @jjtang1985, I think some tests might be missing (also before this PR...):
I think the sort you introduced doesn't necessarily do what you want it to do. I think it makes sense if the named imports within one import are sorted, but all imports is going to be difficult to achieve...

packages/odata-v2/src/common.ts
tomfrenken: We thought about this too, our thought was that it's cleaner if the clients don't have to import from both packages, that's why we re-exported it instead.

PR #1944, opened 2022-01-06
fix: missing token for OAuth2Password

jjtang1985:
Related to: https://github.com/SAP/cloud-sdk-backlog/issues/510

The SDK receives the destination object from the destination service successfully for this auth type, including auth token.
However, the token is not added to the header of the real request.

This PR fixes the missing token.

Follow up:
- check the token in the header for all the auth type, which might introduce breaking changes (v2 only implementation), as the `AuthenticationType` is a type, which is gone in runtime. Use e.g., enum as alternatives.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 
 ## Fixed Issues
 
--
+- [core] Fix `OAuth2Password` auth type for the missing token in the header. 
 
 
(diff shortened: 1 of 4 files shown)


FrankEssenberger (approved):
LGTM can you fix the two unrelated github code checks of unused destinationService -> little scope creep.

PR #1945, opened 2022-01-06
chore: replace moment in cache with Date and number

deekshas8:

Replace moment lib using in connectivity pkg for caching with Date and number. 

Closes SAP/cloud-sdk-backlog#490.

### DIFF
--- a/packages/connectivity/src/scp-cf/cache.ts
+++ b/packages/connectivity/src/scp-cf/cache.ts
@@ -125,9 +125,9 @@ function inferExpirationTime(
 function inferExpirationTimeFromDate(expirationTime: DateInputObject): number {
   const currentDate = new Date();
   const milliseconds =
-    ((expirationTime?.hours ?? 0) * 60 * 60 * 1000) +
-    ((expirationTime?.minutes ?? 0) * 60 * 1000) +
-    ((expirationTime?.seconds ?? 0) * 1000) +
+    (expirationTime?.hours ?? 0) * 60 * 60 * 1000 +
+    (expirationTime?.minutes ?? 0) * 60 * 1000 +
+    (expirationTime?.seconds ?? 0) * 1000 +
     (expirationTime?.milliseconds ?? 0);
 
   return currentDate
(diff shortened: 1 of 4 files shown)


packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -106,11 +104,70 @@ function isExpired<T>(item: CacheEntry<T>): boolean {
   if (item.expires === undefined) {
     return false;
   }
-  return !item.expires.isAfter(moment());
+  // return !item.expires.isAfter(moment());
```
FrankEssenberger, line 107: remove comment

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -106,11 +104,70 @@ function isExpired<T>(item: CacheEntry<T>): boolean {
...
 function inferExpirationTime(
-  expirationTime: moment.MomentInputObject | undefined
-): moment.Moment | undefined {
-  return expirationTime ? moment().add(expirationTime) : undefined;
+  expirationTime: DateInputObject | undefined
+): number | undefined {
+  return expirationTime ? inferExpirationTimeFromDate(expirationTime) : undefined;
+}
+
+interface DateInputObject {
```
FrankEssenberger, line 117: As I see it this is our API and we use it like: 
```new Cache<Destination>({ hours: 0, minutes: 5, seconds: 0 });```
So I would also only support times up to hours. I would also only allow for one keyword for one time property and pick the one with the plural `s` i.e. `hours,minutes,seconds`

FrankEssenberger: @marikaner @jjtang1985 @tomfrenken what is your opinion on this?

FrankEssenberger (approved):
LGTM from the code. I added my opinion on the interface. Perhaps other colleagues can also comment.

packages/connectivity/src/scp-cf/cache.ts
marikaner: I actually suggested to use milliseconds and just a number, so that the default consistent with the specific expiration times (which are also just a number and ms). We agreed to do this on a follow up.

marikaner (approved):
LGTM

PR #1952, opened 2022-01-07
chore: Fixes some issues in the batch

FrankEssenberger:
Adds typeguards `isSuccess` and `isReadResponse` on the batch reponse to properly use the `as` method.

relates SAP/cloud-sdk-backlog#494.

PR #1954, opened 2022-01-07
chore: Add batch entry to service

FrankEssenberger:
Prerequisite: #1952 
Closes SAP/cloud-sdk-backlog#494.

After the #1952 in merged this can be merged. It adds a `batch` entry to the service object.


marikaner (approved):
LGTM, on first sight I considered whether I like returning `typeof batch`, but after a quick consideration, I think this is quite elegant 😊

PR #1958, opened 2022-01-07
fix: move `odata-common` to dev dependency

jjtang1985:
The `@sap-cloud-sdk/odata-common` is used as dependency in the `generator` package, while only a type is used.
As the type has nothing to do with runtime, I move it to dev dependency.

### DIFF
--- a/packages/generator/package.json
+++ b/packages/generator/package.json
@@ -36,7 +36,6 @@
   },
   "dependencies": {
     "@sap-cloud-sdk/generator-common": "^1.50.0",
-    "@sap-cloud-sdk/odata-common": "^1.50.0",
     "@sap-cloud-sdk/util": "^1.50.0",
     "@sap/edm-converters": "~1.0.21",
     "@types/fs-extra": "^9.0.1",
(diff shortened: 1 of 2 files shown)


marikaner:
But you can use the package programatically, right?

jjtang1985:
> But you can use the package programatically, right?

Yes. But we should use it as `dependency` only when it's needed, no matter as a generator user or a SDK developer.
I feel this is not the case for the time being.

FrankEssenberger:
hm, then change is in principle an improvement because it will reduce the generator size is installed. However, I am afraid that it could lead to errors in the future. Imagine I include a non type import of OData common. Will the lerna publish fail or will it go through?

jjtang1985:
> hm, then change is in principle an improvement because it will reduce the generator size is installed. However, I am afraid that it could lead to errors in the future. Imagine I include a non type import of OData common. Will the lerna publish fail or will it go through?

In general, the odata common dependency was introduced for limited use cases and I would not expect further non-type imports.
In order to make sure we don't make mistakes, I removed the dependency completely and only keep it in the test packages.

florian-richter (approved):
LGTM

PR #1961, opened 2022-01-10
fix: move `@types/jsonwebtoken` to type test as dev dependency

jjtang1985:
The `@types/jsonwebtoken` is used as dependency in the `connectivity` package, while is is a compile time dependency.

### size
The install size of 
- `@sap-cloud-sdk/connectivity@2.0.0-beta.1-20220113152822.0` is now 8.79MB.
- `@types/jsonwebtoken` is now 1.58MB.

### major changes
- I copied some instances that are necessary
- The `@types/jsonwebtoken` is moved to the test package for testing the compatibility.

### DIFF
--- a/packages/connectivity/package.json
+++ b/packages/connectivity/package.json
@@ -42,7 +42,6 @@
     "opossum": "^6.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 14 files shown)


florian-richter (approved):
LGTM

PR #1962, opened 2022-01-10
fix: registerDestination with different parameter type

tomfrenken:
With this PR the `registerDestination` function expects a different type than before, it now expects `DestinationWithName` which notifies a user at design type of a mistake.

### DIFF
--- a/packages/connectivity/src/scp-cf/destination/destination-from-env.spec.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-from-env.spec.ts
@@ -209,7 +209,7 @@ describe('env-destination-accessor', () => {
 });
 
 describe('registerDestination', () => {
-  const mockDestination: Destination = {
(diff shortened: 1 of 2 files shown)


packages/connectivity/src/scp-cf/destination/destination-from-env.spec.ts
```diff
@@ -251,17 +251,6 @@ describe('registerDestination', () => {
     ).resolves.toMatchObject(mockDestinationFromEnv);
   });
 
-  it('should throw an exception if a property of the destination is missing', () => {
```
marikaner, line 254: Why?

tomfrenken: I think its impossible to test for compile errors, that's why I removed it instead (since the test can't run anymore)

PR #1968, opened 2022-01-11
fix: Remove warning when using environment destinations

tomfrenken:
This PR is a follow-up to our CAP meeting, where we decided to remove the warning.
It removes the warning and adapts the tests to it.

### DIFF
--- a/packages/connectivity/src/scp-cf/destination/destination-from-env.spec.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-from-env.spec.ts
@@ -101,7 +101,7 @@ describe('env-destination-accessor', () => {
       mockDestinationsEnv(destinationFromEnv);
 
       const logger = createLogger('env-destination-accessor');
-      const warnSpy = jest.spyOn(logger, 'info');
+      const infoSpy = jest.spyOn(logger, 'info');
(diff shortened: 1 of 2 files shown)


tomfrenken:
I just saw that we created an issue for this [here](https://github.com/SAP/cloud-sdk-backlog/issues/513), so I guess we still have to decide which approach we chose, don't merge until its decided then.

tomfrenken:
Since we have now decided to use this approach, it's fine to merge again.

packages/connectivity/src/scp-cf/destination/destination-from-env.spec.ts
```diff
@@ -101,11 +101,11 @@ describe('env-destination-accessor', () => {
       mockDestinationsEnv(destinationFromEnv);
 
       const logger = createLogger('env-destination-accessor');
-      const warnSpy = jest.spyOn(logger, 'warn');
+      const warnSpy = jest.spyOn(logger, 'info');
```
florian-richter, line 104:
```suggestion
      const infoSpy = jest.spyOn(logger, 'info');
```

florian-richter (approved):
Consider my tiny suggestion, otherwise LGTM.

PR #1975, opened 2022-01-13
fix: Batch `as` function deserializer not set

FrankEssenberger:
in the batch request the deserializer property is was not set.


packages/odata-common/src/request-builder/batch/batch-request-builder.ts
```diff
@@ -37,6 +39,9 @@ export class BatchRequestBuilder<
     )[] // readonly entityToConstructorMap: Record< //   string, //   EntityApi<EntityBase, DeSerializersT> // >
   ) {
     super(new ODataBatchRequestConfig(defaultServicePath));
+    this.deSerializers = Object.values(
+      this.getEntityToApiMap()
+    )[0].deSerializers;
```
florian-richter, line 44: Could this fail? Maybe it's safer to use `.find()` instead of accessing by index.

test-packages/integration-tests/test/auth-flows/auth-flow.spec.ts
```diff
@@ -49,9 +49,9 @@ describe('OAuth flows', () => {
     });
 
     const result = await BusinessPartner.requestBuilder()
-      .getAll()
-      .top(1)
-      .execute(destination!);
+        .getAll()
+        .top(1)
+        .execute(destination!);
```
florian-richter, line 54: Why was this indentation changed? Has prettier changed some defaults or was it faulty beforehand?

packages/odata-common/src/request-builder/batch/batch-request-builder.ts
FrankEssenberger: in principle if there are no request the whole thing does nothing and the undefined will not hurt but lets use find because it is always nicer than array access

test-packages/integration-tests/test/auth-flows/auth-flow.spec.ts
FrankEssenberger: unintentional. Reverted

florian-richter (approved):
LGTM

PR #1979, opened 2022-01-14
improve: executeHttpRequest for hiding destination/destination property OriginOptions

jjtang1985:
Closes SAP/cloud-sdk-backlog#514.

The implementation of `executeHttpRequest` is changed to support OriginOptions:
- custom
- request config
The destination related OriginOptions are handled under the hood and users do not have to provide them.
So the headers option looks like:
```ts
        headers: {
          custom: {
            authorization: 'custom-auth-header'
          },
          requestConfig: {
            'if-match': 'ETag'
          }
        }
```


packages/http-client/src/http-client-types.ts
```diff
@@ -38,14 +38,26 @@ export type HttpRequestConfig = HttpRequestConfigBase & {
 };
 
 /**
- * @internal
+ * This interface is similar to [[HttpRequestConfig]], whilst the `headers` and `params` can be defined with "origin" information.
```
FrankEssenberger, line 41: I whilst a proper word? Never used it.

packages/http-client/src/http-client.spec.ts
```diff
@@ -278,6 +265,44 @@ authorization:*******
 sap-client:001`);
     });
 
+    it('warn when custom headers are used', async () => {
```
FrankEssenberger, line 268:
```suggestion
    it('warns when custom headers are used', async () => {
```

FrankEssenberger: Minor: To be precise it logs as info and does not warn.

packages/http-client/src/http-client.ts
```diff
@@ -94,22 +90,109 @@ export function addDestinationToRequestConfig<T extends HttpRequestConfig>(
...
+    splitRequestConfig(requestConfig);
+  const headersDestination = await buildHeaders(destination);
+  const headersAndQueryParametersDestinationProperty: Pick<
+    Destination,
+    'headers' | 'queryParameters'
+  > = getAdditionalHeadersAndQueryParameters(
+    (destination.originalProperties as DestinationConfiguration) || {}
+  );
+
+  const mergedQueryParameter = mergeOptionsWithPriority({
```
FrankEssenberger, line 147: Why the method `mergeOptionsWithPriority` is called twice?

FrankEssenberger, line 147: Here a lot is going on, perhaps you can wrap it in two parts for query parameter and header? This would also suggest a split of `headersAndQueryParametersDestinationProperty` in the two cases.

FrankEssenberger (changes requested):
LGTM in general. Please anwser the questions I had, consider the refactoring suggestion and we can merge it.

packages/http-client/src/csrf-token-header.ts
```diff
@@ -66,11 +70,16 @@ function makeCsrfRequest<T extends HttpRequestConfig>(
...
     method: 'head',
     ...requestConfig,
-    headers: buildCsrfFetchHeaders(requestConfig.headers),
-    url: requestConfig.url
+    params: {
+      requestConfig: requestConfig.params || {}
+    },
+    headers: {
+      custom: buildCsrfFetchHeaders(requestConfig.headers),
+      requestConfig: requestConfig.headers || {}
```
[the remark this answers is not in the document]
jjtang1985: You are right, this second parameter in this function `requestConfig` is actually the one merged from all origins.
So I'll use `custom`.
The default `x-csrf-token: 'Fetch'` is only added, when the value is missing. When user would like to provide this value, the whole csrf token handling is skipped as it's provided.

packages/http-client/src/http-client-types.ts
```diff
@@ -89,7 +101,19 @@ export interface HttpRequestOptions {
...
+ */
 export interface OriginOptions {
+  requestConfig: Record<string, string>;
+  custom?: Record<string, string>;
+}
+
+/**
+ * @internal
+ */
+export interface OriginOptionsInternal {
```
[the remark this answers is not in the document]
jjtang1985: The two types has slight differences, making it hard for e.g., `extends`/`Omit`.
The internal type `OriginOptionsInternal` has everything optional, as we support optional, while the external type `OriginOptions` has mandatory key: `requestConfig`.

FrankEssenberger (approved):

PR #1986, opened 2022-01-17
fix: move isolation strategy to Destination API

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#465.

The isolation strategy was part of the `cahcingOptions` which is missleading because it is only meaningful for the destination cache. Adjusted the API and put the `isolationStrategy` to the destination options.

I also used the `fetchDestinationOptions` along the way of our fetch methods.

### DIFF
--- a/packages/connectivity/src/scp-cf/destination/destination-service-types.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-service-types.ts
@@ -194,7 +194,7 @@ export type DestinationRetrievalOptions = CachingOptions &
   ResilienceOptions & {
     /**
      * The isolation strategy used for caching destinations. For the available options, see [[IsolationStrategy]].
-     * By default, IsolationStrategy.Tenant is set.
+     * By default, IsolationStrategy.Tenant_User is set.
      */
     isolationStrategy?: IsolationStrategy;
   };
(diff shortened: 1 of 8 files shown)


packages/connectivity/src/scp-cf/destination/destination-service-types.ts
```diff
@@ -190,7 +190,14 @@ export interface DestinationCertificate {
...
  */
-export type DestinationRetrievalOptions = CachingOptions & ResilienceOptions;
+export type DestinationRetrievalOptions = CachingOptions &
+  ResilienceOptions & {
+    /**
+     * The isolation strategy used for caching destinations. For the available options, see [[IsolationStrategy]].
+     * By default, IsolationStrategy.Tenant is set.
+     */
+    isolationStrategy?: IsolationStrategy;
+  };
```
jjtang1985, line 200: It seems the line 203 is misleading, which should be `Destination` instead of `DestinationFetchOptions`.

packages/connectivity/src/scp-cf/destination/destination-service-types.ts
```diff
@@ -190,7 +190,14 @@ export interface DestinationCertificate {
 /**
  * Options to use while fetching destinations. Encompasses both [[DestinationCachingOptions]] and [[ResilienceOptions]] interfaces.
  */
-export type DestinationRetrievalOptions = CachingOptions & ResilienceOptions;
+export type DestinationRetrievalOptions = CachingOptions &
+  ResilienceOptions & {
+    /**
+     * The isolation strategy used for caching destinations. For the available options, see [[IsolationStrategy]].
+     * By default, IsolationStrategy.Tenant is set.
```
jjtang1985, line 197:
```suggestion
     * By default, IsolationStrategy.Tenant_User is set.
```

jjtang1985 (changes requested):
LGTM.
Please consider my comments.

jjtang1985 (approved):
LGTM

PR #1994, opened 2022-01-19
Add saml assertion flow and example

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#530.

@jjtang1985 asked to add the bare SamlAssertion flow to the test. I thought it would be good to add the flow to the `getDestiantion` and auth header building. In most cases it is not meaningful to use the saml assertion directly to the target system so I added a warn statement.

### DIFF
--- a/test-packages/integration-tests/test/auth-flows/auth-flow.spec.ts
+++ b/test-packages/integration-tests/test/auth-flows/auth-flow.spec.ts
@@ -41,7 +41,7 @@ describe('OAuth flows', () => {
 
     expect(destination!.authTokens![0].type).toBe('SAML2.0');
     expect(destination!.authTokens![0].value).toBeDefined();
-  });
+  },60000);
 
   xit('OAuth2Password: Fetches destination and destination service has token', async () => {
     const destination = await getDestination({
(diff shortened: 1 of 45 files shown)


test-packages/integration-tests/test/auth-flows/auth-flow.spec.ts
```diff
@@ -51,7 +63,7 @@ describe('OAuth flows', () => {
     const result = await BusinessPartner.requestBuilder()
       .getAll()
       .top(1)
-      .execute(destination!);
+      .execute(destination! as any);
```
jjtang1985, line 66: Could you please share why `as any` is needed here?

jjtang1985 (commented):
I left some questions.

test-packages/integration-tests/test/auth-flows/auth-flow.spec.ts
FrankEssenberger: I added a TODO in the beginning of the file. The reason is that we install the business partner vdm, which is based on core 1.53. On the other had I want to use the `getDstination` from the local sources to debug i.e. `connectivity` varion 2.0.
I updated the auth types of the destination which makes the types used by the `getDestination` and `execute` incompatible. This will disappear once we updated the VDM.

jjtang1985: This means, we test 1.53.x client with 2.0 core modules, which might be tricky.
As this is only the case for this file, it's still fine.

jjtang1985 (approved):
LGTM.
Please consider creating a ticket for the TODO to save a refinement item.

PR #1997, opened 2022-01-20
Add default (de-)serializers to the request builders

marikaner:
Users should not **have** to write out `DefaultDeSerializers` at any time. Before this PR, in some cases, they would have to do this. Therefore I added defaults to the types

Relates to SAP/cloud-sdk-backlog#515.


FrankEssenberger (approved):
LGTM

PR #1998, opened 2022-01-20
chore: Expose changeset

marikaner:
The changeset was not part of the service for some reason, added it, so that it is consistent with `batch`.

Relates to SAP/cloud-sdk-backlog#515.


FrankEssenberger (approved):
LGTM

PR #2000, opened 2022-01-20
chore: Export and, or, not from odata-vX packages

marikaner:
This exports the and, or and not functions from the odata version specific packages instead of common.
Additionally, those are not marked as @internal anymore.

Closes SAP/cloud-sdk-backlog#515.

### DIFF
--- a/packages/odata-common/src/filter/filterable.ts
+++ b/packages/odata-common/src/filter/filterable.ts
@@ -46,7 +46,6 @@ export type Filterable<
  * @typeparam EntityT - Type of the entity filter on.
  * @param expressions - Filterables to be combined with logical `and`.
  * @returns The newly created FilterList.
- * @internal
  */
 export function and<
   EntityT extends EntityBase,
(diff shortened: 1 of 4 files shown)


deekshas8 (approved):
LGTM. 
Minor Comment: The public-api check currently fails as there is an `*` in the index.ts of odata-v2/v4 packages (exports from common.ts)

jjtang1985 (approved):
LGTM for as an example.

I checked the readme of `odata-common`, where I see `# @sap-cloud-sdk/odata-common/internal`.
I feel we should remove "/internal" and mention something like "usually you should use `odata-v2`/`v4` for 99% of the use cases".

marikaner:
> LGTM. Minor Comment: The public-api check currently fails as there is an `*` in the index.ts of odata-v2/v4 packages (exports from common.ts)

But the check is green isn't it? It is also green on the 2.0 branch.

deekshas8:
> > LGTM. Minor Comment: The public-api check currently fails as there is an `*` in the index.ts of odata-v2/v4 packages (exports from common.ts)
> 
> But the check is green isn't it? It is also green on the 2.0 branch.

Yes, the step fails but is still overall successful. Will investigate this.
(node:3647) UnhandledPromiseRejectionWarning: Error: There is a '*' in /home/runner/work/cloud-sdk-js/cloud-sdk-js/packages/odata-v2/src/index.ts

PR #2001, opened 2022-01-20
chore: add xor type for Destination & DestinationFetchOptions

deekshas8:
Closes SAP/cloud-sdk-backlog#512.

Currently, the user can pass properties from both `Destination` and `DestinationFetchOptions` types together.. Created new type to strictly allow either `Destination` or `DestinationFetchOptions` as input to `executeHttpRequest`, `execute`, `buildCsrfHeaders`, etc methods.


packages/connectivity/src/scp-cf/destination/destination.ts
```diff
@@ -324,9 +324,17 @@ const configMapping: Record<string, keyof Destination> = {
...
-  destination: Destination | DestinationFetchOptions
+  destination: DestinationOrFetchOptionsXOR
 ): string {
   return isDestinationFetchOptions(destination)
     ? `Could not find a destination with name "${destination.destinationName}"! Unable to execute request.`
     : 'Could not find a destination to execute request against and no destination name has been provided (this should never happen)!';
 }
+
+type Without<T> = { [P in keyof T]?: never };
+type XOR<T, U> = (Without<T> & U) | (Without<U> & T);
```
marikaner, line 335: According to our [styleguide](https://github.com/SAP/cloud-sdk-js/blob/2.0/STYLEGUIDE.md#use-camel-case-for-acronyms) I think we should call this `Xor` - looks a bit weird, but then it would be consistent.

packages/connectivity/src/scp-cf/destination/destination.ts
```diff
@@ -324,9 +324,17 @@ const configMapping: Record<string, keyof Destination> = {
...
+
+export type DestinationOrFetchOptionsXOR = XOR<
```
marikaner, line 337: Same here, however I think the Xor is not needed, I think a simple `DestinationOrFetchOptions` would suffice.

packages/connectivity/src/scp-cf/destination/destination.ts
marikaner: BTW - if there are two keys that have the same name, this would still work, wouldn't it? (Unlikely, but I am asking out of interest).

marikaner (changes requested):
Generally looks good to me, but I think we should add at least one type test, that checks that this really does not work with a combined type (and maybe two more to check that it works for the individual types).

packages/connectivity/src/scp-cf/destination/destination.ts
deekshas8: I tried this, and it doesn't work for keys with same name. Probably because in `Without<T>`, that property would be of type `never`. And it remains `never` in the resultant type `Without<T> & U` as well. 

It works when I use `Exclude` instead:
```
type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
type Xor<T, U> = (Without<T, U> & U) | (Without<U, T> & T);

export type DestinationOrFetchOptions = Xor<
  Destination,
  DestinationFetchOptions
>;
```

Should we consider handling this?

marikaner: I think this won't be necessary, but maybe add a comment, so that if this comes up in the future, we are aware?

marikaner (approved):
Maybe still add a comment, but LGTM 😊

packages/connectivity/src/scp-cf/destination/destination.ts
```diff
@@ -374,9 +374,20 @@ const configMapping: Record<string, keyof Destination> = {
...
   return isDestinationFetchOptions(destination)
     ? `Could not find a destination with name "${destination.destinationName}"! Unable to execute request.`
     : 'Could not find a destination to execute request against and no destination name has been provided (this should never happen)!';
 }
+
+type Without<T> = { [P in keyof T]?: never };
+/**
+ * XOR of two types containing keys with different names.
+ */
+type Xor<T, U> = (Without<T> & U) | (Without<U> & T);
```
jjtang1985, line 388: minor: Perhaps we can move them to util, as this is a general type definition.

jjtang1985 (approved):
LGTM.
I left one minor comment.
You decide.

PR #2005, opened 2022-01-21
chore: make allowed licences more strict

FrankEssenberger:
In the discussion about code scanning tools to used, I realized that the list of licences is a bit too broad. Luckily there we did not use any of the GPL in production code, only in some test libs.

### DIFF
--- a/scripts/check-licenses.ts
+++ b/scripts/check-licenses.ts
@@ -4,7 +4,7 @@ import { createLogger } from '@sap-cloud-sdk/util';
 
 const logger = createLogger('check-licenses');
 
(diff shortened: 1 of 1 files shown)


jjtang1985 (approved):
LGTM.
Maintaining such list by our own might be risky.
Is there any docs from e.g., OSPO or somewhere so we don't have to reinvent the wheel?

FrankEssenberger:
We are currently in a discussion with the colleagues what tool is allowed. I guess whitesource and blackduck are the recommendations, but they are not fun to use.

PR #2017, opened 2022-01-25
chore: redo isolation strategy.

FrankEssenberger:
Relatest to SAP/cloud-sdk-backlog#541

This PR improved the `IsolationStrategy`. We have moved this option to the destination retrival, away from the general cache, which makes a few changes possible:
- The meaningful values are `Tenant` and `TenantUser`. For a destination it does not make sense in a multi-tenant scenario to isolate on `User` level or have `NoIsolation`. If you have a single tenant app you can just go with `Tenant`. So `User` and `NoIsolation` are removed.
- The default value for the IsolationStrategy is based on the other inputs now: If a JWT is provided UserTenant is assumed (highest isolation). If no JWT is provided a user independent service token is used so IsolationStrategy `Tenant` is the only meaningful choice, because you do not have a user.
- If a `IsolationStrategy` is given to the `getDestination()` this enables the cache. No need to provided `useCache: true`
- The destination-service caller also had a cache, which only makes sense for the Destination**s** endpoints we you get all destinations from a tenant, without the auth flow. This data is rather stable and can be cached. The actual destination by name call should not be cached, since the DestinationRetriver does the caching there already.

Please review this carefully because the caching is a security relevant thing.

### DIFF
--- a/packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
@@ -31,7 +31,6 @@ import {
   onPremisePrincipalPropagationMultipleResponse
 } from '../../../../../test-resources/test/test-util/example-destination-service-responses';
 import { decodeJwt, wrapJwtInHeader } from '../jwt';
-import { IsolationStrategy } from '../cache';
 import { destinationServiceCache } from './destination-service-cache';
 import { getDestination } from './destination-accessor';
 import {
@@ -39,7 +38,11 @@ import {
   alwaysSubscriber,
   subscriberFirst
 } from './destination-selection-strategies';
-import { destinationCache, getDestinationCacheKey } from './destination-cache';
+import {
+  destinationCache,
+  getDestinationCacheKey,
+  IsolationStrategy
+} from './destination-cache';
 import {
   AuthenticationType,
   Destination,
@@ -184,40 +187,13 @@ describe('destination cache', () => {
 
       const c1 = getSubscriberCache(IsolationStrategy.Tenant);
       const c2 = getProviderCache(IsolationStrategy.Tenant);
-      const c3 = getSubscriberCache(IsolationStrategy.User);
-      const c4 = getProviderCache(IsolationStrategy.No_Isolation);
       const c5 = getSubscriberCache(IsolationStrategy.Tenant_User);
       const c6 = getProviderCache(IsolationStrategy.Tenant_User);
 
       expect(c1).toBeUndefined();
       expect(c2).toBeUndefined();
-      expect(c3).toBeUndefined();
-      expect(c4).toBeUndefined();
       expect(c5!.url).toBe('https://subscriber.example');
       expect(c6).toBeUndefined();
-    });
-
-    it('retrieved  provider destinations are cached using only destination name in "NoIsolation" type', async () => {
-      await getDestination({
-        destinationName: 'ProviderDest',
-        jwt: subscriberUserJwt,
-        useCache: true,
-        isolationStrategy: IsolationStrategy.No_Isolation,
-        cacheVerificationKeys: false,
-        iasToXsuaaTokenExchange: false
-      });
-
-      const c1 = getSubscriberCache(IsolationStrategy.No_Isolation);
-      const c2 = getProviderCache(IsolationStrategy.No_Isolation);
-      const c3 = getSubscriberCache(IsolationStrategy.User);
-      const c4 = getSubscriberCache(IsolationStrategy.Tenant);
-      const c5 = getSubscriberCache(IsolationStrategy.Tenant_User);
-
-      expect(c1).toBeUndefined();
-      expect(c2!.url).toBe('https://provider.example');
-      expect(c3).toBeUndefined();
-      expect(c4).toBeUndefined();
-      expect(c5).toBeUndefined();
     });
 
     it('caches only subscriber if the destination names are the same and subscriber first', async () => {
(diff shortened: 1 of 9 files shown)


florian-richter (commented):
Some clean up is necessary. I am unsure if the restriction to only Tenant and TenantUser could come back as a problem, but I understand the reasoning and can't come up a scenario where it would be a problem.

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -199,9 +199,11 @@ class DestinationFromServiceRetriever {
...
   ) {
     const defaultOptions = {
-      isolationStrategy: IsolationStrategy.Tenant_User,
+      isolationStrategy: getIsolationStrategy(
+        subscriberToken?.userJwt?.decoded
+      ),
       selectionStrategy: subscriberFirst,
-      useCache: false,
+      useCache: !!options.isolationStrategy,
       ...options
```
deekshas8, line 207: This operation is done again in the next line

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -317,6 +293,105 @@ describe('destination cache', () => {
...
+        IsolationStrategy.Tenant_User
+      );
+      await expect(
+        getDestination({
+          destinationName: destName,
+          isolationStrategy: IsolationStrategy.Tenant_User,
+          jwt: subscriberServiceToken,
+          iasToXsuaaTokenExchange: false
+        })
+      ).rejects.toThrowError(/Failed to fetch instance destinations./);
```
deekshas8, line 388: I believe `getDestination` in this case will ultimately call `fetchInstanceDestinations` and `fetchSubaccountDestinations` in parallel and depending on which call fails first might lead to a different error. (`Failed to fetch subaccount destinations.`)

deekshas8 (commented):
Left 2 comments. In general, the changes look good to me, but I am not sure of all security related aspects.

FrankEssenberger:
> Some clean up is necessary. I am unsure if the restriction to only Tenant and TenantUser could come back as a problem, but I understand the reasoning and can't come up a scenario where it would be a problem.

I agree on this point. It could also be a bit misleading for a user. Assume you have a single tenant app and have a destination with fixed credentials like `Basic`. You want to configure the cache with the weakest isolation. Also in this case you have to set `Tenant`, which will do the trick because there is only one tenant in your case. 

But I did not have a better idea. Good documentation is needed.

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -317,6 +293,106 @@ describe('destination cache', () => {
...
+      const actual = await getDestination({
+        destinationName: destName,
+        isolationStrategy: IsolationStrategy.Tenant_User,
+        jwt: subscriberUserJwt,
+        iasToXsuaaTokenExchange: false
+      });
+      expect(actual).toEqual(destinationOne);
+    });
+
+    it('ignores cache if isolation requires user JWT but the JWT is not provided', async () => {
```
jjtang1985, line 373: maybe adding a case:
'ignores cache if isolation requires tenant info but the tenant info is missing in the jwt'

packages/connectivity/src/scp-cf/destination/destination-service-cache.ts
```diff
@@ -1,27 +1,25 @@
...
 const DestinationServiceCache = (cache: Cache<Destination[]>) => ({
   retrieveDestinationsFromCache: (
     targetUrl: string,
-    decodedJwt: JwtPayload,
-    isolationStrategy?: IsolationStrategy
+    decodedJwt: JwtPayload
   ): Destination[] | undefined =>
     cache.get(
-      getDestinationCacheKeyService(targetUrl, decodedJwt, isolationStrategy)
+      getDestinationCacheKey(decodedJwt, targetUrl, IsolationStrategy.Tenant)
```
jjtang1985, line 12: For this line and the line 20, I feel users lose some freedom for switching between tenant and tenant+user.
As the current know use cases do not need this flexibility, I'm happy to make it simple line now.

test-resources/test/test-util/xsuaa-service-mocks.ts
```diff
@@ -44,20 +44,23 @@ export function mockClientCredentialsGrantWithCertCall(
...
-  accessToken: string,
-  clientId: string
+  times: number,
+  accessTokenResponse: string,
+  accessTokenAssertion: string,
+  creds: ServiceCredentials
 ) {
   return nock(uri, {
-    reqheaders: xsuaaRequestHeaders({ authorization: `Bearer ${accessToken}` })
+    // reqheaders: xsuaaRequestHeaders({ authorization: `Bearer ${accessTokenResponse}` })
```
jjtang1985, line 53: left over?

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -33,6 +30,11 @@ type DestinationCircuitBreaker<ResponseType> = CircuitBreaker<
   AxiosResponse<ResponseType>
 >;
 
+type DestionsServiceOptions = ResilienceOptions &
```
jjtang1985, line 33:
```suggestion
type DestinationsServiceOptions = ResilienceOptions &
```

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -33,6 +30,11 @@ type DestinationCircuitBreaker<ResponseType> = CircuitBreaker<
...
+  Pick<DestinationFetchOptions, 'useCache'>;
+type DestionServiceOptions = ResilienceOptions &
```
jjtang1985, line 35:
```suggestion
type DestinationServiceOptions = ResilienceOptions &
```

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -164,29 +167,13 @@ export async function fetchDestination(
...
   tokens: AuthAndExchangeTokens,
-  options: DestinationFetchOptions
+  options: DestionServiceOptions
 ): Promise<Destination> {
   const targetUri = `${destinationServiceUri.replace(
     /\/$/,
     ''
   )}/destination-configuration/v1/destinations/${options.destinationName}`;
 
-  if (options?.useCache) {
```
jjtang1985, line 174: Maybe I miss something.
Could you please point out the potential tests about `useCache` changes.

Also, I see the description below, does the test also cover this? 

> If a IsolationStrategy is given to the getDestination() this enables the cache. No need to provided useCache: true

jjtang1985 (changes requested):
The implementation looks good.
Please consider my comments and adjust the changelog.

packages/connectivity/src/scp-cf/destination/destination-cache.ts
```diff
@@ -10,6 +10,15 @@ const logger = createLogger({
   messageContext: 'destination-cache'
 });
 
+/**
+ * Enumerator that selects the isolation type of destination in cache.
+ * The used isolation strategy is either `Tenant` or `Tenant_User` because we want to get results for subaccount and provider tenants which rules out no-isolation or user isolation.
+ */
+export enum IsolationStrategy {
```
marikaner, line 17: @MatKuhr FYI

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -578,3 +578,11 @@ Possible alternatives for such technical user authentication are BasicAuthentica
     );
   }
 }
+
+function getIsolationStrategy(jwt: JwtPayload | undefined): IsolationStrategy {
```
marikaner, line 582: Maybe:
```suggestion
function getDefaultIsolationStrategy(jwt: JwtPayload | undefined): IsolationStrategy {
```

packages/connectivity/src/scp-cf/destination/destination-service-cache.ts
marikaner, line 12: I am also a bit surprised to see this. Wouldn't it be more "secure" if users could specify tenant_user here?

Also, if we leave it as is - doesn't that make the third parameter of `getDestinationCacheKey` obsolete?

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -48,7 +50,7 @@ let circuitBreaker: DestinationCircuitBreaker<
 export function fetchInstanceDestinations(
   destinationServiceUri: string,
   jwt: string,
-  options?: DestinationOptions
+  options?: DestionsServiceOptions
```
marikaner, line 53: There is still a typo here.

marikaner (commented):
My 2 cents

packages/connectivity/src/scp-cf/destination/destination-cache.ts
MatKuhr: Greetings from the Java world where we have: [ResilienceIsolationMode](https://github.wdf.sap.corp/MA/sdk/blob/develop/cloudplatform/resilience/src/main/java/com/sap/cloud/sdk/cloudplatform/resilience/ResilienceIsolationMode.java#L11) which is used as documented [here](https://sap.github.io/cloud-sdk/docs/java/features/resilience#multi-tenancy).

However, this is not directly what customers can configure on destination retrieval. There customers have limited options on choosing on which tenant a request is made. See the [docs on that](https://sap.github.io/cloud-sdk/docs/java/features/connectivity/sdk-connectivity-destination-service#retrieval-strategy-options).


But I think our situation is somewhat different since we pull the tenant/user from some static context information (ThreadContext) which is [automagically created](https://sap.github.io/cloud-sdk/docs/java/features/multi-tenancy/multi-tenancy-thread-context#how-is-a-thread-context-created) by the SDK.

packages/connectivity/src/scp-cf/destination/destination-service.ts
FrankEssenberger: The tests are under `caching options` in the `destination-cache.spec.ts`

packages/connectivity/src/scp-cf/destination/destination-service-cache.ts
FrankEssenberger: The cache is for the destinations only so the call where you get multiple destinations for the a given tenant. For this Tenant is the only meaningful isloation, because it does not depend on user. I wanted to use the destination cache so I passed the argument.

MatKuhr: Are you aware of https://github.com/SAP/cloud-sdk/issues/496 WRT user isolated caching? 

Or in general, is your cache upper-bounded already by the JWT lifetime (or any other authentication expiration time)?

FrankEssenberger: If the destination contains a token with some lifetime (expiresIn) the cache entry should go away the [token is expired](https://github.com/SAP/cloud-sdk-js/blob/8366f5d36d0de0edac0a15ab70b0ff02178d8635/packages/connectivity/src/scp-cf/destination/destination-cache.ts#L109) or after the default expiration time.

MatKuhr: ok great, in Java we currently don't have that (yet) unfortunately

marikaner (approved):

jjtang1985 (approved):
LGTM

PR #2018, opened 2022-01-25
chore: Remove references to internal from odata packages

marikaner:
Remove references to internal from the odata specific packages.

Closes SAP/cloud-sdk-backlog#535.


packages/odata-common/README.md
```diff
@@ -3,7 +3,7 @@
 <a href="https://sap.com/s4sdk"><img src="https://help.sap.com/doc/2324e9c3b28748a4ae2ad08166d77675/1.0/en-US/logo-with-js.svg" alt="SAP Cloud SDK for JavaScript Logo" height="122.92" width="226.773"/></a>
 <!-- sap-cloud-sdk-logo-stop -->
 
-# @sap-cloud-sdk/odata-common/internal
+# @sap-cloud-sdk/odata-common
```
jjtang1985, line 6: 👍 
Should we add a note to the users, saying `odata-vX` should be used as a complete library instead of this internal package?

packages/odata-common/src/response-transformer.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * @internal
+ * Throws an error.
```
jjtang1985, line 2: 😃

packages/odata-common/src/selectable/field.ts
```diff
@@ -1,25 +1,7 @@
-import BigNumber from 'bignumber.js';
 import { EntityBase, Constructable } from '../entity-base';
-import { Time } from '../time';
-
-/**
- * Union type to represent all possible types of a field.
- * @internal
- */
-export type FieldType =
```
jjtang1985, line 9: Could you please explain more about this change and the corresponding other adjustment?

jjtang1985 (commented):
Looks good.
Perhaps you can add more details about the FieldType removal.

packages/odata-common/README.md
marikaner: Makes sense.

packages/odata-common/src/selectable/field.ts
marikaner: As we not have possibly custom (de-)serializers this types does not make sense anymore. It was already removed in most places, but forgotten here. I should have done this on a separate PR ;)

jjtang1985 (approved):
LGTM

jjtang1985:
@marikaner , please resolve the conflicts

FrankEssenberger (approved):
LGTM - I think the compiler would fail if the object in `common.ts` are not exported because the `internal` path is removed.

PR #2019, opened 2022-01-25
fix: v2 pipeline

jjtang1985:
The e2e-test is missing as prerequisites of v2 canary release.


deekshas8 (approved):
LGTM, I added it to build workflow and missed it here

PR #2025, opened 2022-01-27
test: using execa `6.0.0` in SDK without additional configuration changes

jjtang1985:
This PR shows, why the SDK cannot use execa 6.0.0, as an ESM for both TS and JS without any further configurations.

### DIFF
--- a/test-execa-js.js
+++ b/test-execa-js.js
@@ -1,0 +1,7 @@
+import { execa } from 'execa';
+
+async function main() {
+  const ls = await execa('ls');
+}
+
+main();
--- a/test-execa.js
+++ b/test-execa.js
@@ -1,0 +1,53 @@
+"use strict";
+var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
+    function adopt(value) { return value instanceof P ? value : new P(function (resolve) { resolve(value); }); }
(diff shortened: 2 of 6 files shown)


test-execa.js
```diff
@@ -0,0 +1,53 @@
...
+                    if (t[2]) _.ops.pop();
+                    _.trys.pop(); continue;
+            }
+            op = body.call(thisArg, _);
+        } catch (e) { op = [6, e]; y = 0; } finally { f = t = 0; }
+        if (op[0] & 5) throw op[1]; return { value: op[0] ? op[1] : void 0, done: true };
+    }
+};
+exports.__esModule = true;
+var execa_1 = require("execa");
```
jjtang1985, line 39: This is transpired from the ts file, while `require is not allowed`

test-execa-js.js
```diff
@@ -0,0 +1,7 @@
+import { execa } from 'execa';
```
jjtang1985, line 1: Using `import` with a js file is not working directly due to the error.

```
(node:62143) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
import { execa } from 'execa';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

PR #2026, opened 2022-01-27
feature: Add support for proxy type PrivateLink

tomfrenken:
This PR introduces support for proxy type PrivateLink.

Relates to SAP/cloud-sdk-backlog#450.

### DIFF
--- a/packages/connectivity/src/scp-cf/destination/destination-from-service.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-from-service.ts
@@ -416,6 +416,7 @@ Possible alternatives for such technical user authentication are BasicAuthentica
           this.subscriberToken?.userJwt
         );
       case ProxyStrategy.INTERNET_PROXY:
+      case ProxyStrategy.PRIVATELINK_PROXY:
         return addProxyConfigurationInternet(destination);
       case ProxyStrategy.NO_PROXY:
         return destination;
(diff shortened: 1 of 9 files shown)


packages/connectivity/src/scp-cf/destination/proxy-util.ts
```diff
@@ -29,6 +29,12 @@ export function proxyStrategy(destination: Destination): ProxyStrategy {
     );
     return ProxyStrategy.ON_PREMISE_PROXY;
   }
+
+  if (destination.proxyType === 'PrivateLink') {
+    logger.info('PrivateLink destination proxy settings will be used.');
```
FrankEssenberger, line 34: In Java SDK they also mention that this proxy type is not supported in local CI/CD enviroment. I would also mention this here.

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -374,10 +374,6 @@ Possible alternatives for such technical user authentication are BasicAuthentica
       authHeaderJwt: serviceJwt.encoded, // token to get destination from service
       exchangeHeaderJwt: this.subscriberToken.userJwt.encoded // token considered for user and tenant
     };
-
```
FrankEssenberger, line 377: why this was removed?

FrankEssenberger (changes requested):
Lookd good in general. Please answer the question I left.

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
tomfrenken: Good catch, that was by accident 😅

tomfrenken: Oh nvm, the reason this was removed, is because it can't be reached (as indicated by my IDE), since there is no continued for the return statement above.

FrankEssenberger: oh yes this is correct

FrankEssenberger (approved):

PR #2028, opened 2022-01-27
Fix public api check

marikaner:
- Fixes the public API check to fail when there are errors. 
- Adds support for referenced star exports.
- Disallows re-exports of internal API.

Prerequisite: #2018.
Closes SAP/cloud-sdk-backlog#535.


packages/generator/src/entity/namespace.ts
```diff
@@ -117,7 +123,6 @@ export function getPropertyFieldType(
 
   if (prop.isComplex) {
     return 'Comple';
```
deekshas8, line 125:
```suggestion
    return 'Complex';
```

deekshas8 (commented):
The new checks look good. Added two minor comments.

scripts/check-public-api.ts
```diff
@@ -247,28 +222,62 @@ export function parseBarrelFile(fileContent: string, regex: RegExp): string[] {
...
+        .join(', ')} exported in '${filePath}'.`
+    );
+  }
+}
+
+export async function parseIndexFile(filePath: string): Promise<string[]> {
+  const cwd = dirname(filePath);
+  const fileContent = await readFile(filePath, 'utf-8');
+  checkInternalReExports(fileContent, filePath);
+  const localExports = parseBarrelFile(fileContent, /\{([\w,]+)\}from'\./g);
```
deekshas8, line 243: I believe the regex here can be assigned to const `regexExportedIndex` declared above and the variable used here instead

scripts/check-public-api.ts
```diff
@@ -8,12 +8,16 @@ import {
...
   transpileDirectory
 } from '@sap-cloud-sdk/generator-common/internal';
+
 const { readFile, lstat, readdir } = promises;
+
 const logger = createLogger('check-public-api');
+
 const pathToTsConfigRoot = join(__dirname, '../tsconfig.json');
 const pathRootNodeModules = resolve(__dirname, '../node_modules');
 export const regexExportedIndex = /\{([\w,]+)\}/g;
```
deekshas8, line 18: Perhaps:  `export const regexExportedIndex = /\{([\w,]+)\}from'\./g;`
Sorry if my comment wasn't clear before. The regex `/\{([\w,]+)\}from'\./g` itself was correct as it ignores the public re-exports(from '@test-package/') and only considers from './some-file'.

Just wanted to used this variable since its used in the tests also.

marikaner:
I wonder why this is still failing all of the comments the checks complain about are marked internal. And it doesn't fail on the 2.0 branch.

scripts/check-public-api.ts
marikaner: Yes, this wasn't on purpose. Should be ok now.

deekshas8 (approved):
LGTM.

PR #2041, opened 2022-01-28
chore: Use xssec also for validation

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#479.

Replace the custom verification key implementation with the built-in one by xssec.


packages/connectivity/src/scp-cf/jwt.spec.ts
```diff
@@ -82,17 +82,15 @@ describe('jwt', () => {
...
 
     it('succeeds and decodes for correct key', async () => {
-      nock(jku).get('/').reply(200, responseWithPublicKey());
-
+      nock(jku).get('/').times(999).reply(200, responseWithPublicKey());
       await expect(
         verifyJwt(signedJwtForVerification(jwtPayload, jku))
       ).resolves.toEqual(jwtPayload);
-    });
+    }, 600000);
```
florian-richter, line 93: Is this intended to stay? Running it 999 times may slow down tests a bit much, is it not?

florian-richter (commented):
@FrankEssenberger could you give some more context. Since there are no release notes attached, is this a change that does not change user interaction? Anything reviewers should check specifically?

FrankEssenberger:
> @FrankEssenberger could you give some more context. Since there are no release notes attached, is this a change that does not change user interaction? Anything reviewers should check specifically?

I added also @jjtang1985 since he wrote most of the caching. What should be ensure is the behavior, that per default the keys are cached. Also that it fails if it can not be verified should be behave the same. The error messages will differ since the `xssev` throws the errors now.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 ## Improvements
 
 - [connectivity] Add details to error message for missing "URL" properties in destinations.
+- [connectivity] Use the cache of `xssec` library to cache verification certificates.
```
jjtang1985, line 26: I guess you need to merge the main branch, as other changelog items were released.

packages/connectivity/src/scp-cf/jwt.spec.ts
```diff
@@ -107,13 +105,15 @@ describe('jwt', () => {
...
+        verifyJwt(signedJwtForVerification(jwtPayload, jku), {
+          cacheVerificationKeys: false
+        })
       ).rejects.toMatchObject({
-        message: 'Failed to verify JWT. Could not retrieve verification key.',
+        message: 'Failed to verify JWT.',
         cause: {
           message:
-            'No verification keys have been returned by the XSUAA service.'
+            'Obtained token keys from UAA, but key with requested keyID "https://my-jku-url.authentication.sap.hana.ondemand.comkey-id-1" still not found in cache.'
```
jjtang1985, line 116: I know this comes from xssec, but if I disable the cacheVerificationKeys, and the error message still says the key is not found in cache, I would be confused.

jjtang1985 (changes requested):
Looks good to me in general.
I left some comments.

jjtang1985:
> I added also @jjtang1985 since he wrote most of the caching. What should be ensure is the behavior, that per default the keys are cached. Also that it fails if it can not be verified should be behave the same. The error messages will differ since the xssev throws the errors now.

Actually for verification key, @marikaner wrote most of the code. Please have a look, as this is a sensitive topic.

jjtang1985 (approved):

PR #2042, opened 2022-01-28
chore: change fetchCsrfToken default value to true

deekshas8:
Closes SAP/cloud-sdk-backlog#524.

Changed the default value of `fetchCsrfToken` to true for the generic http client for non-GET requests.( note: this was already true by default in the odata and openapi clients). The user can still override this by passing false in the `executeHttpRequest` method.

### DIFF
--- a/packages/http-client/src/http-client.spec.ts
+++ b/packages/http-client/src/http-client.spec.ts
@@ -689,6 +689,7 @@ sap-client:001`);
       );
     });
 
+    // Axios requests are not redirected via proxy
     xit('test axios proxy redirect', () => {
       const proxyConfiguration: ProxyConfiguration = {
         host: 'localhost',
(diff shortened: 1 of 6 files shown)


packages/http-client/src/http-client.spec.ts
```diff
@@ -678,6 +688,33 @@ sap-client:001`);
         })
       );
     });
+
+    xit('test axios proxy redirect', () => {
```
jjtang1985, line 692: I guess you might have created a follow up for this.
Could you please add some comments about the purpose of this disabled test and maybe link to the ticket?

jjtang1985 (changes requested):
LGTM.
Please also update the changelog.

packages/http-client/src/http-client.spec.ts
deekshas8: Added a comment to the test. Have created a [ticket](https://github.com/SAP/cloud-sdk-backlog/issues/560) and linked this test

jjtang1985 (approved):
LGTM

PR #2062, opened 2022-01-31
fix: use the cache to store destination

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#557

Include a forth destination origin: register destination. Which is not using the environment variables. We learned in sec val that the use of environment variables should be minimized if it could contain sensitive information like destinations.

The implementation uses a separate cache with infinite lifetime.


packages/connectivity/src/scp-cf/destination/destination-from-register.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+    logger.info(`Successfully retrieved destination '${options.destinationName}' from registered destinations.`);
+
+    return destination &&
+    proxyStrategy(destination) === ProxyStrategy.INTERNET_PROXY
+        ? addProxyConfigurationInternet(destination)
+        : destination;
+}
+
+function isolationStrategt(options?: RegisterDestinationOptions): IsolationStrategy{
```
jjtang1985: I guess we should document this:
- as api doc
- on the documentation portal

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
```diff
@@ -0,0 +1,104 @@
...
+export type DestinationWithName = Omit<Destination, 'name'> & { name: string };
+
+function destinationHasName(
+    destination: Destination
+): destination is DestinationWithName {
+    return !!destination.name;
+}
+
+/**
+ * Internal
```
jjtang1985, line 59:
```suggestion
 * @internal
```

packages/connectivity/src/scp-cf/destination/destination-from-register.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+import { IsolationStrategy } from '../cache';
+import {
+  DestinationWithName,
+  registerDestination,
+  registerDestinationCache,
+  searchRegisteredDestination
+} from './destination-from-register';
+import { getDestination } from './destination-accessor';
+// import {setTestDestination} from "@sap-cloud-sdk/test-util";
+import { Destination } from './destination-service-types';
```
jjtang1985, line 20:
```suggestion
import { Destination } from './destination-service-types';
```

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
```diff
@@ -0,0 +1,104 @@
...
+
+function isolationStrategt(options?: RegisterDestinationOptions): IsolationStrategy{
+    if(options?.isolationStrategy){
+        return options.isolationStrategy;
+    }
+    const hasUserId = options?.jwt && userId(decodeJwt(options.jwt));
+    return  hasUserId ? IsolationStrategy.Tenant_User : IsolationStrategy.Tenant;
+}
+
+function decodedJwt(options?: RegisterDestinationOptions): Record<string,any>{
```
jjtang1985: I feel this method or name is a bit confusing without checking implementation details, as we build a decoded jwt in line 80.

Now, I understand.
As `cacheRetrievedDestination` accepts `Record`, we build a partial jwt with the `zid` in line 80.
- We can also consider supporting e.g. {userId: string, tenantId: string} as a potential parameter for `cacheRetrievedDestination`. However, the original parameter is already `Record`, which needs type guard 😞 
- Maybe we can also have a type Zid {zid: string} and rename the function to `decoodedJwtOrZid`?

None of the the approach I feel better.
Perhaps adding some docs would be enough, saying we build an object containing `zid`, so it can be used for the cache key.

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
```diff
@@ -0,0 +1,104 @@
...
+    const providerTenantId = parseSubdomain(getXsuaaServiceCredentials(options?.jwt).url);
+
+    return options?.jwt ? decodeJwt(options.jwt) : { zid:providerTenantId };
+}
+
+function destinationAuthToken(
```
jjtang1985, line 105: This looks the same as the one in the destination-from-env?

packages/connectivity/src/scp-cf/destination/destination-from-register.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+    const warnSpy = jest.spyOn(logger, 'warn');
+    await getDestination({ destinationName: 'FORWARD-TOKEN-DESTINATION' });
+    expect(warnSpy).toHaveBeenCalledWith(
+      expect.stringMatching(
+        /Option 'forwardAuthToken' was set on destination but no token was provided to forward./
+      )
+    );
+  });
+
+  it('infos if  registered destination is retrieved.', async () => {
```
jjtang1985, line 158:
```suggestion
  it('infos if registered destination is retrieved.', async () => {
```

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -44,6 +45,7 @@ export async function getDestination(
 ): Promise<Destination | null> {
   return (
     searchEnvVariablesForDestination(options) ||
+    searchRegisteredDestination(options) ||
```
jjtang1985, line 48: Now, this list/order grows, which needs to be documented so the priority is clear when reaching conflics.

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
FrankEssenberger: ups, yes this should go away for env varialbes because these are for testing now.

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
FrankEssenberger: I added a test, and yes the docs need to be adjusted to include the new `registerDestination` method and also make the line clear when to use what.

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
deekshas8:
```suggestion
function isolationStrategy(options?: RegisterDestinationOptions): IsolationStrategy{
```

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
```diff
@@ -0,0 +1,140 @@
...
+  return hasUserId ? IsolationStrategy.Tenant_User : IsolationStrategy.Tenant;
+}
+
+/**
+ * This method either decodes the given JWT. If the JWT is not given it will use the subdomain if the XSUAA and create a Object with zid this subdomain.
+ * This is then passed on to build the cache key.
+ * @param options - Options passed to register the destination containing the jwt.
+ * @returns The decoded JWT or a dummy JWT containing the tenant identifier (zid)
+ */
+function decoodedJwtOrZid(
```
deekshas8, line 109:
```suggestion
function decodedJwtOrZid(
```

packages/connectivity/src/scp-cf/destination/destination-from-register.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+  beforeAll(() => {
+    mockServiceBindings();
+  });
+
+  afterEach(() => {
+    registerDestinationCache.clear();
+    unmockDestinationsEnv();
+  });
+
+  it('registers destiantion and retrieves it', async () => {
```
tomfrenken, line 41:
```suggestion
  it('registers destination and retrieves it', async () => {
```

packages/connectivity/src/scp-cf/destination/destination-from-register.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+    // parseDestination({}as any)
+    registerDestination(testDestination, { jwt: providerServiceToken });
+    const actual = await getDestination({
+      destinationName: testDestination.name,
+      jwt: providerServiceToken
+    });
+    expect(actual).toEqual(testDestination);
+  });
+
+  it('retunrs undefined if destination key is not found', async () => {
```
tomfrenken, line 59:
```suggestion
  it('returns undefined if destination key is not found', async () => {
```

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -21,7 +22,7 @@ import {
  * @returns A promise resolving to the requested destination on success.
  */
 export async function useOrFetchDestination(
-  destination: DestinationOrFetchOptions
+  destination: Destination | DestinationFetchOptions
```
marikaner, line 25: Shouldn't this use the new `DestinationOrFetchOptions` type?

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
marikaner: General question: I think this makes sense for explicitness, but wouldn't it have sufficed to `getDestinationFromDestinationService`?

packages/connectivity/src/scp-cf/destination/destination-from-env.spec.ts
```diff
@@ -65,39 +58,8 @@ describe('env-destination-accessor', () => {
...
-      expect(warnSpy).toHaveBeenCalledWith(
-        expect.stringMatching(
-          /Option 'forwardAuthToken' was set on destination but no token was provided to forward./
-        )
-      );
-    });
-
-    it('warns if destination are read from enviorment and forwardAuthToken is not enabled.', async () => {
+  describe('getDestinationsFromEnv()', () => {
+    it('infos if destination are read from enviorment and forwardAuthToken is not enabled.', async () => {
```
marikaner, line 62:
```suggestion
    it('informs if destination are read from environment and forwardAuthToken is not enabled', async () => {
```

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
```diff
@@ -0,0 +1,140 @@
+import { createLogger } from '@sap-cloud-sdk/util';
```
marikaner, line 1: Minor: The file name "from register" is a bit weird. I am thinking of a cash register, when reading that. Consider "from registration" - but we can also change that later - maybe "from" is not a good idea for this.

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
```diff
@@ -0,0 +1,140 @@
...
+ * @param destination - A destination to add to the `destinations` cache
+ * @param options - Options how to cache the destination
+ */
+export function registerDestination(
+  destination: DestinationWithName,
+  options?: RegisterDestinationOptions
+): void {
+  if (!destination.name || !destination.url) {
+    throw Error(
+      'The registerDestination function requires a destination name and url.'
```
marikaner, line 45: I would make this less technical -
```suggestion
      'Registering destinations requires a destination name and url.'
```

packages/connectivity/src/scp-cf/destination/destination-cache.ts
```diff
@@ -10,7 +10,29 @@ const logger = createLogger({
   messageContext: 'destination-cache'
 });
 
-const DestinationCache = (cache: Cache<Destination>) => ({
+export interface DestinationCacheType {
```
marikaner, line 13: General: I think we should make a class out of this. Duplicating type and implementation doesn't make much sense to me. But we can do that later, I guess.

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  registerDestinationCache.cacheRetrievedDestination(
+    decodedJwtOrZid(options),
+    destination,
+    isolationStrategy(options)
+  );
+}
+
+export type DestinationWithName = Omit<Destination, 'name'> & { name: string };
```
marikaner, line 56: Seems weird. If this solves some technical issue I think we should add a comment.

marikaner (commented):
I was surprised to see this in a separate cache, but I think that this is actually cleaner. However, I was wondering - how much duplicate code did that introduce? (I am thinking of forwardAuthToken handling e.g.).

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
FrankEssenberger: no becaue the registeredDestination uses a different cache without expiration at the moment. Since the current implementation did not allow for entry specific life time

FrankEssenberger: Perhaps this is not true and you could set it with undefined and it will live longer than the default value but than I would say this is a bug

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
FrankEssenberger: The idea is to enforce the name to be present - is there a type for that ?

packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  registerDestinationCache.cacheRetrievedDestination(
+    decodedJwtOrZid(options),
+    destination,
+    isolationStrategy(options)
+  );
+}
+
+export type DestinationWithName = Omit<Destination, 'name'> & { name: string };
```
jjtang1985, line 56:
```suggestion
export type DestinationWithName = Destination & { name: string };
```
This should make it required.

jjtang1985 (approved):
LGTM
I left some a minor suggestion.
Please fix the merge conflicts

packages/connectivity/src/scp-cf/destination/destination-from-register.ts
marikaner: Hmm. maybe something with `Require`...

PR #2079, opened 2022-02-01
chore: Add support for dual modules

marikaner:
This is my initial state for dual modules.
With this, our packages are still transpiled as cjs, but wrapped to mjs (using gen-esm-wrapper), so that users can already transition to esm.
The entrypoints index + internal are then exposed through "conditional exports" (exports in package.json).

What I did not check yet, is whether packaging still works or whether we have to add some more files to the exports.

Closes SAP/cloud-sdk-backlog#547.


packages/connectivity/package.json
```diff
@@ -15,8 +15,19 @@
   "publishConfig": {
     "access": "public"
   },
+  "exports": {
```
jjtang1985, line 18: As a breaking change candidate, please adjust the compatibility notes.

jjtang1985 (approved):
Please consider the changelog and let's test it e2e.

jjtang1985:
@marikaner I'll merge it so I can make a test. The missing changelog item can be added later.

PR #2080, opened 2022-02-01
fix: re-export destination selection strategy

jjtang1985:
Based on the feedback from CAP, the `DestinationSelectionStrategy` needs to be exported.


florian-richter (approved):
LGTM

PR #2081, opened 2022-02-01
chore: Add blackduck scans to our pipeline

tomfrenken:
This PR adds blackduck to our pipeline, check out [my fork](https://github.com/tomfrenken/cloud-sdk-js/actions) to see how it works.

In Blackduck, each version is determined by the module version in Lerna and the github_ref_name, aka. the branch name.
So the main branch will look like `1.57.2-main` while branches may look like `1.57.2-my-branch`.

Additionally, it doesn't scan devDependencies.

Closes SAP/cloud-sdk-backlog#549.


tomfrenken:
I guess the PR number is just as sufficient, so no problem imo.

jjtang1985 (commented):
Have you created some negative tests so we know the PRs can be blocked when positive findings are detected?

.github/workflows/blackduck.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          path: |
+            node_modules
+            */*/node_modules
+          key: ${{ matrix.os }}-${{ hashFiles('**/yarn.lock') }}
+      - run: yarn install --frozen-lockfile
+      - run: sudo apt-get install jq
+      - run: echo "project_version=$(lerna ls -l --json | jq '.[0].version' | tr -d '"')-${GITHUB_REF_NAME}" >> $GITHUB_ENV
+      - name: Blackduck Scan
+        uses: SAP/project-piper-action@master
```
florian-richter, line 33: I opened an issue in the piper repo to ask them to rename their default branch so we can avoid the "In Solidarity / Inclusive Language" finding for this line.

Feel free to upvote 😄 https://github.com/SAP/project-piper-action/issues/119

florian-richter: FYI @fwilhe recommended to use tag 1.0 or branch v1 as a workaround until the default branch is renamed.

tomfrenken: Is the delta between master and v1 small enough that it doesn't matter that much?

tomfrenken:
> Have you created some negative tests so we know the PRs can be blocked when positive findings are detected?

Before I excluded the dev dependencies the pipeline on my fork was failing, I guess I could try out another productive dependency to make sure it also fails on regular dependencies.

.github/workflows/blackduck.yml
florian-richter: `releases/v1` is 1 commit ahead and 68 commits behind. Last commit is 22.9.2020 and on `master` it's 19.7.2021... so I can't say for sure...

fwilhe: > Is the delta between master and v1 small enough that it doesn't matter that much?

As far as I can tell (cf commit history) the changes in `master` only relate to dependency updates and rewrite to use the library that was newly created. There might have been a breaking change in the config format in that "use the lib" rewrite, I'm not sure about that, but functionally `v1` should be equivalent.

I'd be happy to help with adaptations if required, but I lost commit access to the repo so I can't just push a `main` branch myself anymore.

fwilhe: Another option of course would be to reference a commit id in your workflow which anyway probably is a good idea to ensure the version you use does not change unnoticed.

tomfrenken: Alright, I've switched to the commit, that should resolve it 👍

tomfrenken:
While testing blackduck on my fork, I've had 3 findings:

1. After adding a vulnerable dependency, which had a severity of `high` according to dependabot alerts, was not deemed a security risk in blackduck
2. After adding my own test package with a proprietary license, it wasn't always detected by blackduck, so somewhat flaky
3. Blackduck sometimes had performance issues, roughly 3-4/66 so about 5% were stuck.

.github/workflows/blackduck.yml
```diff
@@ -0,0 +1,45 @@
+name: blackduck-scan
+
+on:
+  pull_request: ~
+  push:
+    branches: [2.0]
```
florian-richter, line 6: Should we add a comment or follow up ticket to change this to `main` once we merge 2.0 into it?

florian-richter (approved):
Looks good!

.github/workflows/blackduck.yml
tomfrenken: Yeah good point, perhaps we should add it to whatever ticket we have to switch the default branch back to main.

PR #2092, opened 2022-02-02
chore:remove mjs config

jjtang1985:
This PR 
- keeps the hybrid mode introduced here: #2079 
- removes the `.mjs` related config

so it still have the flexibility to add `.mjs` support (non-breaking) in the future with minimal risks/changes.

The rest of the config should still be considered as breaking, that's why having it before 2.0 is necessary (or wait until 3.0)
```
  "exports": {
    ".": {
      "require": "./dist/index.js"
    },
    "./internal": {
      "require": "./dist/internal.js"
    }
  },
```

### Test coverage
- PoC local unit tests
- Internal vdm e2e tests

### Reference doc
- https://nodejs.org/api/packages.html#package-entry-points

### DIFF
--- a/packages/connectivity/package.json
+++ b/packages/connectivity/package.json
@@ -17,11 +17,9 @@
   },
   "exports": {
     ".": {
-      "import": "./dist/index.mjs",
       "require": "./dist/index.js"
     },
     "./internal": {
-      "import": "./dist/internal.mjs",
       "require": "./dist/internal.js"
     }
   },
--- a/packages/generator-common/package.json
+++ b/packages/generator-common/package.json
@@ -17,11 +17,9 @@
   },
   "exports": {
     ".": {
-      "import": "./dist/index.mjs",
       "require": "./dist/index.js"
     },
     "./internal": {
-      "import": "./dist/internal.mjs",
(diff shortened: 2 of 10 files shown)


marikaner (approved):
I think we can also remove:
```json
   "internal.js",
   "internal.d.ts"
```

from the package.json

FrankEssenberger:
Could you explain how this keeps the hybrid mode. I see that the wrap-esm commands are still there, but when you remove the exports, what is still visible for the user?

marikaner:
> Could you explain how this keeps the hybrid mode. I see that the wrap-esm commands are still there, but when you remove the exports, what is still visible for the user?

This does not keep the dual mode. But good point, if we remove the .mjs files, we should also remove the wrap commands.

FrankEssenberger:
Then I find the PR description by junjie mislieading.

artemkovalyov (changes requested):
We keep CJS for at least the next 6 months. Reasons:
- no immediate demand
- lack of tests
- last minutes issue
- not communicated to customers beforehand

Let's plan it accordingly for 3.0 and announce it around March or when the scope for 3.0 is ready.

PR #2093, opened 2022-02-02
Generator fixes

marikaner:
Add dependencies to generator + recommendation if the odata package was not installed.
The idea is: This will work for the local generator, but if someone is using a globally installed generator, they must install the odata packages.

I am not too happy about the tests, but I think this is a general problem in the odata generator...

Closes SAP/cloud-sdk-backlog#562.

### DIFF
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -98,6 +98,8 @@ export async function generate(options: GeneratorOptions): Promise<void> {
 
 /**
  * @internal
+ * @param projectAndServices - Generated project with services.
+ * @returns An error message with a recommendation to install specific SDK packages.
  */
 export function getInstallODataErrorMessage(
   projectAndServices: ProjectAndServices
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -1,6 +1,10 @@
 import { PathLike } from 'fs';
 import { resolve, dirname } from 'path';
-import { createLogger, splitInChunks } from '@sap-cloud-sdk/util';
+import {
+  createLogger,
+  ErrorWithCause,
+  splitInChunks
+} from '@sap-cloud-sdk/util';
 import { emptyDirSync } from 'fs-extra';
 import {
   Directory,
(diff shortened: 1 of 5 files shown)


packages/generator/src/generator.ts
```diff
@@ -75,12 +79,44 @@ export async function generate(options: GeneratorOptions): Promise<void> {
...
-      (all, chunk) => all.then(() => transpileDirectories(chunk)),
-      Promise.resolve()
-    );
+    try {
+      await chunks.reduce(
+        (all, chunk) => all.then(() => transpileDirectories(chunk)),
+        Promise.resolve()
+      );
+    } catch (err) {
+      if (err.message?.includes('error TS2307')) {
```
FrankEssenberger, line 88: I would like to have a hidden boolean option `failOnDependency` to enable or disable this. In our api hub pipeline it is convinient to see a failing transpilation. Also for other E2E test this could become in handy

packages/generator/src/generator.ts
```diff
@@ -75,12 +79,44 @@ export async function generate(options: GeneratorOptions): Promise<void> {
...
+): string {
+  const hasV2 = projectAndServices.services.some(
+    service => service.oDataVersion === 'v2'
+  );
+  const hasV4 = projectAndServices.services.some(
+    service => service.oDataVersion === 'v4'
+  );
+
+  if (hasV2 && hasV4) {
+    return 'Did you forget to install "@sap-cloud-sdk/odata-v2" and "@sap-cloud-sdk/odata-v4"?';
```
FrankEssenberger, line 113: Does the installation of `odata-xy` also lead to installation of `connectivity` package?

FrankEssenberger (changes requested):
See my comment about a parameter

packages/generator/src/generator.ts
marikaner: This will fail either way. It just adds another log message.

packages/generator/src/generator.ts
marikaner: Si, claro.

packages/generator/src/service/package-json.ts
```diff
@@ -46,15 +46,15 @@ export async function packageJson(
             : {})
         },
         dependencies: {
-          '@sap-cloud-sdk/odata-common/internal': `^${await getSdkVersion()}`,
+          '@sap-cloud-sdk/odata-common': `^${await getSdkVersion()}`,
```
jjtang1985, line 49: Here is a potential candidate of missing test coverage.

jjtang1985 (approved):
LGTM.
I left one comment about test coverage.

packages/generator/src/service/package-json.ts
marikaner: Done!

jjtang1985 (approved):

FrankEssenberger (approved):

PR #2095, opened 2022-02-02
fix: Fix generator batch request imports when there are no entities

marikaner:
The generator currently fails on service generation when there are no entities in the service.
This PR adds them only if necessary.


FrankEssenberger (approved):
LGTM

jjtang1985 (approved):
I searched for the `without-ts-morph` folder, batch is now covered.

PR #2101, opened 2022-02-03
chore: fix linter warning

FrankEssenberger:
There were a few linter warnings - this PR fixes it.


tomfrenken (approved):
LGTM

tomfrenken:
Some snapshot unit tests seem to fail because of the new 2.0 release, you could probably fix them with this PR as well since they seem to be rather easy to fix.

PR #2107, opened 2022-02-03
fix: changelog after 2.0

jjtang1985:
Our release scripts relies on the `CHANGELOG.md`.
This PR:
- renames the old `CHANGELOG.md` to `CHANGELOG-v1.md`, which might not be necessary, as it's anyway on the current `main` branch
- renames the `CHANGELOG-2.md` to `CHANGELOG.md`, and also follows the same format of the old `CHANGELOG.md`.


FrankEssenberger (approved):
Perhaps use `git mv` next time so that the transition is clearer.

PR #2120, opened 2022-02-07
chore: fix todo 1584

FrankEssenberger:
Removed the if statement after testing all S/4 Cloud and OP oData services with the latest version of the generator.


marikaner (approved):
LGTM

PR #2141, opened 2022-02-08
chore: move long running tests to e2e tests

tomfrenken:
This PR moves our `generator-cli` tests to the `e2e` package, as they are long running.

Before moving them I tried to use both `require(module)(params)` syntax as well as`mock-fs`.

`require` had the problem that `jest` was always the first parameter, which doesn't work with the cli.
`mock-fs` had the problem that I needed to mock pretty much everything, for example, the `package.json` in `generator-common` would need to be mocked, as well as the `odata-v2` package.

Closes SAP/cloud-sdk-backlog#480.


jjtang1985 (approved):
LGTM.
Actually, for the time being, the e2e test pipeline is the quickest one.

PR #2142, opened 2022-02-08
chore: Implement Blackduck retry logic

tomfrenken:
This PR implements a very simple retry logic for our Blackduck scans, as it seems we hit a performance limit with dependabot's PRs, because they are all opened simultaneously.

This should hopefully be fixed with a simple retry.


jjtang1985 (commented):
I feel this will not work. All the failing pipeline will start together, after 1 min.

tomfrenken:
> I feel this will not work. All the failing pipeline will start together, after 1 min.

Yes, however I think not all pipelines fail, only a fraction, and that fraction will then restart together after 1 minute

tomfrenken:
This could potentially still hit the performance threshold though I guess 🤔

jjtang1985 (approved):
Let's merge and test it, so we have some statistics.

tomfrenken:
Related to SAP/cloud-sdk-backlog#574.

PR #2153, opened 2022-02-09
fix: 2.0 api doc

jjtang1985:
- Links to new packages of 2.0 were missing, which were fixed.
- Some 1.x.y docs were only pushed to the 1.0 branch and the doc repo, but missing the 2.0 branch. This is also fixed.


FrankEssenberger (approved):
LGTM - I see some 1.X.Y docwhich was missing on the main branch as explained by Junjie and also in the index there is some tsdoc now.

PR #2154, opened 2022-02-09
chore: todo remove includeFactoryName function

deekshas8:
Closes SAP/cloud-sdk-backlog#521.

This PR removes the `includeFactoryName` method and the `factoryName` property from the `VdmComplexType`

### DIFF
--- a/packages/generator/src/edmx-to-vdm/v2/service-entities.ts
+++ b/packages/generator/src/edmx-to-vdm/v2/service-entities.ts
@@ -1,7 +1,6 @@
 import { ServiceMetadata } from '../../edmx-parser';
 import { VdmServiceEntities } from '../../vdm-types';
 import { ServiceNameFormatter } from '../../service-name-formatter';
-import { includeFactoryName } from '../common';
 import { generateFunctionImportsV2 } from './function-import';
 import { generateComplexTypesV2 } from './complex-type';
 import { generateEntitiesV2 } from './entity';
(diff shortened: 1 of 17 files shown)


packages/generator/src/edmx-to-vdm/v2/service-entities.ts
```diff
@@ -27,7 +26,7 @@ export function getServiceEntitiesV2(
   );
 
   return {
-    complexTypes: includeFactoryName(complexTypes, formatter),
+    complexTypes,
```
FrankEssenberger, line 29: I am wondering why none of the generated code changes. It looks like this method was called and it should add some factoryName property. Was this not tested in the past?

deekshas8: There were tests in the [service-generator.spec.ts](https://github.com/SAP/cloud-sdk-js/pull/2154/files/2a6797ee378a0379ccf042916bae2d312fb6e880#diff-20c243404ffec77cc26f8969746f95b9037a181db856dc6fae1d6fcbac9dc845) to check that property `factoryName` has value. 

But the property wasn't used anywhere and explicitly omitted `Omit<VdmComplexType, 'factoryName'>[]`. 
The actual factory generation code might have been removed before (as we don't have `createComplexType_1 `class/function in the resultant client anymore) and this was left behind.

PR #2158, opened 2022-02-10
feat: Add sanitizeRecord function and sanitize http-client debug log

florian-richter:
To prevent accidental leaking of sensitive data, I introduced the `sanitizeRecord` function which is part of the `cloud-sdk-logger`. It searches keys for a number of potentially sensitive keys and replaces the values with a configurable placeholder.

Closes SAP/cloud-sdk-backlog#558.

### DIFF
--- a/packages/util/src/logger/cloud-sdk-logger.spec.ts
+++ b/packages/util/src/logger/cloud-sdk-logger.spec.ts
@@ -6,6 +6,7 @@ import {
   getGlobalLogLevel,
   getLogger,
   muteLoggers,
+  sanitizeRecord,
   setGlobalLogLevel,
   setLogLevel,
   unmuteLoggers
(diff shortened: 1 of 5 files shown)


packages/util/src/logger/cloud-sdk-logger.ts
```diff
@@ -181,6 +181,57 @@ export function getGlobalLogLevel(): string | undefined {
...
+export function sanitizeRecord<T = any>(
+  input: Record<string, T>,
+  replacementString = '<DATA NOT LOGGED TO PREVENT LEAKING SENSITIVE DATA>',
+  sensitiveKeys: string[] = defaultSensitiveKeys
+): Record<string, T> {
+  const normalizedKeys = sensitiveKeys.map(key => key.toLowerCase());
+  return Object.fromEntries(
+    Object.entries(input).map(([inputKey, value]) => {
+      if (
+        normalizedKeys.some(normalizedKey =>
```
FrankEssenberger, line 215: isn't there a more simple way to check `inputKey` is in the array `normalizedKeys`. They are all string so could you use something like: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

packages/util/src/logger/cloud-sdk-logger.spec.ts
```diff
@@ -240,6 +241,125 @@ describe('Cloud SDK Logger', () => {
...
+        cookie: 'csrfToken=secret'
+      };
+      const output = {
+        access_token: '<DATA NOT LOGGED TO PREVENT LEAKING SENSITIVE DATA>',
+        my_access_token: '<DATA NOT LOGGED TO PREVENT LEAKING SENSITIVE DATA>',
+        authentication: '<DATA NOT LOGGED TO PREVENT LEAKING SENSITIVE DATA>',
+        service_authentication:
+          '<DATA NOT LOGGED TO PREVENT LEAKING SENSITIVE DATA>',
+        authorization: '<DATA NOT LOGGED TO PREVENT LEAKING SENSITIVE DATA>',
+        Authorization: '<DATA NOT LOGGED TO PREVENT LEAKING SENSITIVE DATA>',
```
FrankEssenberger, line 323: Add an exception for the "credentials". Perhaps you can have a look if codeql has a value for white listed passwords or a way to add these value to our run.

FrankEssenberger (approved):
LGTM - only some minor question and potential refactoring idea. Besides that function look good and tests are also covering the relevant cases and functionality.

packages/util/src/logger/cloud-sdk-logger.ts
florian-richter: includes would only check if the exact same string is included. This implementation checks if any recordKey has a substring that matches the sensistiveKeys

FrankEssenberger: ah ok.

packages/util/src/logger/cloud-sdk-logger.spec.ts
FrankEssenberger: In the discussion with the GHAS people they gave arguments to run the scan also on test code. So I guess we have to manually exclude the false positive then.

Issue #2159, opened 2022-02-11
How to solve "Property 'URL' of destination configuration must not be undefined."


marikaner:
A user got the following error, while their destination had a URL on it:

`Error during request to remote service: Failed to fetch subaccount destinations. Caused by: Property 'URL' of destination configuration must not be undefined.`

marikaner:
I asked the user to check their destinations (get a service token, then get all destinations) and check if there is any destination, that does not have a URL.
The SAP Cloud SDK verifies all destinations, even if they are not referenced by your code.

PR #2160, opened 2022-02-11
chore: Improve error message on missing URL

marikaner:
This improves the error message for missing URLs on destinations. It is confusing that this checks other destinations as well, therefore to understand that this is not necessarily about the given destination, I added the name of the destination (if existent).
Closes SAP/cloud-sdk-backlog#584 + #2159.


FrankEssenberger (approved):
LGTM. I just mention this issue: https://github.com/SAP/cloud-sdk-backlog/issues/195 because this would remove the fetching of all the subaccount and instance destinations.

Idea: What if we would not throw errors if an other then the requested destination is broken? Just an idea.

marikaner:
> LGTM. I just mention this issue: [SAP/cloud-sdk-backlog#195](https://github.com/SAP/cloud-sdk-backlog/issues/195) because this would remove the fetching of all the subaccount and instance destinations.
> 
> Idea: What if we would not throw errors if an other then the requested destination is broken? Just an idea.

I like that idea. I will merge this PR now, but should we have a follow up right now or the ticket you mentioned?

PR #2174, opened 2022-02-14
fix: Blackduck performance threshold

tomfrenken:
This PR switches our Blackduck scans from a per-branch scan to nightly scans, to avoid hitting the performance threshold.

Closes SAP/cloud-sdk-backlog#574.

### DIFF
--- a/.github/workflows/blackduck.yml
+++ b/.github/workflows/blackduck.yml
@@ -42,21 +42,3 @@ jobs:
         env:
           PIPER_token: ${{ secrets.BLACKDUCK_TOKEN }}
           PROJECT_VERSION: ${{ env.project_version }}
-      - name: Sleep on retry
-        if: ${{ failure() }}
-        run: sleep 60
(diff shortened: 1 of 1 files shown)


.github/workflows/blackduck.yml
```diff
@@ -1,9 +1,8 @@
 name: blackduck-scan
 
 on:
-  pull_request: ~
-  push:
-    branches: ['main']
+  schedule:
```
jjtang1985, line 4: Maybe we can add a manual trigger so we can make a scan before a release.

jjtang1985: You can find the trigger in the release pipeline.

tomfrenken: Ah good idea.

tomfrenken: ✅

florian-richter (approved):
LGTM

PR #2175, opened 2022-02-14
fix: fix url encoding in getByKey

deekshas8:
Closes SAP/cloud-sdk-backlog#548.

### DIFF
--- a/packages/odata-common/src/de-serializers/uri-value-converter.ts
+++ b/packages/odata-common/src/de-serializers/uri-value-converter.ts
@@ -14,6 +14,7 @@ export function isInfOrNan(value: string | number): boolean {
 }
 
 /**
+ * As per OData URL Syntax rules, single quotes within string literals must be represented as two consecutive single quotes.
  * @internal
  */
 export function convertToUriForEdmString(value: any): string {
--- a/packages/odata-common/src/de-serializers/uri-value-converter.ts
+++ b/packages/odata-common/src/de-serializers/uri-value-converter.ts
@@ -15,6 +15,7 @@ export function isInfOrNan(value: string | number): boolean {
 
 /**
  * As per OData URL Syntax rules, single quotes within string literals must be represented as two consecutive single quotes.
+ * encodeURIComponent does not encode single quotes on the high level get-filter and get-resource-path.
  * @internal
  */
 export function convertToUriForEdmString(value: any): string {
--- a/packages/odata-common/src/uri-conversion/get-resource-path.ts
+++ b/packages/odata-common/src/uri-conversion/get-resource-path.ts
@@ -102,15 +102,6 @@ export function createGetResourcePathForKeys(
     return keys;
   }
 
-  /**
-   * Returns an entities keys as key-value pairs.
-   * encodeURIComponent does not escape single quote character.
-   * If value contains ', the uriConverter method encodes the single quote.
-   * @param key - Name of a key property of an entity.
-   * @param value - Value of the given key.
-   * @param schema - Schema of an entity.
-   * @returns Key and encoded value as formatted string.
-   */
   function keyToOData(
     key: string,
     value: any,
--- a/packages/odata-common/src/uri-conversion/get-resource-path.ts
+++ b/packages/odata-common/src/uri-conversion/get-resource-path.ts
@@ -102,15 +102,15 @@ export function createGetResourcePathForKeys(
     return keys;
   }
 
-/**
- * Returns an entities keys as key-value pairs.
- * encodeURIComponent does not escape single quote character.
- * If value contains ', the uriConverter method encodes the single quote.
- * @param key - Name of a key property of an entity
- * @param value - Value of the given key
- * @param schema - schema of an entity
- * @returns key and encoded value as formatted string
- */
+  /**
+   * Returns an entities keys as key-value pairs.
+   * encodeURIComponent does not escape single quote character.
+   * If value contains ', the uriConverter method encodes the single quote.
+   * @param key - Name of a key property of an entity.
+   * @param value - Value of the given key.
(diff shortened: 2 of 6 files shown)


packages/odata-common/src/uri-conversion/get-resource-path.ts
```diff
@@ -108,7 +108,7 @@ export function createGetResourcePathForKeys(
     schema: Record<string, any>
   ): string {
     const edmType = schema[upperCaseSnakeCase(key)].edmType;
-    return `${key}=${uriConverter(value, edmType)}`;
+    return `${key}=${encodeURIComponent(uriConverter(value, edmType))}`;
```
FrankEssenberger, line 111: You encode now all parameters which makes sense I guess. However, on the de-serializer level there is also the [serializeToUri](https://github.com/SAP/cloud-sdk-js/blob/701002de058d658c2204517e065cabef8eb13d9c/packages/odata-common/src/de-serializers/default-de-serializers.ts#L102) property. I wonder where the encoding should go. There are arguments in both direction:
- If it goes to the serialize, users can overwrite it.
- If we have it here we ensure all properties are encoded

Perhaps the colleagues @marikaner @florian-richter @jjtang1985 @tomfrenken can also give their opinion. No matter where we put it, it should be then either here or in the de-serializer.

packages/odata-v2/src/uri-conversion/get-resource-path.spec.ts
```diff
@@ -66,4 +66,15 @@ describe('get resource path', () => {
       'There are too many key properties. Ignoring the following keys: StringProperty'
     );
   });
+
+  it('URI encodes keys with special characters', () => {
```
FrankEssenberger, line 70: perhaps can you add one test for the quote encoding, because this was en error in the past - see here for good examples: https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html

packages/odata-common/src/uri-conversion/get-resource-path.ts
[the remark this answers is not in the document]
jjtang1985: From the encoding [ADR](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/adr/0024-url-encoding.md#entity-path), we assume, the key do not have to be encoded.

deekshas8: @florian-richter , @FrankEssenberger : The URI serialize methods in de-serializers were also my first thought, but that resulted in filter expressions being encoded twice (linked [PR](https://github.com/SAP/cloud-sdk-js/pull/1283)). I think if we want to use serializeToUri (or similar) methods , somehow changes from this PR will have to be adjusted. Please correct me if my understanding is wrong.

Regarding encoding the keys, this [adr](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/adr/0024-url-encoding.md) mentions that we assume we don't need to do it unless we get feedback from users, so I left it out.

jjtang1985: I got lost a bit.
I feel we need some diagrams so we can clarify all these layers.
I see this is a refactoring ticket and we need to very the test coverage.

FrankEssenberger: So I have been thinking about this issue and I think it is correct the way it is done to put the uri compoent rather high in the get-ref-path method. Doing so it is analogous to the `get-filter.ts`. Also note the important detail, that the encodeURICompoent does not touch `'`. This is only relevant for the value of the filter or key, because the valid odata String is `key='myvalue'`. If the myvalue contains `'` these need to be encoded which is done my the [serializeToUri](https://github.com/SAP/cloud-sdk-js/blob/701002de058d658c2204517e065cabef8eb13d9c/packages/odata-common/src/de-serializers/default-de-serializers.ts#L102).

FrankEssenberger: Deeksha will add a comment that one remembers this detail and also two more tests to cover the `/` and `'` isssue.

packages/odata-common/src/de-serializers/uri-value-converter.ts
```diff
@@ -14,6 +14,7 @@ export function isInfOrNan(value: string | number): boolean {
 }
 
 /**
+ * As per OData URL Syntax rules, single quotes within string literals must be represented as two consecutive single quotes.
```
FrankEssenberger, line 17: Perhaps add a second sentence that this is not considered by the `encodeURIComponent` on the high level `get-filter` and `get-resource-path`.

packages/odata-v2/src/request-builder/get-by-key-request-builder.spec.ts
```diff
@@ -15,6 +15,21 @@ import { GetByKeyRequestBuilder } from './get-by-key-request-builder';
...
+        KeyPropertyString: 'DEV?TEST06'
+      });
+      const expected =
+        /^\/testination\/sap\/opu\/odata\/sap\/API_TEST_SRV\/A_TestEntity\(KeyPropertyGuid=guid'\w{8}-\w{4}-\w{4}-\w{4}-\w{12}',KeyPropertyString='DEV%3FTEST06'\)\?\$format=json$/;
+
+      const actual = await new GetByKeyRequestBuilder(testEntityApi, {
+        KeyPropertyGuid: entity.keyPropertyGuid,
+        KeyPropertyString: entity.keyPropertyString
+      }).url(defaultDestination);
+      expect(actual).toMatch(expected);
```
marikaner, line 30: General recommendation: I think it would be nice to use `toMatchInlineSnapshot` here.

PR #2181, opened 2022-02-15
feat: Allow specifying the license for generated clients

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#583.

we need an option to set the license for our generated clients. This PR adds the CLI option to set the license propert in the generate `package.json`.

### DIFF
--- a/packages/generator-common/src/file-writer/package-json.ts
+++ b/packages/generator-common/src/file-writer/package-json.ts
@@ -1,0 +1,10 @@
+/**
+ * @internal
+ */
+export interface PackageJsonOptions {
+  npmPackageName: string;
+  version: string;
+  sdkVersion: string;
+  description: string;
+  license?: string;
+}
--- a/packages/generator-common/src/file-writer/package-json.ts
+++ b/packages/generator-common/src/file-writer/package-json.ts
@@ -8,3 +8,24 @@ export interface PackageJsonOptions {
   description: string;
   license?: string;
 }
+
+export function packageJsonBase(
+  options: PackageJsonOptions
+): Record<string, any> {
+  return {
+    name: options.npmPackageName,
+    version: options.version,
+    description: options.description,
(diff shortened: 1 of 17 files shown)


packages/generator-common/src/file-writer/package-json.ts
```diff
@@ -0,0 +1,31 @@
...
+  license?: string;
+}
+
+export function packageJsonBase(
+  options: PackageJsonOptions
+): Record<string, any> {
+  return {
+    name: options.npmPackageName,
+    version: options.version,
+    description: options.description,
```
florian-richter, line 18: This should have a sensible default value: `An [OData/OpenAPI] client for <packageName> generated by the SAP Cloud SDK v<SDK VERSION>`

packages/openapi-generator/src/options/options.ts
```diff
@@ -89,6 +89,13 @@ export const generatorOptions = {
       'When enabled, a `package.json`, that specifies dependencies and scripts for transpilation and documentation generation is generated.',
     default: false
   },
+  licenseInPackageJson: {
+    string: true,
+    description:
+      "License to be used on the generated package.json. Only considered is 'packageJson' is enabled.",
```
florian-richter, line 95:
```suggestion
      "License to be used on the generated package.json. Only considered if 'packageJson' is enabled.",
```

packages/generator-common/src/file-writer/package-json.ts
FrankEssenberger: The two generators provide their description. What would be the value to have a default there, if it is not used under any circumstance?

florian-richter (approved):
LGTM

packages/generator-common/src/file-writer/package-json.ts
FrankEssenberger: added issue to clean this up: https://github.com/SAP/cloud-sdk-backlog/issues/604

Issue #1737, opened 2021-10-26
getAxiosConfigWithDefaultsWithoutMethod sets paramsSerializer without proper encoding


richardscholten:
In our code we use the `buildAxiosRequestConfig` to get a preconfigured `AxiosRequestConfig`. It calls this method that does not encode the query parameters keys or values:

https://github.com/SAP/cloud-sdk-js/blob/962e0437bd63a0551dc6ff8255313d94c126b8f3/packages/http-client/src/http-client.ts#L231-L244

This results in this error when a query parameter for example contains spaces:

```
TypeError [ERR_UNESCAPED_CHARACTERS]: Request path contains unescaped characters
    at new ClientRequest (_http_client.js:147:13)
    at Object.request (https.js:316:10)
```

It is not clear to me why this `paramsSerializer` is set while Axios itself already encodes parameters by default:

https://github.com/axios/axios/blob/master/lib/helpers/buildURL.js

**To Reproduce**

1. Call `buildAxiosRequestConfig` to get a AxiosRequestConfig.
2. Execute a GET request with Axios using a parameter containing disallowed characters (spaces for example) and the config from step 1.
3. Confirm that the query parameters are indeed not encoded.
4. See that ERR_UNESCAPED_CHARACTERS is thrown.

**Expected behavior**

Query parameters are encoded and request succeeds.

**Used Versions:**

- node version via `node -v`: v12.21.0
- npm version via `npm -v`: 6.14.11
- SAP Cloud SDK version you used as dependency: 1.51.0

**Code Examples**

```
const requestConfig = await buildAxiosRequestConfig(destination);
requestConfig.params = {
  test: 'param with spaces' 
}
const response = await axios.get('...some url...', requestConfig);
```

**Log file**

```
TypeError [ERR_UNESCAPED_CHARACTERS]: Request path contains unescaped characters
    at new ClientRequest (_http_client.js:147:13)
    at Object.request (https.js:316:10)
    at RedirectableRequest._performRequest (...\node_modules\follow-redirects\index.js:272:24)
    at new RedirectableRequest (...\node_modules\follow-redirects\index.js:61:8)
    at Object.request (...\node_modules\follow-redirects\index.js:463:14)
    at dispatchHttpRequest (...\node_modules\axios\lib\adapters\http.js:202:25)
    at new Promise (<anonymous>)
    at httpAdapter (...\node_modules\axios\lib\adapters\http.js:46:10)
[...]

PR #2183, opened 2022-02-15
fix: encoding of http client

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#414.

Adds encoding to our HTTP client of the query parameter. Custom parameter are not encoded. 
This relates to this issue: https://github.com/SAP/cloud-sdk-js/issues/1737


packages/http-client/src/http-client.ts
```diff
@@ -176,18 +203,20 @@ async function getMergedHeaders(
...
   ).queryParameters;
-  return mergeOptionsWithPriority({
+  const parametersFromAllOrigins = {
     ...paramsOriginOptions,
     destinationProperty: queryParametersDestinationProperty,
     destination: destination.queryParameters
-  });
+  };
+  const encodedParameters = getEncodedParameters(parametersFromAllOrigins);
+  return mergeOptionsWithPriority(encodedParameters);
```
florian-richter, line 219: Should you encode first and merge after? I would assume the other way around would be better (avoid unnecessary encoding, although unlikely it could prevent issues with merging caused by encoding)

packages/http-client/src/http-client.ts
```diff
@@ -121,6 +122,32 @@ export function execute<ReturnT>(executeFn: ExecuteHttpRequestFn<ReturnT>) {
...
+function getEncodedParameters(
+  parameters: OriginOptionsInternal
+): OriginOptionsInternal {
+  return {
+    custom: parameters.custom,
+    requestConfig: parameters.requestConfig, // sdk parameters are encoded
+    destinationProperty: encodeRequestQueryParameters(
+      parameters.destinationProperty
+    ),
+    destination: encodeRequestQueryParameters(parameters.destination)
```
florian-richter, line 147:
```suggestion
    ...parameters, // sdk parameters are encoded
    destinationProperty: encodeRequestQueryParameters(
      parameters.destinationProperty
    ),
    destination: encodeRequestQueryParameters(parameters.destination)
```
This would be slightly shorter... probably no difference otherwise.

packages/http-client/src/http-client.ts
```diff
@@ -176,18 +203,20 @@ async function getMergedHeaders(
   });
 }
 
-function getMergedParameters(
+function getMergedAndEncodedParameter(
   destination: Destination,
   paramsOriginOptions?: OriginOptions
-): Record<string, string> | undefined {
+): Record<string, any> | undefined {
```
florian-richter, line 209: Is this intended? I don't see how this is caused by this change...

packages/http-client/src/http-client.ts
```diff
@@ -121,6 +122,32 @@ export function execute<ReturnT>(executeFn: ExecuteHttpRequestFn<ReturnT>) {
...
 }
 
+function encodeRequestQueryParameters(
+  parameter: Record<string, string> | undefined
+): Record<string, string> | undefined {
+  if (parameter) {
+    return Object.keys(parameter).reduce(
+      (encodedParams, key) => ({
+        ...encodedParams,
+        [key]: encodeURIComponent(parameter[key])
```
florian-richter, line 132:
```suggestion
        [encodeURIComponent(key)]: encodeURIComponent(parameter[key])
```
Could it be possible to have a special char in the key? Otherwise this should be fine.

packages/http-client/src/http-client.ts
```diff
@@ -121,6 +122,32 @@ export function execute<ReturnT>(executeFn: ExecuteHttpRequestFn<ReturnT>) {
...
+      }),
+      {}
+    );
```
florian-richter, line 135:
```suggestion
    return Object.fromEntries(Object.entries(parameter).map(
      ([key, value]) => 
        [encodeURIComponent(key), encodeURIComponent(parameter[key])]
      })
    ));
```
I like `Object.entries` better and I think `.fromEntries` and `.map` are nicer than `.reduce`.

florian-richter (approved):
Some comments but looks good overall. The empty line, the type and encoding the key should be considered. The other stuff is optional.

packages/http-client/src/http-client.ts
FrankEssenberger: I agrre the `fromEntries` is more clear then a reduce.

packages/http-client/src/http-client.ts
FrankEssenberger: good point I adjusted the comment that custom should not be encoded and sdk parameters are encoded.

packages/http-client/src/http-client.ts
```diff
@@ -121,6 +122,31 @@ export function execute<ReturnT>(executeFn: ExecuteHttpRequestFn<ReturnT>) {
...
+      ])
+    );
+  }
+}
+
+function getEncodedParameters(
+  parameters: OriginOptionsInternal
+): OriginOptionsInternal {
+  return {
+    ...parameters, // sdk parameters are encoded and custom parameters should be taken as they are
```
jjtang1985, line 142: As SDK developer, we use `requestConfig` for SDK parameters.
However, other users (like CAP and the issue author) might miss the encoding, as SDK will not handle it for them.

packages/http-client/src/http-client.spec.ts
```diff
@@ -759,6 +759,46 @@ sap-client:001`);
...
+    it('should encode query parameters excluding the custom ones.', async () => {
+      const httpRequestConfigWithOrigin: HttpRequestConfigWithOrigin = {
+        method: 'get',
+        url: '/api/entity',
+        params: {
+          custom: {
+            customParam: 'Not/Encoded'
+          },
+          requestConfig: {
+            // Authorization: 'reqConfigAuth',
```
jjtang1985, line 771: Please add a test case for the requestConfig, so it's covered.

jjtang1985 (changes requested):
It seems the we miss the encoding for requestConfig.

packages/http-client/src/http-client.ts
FrankEssenberger: what do you think?

jjtang1985: Nice alternatives, that worth a separate PR.
However, without that PR, we might want to align with CAP/users about the current situation, because this PR might no solve the original issue.
We can also handle this PR now and release it together with this PR.

FrankEssenberger: I think I will include it in this PR. The changes should be not too much. Here just two files are changed.

FrankEssenberger:
Documentation: https://github.com/SAP/cloud-sdk/pull/707

packages/http-client/src/http-client.ts
```diff
@@ -121,6 +123,61 @@ export function execute<ReturnT>(executeFn: ExecuteHttpRequestFn<ReturnT>) {
   };
 }
 
+function encodeRequestQueryParameters(
```
jjtang1985, line 126: Minor: Maybe "request is not needed?"
```suggestion
function encodeQueryParameters(
```

packages/http-client/src/http-client.ts
```diff
@@ -121,6 +123,61 @@ export function execute<ReturnT>(executeFn: ExecuteHttpRequestFn<ReturnT>) {
...
+        key,
+        parameterEncoder(value)
+      ])
+    );
+  }
+
+  // If the parameterEncoder is undefined the executeHttpRequest() was used directly -> encode the request Parameters as wel
+  const requestParameterEncoder = !parameterEncoder
+    ? encodeRequestQueryParameters
+    : encodeTypedClientRequest;
```
jjtang1985, line 169:
```suggestion
  const requestParameterEncoder = parameterEncoder
    ? encodeTypedClientRequest
    : encodeRequestQueryParameters;
```

packages/http-client/src/http-client.ts
```diff
@@ -121,6 +123,61 @@ export function execute<ReturnT>(executeFn: ExecuteHttpRequestFn<ReturnT>) {
...
+export const encodeTypedClientRequest: ParameterEncoder = (
+  params: Record<string, any>
+) => params;
+
+function getEncodedParameters(
+  parameters: OriginOptionsInternal,
+  requestConfig: HttpRequestConfigWithOrigin
+): OriginOptionsInternal {
+  const { parameterEncoder } = requestConfig;
+  // Custom Parameter encoder given->use it
```
jjtang1985, line 153: The 3 conditions are not straight forward.
Maybe we can make it like:
```
if(isCustomParameterEncoder()){
   return ...
}

if(isTypedClinetRequest()){
    return ...
}
```

packages/http-client/src/http-client.ts
```diff
@@ -121,6 +123,61 @@ export function execute<ReturnT>(executeFn: ExecuteHttpRequestFn<ReturnT>) {
...
+    return Object.fromEntries(
+      Object.entries(parameters).map(([key, value]) => [
+        key,
+        parameterEncoder(value)
+      ])
+    );
+  }
+
+  // If the parameterEncoder is undefined the executeHttpRequest() was used directly -> encode the request Parameters as wel
+  const requestParameterEncoder = !parameterEncoder
```
jjtang1985, line 167: When reaching this line, `parameterEncoder` should be `encodeTypedClientRequest`.
Then why should we check it in this line?

jjtang1985:
Please adjust the changelog.

packages/http-client/src/http-client.ts
```diff
@@ -121,6 +123,70 @@ export function execute<ReturnT>(executeFn: ExecuteHttpRequestFn<ReturnT>) {
...
+ * This method does nothing and is only there to indicated that the call was made by Odata or OpenApi client and encoding is already done on filter and key parameters.
+ * @param params - Parameters which are returned
+ * @returns The parameters as they are without encoding.
+ * @internal
+ */
+export const encodeTypedClientRequest: ParameterEncoder = (
+  params: Record<string, any>
+) => params;
+
+type foo = keyof OriginOptionsInternal;
```
jjtang1985, line 136: foo 😄

jjtang1985 (approved):
Please rename the foo.

PR #2190, opened 2022-02-17
chore: add destination type doc

jjtang1985:
There are some restrictions when using on-prem destinations.
I added some docs for that.


florian-richter (approved):
LGTM

PR #2214, opened 2022-02-25
chore: Add docs plugin

marikaner:
This adds a docs plugin for better checking of our rules.

The currently used `valid-jsdoc` rule is outdated and will be soon removed (see [end of life note](https://eslint.org/blog/2018/11/jsdoc-end-of-life)).
The recommendation is to use the [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc), which actually also works for TS (and clojure).

The only feature, that it does not cover, AFAICS is type parameters ([feature-reuqest](https://github.com/gajus/eslint-plugin-jsdoc/issues/844)). 
Closes SAP/cloud-sdk-backlog#580.

@deekshas8 Those are the remaining TODOs:
- [x] Merge #2215
- [x] Remove the existing rules around docs (`valid-jsdoc` and `@microsoft/tsdoc` => also from the dependencies)
- [x] Add these rules to the eslint config
- [x] Check whether you agree with the config

The following PR should be merged after this PR (#2214) is closed:
- [ ] #2223

We should keep some of the rules as warnings, but slowly start fixing them and change them to errors later.


deekshas8:
The current config looks good to me, except the below issues:

- Regarding @typeparam, I added this custom tag to structuredTags and then updated settings of some rules to take it into consideration, but couldn't get it to work (make it required, syntax etc.). I also tried with `@template` which is valid in both jsdoc/tsdoc. But with no success.

- In our pipeline, we run eslint with --fix option. This adds all the missing docs and then runs all rules on them which, as expected, results in lots of errors. To stop this, atleast locally, I ran --fix --fix-type layout.

packages/connectivity/src/index.ts
```diff
@@ -1,6 +1,5 @@
-/* eslint-disable tsdoc/syntax */
 /**
- * [[include:connectivity/README.md]]
+ * [[include:connectivity/README.md]].
```
marikaner, line 2: There should not be any fullstop here. Can we add an exception?

packages/odata-v4/src/uri-conversion/uri-value-converter.ts
```diff
@@ -1,7 +1,4 @@
 import { createUriConverter } from '@sap-cloud-sdk/odata-common/internal';
 import { defaultDeSerializers } from '../de-serializers';
 
-/**
- * internal
- */
 export const uriConverter = createUriConverter(defaultDeSerializers);
```
marikaner, line 4: The previous comment was obvisouly broken, but shouldn't it still complain about missing `@internal` if it is internal or any documentation if there is no documentation?
Or is this something that still needs to be done?

marikaner (commented):
The imports for readme's should not have fullstops in the end. Other than that I just had a minor question, that I guess I already answered myself ;)

packages/odata-v4/src/uri-conversion/uri-value-converter.ts
deekshas8: Yes, this still needs to be done. Actually the require-jsdoc rule be default only checks functions and classes and doesn't warn for missing docs on interface, enum, types, etc. I'll add more `contexts` to expand it.

packages/connectivity/src/index.ts
deekshas8: Ok, I wasn't aware of it. I'll look into it.

deekshas8: I turned the `require-description-complete-sentence` rule off for index.ts file in `.eslintrc.js`

packages/eslint-config/index.js
```diff
@@ -179,16 +179,46 @@ module.exports = {
...
+    'jsdoc/check-indentation': 'error',
+    'jsdoc/check-param-names': 'warn',
+    'jsdoc/check-syntax': 'error',
+    'jsdoc/multiline-blocks': 'error',
+    'jsdoc/newline-after-description': ['error', 'never'],
+    'jsdoc/no-bad-blocks': 'error',
+    'jsdoc/no-defaults': 'error',
+    'jsdoc/no-types': 'error',
+    'jsdoc/require-asterisk-prefix': 'error',
+    'jsdoc/require-description-complete-sentence': [
```
marikaner, line 192: Would this rule work if we just disable it for the `index.ts` files where we describe packages?

marikaner (commented):
I would like to make a conscious decision on whether we add this to the eslint package or not. Either way, we need an entry in the changelog 😊

packages/eslint-config/index.js
deekshas8: Yes, it works for everything thats publicly exported

packages/eslint-config/index.js
```diff
@@ -179,16 +179,46 @@ module.exports = {
...
+      'warn',
+      { abbreviations: ['e.g.', 'i.e.', 'aka.', 'etc.'] }
+    ],
+    'jsdoc/require-description': ['error', { exemptedBy: ['experimental'] }],
```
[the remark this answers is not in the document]
florian-richter: I would assume we could put a compatibility note to overwrite the rule to get the legacy behavior... Or even update the eslint-plugin separately to 3.0 would be fine in my eyes.

.eslintrc.js
```diff
@@ -38,10 +37,15 @@ module.exports = {
         'import/no-internal-modules': 'off',
         'no-unused-expressions': 'off'
       }
+    },
+    {
+      files: ['**/index.ts'],
+      rules: {
+        'jsdoc/require-description-complete-sentence': 'off'
```
florian-richter, line 44: This is off but changed in a lot of comments below. Is that intended?

florian-richter:
> Should be merged after this PR is closed:
> 
> * [ ]  [chore: remove eslint disable comments for @internal #2223](https://github.com/SAP/cloud-sdk-js/pull/2223)

Seems like #2223 and #2214 have a circular dependency. I guess this one should be merged first, right?

.eslintrc.js
deekshas8: Yes, this was intended. The rule `require-description-complete-sentence` adds a `.` even in `index.ts` files which was not needed. The rule itself has no way of adding exceptions for specific cases, so I had to turn it off this way.

florian-richter (approved):
LGTM ⚠️ but there needs to be CHANGELOG entry with the required compatibility note if we want to release this as 2.x.y ⚠️

PR #2221, opened 2022-03-01
chore: fix doc issues

deekshas8:
Closes SAP/cloud-sdk-backlog#ISSUENUMBER.

PR #2228, opened 2022-03-02
chore: fix doc comments

deekshas8:
This PR fixes more issues in doc comments completeness. This is a follow-up on #2221


packages/odata-common/src/request-builder/delete-request-builder-base.ts
```diff
@@ -49,7 +49,7 @@ export abstract class DeleteRequestBuilderBase<
 
   /**
    * Instructs the request to force an overwrite of the entity by sending an 'If-Match: *' header instead of sending the ETag version identifier.
-   * @returns this The request itself to ease chaining while executing the request
+   * @returns This The request itself to ease chaining while executing the request.
```
jjtang1985, line 52:
```suggestion
   * @returns The request itself to ease chaining while executing the request.
```

packages/odata-v4/src/uri-conversion/uri-value-converter.ts
```diff
@@ -2,6 +2,6 @@ import { createUriConverter } from '@sap-cloud-sdk/odata-common/internal';
 import { defaultDeSerializers } from '../de-serializers';
 
 /**
- * internal
+ * @internal
```
jjtang1985, line 5: I guess this is not internal, as I see it's exported in the index.

packages/util/src/equal.ts
```diff
@@ -1,7 +1,7 @@
 import { isNullish } from './nullish';
 // eslint-disable-next-line valid-jsdoc
 /**
- * hidden
+ * Hidden.
```
jjtang1985, line 4:
```suggestion
 * @hidden.
```
See doc: https://typedoc.org/guides/doccomments/#%40hidden-and-%40ignore
Please also adjust other occurrences.

packages/odata-v4/src/filter-functions.ts
```diff
@@ -152,7 +152,7 @@ export function hasSubsequence<
   return filterFunction('hassubsequence', 'boolean', subsequence, sequence);
 }
 /**
- * TODO: totalseconds, time
+ * TODO: totalseconds, time.
```
jjtang1985, line 155: General question: Is it a manual effort to add the "."?

deekshas8: No, the jsdoc rule `jsdoc/require-description-complete-sentence` does this automatically. It was added in the other PR #2214 . I fixed it in a separate PR so its easier to review

packages/odata-v4/src/uri-conversion/uri-value-converter.ts
deekshas8: Thanks, I thought it was a typo.

packages/util/src/equal.ts
deekshas8: We export everything from util package, so I am not sure if we need `@hidden` at all, atleast here. I will remove this and adjust in other places

jjtang1985: This hidden is for hiding the api for doc generation.
We can also remove the "hidden" and add valid docs.

packages/odata-v4/src/filter-functions.ts
jjtang1985: I guess you enabled this rule then run the lint fix.
The rule was removed for making the pipeline happy in the previous PR, right?
Do we plan to add the rule in the future when everything is fixed?

PR #2279, opened 2022-03-18
fix: reduce useless exports from 2.0 clients

jjtang1985:
Closes SAP/cloud-sdk-backlog#602.

- Stop exporting service classes (e.g., `BusinessPartnerService`) from generated clients, use `businessPartnerService()` instead.
- Stop exporting API classes (e.g., `BusinessPartnerAPI`) from generated clients, use `businessPartnerService().businessPartnerApi` instead.


FrankEssenberger (commented):
I wonder what is our general policy. If me method `foo():bar` is exported should the return type `bar` be exported as well or is the type inference good enough in our opinion? The same question holdes for funciton arguments `foo(a:bar)` should the type be exported? In this case since TypeScript has this access type checking it could become necessary to type a parameter object outside the acutal call. We should perhaps align on this and also do it with all our exports of the SDK libs.

florian-richter (approved):
Maybe mention `ReturnType<typeof myFunction>` and `Parameters<typeof myFunction>` in the compatibility section of the release notes for those that need the types that are no longer exported.

PR #2290, opened 2022-03-22
fix: generator compilation error

jjtang1985:
E2E generator tests failed due to the error below, because we use `@types/opossum` code in the `d.ts` file, while the dependency is a dev dependency.

```
[2022-03-21T16:59:39.858Z] ERROR    (generator-cli): ErrorWithCause: Generation of services failed.
    at /.../node_modules/@sap-cloud-sdk/generator/dist/generator-cli.js:22:18
Caused by:
Error: Compilation Errors:
/.../node_modules/@sap-cloud-sdk/connectivity/dist/scp-cf/resilience-options.d.ts:1:27 - error TS7016: [object Object]
    at transpileDirectory (/.../node_modules/@sap-cloud-sdk/generator-common/dist/compiler.js:35:15)
    at async Promise.all (index 0)
    at async generate (/.../node_modules/@sap-cloud-sdk/generator/dist/generator.js:52:13)
Waiting for the debugger to disconnect...
npm ERR! code ELIFECYCLE

```

This PR removes the type usage.

### DIFF
--- a/packages/connectivity/src/scp-cf/resilience-options.ts
+++ b/packages/connectivity/src/scp-cf/resilience-options.ts
@@ -46,8 +46,9 @@ export function timeoutPromise<T>(timeout: number): Promise<T> {
 
 /**
  * This is partially copied from CircuitBreaker.Options of `@types/opossum`.
+ * @internal
  */
-interface CircuitBreakerOptions {
+export interface CircuitBreakerOptions {
   timeout?: number | false | undefined;
   errorThresholdPercentage?: number | undefined;
   volumeThreshold?: number | undefined;
--- a/packages/connectivity/src/scp-cf/resilience-options.ts
+++ b/packages/connectivity/src/scp-cf/resilience-options.ts
@@ -1,5 +1,3 @@
-import CircuitBreaker from 'opossum';
-
 export interface ResilienceOptions {
   /**
    * A boolean value that indicates whether to execute request to SCP-CF services using circuit breaker.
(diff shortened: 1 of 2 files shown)


packages/connectivity/src/scp-cf/resilience-options.ts
```diff
@@ -1,5 +1,3 @@
-import CircuitBreaker from 'opossum';
```
FrankEssenberger, line 1: Could you use `import type` to overcome the issue as well?

FrankEssenberger:
I do not understand the issue, because opossum is the circuit breaker of the `connectivity` we use, so if `npm install` is done it should be part of the dependencies and not dev-dependencies. 

Ah, I think we put all `@types` as dev, because once we ship we ship the transpiled sources, but if a customer locally does the compile due to the generation we need it. So in this case I think we should add @types/opossum as real dependency? This is a matter for discussion @marikaner @florian-richter

jjtang1985:
> So in this case I think we should add @types/opossum as real dependency?

In this case, as internal api, I would not use it as dependency, given the size is [install size](https://packagephobia.com/result?p=@types/opossum)

In general, this might lead to a general discussion about all the packages `@types/*` and define the use cases when we should introduce them as dependency.

FrankEssenberger (approved):
I added a test to ensure, that we stay compatible with the opossum type. Please ensure that the test is ok for you, perhaps there is a more elegant way to test it. I did not want to use the `type-tests` because they are so far away, but one could use them of course.

I guess duplicating the code is the best solution in this special case to save the installation space. This shows also how small the community of TS users is which transpile thier code. They would have faced the same issue if they use the SDK I guess.

jjtang1985:
> I added a test to ensure, that we stay compatible with the opossum type. Please ensure that the test is ok for you, perhaps there is a more elegant way to test it. I did not want to use the type-tests because they are so far away, but one could use them of course.

Thanks for the additional compatibility tests, which makes sense.
We also have `circuit-breaker.spec.ts`, which tests the functional requirement of circuit breaker (opossum, in our case).

PR #2348, opened 2022-04-08
Display weekly download numbers

marikaner:
Add a script to log weekly download numbers.


scripts/npm-stats.ts
```diff
@@ -0,0 +1,53 @@
...
+import { JSDOM } from 'jsdom';
+
+async function getStatsForPackage(packageName) {
+  const { data } = await axios.get(
+    `https://www.npmjs.com/package/${packageName}?activeTab=versions`
+  );
+
+  const dom = new JSDOM(data);
+  const ul = dom.window.document.querySelector(
+    '.a0dff0b1 ~ .c495d29d ~ .a0dff0b1 ~ .c495d29d'
```
florian-richter, line 11:
```suggestion
    '#tabpanel-versions > * > ul:last-child'
```

I find this a bit more speaking (and hopefully less likely to break)

scripts/npm-stats.ts
```diff
@@ -0,0 +1,53 @@
...
+  const res: any[] = [];
+
+  for (let i = 1; i < versionLines.length; i++) {
+    const line = versionLines.item(i);
+    const [version, _, downloads] = line.children;
+    res.push({
+      version: version.textContent,
+      downloads: parseInt(downloads.textContent.replace(/,/g, ''))
+    });
+  }
```
florian-richter, line 23:
```suggestion
 Array.from(versionLines).map(line => line.children).map(([version,_,dl]) => ({ version: version.textContent, downloads: parseInt(dl.textContent.replace(/,/g, '')) }))
```
Maybe some cleanup needed but the `.map()` chain seems nicer to me

scripts/npm-stats.ts
jjtang1985, line 11: 🤔 This might need further improvement as a follow-up

scripts/npm-stats.ts
```diff
@@ -0,0 +1,53 @@
...
+      ...(await stats),
+      [packageName]: await getStatsForPackage(packageName)
+    }),
+    {}
+  );
+}
+
+async function main() {
+  // eslint-disable-next-line no-console
+  console.log(await getAllStats());
```
florian-richter, line 50: Could we create a GH repo where we upload the stats every day? Maybe using this: https://github.com/octokit/octokit.js

florian-richter (approved):
Works! But some improvement suggestions:

PR #2358, opened 2022-04-11
fix: support multiple path params within a path part

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#52.

This PR **enables** the following cases:
1. `/path/{param1}:{param2}`
2. `/path/{param1}abcdef{param2}`
3. `/path/{param1}{param2}`
4. `/path/abc{param1}123/{param2}`
5. `/path/{{{param}` (the innermost will be matched)

This PR **removes** check for the following cases:
1. `/path/value?{param}`: This is illegal but not checked. There should be theoretically no question mark in the path since all query parameters should be defined elsewhere. Despite that, I found nothing related to this problem in the OpenAPI specification, nor did the official [swagger-editor](https://editor.swagger.io/) or [swagger-parser](https://github.com/swagger-api/swagger-parser) checks this. 

This PR **prohibits** the following cases:
1. `/path/{param}` with path parameter `param: #test`. As mentioned in [openapi specification](https://spec.openapis.org/oas/v3.1.0#path-templating), characters `?`, `#` and `/` are not allowed to be included in the value of path parameter. (RFC3986)
```
The generic syntax uses the slash ("/"), question mark ("?"), and
number sign ("#") characters to delimit components that are
significant to the generic parser's hierarchical interpretation of an
identifier. (RFC3986)
```
2. `/path/{pa/ram}` is valid but has **no** placeholder. Same as the first case, `?`, `#`, and `/` are used as delimiters, which will break the curly brackets.

Please let me know if anything is unclear. I appreciate your suggestions.


packages/openapi/src/openapi-request-builder.ts
```diff
@@ -164,23 +164,36 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
-            throw new Error(
-              `Cannot execute request, no path parameter provided for '${paramName}'.`
-            );
-          }
-          return encodeURIComponent(paramValue);
-        }
-        return part;
-      })
-      .join('/');
+    // Check if there is still curly bracket in the replaced path pattern
```
ZhongpinWang, line 186: Please let me know if check at this step is needed. Currently, if there is any `{...}` left, an error will be thrown.

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -72,82 +72,81 @@ export function getRelevantParameters(
...
+        `Could not find placeholder for path parameter '${pathParameter.originalName}'.`
+      );
+    }
+    // Temporarily using TEMP_LEFT_CURLY_BRACKET and TEMP_RIGHT_CURLY_BRACKET to avoid re-substitution
+    // Check test 'parses path template for parameters' for the edge case input
+    pathPattern = pathPattern.replace(
+      new RegExp(`{${pathParameter.originalName}}`, 'g'),
+      `TEMP_LEFT_CURLY_BRACKET${pathParameter.name}TEMP_RIGHT_CURLY_BRACKET`
+    );
+  }
```
marikaner, line 96: I am not a fan of `TEMP_LEFT_CURLY_BRACKET` and I don't think that this part was the problem in the old implementation. The parameter names are unique, aren't they?

The old implementation did the following:
1. split string by '/'
2. filter parts that have placeholders
4. check that every part includes at most one `{placeholders}`

(with some other steps between and after)

I think the problem starts in step 1.
Instead of restricting ourselves to split this into parts, can't we just reuse the regex we use for testing whether something is a placeholder and just use it globally?

marikaner (changes requested):
I am not a fan of the `TEMP_LEFT_CURLY_BRACKET` (and right) and tried to explain it my comment. Let's talk about it in case it is not clear enough.

packages/openapi-generator/src/parser/operation.ts
ZhongpinWang: I admit that it is indeed not elegant. It is used due to the replacement of `originalName` with `name`. Consider the following case:
pathPattern: `/path/{param_old}/{param}`
path parameter `param1`: `originalName: param_old` and `name: param`
path parameter `param2`: `originalName: param` and `name: param_new`

Then after executing this function, the pathPattern will be `/path/{param_new}/{param_new}` instead of `/path/{param}/{param_new}`.

This is essentially the test case `parses path template for parameters` in `operation.spec.ts`.

In the previous implementation, {placeholder} is extracted by using split('/'). But I guess we can also use regex `/{[^/?#{}]+}/` to match all the brackets as well and then sort them somehow.

ZhongpinWang:
In the new implementation, I did not check if there are redundant path parameters provided, which do not match any placeholders in the path pattern. I think it is of course better to have it there, but I wonder if it is necessary since there is essentially no difference with respect to the functionality of the generated typescript code.

ZhongpinWang:
> In the new implementation, I did not check if there are redundant path parameters provided, which do not match any placeholders in the path pattern. I think it is of course better to have it there, but I wonder if it is necessary since there is essentially no difference with respect to the functionality of the generated typescript code.

On second thought, this check could help users find out potential errors. Thus, I added it to the new implementation: [fix: check if path parameter matches any placeholder](https://github.com/SAP/cloud-sdk-js/pull/2358/commits/24a128e6ada76720bb36297ef7eed8c63f3dab01).

marikaner (changes requested):
Some first thoughts, I will have to take a second look

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -164,23 +164,66 @@ export class OpenApiRequestBuilder<ResponseT = any> {
 
   private getPath(): string {
     const pathParameters = this.parameters?.pathParameters || {};
+    const path = this.pathPattern;
+
+    // Get the innermost curly bracket pairs with non-empty and legal content as placeholders.
+    const placeholders = path.match(/{[^/?#{}]+}/g);
+    if (!placeholders) {
```
marikaner, line 171: Please correct me if I'm wrong, but the placeholder names are unique, correct? That is how you should be able to match them exactly without sorting them. As these checks have been done in the generator there is no need to do them on request building (except the one checking for invalid values).

With this your code could be much simpler

- for every placeholder
  - get the path parameter by name
  - check that the value is valid
  - replace it in the path
- return the resulting path

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -64,102 +59,86 @@ export function getRelevantParameters(
...
-      `Path pattern '${pathPattern}' is invalid or not supported.`
-    );
+  if (!placeholders) {
+    // No placeholder or path parameter found.
+    if (pathParameters.length > 0) {
+      throw new Error(
+        `Could not find placeholder for path parameter '${pathParameters
+          .map(pathParameter => pathParameter.originalName)
+          .join("', '")}'.`
+      );
```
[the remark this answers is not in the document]
ZhongpinWang: > 1. This will give you an error like this for multiple parameters: `Could not find placeholder for path parameter 'param1', 'param2'.`
>    Maybe I am a little too strict here, but in that case it should be plural, don't you think?

Maybe the best way to do this is to print `parameter(s)`. 😆 

> 2. If all path parameters are missing in the placeholders I will get the message above, but if 2 out of 3 are missing I will just get this message: `Could not find path parameter for placeholder 'param2'.` I find this a bit inconsistent.
> 
> Could we check this once (see below) + add tests for those two cases?

Sorry, and ignore the comment in the code. I somehow wrote it wrong.

The above if-conditions mean: **no placeholder found** && **path parameter found** => could not find placeholder for path parameter. <=> iterate through path parameters and check if there is a corresponding placeholder.

So it basically means all placeholders are missing in the path parameters, not the other way around. 

In the other direction `Could not find path parameter for placeholder` <=> iterate through placeholders and check if there is a corresponding path parameter.

I hope I am not confusing myself. 😄

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -64,102 +59,86 @@ export function getRelevantParameters(
...
+    }
+    return pathPattern;
   }
 
-  return placeholders.map(placeholder => {
+  const sortedPlaceholders = placeholders.map(placeholder => {
```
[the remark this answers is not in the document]
ZhongpinWang: You are right about this. Thanks!

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -64,102 +59,86 @@ export function getRelevantParameters(
...
+    if (
+      !placeholders.find(
+        placeholder => placeholder.slice(1, -1) === pathParameter.originalName
+      )
+    ) {
+      throw new Error(
+        `Could not find placeholder for path parameter '${pathParameter.originalName}'.`
+      );
+    }
+  }
```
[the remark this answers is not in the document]
ZhongpinWang: I applied your suggestion. Thanks!

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -64,102 +59,86 @@ export function getRelevantParameters(
...
+  while (staticParts.length > 0 && sortedPlaceholders.length > 0) {
+    newPathPattern.push(staticParts.shift()!, sortedPlaceholders.shift()!);
+  }
+  if (staticParts.length > 0) {
+    newPathPattern.concat(staticParts);
+  } else {
+    newPathPattern.concat(sortedPlaceholders);
+  }
+
+  return newPathPattern.join('');
```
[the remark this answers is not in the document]
ZhongpinWang: > 1. How do you know that the static part comes first?
I double-checked it. In case someone is also interested, according to the OpenAPI 3.0.3 documentation

> [...] The field name MUST begin with a forward slash (/). [...]

See more details on [https://swagger.io/specification/#patterned-fields](https://swagger.io/specification/#patterned-fields).

> 2. In JS you can just write `if (staticParts.length)`. Take a look at our [style guide](https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-truthyfalsy-checks-where-possible).

Thank you a lot for this. I was not aware of this style guide. It is very helpful!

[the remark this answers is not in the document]
ZhongpinWang: Super elegant solution. I really like it.

packages/openapi/src/openapi-request-builder.ts
ZhongpinWang: Done. 
Since we can assume that users will always use our `openapi-generator`, there is no need to add checks anymore.

packages/openapi-generator/src/parser/operation.ts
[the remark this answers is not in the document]
ZhongpinWang: There were some complications though. I think we should use `missingPlaceholders` and `originalParameterNames.filter()` since we want to say `could not find placeholder...`. And `includes()` should be negated.

Still, I like the idea because it checks for all missing placeholders. 👍

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -164,23 +164,26 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
-          return encodeURIComponent(paramValue);
-        }
-        return part;
-      })
-      .join('/');
+    // Get the innermost curly bracket pairs with non-empty and legal content as placeholders.
+    const placeholders = path.match(/{[^/?#{}]+}/g);
+    if (!placeholders) {
+      return path;
+    }
```
marikaner, line 173: You can shorten this if you just do the following:
```suggestion
    const placeholders = path.match(/{[^/?#{}]+}/g) || [];
```

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -164,23 +164,26 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
+
+    placeholders.map(placeholder => {
```
marikaner, line 175: If you don't return anything, you don't need to map 😊 
```suggestion
    placeholders.forEach(placeholder => {
```

marikaner: You could also use reduce, then you wouldn't have to make "path" modifiable.
```ts
return placeholders.reduce((path, placeholder) => {
  return path.replace(placeholder, encodeURIComponent(parameterValue));
}, this.pathPattern);
```

ZhongpinWang: Wonderful solution with `reduce()`! Thanks!

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -164,23 +164,26 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
+      const strippedPlaceholder = placeholder.slice(1, -1);
+      const parameterValue = pathParameters[strippedPlaceholder];
+      if (/[/?#]+/.test(parameterValue)) {
+        throw new Error(
+          `Cannot execute request, the value of the path parameter '${strippedPlaceholder}' must not contain '/', '?', or '#'. (RFC3986)`
+        );
+      }
+      path = path.replace(placeholder, encodeURIComponent(parameterValue));
```
marikaner, line 183: Do we need the check this if we encode the component afterwards? That's exactly why we're doing it, isn't it?

packages/openapi-generator/src/parser/operation.ts
```diff
@@ -72,94 +67,77 @@ export function getRelevantParameters(
...
+    // No placeholder found
+    if (pathParameters.length) {
+      // Path parameter found
+      throw new Error(
+        `Could not find placeholder for path parameter(s) ${pathParameters
+          .map(pathParameter => `'${pathParameter.originalName}'`)
+          .join(', ')}.`
+      );
+    }
+    return pathPattern;
```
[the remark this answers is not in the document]
ZhongpinWang: Yes indeed. It was there basically only for null check.. After I applied the `|| []` trick, it works also without this block.

packages/openapi/src/openapi-request-builder.ts
ZhongpinWang: Oh, I guess you are right. I somehow thought that these characters won't be encoded but I was wrong after a double-check. Thanks for the input.

PR #2377, opened 2022-04-14
chore: add memory tests

jjtang1985:
This PR adds an additional test for checking the memory consumption:
- adjust the cap e2e test for bigger entities
- test v1 script with v1 client
- test canary script with canary client
- if the difference is above the threshold, the test will fail


scripts/compare-memory-usage.ts
```diff
@@ -0,0 +1,18 @@
...
+  const canaryMemoryUsage = parseInt(process.argv[3]);
+  const percentageBasedOnV1 = (canaryMemoryUsage / v1MemoryUsage) * 100;
+  if (percentageBasedOnV1 > threshold) {
+    throw new Error(
+      `Test failed: v1 memory usage is ${v1MemoryUsage} and canary memory usage is ${canaryMemoryUsage} (${percentageBasedOnV1} %). Consider change the threshold (${threshold} %) or fix it.`
+    );
+  }
+  /* eslint-disable-next-line no-console */
+  console.log(
+    `Test passed: v1 memory usage is ${v1MemoryUsage} and canary memory usage is ${canaryMemoryUsage} (${percentageBasedOnV1} %).`
```
marikaner, line 14: I think the percentage of the diff might be more interesting than the actual comparison (-10%, e.g.).

test-packages/memory-tests/sdk-canary/test-script.js
```diff
@@ -0,0 +1,115 @@
...
+    .execute(dest);
+  const testEntities105 = await testEntityApi
+    .requestBuilder()
+    .getAll()
+    .execute(dest);
+  const testEntities106 = await testEntityApi
+    .requestBuilder()
+    .getAll()
+    .execute(dest);
+  const testEntities107 = await testEntityApi
```
marikaner, line 37: Don't you like for loops? (or for async probably `.reduce`)

test-packages/memory-tests/sdk-canary/test-script.js
```diff
@@ -0,0 +1,115 @@
...
+      )
+  );
+  console.log(diff.change.size_bytes);
+  return diff.change.size_bytes.toString();
+};
+
+const main = async () => {
+  const result = await test();
+  writeFileSync(resolve(__dirname, './result.txt'), result, 'utf8');
+};
```
marikaner, line 113: Hahaha, I guess this is the workaround for not being able to disable the loggers...? I guess that's ok for now... 🙁

test-packages/memory-tests/sdk-v1/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@sap-cloud-sdk/memory-tests-v1",
+  "version": "0.0.1",
+  "description": "Memory usage tests of the SAP Cloud SDK for JavaScript",
+  "homepage": "https://sap.github.io/cloud-sdk/docs/js/overview-cloud-sdk-for-javascript",
+  "license": "Apache-2.0",
+  "private": true,
+  "repository": "github:SAP/cloud-sdk-js",
+  "scripts": {
+    "generate-vdm": "npx generate-odata-client -i ../../../test-resources/odata-service-specs-e2e/v4 -o .",
```
marikaner, line 10:
```suggestion
    "generate-vdm": "yarn generate-odata-client -i ../../../test-resources/odata-service-specs-e2e/v4 -o .",
```

test-packages/memory-tests/sdk-canary/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@sap-cloud-sdk/memory-tests-canary",
+  "version": "0.0.1",
+  "description": "Memory usage tests of the SAP Cloud SDK for JavaScript",
+  "homepage": "https://sap.github.io/cloud-sdk/docs/js/overview-cloud-sdk-for-javascript",
+  "license": "Apache-2.0",
+  "private": true,
+  "repository": "github:SAP/cloud-sdk-js",
+  "scripts": {
+    "generate-vdm": "npx generate-odata-client -i ../../../test-resources/odata-service-specs-e2e/v4 -o .",
```
marikaner, line 10:
```suggestion
    "generate-vdm": "yarn generate-odata-client -i ../../../test-resources/odata-service-specs-e2e/v4 -o .",
```

test-packages/e2e-tests/srv/test-service.cds
```diff
@@ -21,4 +21,6 @@ service TestService {
   action createTestEntityById(id : Integer) returns TestEntity;
   action createTestEntityByIdReturnId(id : Integer) returns Integer;
   action createTestEntityReturnId(param : TestEntity) returns Integer;
+
+  entity TestEntity50Col as projection on cloudsdk.TestEntity50Col;
```
marikaner, line 25: Minor: 50Col = 50 properties? Maybe there is a better name out there (but maybe not ;) )

test-packages/memory-tests/sdk-v1/test-script.js
```diff
@@ -0,0 +1,93 @@
+const memwatch = require('node-memwatch-new');
+const { writeFileSync } = require('fs');
+const { resolve } = require('path');
+const { TestEntity, TestEntity50Col } = require('./test-service');
+
+const dest = { url: 'http://localhost:4004/' };
```
marikaner, line 6:
```suggestion
const dest = { url: 'http://localhost:4004' };
```

test-packages/memory-tests/sdk-v1/test-script.js
```diff
@@ -0,0 +1,93 @@
...
+
+const test = async () => {
+  // The script is written in js (not ts) and executed with --inspect flag because of this lib.
+  const hd = new memwatch.HeapDiff();
+
+  const testEntities101 = await TestEntity.requestBuilder()
```
marikaner, line 12: Also here, can you iterate?

marikaner (changes requested):
Looks good, I am requesting changes mostly because of the code duplication.

test-packages/memory-tests/sdk-canary/test-script.js
jjtang1985: Here I would like to keep the new entities in the memory.
But you are right, I'll save them in an array in a loop.

scripts/compare-memory-usage.ts
jjtang1985: So you prefer -10% rather than 90%?

test-packages/memory-tests/sdk-canary/test-script.js
```diff
@@ -0,0 +1,115 @@
...
+const { testService } = require('./test-service');
+
+const dest = { url: 'http://localhost:4004/' };
+
+const test = async () => {
+  // The script is written in js (not ts) and executed with --inspect flag because of this lib.
+  const hd = new memwatch.HeapDiff();
+  // It is intended to call execute multiple times without using a loop, so the deserializer, the entity builder and the schema will still be kept in the memory.
+  const testEntityApi = testService().testEntityApi;
+  const testEntities101 = await testEntityApi
```
FrankEssenberger, line 13: As Marika pointed out I would use a loop and keep the testEntities in an array, so that the garbage collector can not clean them away

FrankEssenberger: Also if you use a larger array perhaps you can use a already existing service?

test-packages/memory-tests/sdk-v1/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@sap-cloud-sdk/memory-tests-v1",
+  "version": "0.0.1",
+  "description": "Memory usage tests of the SAP Cloud SDK for JavaScript",
+  "homepage": "https://sap.github.io/cloud-sdk/docs/js/overview-cloud-sdk-for-javascript",
+  "license": "Apache-2.0",
+  "private": true,
+  "repository": "github:SAP/cloud-sdk-js",
+  "scripts": {
+    "generate-vdm": "yarn generate-odata-client -i ../../../test-resources/odata-service-specs-e2e/v4 -o .",
```
FrankEssenberger, line 10: both tests put the code in the same folder - could this lead to conflicts or unexpected behaviour? I would perhaps create two folders with a name indicating canary and v1 and potentially even "memory tes"t in the name.

test-resources/odata-service-specs-e2e/v4/API_TEST_SRV/API-TEST_SRV.edmx
```diff
@@ -46,6 +47,62 @@
                 <Property Name="KeyToTestEntity" Type="Edm.Int32" Nullable="false"/>
                 <Property Name="StringProperty" Type="Edm.String" MaxLength="111"/>
             </EntityType>
+            <EntityType Name="TestEntity50Prop">
```
FrankEssenberger, line 50: perhaps name the spec file `50-properties-memeory-test-service` or so to make clear where or for what it is used. Also feel free to add a comment in the service above the entity to explain why so many entites

test-packages/memory-tests/sdk-v1/test-script.js
```diff
@@ -0,0 +1,93 @@
+const memwatch = require('node-memwatch-new');
+const { writeFileSync } = require('fs');
+const { resolve } = require('path');
+const { TestEntity, TestEntity50Prop } = require('./test-service');
```
FrankEssenberger, line 4: Question: How does this require work - there is no index.js or file/folder `test-service` next to the test-script file?

scripts/compare-memory-usage.ts
```diff
@@ -0,0 +1,18 @@
...
+      `Test failed: v1 memory usage is ${v1MemoryUsage} and canary memory usage is ${canaryMemoryUsage} (${percentageBasedOnV1} %). Consider changing the threshold (${threshold} %) or fix it.`
+    );
+  }
+  /* eslint-disable-next-line no-console */
+  console.log(
+    `Test passed: v1 memory usage is ${v1MemoryUsage} and canary memory usage is ${canaryMemoryUsage} (${percentageBasedOnV1} %).`
+  );
+};
+
+main();
```
FrankEssenberger, line 18: Currently the test-script.js are executed by the memory-test.yml and then they write their result to a file which is then piped into this script. It toook my a few minutes to figure this out and I knew that the memeory workflow is there. I would change it in the following way:
- the `compare-memory-usage.ts` is extended so that it executes the two `test-script.js` files. If you use execa in sequence the heap size should be not affected (not in prallel of course). 
- the `test-script.js` just print and return the used memory.
- the script collects the two values returned by execa execution -> no need to write files to the disk
- evaluation is done.

This would simplify the `memory-tests.yml` and put all the execution steps in one script - please note if that is not possible because the heap tests need a clean isolated execution please ignore this comment.

FrankEssenberger (changes requested):
Some questions but in general nice that we have a test and can be merged after questions are answered.

test-packages/memory-tests/sdk-canary/test-script.js
jjtang1985: > Also if you use a larger array perhaps you can use a already existing service?

I would prefer using a big entity, so you can clearly understand after one request, what's the delta changes in the memory.
If you only use the existing entity, everything seems to be flat.

test-packages/memory-tests/sdk-v1/package.json
jjtang1985: - v1 tests will be generated in `memory-tests/sdk-v1`
- canary tests will be generated in `memory-tests/sdk-canary`

test-packages/memory-tests/sdk-canary/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@sap-cloud-sdk/memory-tests-canary",
+  "version": "0.0.1",
+  "description": "Memory usage tests of the SAP Cloud SDK for JavaScript",
+  "homepage": "https://sap.github.io/cloud-sdk/docs/js/overview-cloud-sdk-for-javascript",
+  "license": "Apache-2.0",
+  "private": true,
+  "repository": "github:SAP/cloud-sdk-js",
+  "scripts": {
+    "generate-vdm": "yarn generate-odata-client -i ../../../test-resources/odata-service-specs-e2e/v4 -o .",
```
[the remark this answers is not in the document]
jjtang1985: No, this will use the one installed from npm.

test-resources/odata-service-specs-e2e/v4/API_TEST_SRV/API-TEST_SRV.edmx
jjtang1985: I extended the spec for the cap e2e tests by adding a new entity.
I can add a comment.

test-packages/memory-tests/sdk-v1/package.json
FrankEssenberger: sorry I missed the `.` as the location. The reason you do not check it it is that it used canary generator right?

scripts/compare-memory-usage.ts
jjtang1985: Discussed

FrankEssenberger (approved):
DIscussed the point. A question regarding the execution why do we have scheduled the job? If we have no code changes (PRs) I would expect the values remain unchanged.

test-packages/memory-tests/sdk-v1/package.json
jjtang1985: Yes, that's the major reason.
Also, we only commit clients for the source code version, so v1 client is not committed.

jjtang1985:
> A question regarding the execution why do we have scheduled the job? If we have no code changes (PRs) I would expect the values remain unchanged.

Because this test is for canary version, which is not related to any PRs.
Running on PRs does not test anything.
Also, in general, we should not change the memory consumption frequently.
Having nightly job should be enough and we can save some resources.

PR #2395, opened 2022-04-20
feat: support custom function import deserialization by transforming the response

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#73.

For function/action imports, it is now possible to deserialize custom responses.

For example:
```json
{
  "d": {
    "GetAttachmentCount": {
      "AttachmentCount": 0
    }
  }
}
```

Since `AttachmentCount` is wrapped by an unexpected attribute `GetAttachmentCount`, it would return `undefined` previously due to the failed match of the entity. 

Now you can use
```ts
 const request = functionImports
      .getAttachmentCount({param})
      .execute(destination, data => data.d.GetAttachmentCount)
```
to transform the response data first in order to deserialize from the correct position. 

Internally, the data will be transformed into
```json
{
  "d": {
    "AttachmentCount": 0
  }
}
```
which is still wrapped by `{ "d": { ... } }` so that the previous deserializer can still work correctly. 

This extension can do a lot more than just change the base. The response data can be altered into any value or form you want by specifying the `dataAccessor`.


packages/odata-common/src/request-builder/action-function-import-request-builder-base.ts
```diff
@@ -27,12 +27,19 @@ export abstract class ActionFunctionImportRequestBuilderBase<
...
+   * @param dataTransformer - A function to transform the received response data.
    * @returns A promise resolving to the requested return type.
    */
-  async execute(destination: DestinationOrFetchOptions): Promise<ReturnT> {
-    return this.executeRaw(destination).then(response =>
-      this.responseTransformer(response.data)
-    );
+  async execute(
+    destination: DestinationOrFetchOptions,
+    dataTransformer?: (data: any) => any
```
marikaner, line 35: Consider renaming this to `dataAccessor`, because it might otherwise be confused with what the original response transformer does.

marikaner (changes requested):
I just left one minor comment, looks very good, I think this is the right way to go. I am wondering why you chose to make the test an integration test. I think this should rather be a unit test (and I saw that there is no test file for this yet, so you would have to create one).

packages/odata-common/src/request-builder/action-function-import-request-builder-base.ts
ZhongpinWang: Good suggestion and done! I am a bit lack in creativity here. Theoretically, I would say `responseTransformer` is the best candidate, but it is used. 😆

ZhongpinWang:
> I just left one minor comment, looks very good, I think this is the right way to go. I am wondering why you chose to make the test an integration test. I think this should rather be a unit test (and I saw that there is no test file for this yet, so you would have to create one).

I removed the integration test with a new unit test. It is much light-weight now.

packages/odata-v2/src/request-builder/function-import-request-builder.spec.ts
```diff
@@ -191,6 +191,26 @@ describe('FunctionImportRequestBuilder', () => {
     ).rejects.toThrow();
   });
 
+  it('returns sigle complex type when using data accessor', async () => {
```
deekshas8, line 194:
```suggestion
  it('returns single complex type when using data accessor', async () => {
```

marikaner (changes requested):
We're almost there, but now the test is in the v2 package while we're testing common functionality. It should be possible to test this in common. @FrankEssenberger or am I wrong?

FrankEssenberger:
> We're almost there, but now the test is in the v2 package while we're testing common functionality. It should be possible to test this in common. @FrankEssenberger or am I wrong?

You are right in principle. However, the [common entity for testing](https://github.com/SAP/cloud-sdk-js/blob/main/packages/odata-common/test/COMMON_SRV.edmx) does not contain a function/action import. To test it in common one would need to add this to the service spec and adjust the [generation script](https://github.com/SAP/cloud-sdk-js/blob/main/packages/odata-common/test/generate-common-entity.ts) so that the function import related things are store either in a seperate file or a in the one where [the rest](https://github.com/SAP/cloud-sdk-js/blob/main/packages/odata-common/test/common-entity.ts) is located.

marikaner:
> > We're almost there, but now the test is in the v2 package while we're testing common functionality. It should be possible to test this in common. @FrankEssenberger or am I wrong?
> 
> You are right in principle. However, the [common entity for testing](https://github.com/SAP/cloud-sdk-js/blob/main/packages/odata-common/test/COMMON_SRV.edmx) does not contain a function/action import. To test it in common one would need to add this to the service spec and adjust the [generation script](https://github.com/SAP/cloud-sdk-js/blob/main/packages/odata-common/test/generate-common-entity.ts) so that the function import related things are store either in a seperate file or a in the one where [the rest](https://github.com/SAP/cloud-sdk-js/blob/main/packages/odata-common/test/common-entity.ts) is located.

I'll create a ticket for us to do this, but I agree that this is out of scope for this task.

marikaner (approved):
👍

PR #2400, opened 2022-04-21
chore: Improve stats script

marikaner:
I missed some comments in the last review. Those are the requested improvements. I did not add any pushing to a different repository, because it is out of scope in my opinion and should be a follow up if necessary.


florian-richter (approved):
LGTM, thanks for adding the changes 😄

PR #2408, opened 2022-04-25
chore: fix code snippet formatting issues

deekshas8:
Closes SAP/cloud-sdk-backlog#673.

Fixed formatting on APIHub of our code snippet example.


FrankEssenberger (approved):
LGTM always remember that most of the old metadata generated here is not used but set in the API hub on ther server side of the web application. But I guess for the code snippests this is not the case.

PR #2409, opened 2022-04-25
chore: Remove peerDeps in generated clients

florian-richter:
The peerDependencies are not necessary as the same dependencies are declared already.


marikaner (approved):
You'll have to regenerate

PR #2417, opened 2022-04-26
fix: type tests

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#654.

The type test was always successful due to an incorrect configuration in `tsconfig.json`. There was this line in the file

```json
"include": ["../../../../test-resources/test/test-util"]
```

Problems are:

1. The relative path should be `../../../test-resources/test/test-util`. 
2. **If we decided to use `include`, don't forget to add `**/*` for the current folder as well, which is set by default if `include` is not used.** The test files were not built at all due to this problem. (I missed this as well and the comment under a [post](https://stackoverflow.com/a/59056128/18544589) from StackOverflow saved my day!!)
3. **Actually, we don't need to include `test-util` right? All the helper files should be linked and compiled automatically. (Or am I wrong?)**

Problem "1" and a lot of `Errors in typescript@local for external dependencies` (~~mostly implicit `any`~~ We use `noImplicitAny: false` to bypass it now) needed to be solved before I could move forward to "2" and "3", because `dtslint` won't run type tests if there is still any error left.

~~**Attention:** In the fix for all these TS errors, you may notice that I used `require('@sap/xssec')` and suppressed the eslint error in `connectivity/src/scp-cf/jwt.ts` and `connectivity/src/scp-cf/xsuaa-service.ts` since this package has no type declaration and `dstlint` keeps complaining about it. I cannot find a proper workaround other than this to make both `eslint` and `dtslint` happy. Please if you have any good suggestions, leave a comment below.~~

After a short discussion with @marikaner, it is decided to not check implicit any, both for variables and package imports.

----

Since [dtslint](https://github.com/microsoft/dtslint) has been archived and moved as a package of the monorepo [DefinitelyTyped-tools](https://github.com/microsoft/DefinitelyTyped-tools), I decided to update the dependency as well.

----

After all these changes, `dtslint` finally works! Most type tests are fixed except for one in `test-packages/type-tests/test/v4/filters.spec.ts`. Check the `TODO` in the file or the follow-up ticket SAP/cloud-sdk-backlog#691 for more details.

An additional self-check for type tests is added in order to make sure that type tests are really working. Since `dtslint` may have changes, which could break the API, this self-check can notify us immediately. (Btw, I am a bit lack of imagination for the naming and phrasing for this checker, so if you have a better suggestion for the name or logs, feel free to leave a comment).


packages/connectivity/package.json
```diff
@@ -35,6 +35,7 @@
     "@sap-cloud-sdk/util": "^2.3.0",
     "@sap/xsenv": "^3.0.0",
     "@sap/xssec": "^3.2.7",
+    "@types/sap__xsenv": "^2.0.3",
```
marikaner, line 38: I believe this could be a dev dependency

packages/connectivity/src/scp-cf/environment-accessor.ts
```diff
@@ -87,7 +87,7 @@ export function getServiceCredentialsList(service: string): any[] {
  * @internal
  */
 export function getServiceList(service: string): Service[] {
-  return xsenv.filterServices({ label: service });
+  return xsenv.filterCFServices({ label: service });
```
marikaner, line 90: Does the old function not exist?
The naming suggests that on k8s this won't work correctly.

packages/connectivity/src/scp-cf/environment-accessor.ts
```diff
@@ -150,10 +150,16 @@ export function getVcapService(): Record<string, any> | null {
...
+    if (error instanceof Error) {
+      throw new ErrorWithCause(
+        "Failed to parse environment variable 'VCAP_SERVICES'.",
+        error
+      );
+    } else {
+      throw new Error(
+        "Failed to parse environment variable 'VCAP_SERVICES' with unknown error cause."
+      );
+    }
```
marikaner, line 162: I wouldn't mention the unknown error cause. Why is this even necessary?

marikaner (changes requested):
I left some first comments, but I have the impression, that the problems that you were facing might be solved if you use [tsd](https://www.npmjs.com/package/tsd) instead of dtslint. I have not used this before but seems like dtslint has some more "features" that we don't need. Also the download numbers on npm are quite low.

ZhongpinWang:
> I left some first comments, but I have the impression, that the problems that you were facing might be solved if you use [tsd](https://www.npmjs.com/package/tsd) instead of dtslint. I have not used this before but seems like dtslint has some more "features" that we don't need. Also the download numbers on npm are quite low.

After the discussion, @marikaner and I decided to use `dtslint` for now. If we encountered any issues in the future, we should consider switching to [tsd](https://www.npmjs.com/package/tsd).

test-packages/type-tests/test/v4/filters.spec.ts
```diff
@@ -2,20 +2,37 @@ import {
...
-// $ExpectType FilterList<TestEntity<DeSerializers<string, boolean, number, BigNumber, number, number, number, number, BigNumber, string, number, number, string, any, Moment, Moment, Duration, Time, any>>, DeSerializers<string, boolean, number, BigNumber, number, number, number, number, BigNumber, string, number, number, string, any, Moment, Moment, Duration, Time, any>>
-and(
-  schema.TO_MULTI_LINK.filter(
-    any(multiLinkSchema.STRING_PROPERTY.equals('test'))
-  )
-);
+
+// -----------
+/*
+TODO: Fix the type issue for the following `and(...)` test
```
FrankEssenberger, line 12: @marikaner I discussed yesterday with @ZhongpinWang that I think the solution of this type issue is beyond this ticket and we should do it as a follow up so that we can get the majority of type tests back in the pipeline quickly.

test-packages/type-tests/self-test/check-if-type-tests-are-working.ts
```diff
@@ -0,0 +1,14 @@
...
+const logger = createLogger('check-if-type-tests-are-working');
+
+try {
+  execSync('dtslint --expectOnly --localTs ../../node_modules/typescript/lib self-test', { stdio: 'ignore' });
+} catch (error) {
+  logger.info('Type tests are working!');
+  process.exit(0);
+}
+
+logger.error('Type tests are not working! Obvious errors were passed!');
```
FrankEssenberger, line 13:
```suggestion
logger.error('Type tests are not working! Expected type error did not lead to a test failure!');
```

test-packages/type-tests/self-test/check-if-type-tests-are-working.ts
```diff
@@ -0,0 +1,14 @@
+import { execSync } from 'child_process';
+import { unixEOL, createLogger } from '@sap-cloud-sdk/util';
+
+const logger = createLogger('check-if-type-tests-are-working');
+
+try {
+  execSync('dtslint --expectOnly --localTs ../../node_modules/typescript/lib self-test', { stdio: 'ignore' });
+} catch (error) {
+  logger.info('Type tests are working!');
```
FrankEssenberger, line 9:
```suggestion
  logger.info('Faulty type tests \'error-test.spec.ts\' fails as expected.');
```

FrankEssenberger (approved):
LGTM - however please also discuss with @marikaner if it is ok to do the one test fix as a follow up.

jjtang1985 (approved):
LGTM
Vote for merging first, to unblock other tickets.

test-packages/type-tests/test/v4/filters.spec.ts
marikaner: Makes sense.

marikaner (approved):
Please check whether the tslint config can be removed, otherwise, great. I like the extra "self-test"!

test-packages/type-tests/test/tslint.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "dtslint/dtslint.json",
+  "extends": "@definitelytyped/dtslint/dtslint.json",
```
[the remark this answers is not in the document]
ZhongpinWang: I removed it. Thanks for pointing it out. 😄

PR #2425, opened 2022-04-27
fix: fix windows test

deekshas8:
Glob package included option `allowWindowsEscape` (from `minimist` package) which always treats `\` in patterns as escapes, not path separators. They have recommended to use `/` as path separators. 
Because of this, the `new GlobSync(pattern)` stopped working as expected and couldn't find any files.

### DIFF
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -1,5 +1,5 @@
 import { PathLike } from 'fs';
-import { resolve, dirname } from 'path';
+import { resolve, dirname, sep, posix } from 'path';
 import {
   createLogger,
   ErrorWithCause,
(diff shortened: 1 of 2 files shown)


packages/generator/src/generator.ts
```diff
@@ -199,7 +199,9 @@ async function generateAdditionalFiles(
     const additionalFilesDir = resolve(
       options.inputDir.toString(),
       options.additionalFiles
-    );
+    )
+      .split(sep)
+      .join(posix.sep);
```
florian-richter, line 204: [info] I would prefer `.replaceAll` but it is only available in node>=15...

florian-richter (approved):

PR #2432, opened 2022-04-29
chore: disable node 12 support

jjtang1985:
Apr. 30 is the last day of node 12 support.
Node [reference](https://nodejs.org/en/about/releases/).


florian-richter:
This was discussed in Slack (I'll send you a link to the conversation in our Slack DMs). My understanding was, that we should keep running node 12 until version 3 as not supporting it would be a breaking change (and not testing it may cause it to break unknowingly).

PR #2435, opened 2022-04-29
fix: Adjust how arguments are passed to jest

florian-richter:
Seems like yarn or turborepo behave differently on windows and need additional `--` to pass the args to the actual executable.

If you have a windows computer, feel free to test the following commands (they work on mac but that didnt help with my previous fix):
- `yarn test:unit --continue -- -- -- --json --outputFile=unit-test-output.json`
- `yarn test:integration -- -- -- --json --outputFile=int-test-output.json`


FrankEssenberger (approved):
I can not verify but the number of `--` gets a bit out of hand I must say even on linux systems....

marikaner (approved):

PR #2442, opened 2022-05-03
chore: disable memory tests from PR builds

jjtang1985:
It was enabled for testing on the initial PR, but should be removed now.


marikaner (approved):
👍

PR #2443, opened 2022-05-03
chore: revert the workspace scripts removal

jjtang1985:
The convenient scripts were removed.
I would like to keep them so I can e.g., start the e2e test app with a shorter command.


marikaner (approved):
👍

PR #2446, opened 2022-05-04
fix: fix tests on windows

deekshas8:
This PR fixes the syntax `--config='../../.prettierrc'` (which does not work on windows) to `--config ../../.prettierrc`.


FrankEssenberger (approved):
Strange but if it works on windows this way....

PR #2447, opened 2022-05-04
fix: canary release

FrankEssenberger:
Currently the canary release does nothing, because there is no [snapshot version id created](https://github.com/SAP/cloud-sdk-js/runs/6285802719?check_suite_focus=true):

image

In this PR I used the [build in function of changeset](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md).

However, this would lead to versions 0.0.0-{tag}-DATETIMESTAMP instead of 2.3.0-{tag}-DATETIMESTAMP. Since the tag is used to identify and all version are changed this should work:

image

However I am not sure if this a beautiful thing to do. On the other hand if we start having separate version on the different packages it could become difficult to take the latest version and add some timestamp postfix since each package potentially has a different version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,7 +129,7 @@ jobs:
         run: |
           echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' >> .npmrc
           date=`date +%Y%m%d%H%M%S`
-          rm .changeset *.md
+          rm .changeset/*.md
           cp canary-release-changeset.md .changeset
           yarn changeset pre enter {$date}
           yarn changeset version
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,7 +128,11 @@ jobs:
       - name: Canary Release
         run: |
           echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' >> .npmrc
-          yarn changeset version --snapshot canary
+          date=`date +%Y%m%d%H%M%S`
+          rm .changeset *.md
+          cp canary-release-changeset.md .changeset
+          yarn changeset pre enter {$date}
+          yarn changeset version
           yarn changeset publish --tag canary
         env:
           NPM_TOKEN: ${{ secrets.NPMJS_ACCESS_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -128,7 +128,7 @@ jobs:
       - name: Canary Release
         run: |
           echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' >> .npmrc
-          date=`date +%Y%m%d%H%M%S`
+          yarn changeset version --snapshot canary
```
jjtang1985, line 131: I would use prerelease:
`yarn changeset pre enter ${date}`

See complete doc [here](https://github.com/changesets/changesets/blob/main/docs/prereleases.md)

jjtang1985 (changes requested):
Please check the prerelease example.

> On the other hand if we start having separate version on the different packages it could become difficult to take the latest version and add some timestamp postfix since each package potentially has a different version.

I'm not a big fan of having this.  Why this can be introduced? because of changeset?

.github/workflows/build.yml
FrankEssenberger: I tried this and it looks promising, however when I do anythong with changeset it wants to bump everything by a major version `changeset status --verbose` tells me everything major, although only three packages are updated with minor. I do not understand this at all here is some explainaition: https://github.com/changesets/changesets/issues/524

Adding the experimental options avoids the major version bump but in the prerelease mode it still wants to add amajor version bump. Really puzzled...

FrankEssenberger: I think, I got it working using the following way:
- Remove all the existing changeset
- Copy a dummy changeset doing a patch version increase on all packages
- Enter the prerelease mode with tag the date as @jjtang1985 suggested
- Update the version 
- The changes should be not checked in of course but this is not the case in the pipeline I think

marikaner (approved):
Seems like a necessary evil. I am not sure that changesets is what we ultimately should use...

FrankEssenberger:
> Seems like a necessary evil. I am not sure that changesets is what we ultimately should use...

I think the tool is good and would avoid bad things from happening in principle, which you only have if you do not bump everything with the same version. So if we just publish always everything with the same version we would not need something fancy.

jjtang1985 (approved):
LGTM, as long as you tested it.
Question: does the new canary version look differently (with different pattern)?
If yes, perhaps we need to release a new stable version to avoid risks.

FrankEssenberger:
>LGTM, as long as you tested it. 

I did locally the steps as in the script and skipped the publish but this should be untouched. 🤞

PR #2451, opened 2022-05-05
fix: canary release 2

FrankEssenberger:
Sorry I had a type with the `{` and `$` sign leading a broken version:

image

Using the [old way ${date}](https://github.com/SAP/cloud-sdk-js/blob/aef7a5d340de0eb2c04bbd2d93cdfe3b29aaec6f/.github/workflows/build.yml#L128) it should work.


jjtang1985 (approved):
LGTM

PR #2453, opened 2022-05-05
fix: canary release 3

FrankEssenberger:
Sorry one more iteration is needed. 
Turn out you can not publish with a custom tag while you are in prerelease mode. 
Since the package.json are adjusted anyway you need to exit pre release mode before:

image


jjtang1985 (approved):
LGTM, in terms of the understanding of the error message.

PR #2458, opened 2022-05-09
fix: Remove `force` flag for `lint:fix` script

florian-richter:
Quote from below with the current status:

> I made some improvements but the bug that Zhongpin found persists. I think this is a problem with turborepo (or the interaction between eslint --fix and turbo...). I opened an issue for turbo here: https://github.com/vercel/turborepo/issues/1220
> 
> This PR will not solve this but may still be worth to be merged in of itself. The remaining issue is an edge case which should happen rarely and can be fixed by running yarn lint:fix --force.

If you are annoyed by the "file mode 100644 -> 100755" changes, you can upvote the [turborepo issue](https://github.com/vercel/turborepo/issues/156). 😉


ZhongpinWang (commented):
I think you can reproduce the problem with the following steps:

1. Run `rm -rf ./node_modules/.cache` in the root.
2. Run `yarn install`.
3. Run `echo '' >> packages/generator/src/internal.ts`, which will append only one newline. 
4. Run `yarn lint:fix` -> Fixed as intended.
5. Run `echo '' >> packages/generator/src/internal.ts` again. 
6. Run `yarn lint:fix` -> Not fixed.

~~Btw, it seems that only deleting the last newline or appending exactly one new line will cause the problem.~~

~~This looks like an edge case that needs to be fixed by turborepo. Maybe we can ignore this and remove `--force` to save some time.~~

(No longer valid, see the next comment)

ZhongpinWang:
I think the issue may not be restricted to newline after all. 

Generally, the issue happens only if we go through the following steps:

1. Run `yarn lint:fix`;
2. Revert changes back;
3. Run `yarn lint:fix` again;
4. Cache will be used since the hash of the state remains the same. Thus, no fix will be done again.

It means that the same state of each package will only be fixed once.

florian-richter:
Good point that the cache may be wrong. I will fix the `outputs` part so the fixed code is cached as well! That should resolve the issue.

florian-richter:
I made some improvements but the bug that Zhongpin found persists. I think this is a problem with turborepo (or the interaction between eslint --fix and turbo...). I opened an issue for turbo here: https://github.com/vercel/turborepo/issues/1220

This PR will not solve this but may still be worth to be merged in of itself. The remaining issue is an edge case which should happen rarely and can be fixed by running `yarn lint:fix --force`.

packages/connectivity/package.json
```diff
@@ -31,7 +31,7 @@
     "test:unit": "jest",
     "coverage": "jest --coverage",
     "lint": "eslint --ext .ts . && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -c",
-    "lint:fix": "eslint --ext .ts . --fix && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w",
+    "lint:fix": "TIMING=1 eslint --ext .ts . --fix --quiet && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w --loglevel error",
```
tomfrenken, line 34: So what does this TIMING=1 do?
Other than that LGTM

tomfrenken (approved):

packages/connectivity/package.json
FrankEssenberger: @florian-richter could you enlighten us?

florian-richter: This was a recommendation in the turbo docs to make linting output nicer. It sets an environment variable for the time of this command which causes eslint to show a table of the longest running eslint tasks. I found it helpful or at least unobtrusive so I kept it.

PR #2464, opened 2022-05-10
fix: re-export variadicArgumentToArray

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#684.


test-packages/test-services-e2e/package.json
```diff
@@ -23,5 +24,10 @@
   "dependencies": {
     "bignumber.js": "^9.0.1",
     "moment": "^2.29.0"
+  },
+  "peerDependencies": {
+    "@sap-cloud-sdk/generator": "^2.3.0",
+    "@sap-cloud-sdk/util": "^2.3.0",
+    "@sap-cloud-sdk/odata-v4": "^2.3.0"
```
florian-richter, line 31: [q] Is this intended?

florian-richter (approved):
The added `depcheck` and `peerDeps` are out-of-scope for this PR. Since the PR is otherwise small and the tests are green, I am okay to merge it like this. In the future, it would be better if this was split into two PRs (... I am aware that I am a bad role model in this regard 🙈)

test-packages/test-services-e2e/package.json
ZhongpinWang: It is added because of the `depcheck`. I think it also makes sense to add it as "peerDependencies" since it will require the package that is using `test-services-e2e` to have these three dependencies, in our case `e2e-test` needs to have these dependencies in order to use the test services.

Issue #1638, opened 2021-09-29
Support Trust Store certificates of Destination service


florian-richter:
An internal user who uses CAP wants to inject the server certificate from S/4HANA which currently fails with following error:

```
[cds] - Error: Error during request to remote service: self signed certificate
```

We are clarifying if/how to configure the necessary certificate in CAP or if this is something the SAP Cloud SDK could pick up automatically from the destination.

Link to internal issue: https://github.wdf.sap.corp/cap/issues/issues/9605

gregorwolf:
As this is a public repository it would be great if you can share the details and not only a link to a SAP internal repository.

florian-richter:
I gave some context in the description above.

FrankEssenberger:
So we started finally working on it and I did some research I can as well give here to provide some more context. If the server uses a self-signed cert the clients will not trust it. In such a case you can:

- Enable trust all -> this is risky and you should not do it. Even in internal or test cases. Man in the middle attacks and somehow dangerous code can end up in produciton
- Include the self-signed CA via NODE_EXTRA_CA_CERTS env variable. The node client will pick it up
- Include the self-signed CA via agent property in [your code](https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions)

We will implement the last one in the SDK and take the value from the destination property [TrustStoreLocation](https://help.sap.com/docs/CP_CONNECTIVITY/cca91383641e40ffbe03bdc78f00f681/e75d7f15b808428e87ae70f494da0b42.html). In the first version we will not support `jks` files like for two way TLS. Only the naitively support formats in node which is `pem` in this case. However, once done you should be able to use the SDK with self-signed certificates if you add them to the destination.

Best
Frank

FrankEssenberger:
Implementation is done and also made one additional E2E test with a colleague needing this for a demo using private links and non CA certificates. It worked even with certificates chains. Closing the issue. If there are bugs in the implementation please open a new issue.

PR #2465, opened 2022-05-10
fix parsing issue for truststore destinations

FrankEssenberger:
I had a session with Harut today and there was some runtime error. I made one test, could reproduce it and fixed it. The reason was that the destination service returns keys with capital properties and we need to parse it. Since I mocked all the requests the error did not occur.

Relates to: #1638


packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -307,20 +314,22 @@ async function callDestinationService(
...
 function getCircuitBreaker<T>(): DestinationCircuitBreaker<
-  DestinationCertificate | DestinationConfiguration | DestinationJson
+  DestinationCertificateJson | DestinationConfiguration | DestinationJson
 > {
   const request: (
     config: AxiosRequestConfig
   ) => Promise<
-    AxiosResponse<DestinationCertificate | DestinationJson | DestinationJson>
+    AxiosResponse<
+      DestinationCertificateJson | DestinationJson | DestinationJson
```
marikaner, line 331:
```suggestion
      DestinationCertificateJson | DestinationJson
```

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -254,18 +255,24 @@ function errorMessageFromResponse(
     : '';
 }
 
+interface DestinationCertificateJson {
+  Type: string;
+  Name: string;
+  Content: string;
+}
```
marikaner, line 262: [nth] This is the same type as the `DestinationCertificate` - I believe it could be closer to each other so that the difference is obvious.
[nth] Also, because the types are almost the same, check this out: https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#key-remapping-via-as.

marikaner (commented):
Looks good, just minor things.

packages/connectivity/src/scp-cf/destination/destination.ts
```diff
@@ -240,7 +240,13 @@ function setDefaultAuthenticationFallback(
     : assoc('authentication', getAuthenticationType(destination), destination);
 }
 
-function parseCertificate(
+/**
+ * Transforms the lower case variables of the destination reposne to lower case.
```
marikaner, line 244:
```suggestion
 * Transforms the lower case variables of the destination response to lower case.
```

packages/connectivity/src/scp-cf/destination/destination.ts
```diff
@@ -240,7 +240,13 @@ function setDefaultAuthenticationFallback(
     : assoc('authentication', getAuthenticationType(destination), destination);
 }
 
-function parseCertificate(
+/**
+ * Transforms the lower case variables of the destination response to lower case.
+ * @internal
+ * @param certificate
+ * @returns The parsed destination certificate.
+ */
```
marikaner, line 248: This seems incomplete and like a copy paste error.

marikaner (approved):
LGTM

packages/connectivity/src/scp-cf/destination/destination.ts
FrankEssenberger: I just checked and the analogous method for destination is also called parseDestiantion

PR #2486, opened 2022-05-18
chore: change log levels

tomfrenken:
This PR changes the log levels of multiple of our functions to `debug` and some of them to `warn`.
Notice that there are still some `info` level loggers when you are reviewing this PR.

Ticket SAP/cloud-sdk-backlog#681.

Please review if the log level changes fit the message context and if the remaining `info` level loggers are necessary.


packages/connectivity/src/scp-cf/environment-accessor.ts
```diff
@@ -177,7 +177,7 @@ export function getEnvironmentVariable(
   if (process.env[name]) {
     return process.env[name];
   }
-  logger.info('Environment variable ' + name + ' is not defined.');
+  logger.warn('Environment variable ' + name + ' is not defined.');
```
jjtang1985, line 180: [nth] The format of the line 171 and line 172 seems to be strange. Could you check locally?

jjtang1985 (changes requested):
Theoretically, this is a breaking change, as some of the logs will disappear, after version bump.
Please adjust the changelog/changesets for the updates.

packages/connectivity/src/scp-cf/environment-accessor.ts
FrankEssenberger: [nth] Also you could take the opportunity to change the `+` to string template.

.changeset/fresh-cups-share.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+'@sap-cloud-sdk/http-client': minor
+---
+
+[Improvement] Default log output on the `info` level has been significantly reduced.
```
marikaner, line 6: [req] Please use the imperative in the changelog. You can find some examples in the changelog file in the root of the project.
@florian-richter will this be moved somewhere else with the new approach?

florian-richter: Nothing changed on the changesets-side of things (other than requiring the tags to be correct later down the line)

jjtang1985 (approved):
LGTM

packages/http-client/src/http-client.ts
```diff
@@ -275,7 +275,7 @@ function splitRequestConfig(requestConfig: HttpRequestConfigWithOrigin): {
 
 function logCustomHeadersWarning(customHeaders?: Record<string, string>) {
   if (customHeaders) {
-    logger.info(
+    logger.warn(
```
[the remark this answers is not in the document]
marikaner: Potentially this could even be debug...

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -273,7 +273,7 @@ class DestinationFromServiceRetriever {
       );
     }
     if (credentials.length > 1) {
-      logger.info(
+      logger.debug(
```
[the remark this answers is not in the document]
FrankEssenberger: I guess if there are multiple destination service instance subscribed it will not hurt in general all will return the subaccount destinations. It is indicating some misconfiguration in most cases so putting it to warn would be also fine to indicate the random behavior. 

=> In favor for warn.

packages/http-client/src/http-client.ts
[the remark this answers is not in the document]
FrankEssenberger: yes debug because the custom headers are sometimes used intentionally to overwrite the default. Also I think the message needs to be adjust, because it will overwrite only if the headers are set by SDK. So `custom headers with highest priorites are: XYZ, they will overwrite headers set by the SDK` or something.

PR #2492, opened 2022-05-20
fix: reuse check

jjtang1985:
When running the check just after checkout, the irrelavant files (node_modules) would not be a problem.
See example [here](https://github.com/fsfe/reuse-action#example-usage).


florian-richter (approved):

marikaner (approved):
👍 Nice catch

PR #2500, opened 2022-05-24
fix: do not fail if files are not there

FrankEssenberger:
We had a small mistake, that if no changeset is there the `rm` fails in the canary release. This adds the `-f` option to it so this does not happen.


deekshas8 (approved):
LGTM

PR #2501, opened 2022-05-24
fix: update cert for e2e test

deekshas8:
Few dependabot PRs failed because of cert expiration. By default, the expiration was 30 days. Renewed cert with 2 yrs expiration.


test-packages/e2e-tests/test/http-agent.spec.ts
```diff
@@ -1,43 +1,42 @@
 import { createServer, Server } from 'https';
 import { promisify } from 'util';
-import {DestinationCertificate} from "@sap-cloud-sdk/connectivity/dist/scp-cf";
-import {Destination} from "@sap-cloud-sdk/connectivity";
-import {executeHttpRequest} from "@sap-cloud-sdk/http-client";
+import { DestinationCertificate } from '@sap-cloud-sdk/connectivity/dist/scp-cf';
```
jjtang1985, line 3: [q] It seems this file was not under the radar of lint, as it did not fail.
How did you fix the lint? Manual fix?
Should we consider adding specs to the lint scope?

deekshas8: I fixed it manually here. 
The package.json in e2e-tests project doesn't have a lint command in scripts. I'll add it.

FrankEssenberger: I think added the certificate here for testing and did not set a very long lifetime. Sorry for that. Please use a self signed certificate with maximum lifetime here.

PR #2508, opened 2022-05-25
fix: add EnumT for v4 service generator

deekshas8:
- Added the `EnumT` type that was missing in service builder for v4. Found this while initially investigating the issue about `and` filter SAP/cloud-sdk-backlog#691
- Fixed filter tests for v2 and added few `and` tests


FrankEssenberger (approved):
LGTM

FrankEssenberger:
[q] @deekshas8 was this related to a customer error beeing reported or to any BLI? If so please link the issue in the back log or create one so that we see the process.

deekshas8:
> [q] @deekshas8 was this related to a customer error being reported or to any BLI? If so please link the issue in the back log or create one so that we see the process.

This was not reported by a customer, but rather relates to SAP/cloud-sdk-backlog#691. I have updated the PR description to reflect that.

PR #2514, opened 2022-05-30
chore: add a script for npm package deprecation

jjtang1985:
This script was used for deprecating all v1 core packages, which will be useful in the future (e.g., deprecating v2 cores).

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,8 +68,7 @@
     "temporal-de-serializers": "yarn workspace @sap-cloud-sdk/temporal-de-serializers",
     "e2e-tests": "yarn workspace @sap-cloud-sdk/e2e-tests",
     "integration-tests": "yarn workspace @sap-cloud-sdk/integration-tests",
-    "type-tests": "yarn workspace @sap-cloud-sdk/type-tests",
-    "deprecate-npm-packages": "ts-node scripts/deprecate-npm-package.ts"
+    "type-tests": "yarn workspace @sap-cloud-sdk/type-tests"
   },
   "devDependencies": {
     "@changesets/cli": "^2.22.0",
--- a/package.json
+++ b/package.json
@@ -68,7 +68,8 @@
     "temporal-de-serializers": "yarn workspace @sap-cloud-sdk/temporal-de-serializers",
     "e2e-tests": "yarn workspace @sap-cloud-sdk/e2e-tests",
     "integration-tests": "yarn workspace @sap-cloud-sdk/integration-tests",
-    "type-tests": "yarn workspace @sap-cloud-sdk/type-tests"
+    "type-tests": "yarn workspace @sap-cloud-sdk/type-tests",
+    "deprecate-npm-packages": "ts-node scripts/deprecate-npm-package.ts"
   },
   "devDependencies": {
     "@changesets/cli": "^2.22.0",
--- a/scripts/deprecate-npm-package.ts
+++ b/scripts/deprecate-npm-package.ts
@@ -1,0 +1,130 @@
+import execa from "execa";
+import semver from "semver";
+const retry =  require("async-retry");
+
+interface DeprecateOptions {
+  deprecateMessage: string;
+  packageNamesToBeDeprecated: string[];
+  versionsToBeDeprecated: string;
+}
+
+const deprecateOptions: DeprecateOptions = {
+  deprecateMessage: "1.x is no longer maintained.",
+  packageNamesToBeDeprecated: [
+    "@sap-cloud-sdk/analytics",
+    "@sap-cloud-sdk/core",
+    "@sap-cloud-sdk/eslint-config",
+    "@sap-cloud-sdk/generator",
+    "@sap-cloud-sdk/generator-common",
+    "@sap-cloud-sdk/openapi-generator",
+    "@sap-cloud-sdk/test-util",
+    "@sap-cloud-sdk/util"
+  ],
+  // todo ^1.17.4-0 for all alpha versions
+  versionsToBeDeprecated: `<1.54.0`,
+  // versionTobeTested: '1.54.0'
+}
+// deprecate command with semver - npm deprecate @sap-cloud-sdk/openapi-generator@"< 1.54.3" "1.x is no longer maintained."
+// check whether a package/version is deprecated - npm view @sap-cloud-sdk/openapi-generator@"< 1.54.2"
+// check all the versions - `npm view @sap-cloud-sdk/util versions`
+async function deprecateNmpPackage(){
+  const responses = await Promise.all(deprecateOptions.packageNamesToBeDeprecated.map(
+    async packageName => {
+      const resVersions = await execa('npm', ['view', packageName, 'versions']);
+      const allVersions = resVersions.stdout.replace(/'|\s|\[|]/g, '').split(',');
+      const uniqueVersionPrefixes = Array.from(new Set(allVersions.map(version => {
+          // Pick prefix of a version. e.g., 1.34.1-20201230083713.13 will be 1.34.1
+          const prefix = version.match(/(1\.\d+\.\d+).*/)
+          let ret = prefix? prefix[1] : undefined;
+          // ret = (ret && semver.gt(ret, '1.36.2')) ? ret : undefined;
+          return ret;
+        }
+      ).filter(e => !!e))) as string[];
+
+      const deprecateVersion = uniqueVersionPrefixes.map(prefix => `~${prefix}-0`).join(' || ');
+      console.log(`[Deprecating]: ${packageName}@${deprecateVersion}`);
+      try {
+        await execa('npm', ['deprecate', `${packageName}@${deprecateVersion}`, deprecateOptions.deprecateMessage]);
+      } catch (e) {
+        // 422 means this version was deprecated before
+        if(!e?.message?.includes('422 Unprocessable Entity')){
+          throw new Error(e);
+        }
+      }
+
+      /**
+        `~1.0.0-0` will match:
+        - `1.0.0-xyz` (canary version)
+        - `1.0.x` (stable version)
+        but NOT 1.0.1-xyz.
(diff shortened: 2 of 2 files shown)


scripts/deprecate-npm-package.ts
```diff
@@ -0,0 +1,96 @@
...
+      const allVersions = resVersions.stdout.replace(/'|\s|\[|]/g, '').split(',');
+      const uniqueVersionPrefixes = Array.from(new Set(allVersions.map(version => {
+          // Pick the prefix of a version. e.g., 1.34.1-20201230083713.13 will be 1.34.1
+          const prefix = version.match(deprecateOptions.versionsToBeDeprecated)
+          return prefix? prefix[1] : undefined;
+        }
+      ).filter(e => !!e))) as string[];
+      // Build a command like `npm deprecate @sap-cloud-sdk/generator@~1.17.4-0 || ~1.18.0-0 || ~1.18.1-0 || ~1.18.2-0`
+      // With `-0` as suffix, it should match all our canary versions for A specific stable version.
+      const deprecateVersion = uniqueVersionPrefixes.map(prefix => `~${prefix}-0`).join(' || ');
```
FrankEssenberger, line 75: also here add a helper method `deprecateCanaryVersions` with the list of strings as input.

scripts/deprecate-npm-package.ts
```diff
@@ -0,0 +1,96 @@
...
+ *  Therefore, to match both versions below and you need a union version (~1.0.0-0 || ~1.0.1-0) or multiple commands:
+ *  - `1.0.0-xyz`
+ *  - `1.0.1-xyz`
+ *  We should not use multiple commands, as the server has certain limit.
+ */
+const deprecateOptions: DeprecateOptions = {
+  /**
+   * Deprecate message for all packages/versions to be deprecated.
+   */
+  deprecateMessage: "1.x is no longer maintained.",
```
FrankEssenberger, line 27: Perhaps use a bit longer message like - `1.x is no longer maintained. Please migrate to the latest version of the SAP Cloud SDK for JS.`

scripts/deprecate-npm-package.ts
```diff
@@ -0,0 +1,96 @@
...
+      const resVersions = await execa('npm', ['view', packageName, 'versions']);
+      const allVersions = resVersions.stdout.replace(/'|\s|\[|]/g, '').split(',');
+      const uniqueVersionPrefixes = Array.from(new Set(allVersions.map(version => {
+          // Pick the prefix of a version. e.g., 1.34.1-20201230083713.13 will be 1.34.1
+          const prefix = version.match(deprecateOptions.versionsToBeDeprecated)
+          return prefix? prefix[1] : undefined;
+        }
+      ).filter(e => !!e))) as string[];
+      // Build a command like `npm deprecate @sap-cloud-sdk/generator@~1.17.4-0 || ~1.18.0-0 || ~1.18.1-0 || ~1.18.2-0`
+      // With `-0` as suffix, it should match all our canary versions for A specific stable version.
```
FrankEssenberger, line 74: [q] If I see it right this script will only deprecate the canary versions - Should the stable versions also be deprecated?

scripts/deprecate-npm-package.ts
```diff
@@ -0,0 +1,96 @@
...
+      const deprecateVersion = uniqueVersionPrefixes.map(prefix => `~${prefix}-0`).join(' || ');
+      console.log(`[Deprecating]: ${packageName}@${deprecateVersion}`);
+      if(deprecateOptions.isProduction) {
+        try {
+          await execa('npm', ['deprecate', `${packageName}@${deprecateVersion}`, deprecateOptions.deprecateMessage]);
+        } catch (e) {
+          // 422 means this version was deprecated before
+          if (!e?.message?.includes('422 Unprocessable Entity')) {
```
FrankEssenberger, line 82: [q] Why you want to fail if a already deprecated is deprecated again? I guess if you rerun the script multiple times due to failures this could become not nice, althought the `npm view versions`  most likely only returns non deprecated version right?

FrankEssenberger (dismissed):
Added two suggestions and some questions

scripts/deprecate-npm-package.ts
jjtang1985: The packages were deprecated.
This PR is for documentation purpose and can be reused in the future.
So this message is just an example.

scripts/deprecate-npm-package.ts
jjtang1985: The dynamic version (`^`/`~`) considers all the stable version, so yes.
Therefore, deprecating stable version is easier.

scripts/deprecate-npm-package.ts
jjtang1985: - when response is NOT 422, it will throw an error as I don't know what happens
- when response is 422, I treat that as successful

> althought the npm view versions most likely only returns non deprecated version right?

It will give me all the versions, including deprecated ones.

scripts/deprecate-npm-package.ts
```diff
@@ -0,0 +1,96 @@
...
+ * Main entry point for deprecating the SDK packages.
+ * Here are some npm command examples, that are related to the deprecation.
+ * [deprecate command with semver] - npm deprecate @sap-cloud-sdk/openapi-generator@"< 1.54.3" "1.x is no longer maintained."
+ * [check whether a package/version is deprecated] - npm view @sap-cloud-sdk/openapi-generator@"< 1.54.2"
+ * [check all existing versions of a package] - `npm view @sap-cloud-sdk/util versions`
+ */
+async function deprecateNmpPackage(){
+  const responses = await Promise.all(deprecateOptions.packageNamesToBeDeprecated.map(
+    async packageName => {
+      const resVersions = await execa('npm', ['view', packageName, 'versions']);
```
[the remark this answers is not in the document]
jjtang1985: It makes sense for readability.
However, as there is not automated tests, and it was only tested by executing it, I would avoid any refactoring for now.

scripts/deprecate-npm-package.ts
jjtang1985: Same here.

scripts/deprecate-npm-package.ts
```diff
@@ -0,0 +1,96 @@
...
+const deprecateOptions: DeprecateOptions = {
+  /**
+   * Deprecate message for all packages/versions to be deprecated.
+   */
+  deprecateMessage: "1.x is no longer maintained.",
+  /**
+   * Package names to be deprecated
+   */
+  packageNamesToBeDeprecated: [
+    "@sap-cloud-sdk/analytics",
```
deekshas8, line 32: It appears the file is not linted correctly, aren't `.ts` files in `scripts/` included?

scripts/deprecate-npm-package.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Main entry point for deprecating the SDK packages.
+ * Here are some npm command examples, that are related to the deprecation.
+ * [deprecate command with semver] - npm deprecate @sap-cloud-sdk/openapi-generator@"< 1.54.3" "1.x is no longer maintained."
+ * [check whether a package/version is deprecated] - npm view @sap-cloud-sdk/openapi-generator@"< 1.54.2"
+ * [check all existing versions of a package] - `npm view @sap-cloud-sdk/util versions`
+ */
+async function deprecateNmpPackage(){
```
deekshas8, line 62: [req] typo

```suggestion
async function deprecateNpmPackage(){
```

scripts/deprecate-npm-package.ts
```diff
@@ -0,0 +1,96 @@
...
+    "@sap-cloud-sdk/openapi-generator",
+    "@sap-cloud-sdk/test-util",
+    "@sap-cloud-sdk/util"
+  ],
+  /**
+   * Regex used as a filter, indicating which versions should be deprecated.
+   * The group is crucial for picking the stable version prefix, which cannot be removed.
+   * Check the doc of the `deprecateOptions` object for more details.
+   */
+  versionsToBeDeprecated: /(1\.\d+\.\d+).*/,
```
deekshas8, line 46: [q] Wouldn't `/(1\.\d+\.\d+)/` be enough to match the version prefix part. How it's done currently is also good for me.

deekshas8 (changes requested):

scripts/deprecate-npm-package.ts
florian-richter: `scripts/` is excluded as it had many false positives especially with JSDoc which is not so necessary in this folder. Maybe we can adjust just a few rules for this folder. I didn't have the time but was annoyed enough by the warnings to disable it... 😅

scripts/deprecate-npm-package.ts
jjtang1985: Yes, you are right.
I will add `^` and therefore use `/^(1\.\d+\.\d+)/`;)

scripts/deprecate-npm-package.ts
jjtang1985: I guess the `scripts/` directory does not belong to any packages, which is not included.
I'll create a ticket.

package.json
```diff
@@ -68,7 +68,8 @@
     "temporal-de-serializers": "yarn workspace @sap-cloud-sdk/temporal-de-serializers",
     "e2e-tests": "yarn workspace @sap-cloud-sdk/e2e-tests",
     "integration-tests": "yarn workspace @sap-cloud-sdk/integration-tests",
-    "type-tests": "yarn workspace @sap-cloud-sdk/type-tests"
+    "type-tests": "yarn workspace @sap-cloud-sdk/type-tests",
+    "deprecate-npm-packages": "ts-node scripts/deprecate-npm-package.ts"
```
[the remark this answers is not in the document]
jjtang1985: makes sense

jjtang1985:
@deekshas8 , feel free to review again.

deekshas8 (approved):
LGTM.

PR #2519, opened 2022-05-31
feat: support comparison operators for missing edm types

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#488.

`Edm.String`, `Edm.Boolean`, and `Edm.Guid` are primitive OData types and should support all six operations. Thus, they are changed from `EdmTypeField` to `OrderableEdmTypeField`. 

I also added `lt`, `le`, `gt`, `ge` for the string filter function, which makes, e.g., the following case possible

```ts
filterFunctions().substring('abcde', 1).greaterOrEqual('bcde');
```


FrankEssenberger (changes requested):
Perhaps I have overseen it but could you add tests that for the primitive types the filter which are supposed to be present are also there.

FrankEssenberger (approved):

PR #2523, opened 2022-06-01
fix: `executeHttpRequest` for accepting the same params as v1.x

jjtang1985:
- [x] overload `executeHttpRequest` by supporting the same parameters like v1.x
- [x] deprecate the old overloading of `executeHttpRequest`, that accepts `xxxOrigin`
- [x] create a new function `executeHttpRequestWithOrigin` that does the same thing as the deprecated function 
- [x] compatibility notes 
- ADR will be update in a separate PR for quicker merge


packages/http-client/src/http-client.ts
```diff
@@ -300,6 +325,46 @@ function logRequestInformation(request: HttpRequestConfig) {
...
+ * @param options - An [[HttpRequestOptions]] of the http request for configuring e.g., csrf token delegation. By default, the SDK will not fetch the csrf token.
+ * @returns A promise resolving to an [[HttpResponse]].
+ */
+export function executeHttpRequest<T extends HttpRequestConfig>(
+  destination: DestinationOrFetchOptions,
+  requestConfig: T,
+  options?: HttpRequestOptions
+): Promise<HttpResponse>;
+/**
+ * @deprecated This overloading is replaced by the function [[executeHttpRequestWithOrigin]].
```
marikaner, line 343: [req] Incorrect word, I think.
```suggestion
 * @deprecated This overload is replaced by the function [[executeHttpRequestWithOrigin]].
```

packages/http-client/src/http-client.ts
```diff
@@ -311,7 +376,9 @@ function logRequestInformation(request: HttpRequestConfig) {
  * @param options - An [[HttpRequestOptions]] of the http request for configuring e.g., CSRF token delegation. By default, the SDK will not fetch the CSRF token.
  * @returns A promise resolving to an [[HttpResponse]].
  */
-export function executeHttpRequest<T extends HttpRequestConfigWithOrigin>(
+export function executeHttpRequestWithOrigin<
```
marikaner, line 379: [nth] In the API doc, consider this:
Instead of "When reaching conflicts, values with high priorities are chosen." Explicitly tell what is going to happen -> custom > requestConfig.

marikaner: [req] Check API doc for http => HTTP

marikaner: [req] Also here, make sure the docs about fetching by default are correct.

marikaner (changes requested):
I requested changes in the docs only.
Please don't forget to fix the log for when there are custom headers. This can also be done on a separate PR. (see BLI for details)

packages/http-client/src/http-client.ts
```diff
@@ -124,6 +125,30 @@ export function execute<ReturnT>(executeFn: ExecuteHttpRequestFn<ReturnT>) {
...
+  let ret: HttpRequestConfigWithOrigin = {
+    ...(requestConfig as HttpRequestConfigBase)
+  };
+  ret = requestConfig.headers
+    ? { ...ret, headers: { requestConfig: {}, custom: requestConfig.headers } }
+    : ret;
+  ret = requestConfig.params
+    ? { ...ret, params: { requestConfig: {}, custom: requestConfig.params } }
+    : ret;
+  return ret;
```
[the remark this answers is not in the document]
jjtang1985: The only difference is that, when `headers` is `undefined`:
- my code will return an object without the `headers` key
- your version will be `{ headers: { requestConfig: {}, custom: undefined} }`
I considered the second one, but feel a bit inconsistent.

packages/http-client/src/http-client.ts
jjtang1985: > Also here, make sure the docs about fetching by default are correct.
Nice catch, this has been changed.

jjtang1985:
> Please don't forget to fix the log for when there are custom headers. This can also be done on a separate PR.

I'll have a separate PR.

.changeset/too-many-holidays.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-cloud-sdk/http-client': minor
+---
+
+[Fixed Issue] Fix the `executeHttpRequest` function, so it accepts the same parameters as version 1.x.
```
marikaner, line 5:
```suggestion
[Fixed Issue] Fix the `executeHttpRequest` function, so it accepts the same parameters as in version 1.
```

packages/http-client/src/http-client.ts
```diff
@@ -301,18 +332,62 @@ function logRequestInformation(request: HttpRequestConfig) {
...
+): Promise<HttpResponse>;
+// eslint-disable-next-line jsdoc/require-jsdoc
+export function executeHttpRequest<
+  T extends HttpRequestConfig | HttpRequestConfigWithOrigin
+>(
+  destination: DestinationOrFetchOptions,
+  requestConfig: T,
+  options?: HttpRequestOptions
+): Promise<HttpResponse> {
+  // eslint-disable-next-line jsdoc/require-jsdoc
```
marikaner, line 365: [req] This comment seems like a mistake?

packages/http-client/src/http-client.ts
```diff
@@ -301,18 +332,62 @@ function logRequestInformation(request: HttpRequestConfig) {
...
 /**
  * Builds a [[DestinationHttpRequestConfig]] for the given destination, merges it into the given [[HttpRequestConfigWithOrigin]]
  * and executes it (using Axios).
  * The [[HttpRequestConfigWithOrigin]] supports defining header options and query parameter options with origins.
- * When reaching conflicts, values with high priorities are chosen.
- * The priorities are defined in the [[origins]].
+ * When reaching conflicts, values are picked in the order of:
+ * 1. `custom`
+ * 2. `destination`
+ * 3. `requestConfig`.
```
marikaner, line 382: [req] There is no destination keyword in the values of`HttpRequestConfigWithOrigin`. Haven't we changed this to be just custom and requestConfig?
[nth] Also, I still think it is unclear what "reaching conflicts" is (I just think it sounds weird) - consider rewording, e.g. "equally named headers and query parameters are prioritized in the following order" (or similar).

test-packages/type-tests/test/http-client.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { executeHttpRequest } from '@sap-cloud-sdk/http-client';
+import { executeHttpRequest, executeHttpRequestWithOrigin } from "@sap-cloud-sdk/http-client";
```
marikaner, line 1: [q] Lint! Why does this PR pass the check?? Are type-tests excluded from the lint checks?
```suggestion
import { executeHttpRequest, executeHttpRequestWithOrigin } from '@sap-cloud-sdk/http-client';
```

packages/http-client/src/http-client.ts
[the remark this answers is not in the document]
marikaner: Why do you think this is inconsistent? The custom keyword will always be there, therefore it seems even more consistent to me ;)
I think the shorter implementation reads much better and has less code duplication. It should technically not be a problem and the only difference.
If you think that this can cause issues, maybe consider this:

```ts
return {
    ...requestConfig,
    headers: { requestConfig: {}, ...requestConfig.headers && { custom: requestConfig.headers } },
    params: { requestConfig: {}, ...requestConfig.params && { custom: requestConfig.params } }
  };
```

But honestly, to me this seems like an unnecessary complication.

marikaner (dismissed):
I think the explantation of the origin might not be correct with the current state. Please check 😊

test-packages/type-tests/test/http-client.spec.ts
jjtang1985: let's enable type tests

packages/http-client/src/http-client.ts
jjtang1985: > 365:3  error  Missing JSDoc comment  jsdoc/require-jsdoc

🤷

jjtang1985: I'll create an issue for them

.changeset/too-many-holidays.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-cloud-sdk/http-client': minor
+---
+
+[Fixed Issue] Fix the `executeHttpRequest` function, so it accepts the same parameters as in version 1.
+[Compatibility Note] Deprecate one overload of the `executeHttpRequest` function, that accepts `HttpRequestConfigWithOrigin` as a parameter. Use the new function `executeHttpRequestWithOrigin` as replacement.
```
marikaner, line 6: [q] Does it make sense to also add an entry for a new feature here: "executeHttpRequestWithOrigin" ?

packages/http-client/src/http-client.ts
```diff
@@ -301,18 +327,61 @@ function logRequestInformation(request: HttpRequestConfig) {
...
+
 /**
  * Builds a [[DestinationHttpRequestConfig]] for the given destination, merges it into the given [[HttpRequestConfigWithOrigin]]
  * and executes it (using Axios).
  * The [[HttpRequestConfigWithOrigin]] supports defining header options and query parameter options with origins.
- * When reaching conflicts, values with high priorities are chosen.
- * The priorities are defined in the [[origins]].
+ * Equally named headers and query parameters are prioritized in the following order:
+ * 1. `custom`
+ * 2. `requestConfig`.
```
marikaner, line 376: [q] Isn't the destination missing now? I mean - not `destination`, but the fact that `requestConfig` would be overwritten by parameters created by the SDK. Will you add a link as discussed?

marikaner (approved):
I left two questions, but I already approved. I think the priority is still not understandable from the API Doc. Therefore I would prefer to have a link to the docs for details.

PR #2527, opened 2022-06-02
feat: customizable cache implementation

deekshas8:
Closes SAP/cloud-sdk-backlog#641

This PR provides a way to enable user-supplied destination-cache implementation (in place of the SDK implementation).
- `clear` method is part of the `CacheInterface`
- The `set` method arguments are updated.
- Added a new interface `DestinationInterface` that extends the existing `CacheInterface`.
- A global setter method `setDestinationCache` can be used to set a custom instance to the `destinationCache`.
- The `test-cache.ts` file contains a custom cache implementation with the new interface `DestinationInterface`. 
- Updated how `cache.set` is called in other caches with the new parameters (`CacheEntry<T>`).


packages/connectivity/src/scp-cf/destination/destination-cache.ts
```diff
@@ -11,6 +11,12 @@ const logger = createLogger({
   messageContext: 'destination-cache'
 });
 
+const _destinationcache: CacheInterface<Destination> = new Cache<Destination>({
```
FrankEssenberger, line 14: I guess the `_destinationcache` is not in line with our naming conventions.

packages/connectivity/src/scp-cf/destination/destination-cache.ts
```diff
@@ -140,16 +146,23 @@ function cacheRetrievedDestination(
...
 
 /**
- * @internal
+ * Set the custom cache instance.
+ * @param cache - Cache instance.
  */
-export const destinationCache = DestinationCache(
-  new Cache<Destination>({ hours: 0, minutes: 5, seconds: 0 })
-);
+export function setDestinationCache<T extends CacheInterface<Destination>>(
```
FrankEssenberger, line 156: [req] We should use the more strict `DestinationInterface` everywhere here so that we are sure that we are compatible with the API contract we communicate to customers.

packages/connectivity/src/scp-cf/destination/destination-cache.ts
FrankEssenberger: [req] see comment on type below

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -1,10 +1,26 @@
...
+  get(key: string | undefined): T | undefined;
+  set(key: string | undefined, item: CacheEntry<T>): void;
+  clear(): void;
+}
+
+/**
+ * Interface to implement custom caching.
+ */
+export interface DestinationInterface extends CacheInterface<Destination> {
+  set(key: string | undefined, item: CacheEntry<Destination>): void;
```
FrankEssenberger, line 17: [q] what is a `undefined` key?

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -67,6 +69,12 @@ const destinationOne: Destination = {
   isTrustingAllCertificates: false
 };
 
+// Cache with expiration time
+const testCacheOne = new TestCache(30);
+
+// Setting the destinationCache with custom class instance
+setDestinationCache(testCacheOne);
```
FrankEssenberger, line 76: [req] I think this will become relevant before merging, but I leave the comment anyway. Think there should be additional test regarding the custom inplemenation so keep the old test but in a separate `describe` do some tests where you set the custom implementation and assert that this implementation is used for all used cases like set,get and clear. This should be possible via spy I guess.

FrankEssenberger (changes requested):
Implementation looks fine for me. Just a few comments questions regarding types. I realize that for this specific case the concrete code example was much better than the abstract ADR discussion. But perhaps this was only the case because we had the discussion before.

FrankEssenberger (approved):
LGTM but please contact the graph colleagues. I also left a comment on the test.

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -806,6 +807,41 @@ describe('destination cache', () => {
       expect(retrieveDestination()).toBeUndefined();
     });
   });
+
+  describe('custom destination cache', () => {
+    // Cache with expiration time
```
FrankEssenberger, line 812: consider also adding a test, where:
- default cache is used -> assert working
- switch to custom one -> assert working
- switch back to default -> still working old state there.

I have this in mind because potentially users want to use the global cache at a certain time or certain destinations and we must ensure that there are no side effects. Besides the obvious that each destination is just in one cache of course.

deekshas8: As discussed, we would need to expose a function that returns the default cache instance which the user will maintain on their end and then switch to, by calling `setDestinationCache` again in case they want to use the default cache.
We can implement it later if this is requested by users.

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -1,18 +1,30 @@
...
-  hasKey(key: string): boolean;
-  get(key: string): T | undefined;
-  set(key: string, item: T, expirationTime?: number): void;
+  get(key: string | undefined): Promise<T | undefined>;
+  set(key: string | undefined, item: CacheEntry<T>): Promise<void>;
+  clear(): Promise<void>;
+}
+
+/**
+ * Interface to implement custom destination caching.
```
FrankEssenberger, line 10: [nth] Perhaps to can create a link to the `setDestinationCache` method with something like: Use the method setDestinationCache to set the custom implementation.

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -1,18 +1,30 @@
...
+ */
+export interface DestinationCacheInterface extends CacheInterface<Destination> {
+  set(key: string | undefined, item: CacheEntry<Destination>): Promise<void>;
```
FrankEssenberger, line 13: [q] I assume you tried this but what happens if you use a conditional type like `type syncOrAsyng<isPromise extends boolean,U> = V extends true ? Promise<U> : U;` and return `syncOrAsyng<isPromise,void>`  for the set and add a second generic argument `isPromise` to the CacheInterface, perhaps `isAsync` is a better name.

FrankEssenberger: The destination cache would be async, but perhaps the other caches could remain sync this way.

FrankEssenberger: @florian-richter @jjtang1985 we also experimented with a conditional type which is either `Promise<T>` or `T` to get the cache for the sync and async case. However, this lead to an additional generic arguments and ugly return types. So I am not conviced what is better:
- having await where it is not needed
- havving a complicated type foo bla....

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -1,7 +1,19 @@
...
+  get(key: string | undefined): T | undefined;
+  set(key: string | undefined, item: CacheEntry<T>): void;
+  clear(): void;
+}
+
+/**
+ * Interface to implement custom destination caching.
+ * To set a custom implementation, call method [[setDestinationCache]] and pass the cache instance.
+ */
+export interface DestinationCacheInterface {
```
FrankEssenberger, line 13: [pp] Perhaps you move the interface to the destination-cache.ts?

packages/connectivity/src/scp-cf/destination/destination-cache.ts
```diff
@@ -140,16 +140,23 @@ function cacheRetrievedDestination(
   const expirationTime = expiresIn
     ? Date.now() + parseInt(expiresIn) * 1000
     : undefined;
-  cache.set(key, destination, expirationTime);
+  cache.set(key, { entry: destination, expires: expirationTime });
 }
 
 /**
- * @internal
+ * Set the destination cache instance.
```
FrankEssenberger, line 147: [nth] Extend the JS doc and explain we have a default but if you want to use some different use this method before the first `getDestination` eitther direct or via `executeHttpRequest` and so on.

FrankEssenberger (approved):
LGTM left two minor comments

packages/connectivity/src/scp-cf/destination/destination-cache.ts
```diff
@@ -49,35 +59,35 @@ export interface DestinationCacheType {
  * @returns A destination cache object.
  * @internal
  */
-export const DestinationCache = (
-  cache: Cache<Destination>
-): DestinationCacheType => ({
-  retrieveDestinationFromCache: (
+export const DestinationCache = <T extends DestinationCacheInterface>(
+  cache: T = new Cache<Destination>({ hours: 0, minutes: 5, seconds: 0 }) as any
```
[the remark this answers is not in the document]
FrankEssenberger: This is actually a good idea. Removing the safety net via `as XYZ` is  never a good idea and type casts indicate always something fishy is going on. @deekshas8 could you add a second cache interface which is just promisifying all the methods from the initial one.

packages/connectivity/src/scp-cf/destination/destination-cache.ts
```diff
@@ -49,35 +109,39 @@ export interface DestinationCacheType {
...
-export const DestinationCache = (
-  cache: Cache<Destination>
-): DestinationCacheType => ({
-  retrieveDestinationFromCache: (
+export const DestinationCache = <T extends DestinationCacheInterface>(
+  cache: T = new DefaultDestinationCache({
+    hours: 0,
+    minutes: 5,
+    seconds: 0
+  }) as any
```
FrankEssenberger, line 117: [req] The `as any` should not be needed anymore

FrankEssenberger: Besides that looks good to me.

deekshas8: Without `any` it gives an error:

`'DefaultDestinationCache' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'DestinationCacheInterface'.`

PR #2529, opened 2022-06-03
fix: TIMING env var is set on windows

deekshas8:
This PR sets the TIMING env var correctly for windows. 
Currently, running the `lint:fix` command gives the following error on windows:
> 'TIMING' is not recognized as an internal or external command

Note: Please test if this also works on mac.


FrankEssenberger (approved):
LGTM - worked also on my mac with cahcing from turbo repo.

PR #2547, opened 2022-06-07
Update GH actions to new node and new actions versions

florian-richter:
Currently the checks fail as puppeteer does not accept node v12. I updated the node versions and the actions versions. Not sure if we need to keep testing with node v12...

### DIFF
--- a/.github/workflows/blackduck.yml
+++ b/.github/workflows/blackduck.yml
@@ -12,14 +12,13 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node-version: [14.x]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - run: git fetch --depth=1
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node-version }}
-      - uses: actions/cache@v2
+          node-version: 14
+      - uses: actions/cache@v3
         id: cache
         with:
           path: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,14 +15,14 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node-version: [12.x, 14.x]
+        node-version: [14, 16]
     steps:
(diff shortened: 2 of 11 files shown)


.github/workflows/blackduck.yml
```diff
@@ -12,14 +12,13 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
```
marikaner, line 14: [req] I guess we don't need the matrix here anymore.

marikaner: => `runs-on: ubuntu-latest`

marikaner (approved):
Requested one change, but already approving.

.github/workflows/blackduck.yml
florian-richter: We use the matrix below as a variable for the name. This could be hardcoded as well, but I wasn't sure if this would break something.

.github/workflows/build.yml
```diff
@@ -13,23 +13,24 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 15
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        node-version: [12.x, 14.x]
+        node-version: [12, 14, 16, 18]
```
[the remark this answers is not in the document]
FrankEssenberger: ok fair. I approve

FrankEssenberger (approved):

jjtang1985:
Here is one related OS issue: https://github.com/SAP/cloud-sdk-js/issues/2566

tomfrenken:
Looking at the customer issue, should we actually use the `--ignore-engines` flag?
I guess it works but would that also be a recommended approach for our users?

FrankEssenberger:
> Looking at the customer issue, should we actually use the `--ignore-engines` flag? I guess it works but would that also be a recommended approach for our users?

I think we should ask the colleagues again to include node 16 in the list of supported version.

.github/workflows/blackduck.yml
marikaner: If it does we can fix it ;)

PR #2551, opened 2022-06-08
chore: fix destination for service binding with options

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#553.

### DIFF
--- a/packages/connectivity/src/scp-cf/destination/destination-from-env.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-from-env.ts
@@ -110,7 +110,7 @@ function validateDestinations(destinations: any[]) {
  */
 export function searchEnvVariablesForDestination(
   options: DestinationFetchOptions
-): Destination | undefined {
+): Destination | null {
   logger.debug('Attempting to retrieve destination from environment variable.');
 
   if (getDestinationsEnvVariable()) {
@@ -130,6 +130,7 @@ export function searchEnvVariablesForDestination(
   }
 
   logger.debug('No environment variable set.');
+  return null;
 }
 
 /**
--- a/packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
@@ -70,9 +70,19 @@ export type DestinationWithName = Destination & { name: string };
 export function searchRegisteredDestination(
   options: DestinationFetchOptions
 ): Destination | null {
+
+  let decodedJwt;
+  // An error could be thrown if no jwt and no xsuaa service exist.
+  try {
+    decodedJwt = decodedJwtOrZid(options)
+  } catch(e) {
+    logger.warn('Failed to decode jwt or create a dummy jwt with `zid` property.');
+    logger.warn(e);
+  }
+
   const destination =
     registerDestinationCache.retrieveDestinationFromCache(
-      decodedJwtOrZid(options),
+      decodedJwt ?? {},
       options.destinationName,
       isolationStrategy(options)
     ) || null;
--- a/packages/connectivity/src/scp-cf/destination/destination-from-vcap.spec.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-from-vcap.spec.ts
@@ -1,13 +1,12 @@
-import { xsuaaBindingMock } from '../../../../../test-resources/test/test-util';
 import { getDestination } from './destination-accessor';
 import { destinationForServiceBinding, searchServiceBindingForDestination } from './destination-from-vcap';
 
 describe('vcap-service-destination', () => {
-  beforeAll(() => {
+  beforeEach(() => {
     mockServiceBindings();
   });
 
-  afterAll(() => {
+  afterEach(() => {
     delete process.env.VCAP_SERVICES;
   });
 
@@ -60,7 +59,7 @@ describe('vcap-service-destination', () => {
     ).toThrowErrorMatchingSnapshot();
   });
 
-  it('finds the service bindings when getting destination', async () => {
+  it('finds the destination when searching for service bindings', async () => {
     const transformationFn = serviceBinding => ({
       url: serviceBinding.credentials.sys
     });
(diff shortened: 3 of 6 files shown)


packages/connectivity/src/scp-cf/destination/destination-from-env.ts
```diff
@@ -110,7 +110,7 @@ function validateDestinations(destinations: any[]) {
  */
 export function searchEnvVariablesForDestination(
   options: DestinationFetchOptions
-): Destination | undefined {
+): Destination | null {
```
FrankEssenberger, line 113: [q] why this change?

.changeset/warm-chicken-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Fixed Issue] Fix `getDestination()` to allow passing an async transform function for a service binding in the `options`. It also enables passing the transform function as a part of the `DestinationOrFetchOptions`, which is required by `execute()`, `executeHttpRequest()`, `executeRaw()`, etc.
```
FrankEssenberger, line 5: I know that the `transformFn` was marked as `@internal` but I would anyway write a compatibility note that the method was renamed and made `async`.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.spec.ts
```diff
@@ -28,34 +34,55 @@ describe('vcap-service-destination', () => {
...
       destinationForServiceBinding('my-custom-service')
-    ).toThrowErrorMatchingSnapshot();
+    ).rejects.toThrowErrorMatchingSnapshot();
+  });
+
+  it('finds the destination when searching for service bindings', async () => {
+    const serviceBindingTransformFn = async (
+      serviceBinding: ServiceBinding
+    ) => ({
+      url: serviceBinding.credentials.sys
```
FrankEssenberger, line 75: [pp] Could you perhaps use a obscure specific name like `http://value-from-transform-function.com`. and use in the expect below as well. Then it becomes clear that the method was used. Or use a `spyOn` with `serviceBindingTransformFn` has been called once because we want to ensure that and not that the value is coming from somewhere else, which could happen is you use something from the serviceBinding.cred.sys.

FrankEssenberger (approved):
LGTM - only minor question and comments

packages/connectivity/src/scp-cf/destination/destination-from-env.ts
florian-richter: Seems to violate our [styleguide](https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-undefined-for-explicit-absence-of-values) in my opinion. So if there is no good reason, this should be reverted.

packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
```diff
@@ -70,9 +70,21 @@ export type DestinationWithName = Destination & { name: string };
 export function searchRegisteredDestination(
   options: DestinationFetchOptions
 ): Destination | null {
+  let decodedJwt;
+  // An error could be thrown if no jwt and no xsuaa service exist.
```
florian-richter, line 74: [q] What do you mean "cloud be thrown"? Could implies to me that there is non-deterministic behavior here (which I don't expect). It should be "will be" if it's implemented like this or "should be" if it is a TODO.

packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
```diff
@@ -70,9 +70,21 @@ export type DestinationWithName = Destination & { name: string };
...
+  try {
+    decodedJwt = decodedJwtOrZid(options);
+  } catch (e) {
+    logger.debug(
+      'Could not retrieve destination from registered destination. Failed to decode jwt or create a dummy jwt with `zid` property.'
```
florian-richter, line 79:
```suggestion
      'Failed to retrieve registered destination, because it was neither possible to decode jwt nor create a dummy jwt with `zid` property.'
```
[pp] The original message seems confusing to me. I tried to rephrase (hopefully keeping the correct meaning).

packages/connectivity/src/scp-cf/destination/destination-from-vcap.spec.ts
florian-richter: Please never use any URLs that are not intended for testing. Rather prefer `example.com/value-from-transform-function` or `value-from-transform-function.example.com` as example.com is owned by IANA explicitly for testing.

florian-richter (changes requested):
Comment and debug message could be improved. Otherwise seems good

packages/connectivity/src/scp-cf/destination/destination-from-env.ts
ZhongpinWang: Thanks for the feedback! 
I changed this because all other three functions in `getDestination()` and also `getDestination` itself are returning `Destination | null` or promise of it. I think the style guide was more referring to the case for function parameters other than return type.
Also, correct me if I am wrong, I feel like the difference between `null` and `undefined` is that `null` means the value is explicitly empty, and `undefined` means the value has not been retrieved and assigned yet. [A post on StackOverflow](https://stackoverflow.com/a/5076962)

packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
ZhongpinWang: Thanks! I think your version is more clear. 👍

.changeset/warm-chicken-allow.md
ZhongpinWang: I was a bit hesitating. I will add a compatibility note for this.

packages/connectivity/src/scp-cf/destination/destination-from-env.ts
florian-richter: I think the styleguide is not only meant for params, but we can have a discussion if it should be changed. I don't remember the arguments for the current way, but I remember there was a discussion back then 😊

florian-richter (approved):

Issue #2558, opened 2022-06-10
Retrieved destination doesn't contain authTokens


thisisevanfox:
Hello all,

since I upgraded the @sap-cloud-sdk package from 1.53.x/1.54.x to 2.4.0, I'm facing the issue that retrieved destination does not include any authTokens although the forwardAuthToken property is set to true.

To clarify the problem, I have attached code and screenshots below of how the application behaved before and after the upgrade.

I am not sure if this is really an issue or if the behaviour is so desired. I wanted to report the behaviour anyway to find out if I am doing something wrong. If the behaviour is as desired as described below, can you tell me what I'm doing wrong or what I need to do differently to get authTokens with the destination?

Kind regards
Johannes
______________________

### Definition of destination in "User-Provided Variables" in the BTP
```
[ {
     "forwardAuthToken" : true,
     "name" : "myDestination",
      "url" : "https://<my-url>.cfapps.eu10.hana.ondemand.com"
} ]
```

### Old behaviour
Used: version: `"@sap-cloud-sdk/core": "1.53.1"`

Calling the `getDestination` method:
```
	import * as coreCloud from "@sap-cloud-sdk/core";
	[...]
        const foundDestination = await coreCloud.getDestination("myDestination", {
             userJwt: <valid-user-token>,
             useCache: true,
	});
	[...]
```
**Response:**
![image](https://user-images.githubusercontent.com/43059689/173042003-475e89fb-d116-43c1-b4f9-24216c6f964c.png)

As you can see authTokens array is filled.

### New behaviour
Used: version: `"@sap-cloud-sdk/connectivity ": "2.4.0"`

Calling the `getDestination` method:
```
	import * as connectivityCloud from "@sap-cloud-sdk/connectivity";
	[...]
        const foundDestination = await connectivityCloud.getDestination({
             destinationName: "myDestination",
             jwt: <valid-user-token>,
             useCache: true,
	});
	[...]
```

**Response:**
![image](https://user-images.githubusercontent.com/43059689/173041882-fdf1e972-5881-4058-be99-8b0196aa6246.png)

As you can see authTokens array is not filled.

###  My finding
I debugged the sources of the two versions of the sdk. And it looks like that the part where the authTokens are added to the destination is not implemented in the new version.

**Old implementation**
![image](https://user-images.githubusercontent.com/43059689/173044179-3e65a9d0-aae5-4377-a528-12a8694d58de.png)

**New implementation**
[...]

PR #2567, opened 2022-06-14
chore: adjust tsdoc for register destination

FrankEssenberger:
Fix a little issue with the TS doc which came up in this issue: https://github.com/SAP/cloud-sdk-js/issues/2558


packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
```diff
@@ -36,7 +36,8 @@ type RegisterDestinationOptions = Pick<
 /**
  * Registers a destination in a cache for later usage.
  *
- * Throws an error if a destination with the same name as the given test destination already exists.
+ * If a destination with the same key is already in the cache it is replaced.
```
deekshas8, line 39:
```suggestion
 * If a destination with the same key is already in the cache, it is replaced.
```

[pp] ...it is overwritten.

packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
```diff
@@ -36,7 +36,8 @@ type RegisterDestinationOptions = Pick<
...
+ * The key is build using the `getDestinationCacheKey` method.
```
deekshas8, line 40:
```suggestion
 * The key is built using the `getDestinationCacheKey` method.
```

marikaner (approved):

PR #2568, opened 2022-06-14
improvement: make `OriginOptions` flexible

jjtang1985:

Make the request config optional and accept the edge case when users provide a normal request config (SDK v1 signature), but is treated as `OriginOptions`.


marikaner (approved):
I just have minor comments.

PR #2575, opened 2022-06-16
Configure devcontainer

fwilhe:
Treat this as a suggestion for now.

This PR configures [devcontainer](https://code.visualstudio.com/docs/remote/containers) which allow developing inside a container in VS Code.
The idea behind this is to allow developing on a machine where only Docker and VS Code are available but no other runtime requirements exist.

This is fully optional, you can simply ignore the devontainer and use your local setup as you did before, but for those who chose the devcontainer it provides a simple way to get a reproducible environment.


FrankEssenberger (approved):

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+# Install chromium via apt because installing via puppeteer will fail on m1 mac otherwise
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends chromium
+
+# [Optional] Uncomment if you want to install an additional version of node using nvm
+# ARG EXTRA_NODE_VERSION=10
+# RUN su node -c "umask 0002 && ./usr/local/share/nvm/nvm.sh && nvm install ${EXTRA_NODE_VERSION}"
+
+# [Optional] Uncomment if you want to install more global node modules
+# RUN su node -c "npm install -g <your-package-list-here>"
```
florian-richter, line 15: [pp] I think basically all our files use a newline at the end.

fwilhe: Fine with me, is this somehow managed by prettier? I think I found the vs code setting for that, but it would be nicer to have that automatically managed.

PR #2577, opened 2022-06-16
fix: rename entity property to _entity

deekshas8:
Closes SAP/cloud-sdk-backlog#745.

The `EntityBuilder` class has protected property `entity`. If an Entity in a service also has a property by the name `entity`, it causes a conflict and results in compilation error:

```
Property 'entity' is protected in type '{ entity: (value: DeserializedType<DeSerializersT, "Edm.String">) =>
EntityBuilderType<RawDataStore<DeSerializersT>, any>; ... 10 more ...; content: (value: DeserializedType<...> | ... 1 more ... |
undefined) => EntityBuilderType<...>; } & EntityBuilder<...>' but public in type '{ entity: (value:
DeserializedType<DeSerializersT, "Edm.String">) => EntityBuilderType<RawDataStore<DeSerializersT>, DeSerializersT>; ... 
10 more ...; content: (value: DeserializedType<...> | ... 1 more ... | undefined) => EntityBuilderType<...>; }'.

51     return entityBuilder(this); 
```

### DIFF
--- a/packages/odata-common/src/entity-base.ts
+++ b/packages/odata-common/src/entity-base.ts
@@ -351,7 +351,7 @@ export function entityBuilder<
   Object.values(entityApi.schema).forEach((field: any) => {
     const fieldName = `${camelCase(field._fieldName)}`;
     builder[fieldName] = function (value) {
-      this.entity[fieldName] = value;
+      this._entity[fieldName] = value;
       return this;
     };
   });
--- a/packages/odata-common/src/entity-builder.ts
+++ b/packages/odata-common/src/entity-builder.ts
@@ -44,11 +44,11 @@ export class EntityBuilder<
   EntityT extends EntityBase,
   DeSerializersT extends DeSerializers
 > {
-  protected entity: EntityT;
+  protected _entity: EntityT;
 
   constructor(private _entityApi: EntityApi<EntityT, DeSerializersT>) {
-    if (!this.entity) {
-      this.entity = new _entityApi.entityConstructor(_entityApi.schema);
+    if (!this._entity) {
+      this._entity = new _entityApi.entityConstructor(_entityApi.schema);
     }
   }
 
@@ -59,7 +59,7 @@ export class EntityBuilder<
    */
   public withCustomFields(customFields: Record<string, any>): this {
     const validCustomFields = this.filterCustomFields(customFields);
-    this.entity = this.entity.setCustomFields(validCustomFields);
+    this._entity = this._entity.setCustomFields(validCustomFields);
     return this;
   }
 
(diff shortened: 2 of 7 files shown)


packages/odata-common/src/entity-base.ts
```diff
@@ -351,7 +351,7 @@ export function entityBuilder<
   Object.values(entityApi.schema).forEach((field: any) => {
     const fieldName = `${camelCase(field._fieldName)}`;
     builder[fieldName] = function (value) {
-      this.entity[fieldName] = value;
+      this._entity[fieldName] = value;
```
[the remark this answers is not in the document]
deekshas8: Yes it will fail if there is a property `_entity`. The reserved words list already has `Entity` but I believe that has no effect on naming of property fields.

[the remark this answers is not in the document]
deekshas8: @FrankEssenberger I added both `entity` and `ENTITY` to the `reservedVdmKeywords` for instance and static property naming. So entity will now be formatted to instance property `entity_1` and static `ENTITY_1` in schema.

packages/odata-common/src/entity-builder.ts
```diff
@@ -44,11 +44,11 @@ export class EntityBuilder<
   EntityT extends EntityBase,
   DeSerializersT extends DeSerializers
 > {
-  protected entity: EntityT;
+  protected _entity: EntityT;
```
jjtang1985, line 47: [req] Please also adjust the test service spec, so we can see how the clients will be generated as well as the compilation.

jjtang1985 (changes requested):
I would like to change the test spec, so we are sure this is tested with e2e scope.

packages/odata-common/src/entity-builder.ts
FrankEssenberger: [req] I think we do not need the `_` andymore if it is in the reserved words right?

jjtang1985: I personally would keep this change so by default, a user would not have `entity_1`.

marikaner:
Again, my 2 cents: I think `_entity` is a good solution, but it is not necessary to add `entity` to the reserved keywords. You should check whether` _entity` has to be added, though.

packages/odata-common/src/entity-builder.ts
jjtang1985: @marikaner , as a PO, you have a task for your opinion.

jjtang1985:
> Again, my 2 cents: I think _entity is a good solution, but it is not necessary to add entity to the reserved keywords. You should check whether _entity has to be added, though.

This can be checked by adjusting the test service spec, so we are sure.

FrankEssenberger (approved):
LGTM

packages/odata-common/src/entity-base.ts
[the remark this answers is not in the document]
FrankEssenberger: I already approved, I had the question why you renamedit with the `_`?

packages/odata-common/src/entity-builder.ts
FrankEssenberger: I adjusted the code and also made one more manual test with the spec file provided by the users and worked with the `_entity`

jjtang1985 (changes requested):
Looks good.
Please add change log.

jjtang1985 (approved):

PR #2582, opened 2022-06-17
chore: record more download statistics

jjtang1985:
I added all the published packages to the list so we understand the trend of:
- our internal download
- `odata-v2` download
- `generator` download
- `connectivity` download


florian-richter (approved):
[q] Are there rate limits we could run into with this change?

jjtang1985:
> Are there rate limits we could run into with this change?

Local tests do not show any limits.

PR #2583, opened 2022-06-17
fix: type issue

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#691.
 
So I give up, I do not understand why it works for V2 and not for v4 on a deep level. The change to the `Entity` in this PR  was not relevant. On the play ground I did the following:
- Create a minimal example for an Entity with a single 1:1 nav prop no enum, complex type what so ever
- Align deserializers for v2 and v4 and types
- Remove one-to-many type also from filterable
- Align entity (version is v2 for both)

The error remains only for v4 not for v2 using the reduced example.

I knew from the investigation and tests that the default value set to any fixes it, but this leaves a bitter taste because I do not see why, WHY, WHYYYY but I have to accept that I do not see the reason.....

### DIFF
--- a/packages/odata-common/src/filter/filterable.ts
+++ b/packages/odata-common/src/filter/filterable.ts
@@ -20,7 +20,7 @@ export type Filterable<
   DeSerializersT extends DeSerializers,
   LinkedEntityApiT extends EntityApi<EntityBase, DeSerializersT> = EntityApi<
     EntityBase,
-    DeSerializersT
+    any
   >
 > =
   | Filter<EntityT, DeSerializersT, any>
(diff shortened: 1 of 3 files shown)


packages/odata-common/src/filter/filterable.ts
```diff
@@ -20,7 +20,7 @@ export type Filterable<
   DeSerializersT extends DeSerializers,
   LinkedEntityApiT extends EntityApi<EntityBase, DeSerializersT> = EntityApi<
     EntityBase,
-    DeSerializersT
+    any
```
FrankEssenberger, line 23: This fixes the issue for v4 but I do not know why....

florian-richter (approved):
I also don't understand this. But if this fixes the tests, I don't see any harm in merging it.

FrankEssenberger:
There was also a customer issue related to this.

Issue #2589, opened 2022-06-19
Dates before the unix epoch do not parse in the generated oData clients


carlonnheim:
**Describe the bug**
The payload converter for Edm DateTime values does not allow negative values. The API fails to handle dates earlier than the unix epoch (1970-01-01) due to this.

**To Reproduce**
Generate an oData client and run a get-query to an instance with a date property earlier than 1970-01-01.

**Expected behavior**
The date is parsed properly.

**Screenshots**
Not applicable.

**Used Versions:**

- node version v16.15.0
- npm version 8.5.5
- SAP Cloud SDK version 1.54.2 (used in a CAP project and CAP still uses v1)

**Code Examples**
The issue is in odata-v2/payload-value-converter.js
```
/**
 * @hidden
 * This function can be used for both Edm.DateTime and and Edm.DateTimeOffset.
 */
function edmDateTimeToMoment(edmDateTime) {
    var _a;
    var dateTimeOffsetComponents = (_a = /^\/Date\((?<ticks>\d+)((?<sign>[+-])(?<offset>\d{4}))?\)\/$/.exec(edmDateTime)) === null || _a === void 0 ? void 0 : _a.groups;
    if (!dateTimeOffsetComponents) {
        throw new Error("Failed to parse edmDateTime: ".concat(edmDateTime, " to moment."));
    }
    var timestamp = (0, moment_1.default)(parseInt(dateTimeOffsetComponents.ticks));
    if (dateTimeOffsetComponents.sign && dateTimeOffsetComponents.offset) {
        var offsetMultiplier = dateTimeOffsetComponents.sign === '+' ? 1 : -1;
        var offsetInMinutes = parseInt(dateTimeOffsetComponents.offset);
        return timestamp.utc().utcOffset(offsetMultiplier * offsetInMinutes);
    }
    return timestamp;
}
```
The regex needs to allow a sign on the ticks value, like so
```
    var dateTimeOffsetComponents = (_a = /^\/Date\((?<ticks>[+-]?\d+)((?<sign>[+-])(?<offset>\d{4}))?\)\/$/.exec(edmDateTime)) === null || _a === void 0 ? void 0 : _a.groups;
```

**Log file**
Not applicable

**Impact / Priority**

Affected development phase: UAT

Impact: Impaired

Timeline: Go-Live is in 3 months.

**Additional context**
[...]

PR #2593, opened 2022-06-20
fix: add negative timestamp support

FrankEssenberger:
Relates to: #2589 

Add support for negative timestamps before 1970-01-01.

### DIFF
--- a/packages/odata-v2/src/de-serializers/converters.ts
+++ b/packages/odata-v2/src/de-serializers/converters.ts
@@ -46,11 +46,15 @@ export function deserializeToMoment(edmDateTime: string): moment.Moment {
     throw new Error(`Failed to parse edmDateTime: ${edmDateTime} to moment.`);
   }
 
-  const timestamp = moment(parseInt(dateTimeOffsetComponents.ticks));
+  const timestamp = moment(parseInt(dateTimeOffsetComponents.signedticks));
 
-  if (dateTimeOffsetComponents.sign && dateTimeOffsetComponents.offset) {
-    const offsetMultiplier = dateTimeOffsetComponents.sign === '+' ? 1 : -1;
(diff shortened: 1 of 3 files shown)


packages/odata-v2/src/de-serializers/converters.ts
```diff
@@ -39,7 +39,7 @@ export function serializeFromTime(value: Time): string {
  */
 export function deserializeToMoment(edmDateTime: string): moment.Moment {
   const dateTimeOffsetComponents =
-    /^\/Date\((?<ticks>\d+)((?<sign>[+-])(?<offset>\d{4}))?\)\/$/.exec(
+    /^\/Date\((?<ticks>-{0,1}\d+)((?<sign>[+-])(?<offset>\d{4}))?\)\/$/.exec(
```
fwilhe, line 42: We do already have a "sign" group. How does that relate to what we're changing here?

fwilhe: Screenshot 2022-06-20 at 15 29 58

Not sure if this is an issue, but I find it at least confusing that we have a negative timestamp where the sign is `+`.

florian-richter:
```suggestion
    /^\/Date\((?<signedticks>-{0,1}\d+)((?<offsetsign>[+-])(?<unsignedoffset>\d{4}))?\)\/$/.exec(
```
[pp] I don't mind the current implementation but maybe this improves clarity?

florian-richter (approved):
LGTM

packages/odata-v2/src/de-serializers/converters.ts
fwilhe: That makes it more clear to me, thanks

fwilhe (approved):

PR #2621, opened 2022-06-28
fix: resolve circular dependencies

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#710.

1. Fix `madge` by passing `tsconfig.json` to it.
2. Fix circular dependencies that can be fixed.
3. Exclude circular dependencies that are hard to fix.
  - `1) connectivity/src/scp-cf/jwt.ts > connectivity/src/scp-cf/environment-accessor.ts`
  - `2) http-client/src/csrf-token-header.ts > http-client/src/http-client.ts`
  - `test-packages`
4. Create a self-test for `madge` and move the self-test of `dtslint` to the same test package.
5. Adjust the workflow to run the self-tests first.
6. Disallow import from `.`, `..`, `../..`, `**/index`.


test-packages/type-tests/package.json
```diff
@@ -8,9 +8,8 @@
   "types": "test",
   "repository": "github:SAP/cloud-sdk-js",
   "scripts": {
-    "test": "yarn test:self-test && yarn test:type",
+    "test": "yarn test:type",
     "test:type": "dtslint --expectOnly --localTs ../../node_modules/typescript/lib test",
```
florian-richter, line 12:
```suggestion
    "test": "dtslint --expectOnly --localTs ../../node_modules/typescript/lib test",
```
[pp] I think we can just always use the `test` without `test:type`.

florian-richter (approved):
LGTM

PR #2625, opened 2022-06-29
chore: Fetch single destination

fwilhe:
Closes SAP/cloud-sdk-backlog#195.

Call the destination by name instead of getting all destinations.


fwilhe:
For reviewing this PR, I think it makes sense to read the functions in this order:

```
getDestination
	getDestinationFromDestinationService
		searchDestinationWithSelectionStrategyAndCache
			searchProviderAccountForDestination
				getProviderDestinationService
					getInstanceAndSubaccountDestinations
						fetchInstanceDestinations
							fetchDestinations
		fetchDestinationWithUserExchangeFlows
			fetchDestinationByToken
				fetchDestination
					fetchDestinationByTokens
		fetchDestinationWithNonUserExchangeFlows
			fetchDestinationByToken
				fetchDestination
					fetchDestinationByTokens
```

fwilhe:
Review note: The compatibility note might be too long and detailled, please provide feedback.

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -544,16 +532,10 @@ Possible alternatives for such technical user authentication are BasicAuthentica
   private async getProviderDestinationService(): Promise<
     DestinationSearchResult | undefined
   > {
-    const provider = await this.getInstanceAndSubaccountDestinations(
+    const destination = await this.fetchDestinationWithoutToken(
       this.providerServiceToken.encoded
     );
-    const destination = this.options.selectionStrategy(
-      {
-        subscriber: emptyDestinationByType,
```
deekshas8, line 552: [req] The variable `emptyDestinationByType` is still declared in the file and should also be removed.

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -212,6 +212,62 @@ export async function fetchCertificate(
...
 
+/**
+ * Fetch a single destination by name without credentials.
+ * @internal
+ */
+export async function fetchDestinationByNameWithoutTokens(
+  destinationServiceUri: string,
+  serviceToken: string,
+  options: DestinationServiceOptions
+): Promise<Omit<Destination, 'AuthAndExchangeTokens'> | undefined> {
```
deekshas8, line 223: [q] In the api reference, I coudn't find a property `AuthAndExchangeTokens` in Destinations, is it intended to be `authTokens`?

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -212,6 +212,62 @@ export async function fetchCertificate(
...
+  const targetUri = `${removeTrailingSlashes(
+    destinationServiceUri
+  )}/destination-configuration/v1/destinations/${
+    options.destinationName
+  }?$skipCredentials=true`;
+
+  const authHeader = wrapJwtInHeader(serviceToken).headers;
```
deekshas8, line 230: [nth] Move the `wrapJwtInHeader` call after the `if` block since it's not needed before that.

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -212,6 +212,62 @@ export async function fetchCertificate(
...
+        destinationServiceCache.cacheRetrievedDestinations(
+          targetUri,
+          decodeJwt(serviceToken),
+          [destination]
+        );
+      }
+      return destination;
+    })
+    .catch((error: AxiosError<any>) => {
+      if (error.response?.status === 404) {
```
deekshas8, line 259: [q] For my understanding, could you please explain why we handle 404 differently and return undefined?

test-packages/integration-tests/test/cache.spec.ts
```diff
@@ -90,18 +92,30 @@ describe('CacheDestination & CacheClientCredentialToken', () => {
...
+      'FINAL-DESTINATION',
+      wrapJwtInHeader(providerServiceToken)
+    );
+    mockSingleDestinationCallSkipCredentials(
+      nock,
+      destinationSingleResponse([destinationAuthFlow]),
+      'FINAL-DESTINATION-AUTH-FLOW',
+      wrapJwtInHeader(providerServiceToken)
+    );
     mockInstanceDestinationsCall(
```
deekshas8, line 108: [q] Is this call still needed here?

test-packages/integration-tests/test/v2/request-builder.spec.ts
```diff
@@ -182,7 +188,8 @@ describe('Request Builder', () => {
         destinationName: 'FINAL-DESTINATION'
       });
 
-    await expect(request).resolves.not.toThrow();
+    await request;
+    // await expect(request).resolves.not.toThrow();
```
deekshas8, line 192: [q] should probably be reverted?

deekshas8 (changes requested):

packages/connectivity/src/scp-cf/destination/destination-service.ts
fwilhe: My interpretation is that we use `undefined` as "The destination the user asked for does not exist", so we don't throw an error, but we also can't return a destination. The api will return 404 if the name of the destination does not exist.

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -283,7 +339,7 @@ async function callDestinationEndpoint(
 ): Promise<AxiosResponse<DestinationJson | DestinationConfiguration>> {
   if (!uri.match(/[instance|subaccount]Destinations|v1\/destinations/)) {
     throw new Error(
-      `callDestinationEndpoint was called with illegal arrgument: ${uri}. URL must be destination(s) endpoint of destination service.`
+      `callDestinationEndpoint was called with illegal argument: ${uri}. URL must be destination(s) endpoint of destination service.`
```
marikaner, line 342: [req] I can see that this was here before, but this is not a good error message. Users do not have access to `callDestinationEndpoint` and most certainly did not call it directly. This error will never happen unless we make a major mistake in our code, but still, I think we can improve the message:
```suggestion
      `Could not call destination service with URL: "${uri}". URL must be an endpoint of the destination service.`
```

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -212,6 +212,62 @@ export async function fetchCertificate(
...
+): Promise<Omit<Destination, 'AuthAndExchangeTokens'> | undefined> {
+  const targetUri = `${removeTrailingSlashes(
+    destinationServiceUri
+  )}/destination-configuration/v1/destinations/${
+    options.destinationName
+  }?$skipCredentials=true`;
+
+  if (options?.useCache) {
+    const destinationsFromCache =
+      destinationServiceCache.retrieveDestinationsFromCache(
```
marikaner, line 232: [q] So we're caching for `$?skipCredentials=true`. Not sure I understand why?

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -212,6 +212,62 @@ export async function fetchCertificate(
...
+        targetUri,
+        decodeJwt(serviceToken)
+      );
+    if (destinationsFromCache && destinationsFromCache.length > 0) {
```
marikaner, line 236: [req] You can shorten this + see our [styleguide](https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-truthyfalsy-checks-where-possible).

```suggestion
    if (destinationsFromCache?.length) {
```

marikaner: [q] How can there be more than 1 destination cached for the same key?

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -212,6 +212,62 @@ export async function fetchCertificate(
   }
 }
 
+/**
+ * Fetch a single destination by name without credentials.
```
marikaner, line 216: [req] Could you explain the use case why fetching without tokens is necessary? I remember there was a discussion about this, but not what the reason was.

marikaner: [pp] This function seems quite long to me. Please consider extracting some functions to make it easier to read.

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -352,6 +322,16 @@ class DestinationFromServiceRetriever {
     return credentials[0];
   }
 
+  private async fetchDestinationWithoutToken(
+    jwt: string
+  ): Promise<Destination | undefined> {
+    return fetchDestinationByNameWithoutTokens(
```
marikaner, line 328: [req] If the outer function is called`...WithoutToken` why is the inner `...WithoutTokenSSSS`?

packages/connectivity/src/scp-cf/destination/destination-service.ts
fwilhe: I'm not sure if there are security implications of omitting or not omitting that field here. Another motivation might be to keep the response "clean" from an api desgin perspective. I don't know the context well enough to judge this, but if no one sees an issue with `Promise<Destination | undefined>` that's fine with me.

fwilhe: > [q] So why not just `Promise<Destination | undefined>` then?

Changed it to that now after discussing with @deekshas8

packages/connectivity/src/scp-cf/destination/destination-service.ts
fwilhe: Comment added

.changeset/cool-gifts-camp.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sap-cloud-sdk/connectivity': patch
+---
+
+[Compatibility Note] When fetching a single destination from destination service, the SDK always loads and parses all available destinations and filters on the name later.
+This behavior has issues when some destinations can not be properly parsed, and it has negative performance implications when only one destination is used but many destinations were configured.
+The new behavior is to only load and parse the specific destination by name.
+The new behavior might have negative performance implications if many destinations are used.
```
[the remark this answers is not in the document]
fwilhe: Rephrased, hope it's good now. I want to keep the balance between providing enough info and not being too verbose.

packages/connectivity/src/scp-cf/destination/destination-service.ts
fwilhe: I don't understand the use-case well enough to answer this question. The underlaying assumption is that in the second call of this destination, the credentials are required?

FrankEssenberger: In the old implementation the `fetchDestinations` also had a cache with a fixed life time of 300.000 ms because the it only returned the fixed data for the destination where the destination service has also an internal cache of 5 minutes. The `skipCredentials=true` is somehow similar as it also returns just the header data but for a single destination instead of the list.

packages/connectivity/src/scp-cf/destination/destination-service.ts
FrankEssenberger: You need to know the auth flow to set the `x-tenant` or `x-user-token` headers correctly - see [here](https://api.sap.com/api/SAP_CP_CF_Connectivity_Destination/resource). Thats why we make the first call to get this information and then the second one. This could only be avoided if the user would provide the flow which is not convenient I would say.

FrankEssenberger:
Replaced by @2789 since auth flow could not be stopped completely.

PR #2633, opened 2022-07-01
Fetch token for destinations from service bindings

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#572.

If necessary a client credential token is fetched when a destination is constructed for the service binding. Support services are currently: workflow, s4-hana-cloud, business-logging, destiantion, saas-registry. Cacheing is added to the service binding destinations as well.

I also adjusted the the cache key for the register destination to use the id instead of the domain. This feels safer because the id is unique for suere. Domain should also be unique but you never know.

### DIFF
--- a/packages/connectivity/src/scp-cf/destination/destination-from-vcap.spec.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-from-vcap.spec.ts
@@ -1,36 +1,125 @@
+import {
+  mockServiceToken,
+  providerUserPayload,
+  xsuaaBindingMock
+} from '../../../../../test-resources/test/test-util';
+import * as tokenAccessor from '../token-accessor';
 import { getDestination } from './destination-accessor';
 import {
   destinationForServiceBinding,
   ServiceBinding
 } from './destination-from-vcap';
+import { destinationCache } from './destination-cache';
 
 describe('vcap-service-destination', () => {
+  const spy = jest.spyOn(tokenAccessor, 'serviceToken');
+
+  beforeAll(() => {
+    mockServiceToken();
+  });
+
   beforeEach(() => {
     mockServiceBindings();
   });
 
   afterEach(() => {
     delete process.env.VCAP_SERVICES;
+    jest.clearAllMocks();
   });
 
   it('creates a destination for the business logging service', async () => {
     await expect(
-      destinationForServiceBinding('my-business-logging')
+      destinationForServiceBinding('my-business-logging', {
+        jwt: providerUserPayload
+      })
     ).resolves.toEqual({
       url: 'https://business-logging.my.example.com',
       authentication: 'OAuth2ClientCredentials',
-      username: 'CLIENT_!_|_!_ID',
-      password: 'PASSWORD'
+      name: 'my-business-logging',
+      authTokens: [expect.objectContaining({ value: expect.any(String) })]
     });
+    expect(spy.mock.calls[0][0]['credentials']['clientid']).toBe(
+      'clientIdBusinessLogging'
+    );
+  });
+
+  it('creates a destination for the destination service', async () => {
+    await expect(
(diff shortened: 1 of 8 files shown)


packages/connectivity/src/scp-cf/destination/destination-from-vcap.spec.ts
```diff
@@ -93,18 +182,56 @@ function mockServiceBindings() {
...
+        'business-logging',
+        'logging',
+        'com.sap.appbasic.businesslogs',
+        'comsapappbasicbusinesslogs'
+      ],
       credentials: {
         writeUrl: 'https://business-logging.my.example.com',
         uaa: {
-          clientid: 'CLIENT_!_|_!_ID',
+          clientid: 'clientIdBusinessLogging',
```
florian-richter, line 199: [q] Does this change remove testing this string with special characters? (assuming special characters are a sensible test case)

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -24,19 +30,45 @@ const logger = createLogger({
  */
 export async function destinationForServiceBinding(
   serviceInstanceName: string,
-  options: DestinationForServiceBindingsOptions = {}
+  options: DestinationForServiceBindingsOptions & {
+    jwt?: JwtPayload;
+    useCache?: boolean;
+  } = {}
```
florian-richter, line 36: [nth] Define `{ jwt?: JwtPayload; useCache?: boolean; }` as a type outside of the function params so it's easier to read the function signature.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -157,20 +207,105 @@ function noServiceBindingFoundError(
...
+  };
+  const token = await serviceToken(service, { jwt });
+  return buildClientCredentialsDesntiona(
+    token,
+    serviceBinding.credentials.endpoints.workflow_odata_url,
+    serviceBinding.name
+  );
+}
+
+function buildClientCredentialsDesntiona(
```
florian-richter, line 282:
```suggestion
function buildClientCredentialsDestination(
```
[req] typo

florian-richter (changes requested):
Some questions and nice-to-haves 😊

packages/connectivity/src/scp-cf/destination/destination-from-vcap.spec.ts
FrankEssenberger: no I just wanted to have different names for each client secret. I think with the string representation we do not have problem or test need for special characters.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -187,10 +322,17 @@ export async function searchServiceBindingForDestination(
 ): Promise<Destination | null> {
   logger.debug('Attempting to retrieve destination from service binding.');
   try {
+    const jwt = options.iss
```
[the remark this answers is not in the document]
FrankEssenberger: good idea. Then all properties have the same name and as the key so object are also smaller.

test-packages/integration-tests/test/auth-flows/auth-flow.spec.ts
```diff
@@ -35,6 +35,36 @@ describe('OAuth flows', () => {
     destinationService = getService('destination');
   });
 
+  xit('creates a destination from service binding and gets a client credentials grant', async () => {
```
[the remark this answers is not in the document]
FrankEssenberger: This was the local E2E against the reals BTP account. We never enable these and in principle the test should be separate for each service but I wanted to keep it a bit shorter and easier to switch it on/off.

florian-richter (approved):

PR #2641, opened 2022-07-04
feat: support send email with proxy type: internet

jjtang1985:
This this the first step of the mail feature:
- create a mail-client package
- support proxy type: internet
- adjust automated local e2e tests
- test auth-flow manually


packages/http-client/src/http-client.ts
```diff
@@ -54,14 +53,10 @@ const logger = createLogger({
...
 ): Promise<DestinationHttpRequestConfig> {
-  const resolvedDestination = await resolveDestination(destination);
-  if (!resolvedDestination) {
-    throw Error(
-      `Failed to resolve the destination '${toDestinationNameUrl(
-        destination
-      )}'.`
-    );
-  }
+  const resolvedDestination = await resolveDestinationWithType(
```
FrankEssenberger, line 56: [q] This looks like a breaking or at least functional change if instead of a exception the logic is continued. Also why is the `buildHeaders` not complaining about the null?

FrankEssenberger: ah I just saw that the resolveDestinationWithType does the check so I guess it is fine.

packages/mail-client/src/mail-client.ts
```diff
@@ -0,0 +1,152 @@
...
+      pass: password
+    },
+    tls: {
+      /**
+       * If true the server will reject any connection which is not
+       * authorized with the list of supplied CAs. This option only has an
+       * effect if requestCert is true.
+       */
+      /** Disable tls config to fix the self signed certificate error. */
+      rejectUnauthorized: false
```
FrankEssenberger, line 87: [req] We should not do this. If I get it right we should people to configure this because this could open up for man in the middle attacks. In the repo they always recomend to use the `rejectUnauthorized` but I think this is not good: https://github.com/nodemailer/nodemailer/issues?q=self+signed. If you have a test could you add the `ca:...` property to the tls with the buffer as shown here: https://nodejs.org/api/tls.html#tlsconnectoptions-callback. If this works we should read the certificates from the cloud platform trust store (see here https://github.com/SAP/cloud-sdk-js/blob/50b9142fc72a67a68816c030c1490773d3c6d2a7/packages/connectivity/src/scp-cf/destination/destination-service.ts#L182) and put it because this is the secure way to do it.

packages/mail-client/src/mail-client.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    const mailOptions1: MailOptions = {
+      from: 'from1@example.com',
+      to: 'to1@example.com'
+    };
+    const mailOptions2: MailOptions = {
+      from: 'from2@example.com',
+      to: 'to2@example.com'
+    };
+    await expect(
+      sendMail(destination, mailOptions1, mailOptions2)
```
FrankEssenberger, line 39: [q] can I also put a `[mail1,mail2]` in the sendMail or will this fail? I think we had a similar issue in the past. We have to keep in mind the big JS user base how do not get type errors so the best thing in my opinion would to overload either `sendMail(destination,mail)` and `sendMail(destiation,[mail1,mail2])` The case with the variadic argument I find not really useful to be honest.

test-packages/e2e-tests/test/mail/mail.spec.ts
```diff
@@ -47,37 +50,48 @@ describe('Mail', () => {
...
+  //   secure: false,
+  //   auth: {
+  //     user: 'user',
+  //     pass: 'pd'
+  //   },
+  //   tls: {
+  //     // disable tls config to fix the self signed certificate error
+  //     rejectUnauthorized: false
+  //   }
+  // });
```
FrankEssenberger, line 81: [q] remove comments?

test-packages/e2e-tests/test/mail/mail.spec.ts
```diff
@@ -47,37 +50,48 @@ describe('Mail', () => {
...
-    const response = await transporter.sendMail(mailOptions[mailOptionIndex]);
-    // eslint-disable-next-line no-console
-    console.log(
-      `...email ${mailOptionIndex}/${mailOptions.length} for subject "${mailOptions[mailOptionIndex].subject}" was sent successfully.`
-    );
-  }
-  transporter.close();
-  // eslint-disable-next-line no-console
-  console.log('SMTP transport connection closed.');
+  ...mailOptions: MailOptions[]
```
FrankEssenberger, line 53: [q] where is the local mail server setup and started? I do not see a before and after step in this test?

test-packages/e2e-tests/test/mail/mail.spec.ts
```diff
@@ -29,15 +32,15 @@ describe('Mail', () => {
         );
       })
     ).toBe(true);
-  });
+  }, 60000);
```
FrankEssenberger, line 35: [q] Looks like the trafic of the local server is stored to some files, which are asserted on. Make sure the test directry is on the gitignore list and also stuff is cleaned up afterwards. Best would be that the mock FS is switched on for this folder and then the content disappears automatically.

packages/connectivity/src/scp-cf/destination/destination-accessor.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    url: 'https://my.system.com',
+    name: 'my-destination',
+    proxyType: 'Internet',
+    username: 'USER_NAME',
+    password: 'password',
+    authentication: 'BasicAuthentication',
+    type: 'HTTP'
+  };
+
+  it('resolves destination when the destination type matches', async () => {
```
FrankEssenberger, line 16: can you add a type with a mocked mail destination and asserting on the relevant parts for a mail destination.

packages/mail-client/src/mail-client.ts
```diff
@@ -0,0 +1,152 @@
...
+  originalProperties: Record<string, any>
+): { host: string; port: number } {
+  const host = originalProperties['mail.smtp.host'];
+  const port = originalProperties['mail.smtp.port'];
+
+  if (!(host && port)) {
+    throw Error(
+      `The destination '${toDestinationNameUrl(
+        destination
+      )}' does not contain host or port information as properties.`
```
FrankEssenberger, line 41: [nth] Perhaps give a hint where this is maintend - destination property bli bla blup.

packages/mail-client/src/mail-client.ts
```diff
@@ -0,0 +1,152 @@
...
+ */
+export async function sendMail<T extends MailOptions>(
+  destination: DestinationOrFetchOptions,
+  ...mailOptions: T[]
+): Promise<MailResponse[]> {
+  const resolvedDestination = await resolveDestinationWithType(
+    destination,
+    'MAIL'
+  );
+  const originalProperties =
```
FrankEssenberger, line 144: [q] As I see we consider user/pwd and smtp.host/port from the destination properties. However there are more properties: 
image
Should we also consider these like the other destination properties? Order should be User > Dest I think right?

FrankEssenberger (dismissed):
Looks good, the only major concern I have the is reject untrusted because I would like to have TLS if desired

packages/mail-client/src/mail-client.ts
FrankEssenberger: We should allow to pass it as an option to switch it off for some intranet/test case.

test-packages/e2e-tests/test/mail/mail.spec.ts
jjtang1985: Here it is:
https://github.com/SAP/cloud-sdk-js/blob/main/test-packages/e2e-tests/package.json#L10

test-packages/e2e-tests/test/mail/mail.spec.ts
jjtang1985: mock fs won't work for this e2e tests scenario, where we have client and server separations.

packages/mail-client/src/mail-client.ts
jjtang1985: - "from" is covered.
- "protocol" is anyway "smtp" for the time being
- "provider" is not well documented, so I'll ignore for now.

test-packages/e2e-tests/test/mail/mail.spec.ts
jjtang1985: I added a todo for cleaning up before merging.

packages/mail-client/src/mail-client.ts
jjtang1985: As discussed, we can work on a separate PR for the certificates.

Notes:
- we did not find the same way like http destination, where you can upload a certificate, meaning it should be passed as a parameter
- the `nodemailer` accepts the following options, which is actually the `SecureContextOptions` of `node`

``` ts
tls: {
      cert: xxx,
      ca: xxx,
      rejectUnauthorized: xxx
    }
```

jjtang1985:
> Looks good, the only major concern I have the is reject untrusted because I would like to have TLS if desired

Discussed with Frank, so the TLS will be handled when working on the on-prem.

packages/mail-client/src/mail-client.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+import { sendMail } from './mail-client';
+import { MailOptions } from './mail-client-types';
+
+describe('mail client', () => {
+  class MockMailer {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    sendMail() {}
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    close() {}
+  }
```
marikaner, line 11: [q] Why is this a class. Wouldn't it be nicer to have an object?
[req] I think you can avoid disabling lint by using `jest.fn()`.

```suggestion
const mailTransport = {
  sendMail: jest.fn(),  close: jest.fn()
};
```

packages/mail-client/src/mail-client.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      to: 'to2@example.com'
+    };
+    await expect(
+      sendMail(destination, mailOptions1, mailOptions2)
+    ).resolves.not.toThrow();
+    expect(spyCreateTransport).toBeCalledTimes(1);
+    expect(spySendMail).toBeCalledTimes(2);
+    expect(spySendMail).toBeCalledWith(mailOptions1);
+    expect(spySendMail).toBeCalledWith(mailOptions2);
+    expect(spyClose).toBeCalledTimes(1);
```
marikaner, line 45: [q] Should we also have a test for sending an attachment?

marikaner (dismissed):
I have many small suggestions and requests, but overall well readable and understandable code 👍

packages/mail-client/src/mail-client.ts
```diff
@@ -0,0 +1,167 @@
...
+
+function getMailDestinationOriginalProperties(
+  destination: Destination
+): Record<string, any> {
+  const originalProperties = destination.originalProperties;
+  if (!originalProperties) {
+    throw Error(
+      `The destination '${toDestinationNameUrl(
+        destination
+      )}' does not contain property configurations.`
```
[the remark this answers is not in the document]
jjtang1985: It's hard to say. The properties information comes from the response of the destination service.

packages/mail-client/src/mail-client.ts
```diff
@@ -0,0 +1,167 @@
...
+  const from = originalProperties['mail.from'];
+  return { from };
+}
+
+async function sendMailWithNodemailer<T extends MailOptions>(
+  destination: Destination,
+  originalProperties: Record<string, any>,
+  ...mailOptions: T[]
+): Promise<MailResponse[]> {
+  const transporter = createTransport(destination, originalProperties);
```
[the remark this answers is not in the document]
jjtang1985: It uses the `nodemailer` function `nodemailer.createTransport` and returns `Transporter`.
I would use `transporter` for reflecting the return type.

packages/mail-client/src/mail-client.ts
```diff
@@ -0,0 +1,167 @@
...
+  const mailOptionsFromDestination =
+    buildMailOptionsFromDestination(originalProperties);
+
+  const response: MailResponse[] = [];
+  for (const mailOptionIndex in mailOptions) {
+    logger.debug(
+      `Sending email ${mailOptionIndex + 1}/${mailOptions.length}...`
+    );
+    response[mailOptionIndex] = await transporter.sendMail({
```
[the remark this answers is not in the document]
jjtang1985: adjusted in the api doc

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+}
+
+/**
+ * Represents an envelope that contains the email addresses of the sender and the recipients.
+ * This interface is compatible with `MimeNode.Envelope` of `nodemailer`.
+ * @experimental This API is experimental and might change in newer versions. Use with caution.
+ */
+export interface Envelope {
+  /**
+   * The email address of the sender or is set to false.
```
marikaner, line 161:
```suggestion
   * E-mail address of the sender or `false` if it should not be shown.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  from: string | false;
+  /**
+   * An array of email addresses of recipients.
```
marikaner, line 165:
```suggestion
   * Array of e-mail addresses of recipients.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  to: string[];
+}
+
+/**
+ * Represents the mail response from the mail server.
```
marikaner, line 171:
```suggestion
 * Represents the response from the mail server.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+ * This interface is compatible with `SMTPTransport.SentMessageInfo` of `nodemailer`.
+ * @experimental This API is experimental and might change in newer versions. Use with caution.
+ */
+export interface MailResponse {
+  /**
+   * An envelope that contains the sender and recipients info.
```
marikaner, line 177:
```suggestion
   * Envelope that contains the sender and recipients info.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  envelope?: Envelope;
+  /**
+   * An array of email addresses that accepted the requests.
```
marikaner, line 181:
```suggestion
   * Array of e-mail addresses that accepted the requests.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  accepted?: (string | Address)[];
+  /**
+   * An array of email addresses that rejected the requests.
```
marikaner, line 185:
```suggestion
   * Array of e-mail addresses that rejected the requests.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  rejected?: (string | Address)[];
+  /**
+   * An array of email addresses that are processing the requests.
```
marikaner, line 189:
```suggestion
   * Array of e-mail addresses that are processing the requests.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+  pending?: (string | Address)[];
+  /**
+   * Other response message from the mail server.
+   */
+  response?: string;
+}
+
+/**
+ * Represents a mail destination configured on the SAP Business Technology Platform.
+ * Some keys of the originalProperties are ignored by the SDK for the time being including `mail.transport.protocol` and `mail.smtp.provider.*`.
```
marikaner, line 200:
```suggestion
 * Currently only `mail.transport.protocol` and `mail.smtp.provider.*` are considered.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+ * @experimental This API is experimental and might change in newer versions. Use with caution.
+ */
+export interface MailDestination {
+  /**
+   * Name of the destination retrieved from SAP Business Technology Platform.
```
marikaner, line 205:
```suggestion
   * Name of the destination.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  name?: string | null;
+  /**
+   * Type of authentication to use.
```
marikaner, line 209:
```suggestion
   * Type of authentication.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   *
```
marikaner, line 210:
```suggestion
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   * Defaults to `NoAuthentication`, unless {@link username} and {@link password} are provided, in which case the default is `BasicAuthentication`.
+   */
+  authentication?: AuthenticationType;
+  /**
+   * Proxy type to specify whether the target resides on-premise (not used).
```
marikaner, line 215:
```suggestion
   * Proxy type to specify whether the target resides on-premise (currently not used).
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  proxyType?: DestinationProxyType;
+  /**
+   * The type of the destination, defaults to 'HTTP'. The SAP Cloud SDK only understands destinations of type 'HTTP' and 'MAIL'.
```
marikaner, line 219:
```suggestion
   * Type of the destination. Defaults to 'HTTP'.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  type?: 'HTTP' | 'LDAP' | 'MAIL' | 'RFC';
+  /**
+   * Further properties of the destination as defined in destination service on SAP Business Technology Platform, possibly empty.
+   */
+  originalProperties?: { [key: string]: any };
+  /**
+   * Host of the mail server.
+   * Parsed from originalProperties['mail.smtp.host'].
```
marikaner, line 228:
```suggestion
   * Based on the additional destination property 'mail.smtp.host'.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  host?: string;
+  /**
+   * Port of the mail server.
+   * Parsed from originalProperties['mail.password'].
```
marikaner, line 233:
```suggestion
   * Based on the additional destination property 'mail.password'.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  port?: number;
+  /**
+   * Optional sender info defined in the destination properties.
+   * Parsed from originalProperties['mail.from'].
```
marikaner, line 238:
```suggestion
   * Based on the additional destination property 'mail.from'.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  from?: string;
+  /**
+   * Username to use for basic authentication, optional if other means of authentication shall be used.
+   * Parsed from originalProperties['mail.user'].
```
marikaner, line 243:
```suggestion
   * Based on the additional destination property 'mail.user'.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   * Parsed from originalProperties['mail.smtp.host'].
+   */
+  host?: string;
+  /**
+   * Port of the mail server.
+   * Parsed from originalProperties['mail.password'].
+   */
+  port?: number;
+  /**
+   * Optional sender info defined in the destination properties.
```
marikaner, line 237:
```suggestion
   * Sender info, e.g. e-mail address.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  from?: string;
+  /**
+   * Username to use for basic authentication, optional if other means of authentication shall be used.
+   * Parsed from originalProperties['mail.user'].
+   */
+  username?: string;
+  /**
+   * Password to use for basic authentication, optional if other means of authentication shall be used.
+   * Parsed from originalProperties['mail.password'].
```
marikaner, line 248:
```suggestion
   * Based on the additional destination property 'mail.password',
```

packages/mail-client/src/mail-client.ts
```diff
@@ -0,0 +1,187 @@
...
+      }" was sent successfully.`
+    );
+  }
+
+  transporter.close();
+  return response;
+}
+
+/**
+ * Sends emails to a target mail server defined in a given destination.
```
marikaner, line 150:
```suggestion
 * Sends e-mails to a target mail server defined in a given destination.
```

packages/mail-client/src/mail-client.ts
```diff
@@ -0,0 +1,187 @@
...
+ * Builds a `Transporter` between the application and the mail server, sends mails sequentially by using the `Transporter`, then closes it.
```
marikaner, line 151:
```suggestion
 * Builds a transport between the application and the mail server, sends mails sequentially by using the transport, then closes it.
```

packages/mail-client/src/mail-client.ts
```diff
@@ -0,0 +1,187 @@
...
+      ...mailOptions[mailOptionIndex]
+    });
+    logger.debug(
+      `...email ${mailOptionIndex + 1}/${mailOptions.length} for subject "${
+        mailOptions[mailOptionIndex].subject
+      }" was sent successfully.`
+    );
+  }
+
+  transporter.close();
```
marikaner, line 145: [pp] I would like to call it `transport` as the function also says `createTransport`.
```suggestion
  const transport = createTransport(mailDestination);
  const mailOptionsFromDestination =
    buildMailOptionsFromDestination(mailDestination);

  const response: MailResponse[] = [];
  for (const mailOptionIndex in mailOptions) {
    logger.debug(
      `Sending email ${mailOptionIndex + 1}/${mailOptions.length}...`
    );
    response[mailOptionIndex] = await transport.sendMail({
      ...mailOptionsFromDestination,
      ...mailOptions[mailOptionIndex]
    });
    logger.debug(
      `...email ${mailOptionIndex + 1}/${mailOptions.length} for subject "${
        mailOptions[mailOptionIndex].subject
      }" was sent successfully.`
    );
  }

  transport.close();
```

marikaner (approved):
LGTM! Let's keep sending multiple e-mails sequential for on premise testing and then try to make it parallel once this is tested.

PR #2649, opened 2022-07-06
feat: add resilience middleware

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#667.


packages/odata-common/src/request/odata-request.ts
```diff
@@ -227,9 +227,20 @@ export class ODataRequest<RequestConfigT extends ODataRequestConfig> {
       throw Error('The destination cannot be undefined.');
     }
 
-    return executeHttpRequest(destination, await this.requestConfig(), {
-      fetchCsrfToken: this.config.fetchCsrfToken
-    }).catch(error => {
+    const executeHttpRequestHandler = async () =>
+      executeHttpRequest(destination, await this.requestConfig(), {
```
FrankEssenberger, line 231: I would pass the middlewares down to the executeHttpRequest in the last argument as indicated in the ADR. Doing so needs to wrap the call only once for all use case: openapi, odata and direct usage.

FrankEssenberger:
This PR contains too many steps. Please reduce it also for easier review and do follow ups:
- do not expose the middle ware yet
- do not include the retry yet
- refactor the code by using a resilience middleware created from the resilience options see ADR with [PR](https://github.com/SAP/cloud-sdk-js/pull/2679)
- In principle your `addResilience` function should be off type MiddleWare
- Simplify the `addResilience` perhaps with the idea mentioned in the [PR](https://github.com/SAP/cloud-sdk-js/pull/2679) seeing the resilience steps as middle wares itself which are easily accumulated. 
- For some requests the `RequestContext` with its URL, method etc is not so obvious but a look at the mocks in the tests shows which URL is called e.g. xssec.

ZhongpinWang:
Superseded by SAP/cloud-sdk-js#2695.

PR #2659, opened 2022-07-08
fix: move xsuaa only when cache is needed

FrankEssenberger:
A small fix the for @mr-flannery we check the XSUAA only when cache is on. Only then we use the URL to build the cache key.


marikaner (approved):
👍

PR #2660, opened 2022-07-08
fix: Improve handling of datetimeoffset, date, time, and duration

florian-richter:
Closes SAP/cloud-sdk-backlog#766.

This improves our handling of OData v4 date and time parsing. It should now be close to the spec but does not cover 100% (marked with comments in the code) and is also more often more lenient (not marked). `moment` is definitely a limiting factor here as it does not support a 1-to-1 mapping (also due to JS shortcoming especially when handling very precise numbers).

Please check if:
- the tests are sensible
- if gaps mentioned in the comments in the tests are accidental or intentional gaps
- the changes to the actual code
 
There should not be any breaking changes, but I will add a compatibility note as it is now more strict in some places.

### TODO:
- [x] Write compatibility note


packages/odata-v4/src/de-serializers/converters.ts
```diff
@@ -4,7 +4,7 @@ import moment from 'moment';
  * @internal
  */
 export function deserializeDateToMoment(date: string): moment.Moment {
-  const parsed = moment.utc(date, 'YYYY-MM-DD', true);
+  const parsed = moment.utc(date, 'Y-MM-DD', true);
```
FrankEssenberger, line 7: [q] The single `Y` means any number of years?

packages/odata-v4/src/de-serializers/converters.ts
```diff
@@ -26,13 +26,13 @@ export function serializeToDate(value: moment.Moment): string {
 export function deserializeDateTimeOffsetToMoment(
   dateTime: string
 ): moment.Moment {
-  const prefix = 'YYYY-MM-DDTHH:mm';
-  // In moment the Z is either Offset from UTC as +-HH:mm, +-HHmm, or Z
-  const validFormats = [`${prefix}Z`, `${prefix}:ssZ`, `${prefix}:ss.SSSZ`];
+  const prefix = 'Y-MM-DDTHH:mm';
+  // In moment the ZZ is either Offset from UTC as +-HH:mm, +-HHmm, or Z
+  const validFormats = [`${prefix}ZZ`, `${prefix}:ssZZ`, `${prefix}:ss.SSSSZZ`];
```
FrankEssenberger, line 31: [q] Why there is a `S` more?

packages/odata-common/src/time.ts
```diff
@@ -5,10 +5,10 @@ export const durationRegexV2 =
   /PT(?<hours>\d{1,2}H)?(?<minutes>\d{1,2}M)?(?<seconds>\d{1,2}S)?/;
 
 /**
- * @internal
+ * @internal See https://www.w3.org/TR/xmlschema11-2/#nt-duDTFrag
  */
 export const durationRegexV4 =
-  /([+-]{1,1})?P(\d{1,2}D)?(T(\d{1,2}H)?(\d{1,2}M)?(\d{1,2}S)?(\d{2,2}\.\d+S)?)?/;
+  /([+-])?P(\d+D)?(T(\d+H)?(\d+M)?(\d+(\.\d+)?S)?)?/;
```
FrankEssenberger, line 11: [nth] a sample string as a comment matching the regex is helpful for me.

packages/odata-v4/src/de-serializers/converters.ts
florian-richter: Yes this means any number interpreted as years. This is the only way that I could find that moment parses >4-digit and negative dates (which are both supported by the spec).

Do you think the error message should be changed or is it better to ignore the edge-cases in the error message for clarity?

packages/odata-v4/src/de-serializers/converters.ts
florian-richter: 1. The extra S means that any number of partial seconds are parsed (but ignored in the result, as moment only supports milliseconds)
2. This would allow this, but works like the workaround mentioned by the user. So there will be a loss of precision if we read the input has a higher precision and write it back to the source system. (see the tests)

packages/odata-v4/src/de-serializers/converters.ts
FrankEssenberger: Better ignore it. In 99% of the cases it will be 2020 or so.

FrankEssenberger (approved):

.changeset/fair-knives-flash.md
```diff
@@ -0,0 +1,10 @@
+---
+'@sap-cloud-sdk/odata-common': patch
+'@sap-cloud-sdk/odata-v4': patch
+'@sap-cloud-sdk/temporal-de-serializers': patch
+'@sap-cloud-sdk/test-services-odata-v2': patch
+'@sap-cloud-sdk/test-services-odata-v4': patch
+'@sap-cloud-sdk/test-services-openapi': patch
```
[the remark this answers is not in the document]
florian-richter: Changesets does it like this. I don't think it's worth the effort to investigate to change it.

packages/odata-v4/src/de-serializers/converters.spec.ts
```diff
@@ -0,0 +1,435 @@
...
+      // });
+      expect(deserializeToTime('00:00:00')).toEqual({
+        hours: 0,
+        minutes: 0,
+        seconds: 0
+      });
+    });
+
+    xit('handles time with timezones', () => {
+      // TODO The spec says this is allowed
```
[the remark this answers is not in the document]
florian-richter: Before this PR, we parse all times no matter if they are valid or not. After the PR (in the current state) we only allow 00:00:00 - 23:59:59.

PR #2668, opened 2022-07-13
Update GH Actions through dependabot

marikaner:
This introduces dependabot for GH Actions


florian-richter (approved):
LGTM (although I had to look up why `directory: '/'` but it says that [here](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot) so it's probably correct)

PR #2686, opened 2022-07-15
Stop ignoring engines

marikaner:
It should be safe to not ignore the engines anymore as edm converters lifted their restriction.

UPDATE: Except for the build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
             node_modules
             */*/node_modules
           key: ubuntu-latest-${{ hashFiles('**/yarn.lock') }}
-      - run: yarn install --ignore-engines --frozen-lockfile
+      - run: yarn install --frozen-lockfile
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -29,7 +29,7 @@ jobs:
             node_modules
             */*/node_modules
           key: ubuntu-latest-${{ hashFiles('**/yarn.lock') }}
-      - run: yarn install --ignore-engines --frozen-lockfile
+      - run: yarn install --frozen-lockfile
```
marikaner, line 32:
```suggestion
      - run: yarn install --ignore-engines --frozen-lockfile
```

marikaner:
> [q] Why not remove the flag from the build?

Because 12 and 18 don't work with some other dependencies that restrict engines... (12 is puppeteer and 18 is detective-postcss, which comes in through madge`).

PR #2690, opened 2022-07-18
feat: JSON spec file takes over the same service YAML one

shojisoma:
Relates https://github.com/SAP/cloud-sdk-backlog/issues/258

open-api generator chooses JSON file when a YAML file for the same service is given. It removes the YAML files and logs a warning message as well.


marikaner:
@shojisoma Thank you for looking into this. As discussed, let's introduce globs instead.

PR #2693, opened 2022-07-19
chore: Make puppeteer a dev dependency

marikaner:
As in the title


shojisoma (approved):
I agree with that because `puppeteer` is only used in `sciprts/npm-stas.ts` that run by `.github/workflows/tests-windows.yaml`. it runs only for the development process.

PR #2695, opened 2022-07-19
feat: transform circuit breaker and timeout to use resilience middleware

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#667.

Abbreviations:
- RM: resilience middleware
- CB: circuit breaker

### Changes:

1. Apply ideas in the ADR, such as types for the middleware and options.
2. Wrap `timeout` and `CB` into **one** reduced RM.
3. Use `id` for the RM. Different RMs should have different CBs. **If a user uses the same `id` but with different RM options for `timeout` and `CB`, an error will be thrown.**
4. Use the new RM for BTP services such as `xsuaa` and `destination-service`, and also `csrf-token-fetching` in http-client.
5. Discard `timeout` in CB. Only use our own `timeout` middleware. Wrap it in the CB will have the same effect.

Example:

```ts
// Create a resilience middleware with id and options
const resilienceMiddleware = resilience({
    id: "my-resilience-1",
    timeout: () => 10000,
    circuitBreaker: () => ({
        errorThresholdPercentage: 50,
        volumeThreshold: 10,
        resetTimeout: 30000
    })
});

// A function that returns a promise
const fn = () => new Promise(resolve => setTimeout(resolve, 1000));

// Wrap the function into the resilience middleware
const wrapped = resilienceMiddleware({
    fn,
    exitChain: false
});

// Call the wrapped function
const result = await wrapped.fn();
```


packages/connectivity/src/scp-cf/resilience/resilience.ts
```diff
@@ -0,0 +1,224 @@
...
+}
+
+export function resilience(resilienceMiddlewareOptions: ResilienceMiddlewareOptions): Middleware<any> {
+  const { id } = resilienceMiddlewareOptions;
+  const middlewareOptionsPair = resilienceMiddlewareMap.get(resilienceMiddlewareOptions.id);
+  if (middlewareOptionsPair) {
+    const { middleware, options } = middlewareOptionsPair;
+    if (!_.isEqualWith(resilienceMiddlewareOptions, options, (a: ResilienceMiddlewareOptions, b: ResilienceMiddlewareOptions) => {
+      return a.circuitBreaker() === b.circuitBreaker() && a.timeout() === b.timeout();
+    })) {
```
marikaner, line 166: [req] I am not a fan of `lodash` and I don't see a need to include it. How is this better than
```suggestion
    if (resilienceMiddlewareOptions.circuitBreaker() === options.circuitBreaker() && resilienceMiddlewareOptions.timeout() === options.timeout()) {
```

packages/connectivity/package.json
```diff
@@ -44,9 +44,11 @@
     "http-proxy-agent": "^5.0.0",
     "https-proxy-agent": "^5.0.0",
     "jsonwebtoken": "^8.5.1",
+    "lodash": "^4.17.21",
```
marikaner, line 47: [req] I don't think we should include lodash now. Please also see my comment below, where this is used.

marikaner (changes requested):
Small inquire for `lodash` before doing the actual review.

packages/connectivity/src/scp-cf/destination/destination-accessor-failure-cases.spec.ts
```diff
@@ -100,7 +108,11 @@ describe('Failure cases', () => {
       await getDestination({
         destinationName,
         jwt: subscriberServiceToken,
-        enableCircuitBreaker: false,
+        resilience: {
+          id: 'getDestination',
+          timeout: defaultResilienceOptions.timeout,
```
FrankEssenberger, line 113: [q] I guess the timeout should also be a function.

packages/connectivity/src/scp-cf/resilience/resilience-options.ts
```diff
@@ -0,0 +1,70 @@
+import {
+  CircuitBreakerOptions,
+  defaultCircuitBreakerOptions
+} from './circuit-breaker-options';
+
+/**
+ * Options to configure resilience when fetching destinations.
+ * @deprecated
```
FrankEssenberger, line 8: [req] Mention the mitigation => use resilience property on XYZ instead.

packages/connectivity/src/scp-cf/resilience/resilience.ts
```diff
@@ -0,0 +1,233 @@
...
+ * TODO: Add JSDoc later.
+ * @param resilienceMiddlewareOptions - TODO: Add JSDoc later.
+ * @returns TODO: Add JSDoc later.
+ * @internal
+ */
+export function createCircuitBreakerMiddleware<T>(
+  resilienceMiddlewareOptions: ResilienceMiddlewareOptions
+): Middleware<T> {
+  const { circuitBreaker } = resilienceMiddlewareOptions;
+  const timeoutMiddlewareFn = (
```
FrankEssenberger, line 82: [q] I guess this is a copy paste error and the name should be different.

packages/connectivity/src/scp-cf/resilience/resilience.ts
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * TODO: Add JSDoc later.
+ * @param resilienceMiddlewareOptions - TODO: Add JSDoc later.
+ * @returns TODO: Add JSDoc later.
+ */
+export function resilience(
+  resilienceMiddlewareOptions: ResilienceMiddlewareOptions
+): Middleware<any> {
+  const { id } = resilienceMiddlewareOptions;
```
FrankEssenberger, line 167: [q] Is this mapped introduced to have separate entries per btp-service e.g. circuit breaker? If so the request context should be used to determine the URL beeing call as well as the tenant and ensure a circuit breaker per service and tenant.

packages/connectivity/src/scp-cf/resilience/resilience-options.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+export interface MiddlewareInOutOptions<T> {
+  fn: RequestHandler<T>;
+  exitChain?: boolean;
+}
+
+/**
+ * TODO: Add JSDoc later.
+ */
+export type Middleware<T> = (
```
FrankEssenberger, line 68: [q] I guess the context should be passed along the middle ware chain.

packages/connectivity/src/scp-cf/resilience/resilience.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/**
+ * TODO: Add JSDoc later.
+ * @param resilienceMiddlewareOptions - TODO: Add JSDoc later.
+ * @returns TODO: Add JSDoc later.
+ */
+export function resilience(
+  resilienceMiddlewareOptions: ResilienceMiddlewareOptions
+): Middleware<any> {
```
FrankEssenberger, line 166: [req] Add a generic type instead of <any>

packages/connectivity/src/scp-cf/destination/destination-accessor-failure-cases.spec.ts
ZhongpinWang: That's right. But the `defaultResilienceOptions.timeout` is already a function.

packages/connectivity/src/scp-cf/resilience/resilience-options.ts
ZhongpinWang: Thanks for the reminder. I added the links to the new options.

packages/connectivity/src/scp-cf/resilience/resilience.ts
ZhongpinWang: Good spot!

packages/connectivity/src/scp-cf/resilience/resilience.ts
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * TODO: Add JSDoc later.
+ * @param fn - TODO: Add JSDoc later.
+ * @param options - TODO: Add JSDoc later.
+ * @returns TODO: Add JSDoc later.
+ */
+export function callWithResilience<T>(
+  fn: RequestHandler<T>,
+  options: ResilienceOptions & { resilience?: ResilienceMiddlewareOptions }
```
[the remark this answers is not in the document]
ZhongpinWang: Theoretically yes. I am just not so sure if we need context (or even function) anymore if `id` can be used. Or if we decide to use context, I will remove the `id` and use `RequestContext` as the key to the map of CB. @FrankEssenberger

packages/connectivity/src/scp-cf/resilience/resilience.ts
ZhongpinWang: Currently, `getClientCredentialsToken`, `getUserToken`, `callDestinationService` and `getCsrfHeaders` (all btp services) are using different CBs. Their `id`s are in the format of `btpService-<function-name>`. I don't know how to use `url` for some of the services like `xsuaa`. We are not calling the service directly using `url` but via the `xssec` API.

Tenant-awareness can be achieved by having a `Map<tenant_id, Map<middleware_id, Middleware>>` (or change the `middleware_id` to RequestContext).

packages/connectivity/src/scp-cf/resilience/resilience-options.ts
ZhongpinWang: Same above.

packages/connectivity/src/scp-cf/resilience/resilience.ts
ZhongpinWang: Done.

PR #2703, opened 2022-07-22
fix: OpenAPI generator should create output directory

shojisoma:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/416

As this [comment](https://github.com/SAP/cloud-sdk-backlog/issues/416#issuecomment-1191386388) about the root cause, the problem is `fs.rmdir(path, { recursive: true })` is used in `generator()`that will be deprecated in future Node versions. 

### Change:

1. Changed `fs.rmdir` to ~~`fs.rm`~~ which will be supported in future versions.
2. Added --clearOutputDir in a test of a file creating process to test this issue.

### Update:
Since `fs.rm` is not supported in Node v12, I use `fs-extra` instead of it.

### DIFF
--- a/packages/openapi-generator/src/generator.ts
+++ b/packages/openapi-generator/src/generator.ts
@@ -8,6 +8,7 @@ import {
   formatJson,
   ErrorWithCause
 } from '@sap-cloud-sdk/util';
+import { emptyDirSync } from 'fs-extra';
 import {
   getSdkMetadataFileNames,
   getSdkVersion,
@@ -44,7 +45,7 @@ import {
 import { tsconfigJson } from './options/tsconfig-json';
 import { sdkMetadata } from './sdk-metadata';
 
-const { rm, mkdir, lstat } = promisesFs;
+const { mkdir, lstat } = promisesFs;
 const logger = createLogger('openapi-generator');
 
 /**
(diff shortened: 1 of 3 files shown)


packages/openapi-generator/src/generator.ts
```diff
@@ -73,7 +74,7 @@ export async function generateWithParsedOptions(
   }
 
   if (options.clearOutputDir) {
-    await rmdir(options.outputDir, { recursive: true });
+    await emptyDirSync(options.outputDir);
```
florian-richter, line 77:
```suggestion
    // Use newer fs.rm() if available and fall back to fs.rmdir() if not available.
    // This can be removed if node 14 is the oldest supported version.
    const rm = promisesFs.rm || promisesFs.rmdir;
    await rm(options.outputDir, { recursive: true });
```
[req] Please avoid adding dependencies unless necessary.

florian-richter (changes requested):

packages/openapi-generator/src/generator.ts
shojisoma: thank you for your review. I replaced `emptyDirSync` with the `fs` methods.

florian-richter (approved):
🚀

PR #2704, opened 2022-07-22
refactor: set validitytime in cache as milliseconds

shojisoma:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/534
Make `Cache` class just accepts milliseconds instead of `DateInputObject` like `new Cache<T>(validityTime?: number)`

### Change
1. Changed `validityTime` to `number`
2. Removed interface `DateInputObject`
3. Adjusted other parts that use Cache class with `DateInputObject`

### DIFF
--- a/packages/connectivity/src/scp-cf/cache.ts
+++ b/packages/connectivity/src/scp-cf/cache.ts
@@ -38,11 +38,11 @@ export class Cache<T> implements CacheInterface<T> {
    * Default validity period for each entry in cache.
    * If `undefined`, all cached entries will be valid indefinitely.
    */
-  private defaultValidityTime: number | undefined;
+  private defaultValidityTimeInMs: number | undefined;
 
-  constructor(validityTime?: number) {
+  constructor(validityTimeInMs?: number) {
     this.cache = {};
-    this.defaultValidityTime = validityTime;
+    this.defaultValidityTimeInMs = validityTimeInMs;
   }
 
   /**
@@ -80,7 +80,7 @@ export class Cache<T> implements CacheInterface<T> {
   set(key: string | undefined, item: CacheEntry<T>): void {
     if (key) {
       const expires =
-        item.expires ?? inferExpirationTime(this.defaultValidityTime);
+        item.expires ?? inferExpirationTime(this.defaultValidityTimeInMs);
       this.cache[key] = { entry: item.entry, expires };
     }
   }
@@ -94,11 +94,11 @@ function isExpired<T>(item: CacheEntry<T>): boolean {
 }
 
 function inferExpirationTime(
-  expirationTime: number | undefined
+  validityTimeInMs: number | undefined
 ): number | undefined {
-  return expirationTime
+  return validityTimeInMs
     ? new Date()
-        .setMilliseconds(new Date().getMilliseconds() + expirationTime)
+        .setMilliseconds(new Date().getMilliseconds() + validityTimeInMs)
         .valueOf()
     : undefined;
 }
--- a/packages/connectivity/src/scp-cf/cache.ts
+++ b/packages/connectivity/src/scp-cf/cache.ts
@@ -96,9 +96,10 @@ function isExpired<T>(item: CacheEntry<T>): boolean {
 function inferExpirationTime(
   validityTimeInMs: number | undefined
 ): number | undefined {
+  const now = new Date();
   return validityTimeInMs
     ? new Date()
-        .setMilliseconds(new Date().getMilliseconds() + validityTimeInMs)
+        .setMilliseconds(now.getMilliseconds() + validityTimeInMs)
         .valueOf()
     : undefined;
 }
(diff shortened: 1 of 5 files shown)


packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -48,9 +38,9 @@ export class Cache<T> implements CacheInterface<T> {
    * Default validity period for each entry in cache.
    * If `undefined`, all cached entries will be valid indefinitely.
    */
-  private defaultValidityTime: DateInputObject | undefined;
+  private defaultValidityTime: number | undefined;
```
florian-richter, line 41:
```suggestion
  private defaultValidityTimeInMs: number | undefined;
```
[req] In my opinion, it is very confusing to just have `number` as a type and not state what unit of measurement is intended

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -48,9 +38,9 @@ export class Cache<T> implements CacheInterface<T> {
...
 
-  constructor(validityTime?: DateInputObject) {
+  constructor(validityTime?: number) {
```
florian-richter, line 43:
```suggestion
  constructor(validityTimeInMs?: number) {
```
[req]

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -104,22 +94,11 @@ function isExpired<T>(item: CacheEntry<T>): boolean {
 }
 
 function inferExpirationTime(
-  expirationTime: DateInputObject | undefined
+  expirationTime: number | undefined
```
florian-richter, line 97:
```suggestion
  validityTimeInMs: number | undefined
```
[req] Add `InMs` to indicate unit of measurement + Consistent naming, parameter should be called `validityTimeInMs`

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -104,22 +94,11 @@ function isExpired<T>(item: CacheEntry<T>): boolean {
...
 ): number | undefined {
   return expirationTime
-    ? inferExpirationTimeFromDate(expirationTime)
+    ? new Date()
+        .setMilliseconds(new Date().getMilliseconds() + expirationTime)
+        .valueOf()
```
florian-richter, line 102: [pp] I like to put `const now = new Date();` above. AFAIK `new` does not bind as strongly as calling a function so you probably intend this: `(new Date()).getMilliseconds()` but actually call this: `new (Date().getMilliseconds())`.

florian-richter (changes requested):

packages/connectivity/src/scp-cf/cache.ts
shojisoma: thank you for your comment. that's right. I added `InMs`

packages/connectivity/src/scp-cf/cache.ts
shojisoma: I added `InMs` here too

packages/connectivity/src/scp-cf/cache.ts
shojisoma: thank you for your comment. I changed the parameter name as your comment for consistent naming.

packages/connectivity/src/scp-cf/cache.ts
shojisoma: thank you for your comment. I moved out `new Date()` as another const variable.

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -104,22 +94,12 @@ function isExpired<T>(item: CacheEntry<T>): boolean {
...
 
 function inferExpirationTime(
-  expirationTime: DateInputObject | undefined
+  validityTimeInMs: number | undefined
 ): number | undefined {
-  return expirationTime
-    ? inferExpirationTimeFromDate(expirationTime)
+  const now = new Date();
+  return validityTimeInMs
+    ? new Date()
```
florian-richter, line 101:
```suggestion
    ? now
```
[req] This also applies here

packages/connectivity/src/scp-cf/destination/destination-cache.ts
```diff
@@ -38,7 +38,7 @@ export interface DestinationCacheInterface {
 export class DefaultDestinationCache implements DestinationCacheInterface {
   cache: Cache<Destination>;
 
-  constructor(validityTime?: DateInputObject) {
+  constructor(validityTime?: number) {
```
florian-richter, line 41:
```suggestion
  constructor(validityTimeInMs?: number) {
```

florian-richter (commented):
A bit more feedback

packages/connectivity/src/scp-cf/cache.ts
shojisoma: thank you for your review again. I forgot to replace it. I did it.

packages/connectivity/src/scp-cf/destination/destination-cache.ts
shojisoma: thank you for your review. I should change here also. I add `InMs` here too.

florian-richter (approved):
👍

PR #2715, opened 2022-07-29
fix: glob pattern in openapi-generator on windows

deekshas8:
The generation of openapi- clients fails on Windows when globs are used as input.
On Windows, `\\` is used as path separator, however only `/` character is used by  glob implementation. This PR adjusts the glob pattern passed as input to the openapi-generator.


florian-richter (approved):
I can't verify locally, but the changes make sense to me.

PR #2730, opened 2022-08-03
feat: allow deserialization to all request builders

shojisoma:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/698

Currently, only the action/function import request builder has `dataAccessor()` to deserialize different response structures. It also makes sense for all other request builders.


FrankEssenberger:
Closed since not meaningful for not funciton imports. See linked BLI for details.

PR #2731, opened 2022-08-03
feat: support sending emails with OnPrem proxy

jjtang1985:
For OnPrem proxy:
1. a socket connection is established, where custom request/response handlers are needed
1. use that socket connection with nodemailer instead of connecting the mail server directly (internet proxy)2. 
1. close the socket + transport 

e2e test:
Application deployed on CF can send mails with OnPrem proxy.


packages/mail-client/src/mail-client.spec.ts
```diff
@@ -44,6 +56,47 @@ describe('mail client', () => {
...
+      name: 'my-destination',
+      type: 'MAIL',
+      authentication: 'BasicAuthentication',
+      proxyType: 'OnPremise',
+      proxyConfiguration: {
+        host: 'smtp.gmail.com',
+        port: 587
+      }
+    };
+    const mailOptions1: MailOptions = {
```
tomfrenken, line 89: [q] Why is it called `mailOptions1`?
Is there also some other `mailOptions2`?
[req] Either way, if there is `mailOptions2` it should probably have a more specific name.

tomfrenken (changes requested):
Some findings/questions, after that probably a short follow-up call.

packages/connectivity/src/scp-cf/connectivity-service-types.ts
```diff
@@ -32,6 +32,11 @@ export interface ProxyConfiguration {
    * HTTP headers to be added to a request if tunneled through the proxy.
    */
   headers?: ProxyConfigurationHeaders;
+
+  /**
+   * A JWT for proxy authorization.
```
FrankEssenberger, line 37: [nth] I would assume any authorization header could be accepted by the server. So perhaps adjust the description to something more general: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization

FrankEssenberger: [q] Also there is a `Proxy-Authorization` property in the `headers` which is added I [here](https://github.com/SAP/cloud-sdk-js/blob/84976bfdc835dd4b4aeadefbc1d3577a33cd43cc/packages/connectivity/src/scp-cf/authorization-header.ts#L144) I thought to the request. Could you double check if you really need this new property.

FrankEssenberger: I assume you do not use the destination for the mail, but still we could use the existing property.

packages/connectivity/src/scp-cf/connectivity-service.spec.ts
```diff
@@ -140,4 +145,28 @@ describe('connectivity-service', () => {
       done();
     });
   });
+
+  it('[MAIL] adds a proxy configuration containing at least the host, the port, and the "proxy-authorization"', async () => {
```
FrankEssenberger, line 149: [pp] I would use a `descirbe` to separate the `mail` related thest. I do not like the `[MAIL]` tag so much

packages/connectivity/src/scp-cf/connectivity-service.ts
```diff
@@ -65,6 +72,24 @@ export function proxyHostAndPort(): HostAndPort {
...
 }
 
+/**
+ * @internal
+ * @returns Socks Proxy Configuration
+ */
+export async function socksProxyHostAndPort(): Promise<ProxyConfiguration> {
+  const service = readConnectivityServiceBinding();
+  const connectivityServiceCredentials = service.credentials;
+  const connectivityServiceToken = await serviceToken('connectivity', {
```
FrankEssenberger, line 82: [req] I think you can pass the service directly to the serviceToken method.

packages/connectivity/src/scp-cf/destination/proxy-util.ts
```diff
@@ -259,6 +259,10 @@ export function proxyAgent(
       return {
         httpsAgent: new HttpsProxyAgent(agentConfig)
       };
+    default:
```
FrankEssenberger, line 262: [nth] I guess here the name `http-proxy-util.ts`is not more appropiate.

packages/mail-client/src/mail-client.spec.ts
```diff
@@ -44,6 +56,47 @@ describe('mail client', () => {
     expect(spySendMail).toBeCalledTimes(2);
     expect(spySendMail).toBeCalledWith(mailOptions1);
     expect(spySendMail).toBeCalledWith(mailOptions2);
-    expect(spyClose).toBeCalledTimes(1);
+    expect(spyCloseTransport).toBeCalledTimes(1);
+  });
+
+  it('[OnPrem] should create transport/socket, send mails and close the transport/socket', async () => {
```
FrankEssenberger, line 62: puh, Junjie is the master spy -#JamesBond

packages/mail-client/src/mail-client.ts
```diff
@@ -122,7 +177,11 @@ async function sendMailWithNodemailer<T extends MailOptions>(
   mailDestination: MailDestination,
   ...mailOptions: T[]
 ): Promise<MailResponse[]> {
-  const transport = createTransport(mailDestination);
+  let socket: SocksSocket | undefined;
+  if (mailDestination.proxyType === 'OnPremise') {
+    socket = await createSocket(mailDestination);
+  }
+  const transport = await createTransport(mailDestination, socket);
```
FrankEssenberger, line 184: [pp] I know the `for` loop was already there from the last PR. I prefer the pattern of a await Promise.all(opts.map.(opt,idx=>{})) I assume that is also stronger parallelized, but if it is intentional to not overload the server keep it.

packages/mail-client/src/mail-client.ts
```diff
@@ -79,14 +87,50 @@ function getCredentials(
...
+    destination: {
+      host: mailDestination.host!,
+      port: mailDestination.port!
+    }
+  };
+  const socketConnection = await SocksClient.createConnection(
+    connectionOptions
+  );
+  const socksSocket = socketConnection.socket as SocksSocket;
+  socksSocket._readableState.readableListening = true;
```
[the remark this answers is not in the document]
jjtang1985: I tested again, and without this change it worked, so I'll remove this line.

packages/connectivity/src/scp-cf/connectivity-service-types.ts
jjtang1985: Yes, you are right when we are talking about a "header", which saved in line 34.
However, when using non-http (e.g., socket) protocols, the header is not applicable any more.
JWT is a common auth mechanism nowadays, so I added this key.

packages/mail-client/src/mail-client.ts
jjtang1985: I discussed with Marika about the sequential v.s. parallel and we decided to have a small ticket (https://github.com/SAP/cloud-sdk-backlog/issues/820) to make the implementation/test easier.
Once we have the e2e tests in the pipeline, we can send emails in parallel.

FrankEssenberger (approved):

PR #2754, opened 2022-08-11
fix: use socks 2.6.2

jjtang1985:
When using `socks@2.7.0`  (with `nodemailer`) for creating socket connection, it failed.
As it's working with `socks@2.6.2`, let's stick to the old version for now.

Also, there is a type fix about port.

### DIFF
--- a/packages/mail-client/src/mail-client.ts
+++ b/packages/mail-client/src/mail-client.ts
@@ -10,7 +10,7 @@ import {
   transformVariadicArgumentToArray
 } from '@sap-cloud-sdk/util';
 import nodemailer, { SentMessageInfo, Transporter } from 'nodemailer';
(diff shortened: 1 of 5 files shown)


packages/mail-client/src/mail-client.ts
```diff
@@ -87,25 +87,32 @@ function getCredentials(
   return { username, password };
 }
 
-async function createSocket(mailDestination: MailDestination): Promise<Socket> {
+/**
+ * @internal
```
tomfrenken, line 91: [pp] It would be helpful if there was a link to this `custom_auth_method: 0x80` thing so that someone can understand it immediately.

tomfrenken (approved):
Just one comment, otherwise LGTM.

PR #2762, opened 2022-08-12
feature: Batch for GET function imports

FrankEssenberger:
Relates  SAP/cloud-sdk-backlog#810.

This adds the function imports to the ReadRequests in batch. The POST related functions and action will go to the `WriteRequests` in a separate PR. Also response data transformation should be discussed. For the entities we have something like `.as(entityApi)` we could do something similar for the function import response. But first the basics.

Some function imports have not serializable entities. We removed the `execute` method for these. This caused type problems since the `MethodRequestBuilder` has an `execute` function. However, since batch does not need an `execute`, I removed it for all related request builders. 

I also realized, that the CAP server expected the path to be relative to the entity, but the trippin worked with an absolute path.


packages/odata-common/src/request-builder/batch/batch-request-builder.ts
```diff
@@ -108,3 +108,28 @@ export class BatchRequestBuilder<
...
+
+function buildMapEntry<DeSerializersT extends DeSerializers>(
+  request: AllBuilderTypes<DeSerializersT>
+): Record<string, EntityApi<EntityBase, DeSerializersT>> {
+  if (isActionFunctionImport(request)) {
+    return {};
+  }
+  return {
+    [request._entityApi.entityConstructor._entityName]: request._entityApi
+  };
```
florian-richter, line 134:
```suggestion
  return isActionFunctionImport(request) ? {} : {
    [request._entityApi.entityConstructor._entityName]: request._entityApi
  };
```
[pp]

packages/generator/src/batch/imports.ts
```diff
@@ -36,5 +40,15 @@ export function importBatchDeclarations(
 }
 
 function getNamedImports(service: VdmServiceMetadata): string[] {
-  return service.entities.map(e => e.className);
+  const complexTypes = service.functionImports
+    .filter(
+      f =>
+        f.returnType.returnTypeCategory === VdmReturnTypeCategory.COMPLEX_TYPE
```
florian-richter, line 46:
```suggestion
      ({returnType}) =>
        returnType.returnTypeCategory === VdmReturnTypeCategory.COMPLEX_TYPE
```
[pp] Destructuring the parameter remove the need for the arbitrarily named `f`

FrankEssenberger: ah, I have to remember deconstructing also in array function

packages/odata-common/src/request-builder/batch/batch-request-builder.ts
```diff
@@ -33,8 +34,9 @@ export class BatchRequestBuilder<
     readonly defaultServicePath: string,
     readonly requests: (
       | BatchChangeSet<DeSerializersT>
-      | GetAllRequestBuilderBase<EntityBase, DeSerializersT>
-      | GetByKeyRequestBuilderBase<EntityBase, DeSerializersT>
+      | Omit<GetAllRequestBuilderBase<EntityBase, DeSerializersT>, 'execute'>
+      | Omit<GetByKeyRequestBuilderBase<EntityBase, DeSerializersT>, 'execute'>
+      | Omit<ActionFunctionImportRequestBuilderBase<any, any>, 'execute'>
```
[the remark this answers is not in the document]
jjtang1985: I guess, because we don't know the parameters/return types here.

packages/odata-v2/src/request-builder/batch-request-builder.spec.ts
```diff
@@ -14,13 +20,58 @@ Content-Type: application/json; charset=utf-8
...
+      'GET /sap/opu/odata/sap/API_TEST_SRV/TestFunctionImportGET HTTP/1.1',
+      'Content-Type: application/json',
+      'Accept: application/json',
+      '',
+      '',
+      `--batch_${boundary}--`,
+      ''
+    ].join('\r\n');
+
+    const bodl =
```
jjtang1985, line 56: [q] I guess this is not used?

jjtang1985 (changes requested):
Looks good in general.
I left some minor suggestions.

test-packages/e2e-tests/test/batch.spec.ts
```diff
@@ -15,6 +17,33 @@ describe('batch', () => {
...
+
+    const [responseSdk, responseInt, responseGetAll] = await batch(
+      returnSapCloudSdk({}),
+      returnInt({ param: 123 }),
+      requestBuilder.getAll().top(1)
+    )
+      .skipCsrfTokenFetching()
+      .withSubRequestPathType('relativeToEntity')
+      .execute(destination);
+    if (responseSdk.isReadResponse()) {
```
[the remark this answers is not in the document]
FrankEssenberger: made an assertion. Unfortunately the type guard does not work with expect(isReadResponse()).toBe(true), I needed to use an exception to have the type correct after the assertion, but I think it is worth the more regorous test.

packages/odata-common/src/request-builder/batch/batch-request-builder.ts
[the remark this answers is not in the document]
FrankEssenberger: yes that is the reason. I changed it to `unknown`for the return type, because we really do not have any base type and for the request config I used the base of the Action/Function Imports  which is the `ODataRequestConfig`

packages/odata-common/src/request-builder/batch/batch-response-parser.ts
```diff
@@ -160,6 +160,12 @@ export function parseHttpCode(response: string): number {
     return parseInt(group[1].toString());
   }
 
+  // The Trippin sample service returns this Unknown Status Code but if the @odata metadata is there we got data.
+  if (response.match('Unknown Status Code') && response.match('@odata')) {
```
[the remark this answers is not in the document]
FrankEssenberger: added a log

packages/generator/src/batch/imports.ts
```diff
@@ -18,6 +19,9 @@ export function importBatchDeclarations(
         'GetByKeyRequestBuilder',
         'ODataBatchRequestBuilder',
         'UpdateRequestBuilder',
+        ...(service.functionImports.length > 0
```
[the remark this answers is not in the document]
FrankEssenberger: I will do the post function imports as a next thing.

packages/odata-common/src/request-builder/batch/batch-change-set.ts
```diff
@@ -17,11 +18,20 @@ export class BatchChangeSet<
...
     readonly boundary: string = `changeset_${uuid()}`
   ) {}
 }
+
+/**
+ *
+ *  Some function imports contain not serializable entities and the execute() method is removed from them.
+ *  Since the execute method is not needed in batch the execute it is removed from all function imports.
+ *
+ * @Internal
```
deekshas8, line 31: [req] The type is marked internal here but exported in index.ts.

test-packages/e2e-tests/test/batch.spec.ts
```diff
@@ -15,6 +17,38 @@ describe('batch', () => {
   beforeEach(async () => deleteEntity(entityKey, destination));
   afterEach(async () => deleteEntity(entityKey, destination));
 
+  it('should execute multiple function imports', async () => {
+    await deleteEntity(1234, destination);
```
deekshas8, line 21: [q] Not sure I understand why we call `deleteEntity` here.

jjtang1985: [req] I guess this is not intended.

packages/odata-v2/src/request-builder/batch-request-builder.spec.ts
```diff
@@ -14,13 +20,56 @@ Content-Type: application/json; charset=utf-8
...
+    const response = await batch(
+      functionImports.testFunctionImportGet({} as any)
+    ).execute({ url: baseUrl });
+    if (response[0].isReadResponse()) {
+      const casted = testFunctionImportGet({} as any).responseTransformer(
+        response[0].body
+      );
+      expect(casted).toEqual('MyText');
+    }
+    jest.restoreAllMocks();
```
jjtang1985, line 70: [pp] I would move it to `afterEach`.

packages/odata-common/src/request-builder/batch/batch-change-set.ts
jjtang1985: Then our pipeline should catch it?

jjtang1985 (approved):
Please double check the `deleteEntity` before merging.

PR #2772, opened 2022-08-17
refactor: unify package description generator

shojisoma:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/604

this unifies 4 package description generators into one.
the 4 generators are in these files

1 & 2. https://github.com/SAP/cloud-sdk-js/blob/main/packages/generator/src/package-description.ts
3. https://github.com/SAP/cloud-sdk-js/blob/main/packages/openapi-generator/src/sdk-metadata/package-description.ts
4. https://github.com/SAP/cloud-sdk-js/blob/8283b44208f4d767a2e135a98dc55832b34b3e51/packages/openapi-generator/src/file-serializer/package-json.ts

#### Changes

- unified 4 package description generators into one below and change the sentence not to use a word `VDM`
```
#packages/generator-common/src/package-description.ts

function packageDescription(
  packageName: string,
  formatter?: (packageName: string) => string
): string {
  const formattedPackageName = formatter ? formatter(packageName) : packageName;
  return `SAP Cloud SDK for JavaScript: Generated client for service ${formattedPackageName}`;
}

```
- remove the 4 generators
- made changes in function calls that call the 4 generators in order to accept a new unified one
- update snapshots and package.jsons relating to the 4 generators


packages/generator/src/sdk-metadata/pregenerated-lib.ts
```diff
@@ -13,6 +10,8 @@ export function getServiceDescription(
   options: GeneratorOptions
 ): string {
   return options.s4hanaCloud
-    ? s4hanaCloudDescription(service.directoryName)
-    : genericDescription(service.directoryName);
+    ? packageDescription(service.directoryName, (directoryNameRaw: string) =>
```
FrankEssenberger, line 13: [q] @marikaner could we perhaps remove the s4hanaCloud case completly and just use the name as is? In the future customers will create their own services anyhow and they will not really care what is in the package.json descirption field of the generated service I guess.

FrankEssenberger (approved):
LGTM - Please consider what Marika says about the s4HanaCould options. Perhaps we can remove even more.

packages/generator-common/src/package-description.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @internal
+ */
+export function packageDescription(
+  packageName: string,
+  formatter?: (packageName: string) => string
+): string {
+  const formattedPackageName = formatter ? formatter(packageName) : packageName;
+  return `SAP Cloud SDK for JavaScript: Generated client for service ${formattedPackageName}`;
```
marikaner, line 9: [req]
@FrankEssenberger I currently find this confusing because in the service mapping we make a difference between service name (e.g. "Business Partner Service") and npm package name (e.g. "@sap/cloud-sdk-vdm-business-partner-service").
This seems to use the package name, which is incorrect.

marikaner: Also, when I look further into the code it seems that this is even derived from the directory name, not the service name nor package name, which I think mixes even more things in here.

packages/generator/src/sdk-metadata/pregenerated-lib.ts
marikaner: I agree. Please see my other comment regarding the `packageDescription` function. I would just use what is in the service mapping / options per service (and if there is nothing this should be filled with a default initially).

packages/openapi-generator/src/generator.ts
```diff
@@ -353,7 +351,10 @@ async function generatePackageJson(
     'package.json',
     packageJson({
       npmPackageName: packageName,
-      description: genericDescription(directoryName),
+      description: packageDescription(
+        directoryName,
+        (directoryNameRaw: string) => directoryNameRaw.split('-').join(' ')
+      ),
```
marikaner, line 357: [nth] Seems like duplicate code to me. With my other comments, I guess this can be made obsolete?

marikaner (changes requested):
I think we should fix where the name of the service comes from. Otherwise looks good for me.

packages/generator/src/sdk-metadata/pregenerated-lib.ts
shojisoma: removed `s4hanaCloud` case

packages/generator-common/src/package-description.ts
shojisoma: package generator and openapi-generator have readme.ts that make README.md. The service name that is retrieved by readme.ts is there in the first line of it. [`VdmServiceMetadata.speakingModuleName`](https://github.com/SAP/cloud-sdk-js/blob/8678c543f5193df838d4b440f232598b2a911a6f/packages/generator/src/service/readme.ts#L15)  and [`openApiDocument.serviceName`](https://github.com/SAP/cloud-sdk-js/blob/8678c543f5193df838d4b440f232598b2a911a6f/packages/openapi-generator/src/file-serializer/readme.ts#L14) are used as the service name in the readme.ts. I only use these 2 for the service name in the package.json description, not `packageName` or `directoryName` as before.

packages/generator/src/sdk-metadata/pregenerated-lib.ts
shojisoma: the service name that is in the service mapping/options per service comes from [`VdmServiceMetadata.speakingModuleName`](https://github.com/SAP/cloud-sdk-js/blob/8678c543f5193df838d4b440f232598b2a911a6f/packages/generator/src/service/readme.ts#L15). I use this for all the service names.

FrankEssenberger (approved):
I agree the speakingName (dashes removed) and serviceName (as in the options) is a bit inconsistent. However, we do not have a serviceName property in the service mapping for oData so for me this is good enough and in the end the value in the descriotion of the generated package.json is not that critical I guess and the important thing was to get rid of the `VDM` word.

FrankEssenberger:
As discussed please add a changeset before merging.

PR #2775, opened 2022-08-18
fix: make jsdoc annotation more strict

FrankEssenberger:
In my last PR #2762 @deekshas8 pointed out that she saw some issue with the API check. I investigated the issue and the resons was that I used `@Internal` which is accepted by jsdoc, but the compiler flag needs to be in lower case to work i.e. `@internal`. 

In this PR I added the rule `jsdoc/check-tag-name` which works for most of the tags. Unfortunately it does not work 100% for internal which is special because it disables the jsdoc requirments partially. With this rule tags like:

`
@pramaFoo @Internal1 @something
`

would be not possible. However, `@Internal` still worked. Therefore I added one eslint regex which could come in handy also in other situation. I added also a compatibility node because users could get warnings. In this case they would need to add their tags to the allowed words.


packages/eslint-config/index.js
```diff
@@ -183,6 +192,7 @@ module.exports = {
     'jsdoc/check-alignment': 'error',
     'jsdoc/check-indentation': 'error',
     'jsdoc/check-param-names': 'error',
+    'jsdoc/check-tag-names' : ['error',{"definedTags": ["packageDocumentation","typeParam","experimental"]}],
```
deekshas8, line 195:
```suggestion
    'jsdoc/check-tag-names' : ['error',{ 'definedTags': ['packageDocumentation', 'typeParam', 'experimental'] }],
```

scripts/check-public-api.ts
```diff
@@ -251,7 +251,7 @@ export async function parseIndexFile(filePath: string): Promise<string[]> {
     /export \* from '([\w/.]+)'/g,
     fileContent
   );
-  const starFileExports = await Promise.all(
+  const starFileExports =  await Promise.all(
```
deekshas8, line 254:
```suggestion
  const starFileExports = await Promise.all(
```

scripts/check-public-api.spec.ts
```diff
@@ -168,6 +169,24 @@ describe('check-public-api', () => {
       mock.restore();
     });
 
+    it('considers',async ()=>{
```
deekshas8, line 172: [q] Isn't this test similar to the one above (which also checks referenced imports?

scripts/check-public-api.ts
deekshas8: [info] The linting is a bit off (also in the .spec file). I don't exactly remember why we exclude scripts/ in .eslintrc

scripts/check-public-api.spec.ts
FrankEssenberger: ups needs to be delete

scripts/check-public-api.spec.ts
```diff
@@ -168,6 +169,24 @@ describe('check-public-api', () => {
       mock.restore();
     });
 
+    it('parses exports types correctly',async ()=>{
```
jjtang1985, line 172: [q] Is this related to the regex rule?

jjtang1985 (approved):
LGTM, as far as you tested it the regex rule.

scripts/check-public-api.spec.ts
FrankEssenberger: no I thought the error was related to export type so I added this extra test in the hope to reproduce the issue. I thought it is good to keep if we change our parsing to provide some extra safety.

PR #2783, opened 2022-08-22
chore: add compatibility note for  #2775 

FrankEssenberger:
Forgot to check in the compability note for #2775


jjtang1985 (approved):
Good point.

PR #2789, opened 2022-08-23
fix: Make parsing forgiving

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#15.

We tried to do it here: #2625. However, this approach failed because the $skipCredentials did not work as I thought. This PR makes now a simple fix, that destination which are not parsable are skipped.


packages/connectivity/src/scp-cf/destination/destination-service.spec.ts
```diff
@@ -50,6 +51,11 @@ const oauth2SamlBearerDestination = {
   tokenServicePassword: 'password'
 };
 
+const borkenDestination = {
```
fwilhe, line 54: typo "borken"

fwilhe:
Should this have a change-log entry?

packages/connectivity/src/scp-cf/destination/destination-service.spec.ts
```diff
@@ -95,6 +101,36 @@ describe('destination service', () => {
       });
     });
 
+    it('returns if some destinations are broken', async () => {
```
fwilhe, line 104:
```suggestion
    it('returns if some destinations are broken - instance destinations', async () => {
```

packages/connectivity/src/scp-cf/destination/destination-service.spec.ts
```diff
@@ -95,6 +101,36 @@ describe('destination service', () => {
...
+      })
+        .get('/destination-configuration/v1/instanceDestinations')
+        .reply(200, response);
+
+      const logger = createLogger({
+        package: 'connectivity',
+        messageContext: 'destination-service'
+      });
+      const debugSpy = jest.spyOn(logger, 'debug');
+      const subaccountDestinations: Destination[] =
```
deekshas8, line 124:
```suggestion
      const instanceDestinations: Destination[] =
```

packages/connectivity/src/scp-cf/destination/destination-service.spec.ts
```diff
@@ -175,6 +211,36 @@ describe('destination service', () => {
       });
     });
 
+    it('returns if some destinations are broken', async () => {
```
fwilhe, line 214:
```suggestion
    it('returns if some destinations are broken - subaccount destinations', async () => {
```

fwilhe: I found it hard to see how the tests are different - maybe the different name makes it more obvious

packages/connectivity/src/scp-cf/destination/destination-service.spec.ts
```diff
@@ -95,6 +101,36 @@ describe('destination service', () => {
...
+      const logger = createLogger({
+        package: 'connectivity',
+        messageContext: 'destination-service'
+      });
+      const debugSpy = jest.spyOn(logger, 'debug');
+      const subaccountDestinations: Destination[] =
+        await fetchInstanceDestinations(destinationServiceUri, jwt, {
+          enableCircuitBreaker: false
+        });
+      expect(subaccountDestinations.length).toBe(2);
```
deekshas8, line 128:
```suggestion
      expect(instanceDestinations.length).toBe(2);
```

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -116,9 +116,16 @@ async function fetchDestinations(
 
   return callDestinationEndpoint(targetUri, headers, options)
     .then(response => {
-      const destinations: Destination[] = response.data.map(d =>
-        parseDestination(d)
-      );
+      const destinations: Destination[] = response.data.reduce((all, d) => {
```
fwilhe, line 119: nitpick:

I think this would be more readable in such a form:

```
      const destinations: Destination[] = response.data.filter(d => canParseDestination(d)).map(d =>
        parseDestination(d)
      );
```

Where `canParseDestination` is a new function that performs the required validations but does not throw an error/exception in the bad case but returns a boolean.

Is there a reason why this would be a bad idea?

packages/connectivity/src/scp-cf/destination/destination-service.spec.ts
```diff
@@ -175,6 +211,36 @@ describe('destination service', () => {
       });
     });
 
+    it('returns if some destinations are broken - subaccount destinations', async () => {
```
jjtang1985, line 214:
```suggestion
    it('only returns valid destinations - subaccount destinations', async () => {
```
[req] also for the other test

jjtang1985 (changes requested):
Looks good. I left some suggestions and please consider the changelog.

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -116,9 +117,19 @@ async function fetchDestinations(
 
   return callDestinationEndpoint(targetUri, headers, options)
     .then(response => {
-      const destinations: Destination[] = response.data.map(d =>
-        parseDestination(d)
-      );
+      const destinations: Destination[] = response.data
+        .filter(dest => {
+          try {
```
fwilhe, line 122: Question: Why use try/catch for control flow here? My suggestion was to make the `validateDestinationConfig` have a bool return type and use that return value for the filter.

I'm open to learn about the advantages of using try/catch here, but to me this seems like using it for control flow which feels odd.

FrankEssenberger: I wanted to reuse the existing method, which throws and exception. If I would write the message new, I would use a bool to fit the filter format.

fwilhe: In that case I would prefer adding a new method with that interface and having it internal, but that's not a hard requirement for me to approve.

fwilhe (approved):
lgtm, but maybe another approval would be good before merge

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -116,9 +133,10 @@ async function fetchDestinations(
 
   return callDestinationEndpoint(targetUri, headers, options)
     .then(response => {
-      const destinations: Destination[] = response.data.map(d =>
-        parseDestination(d)
-      );
+      const destinations: Destination[] = response.data
+        .filter(isParsable)
```
fwilhe, line 137: looks much cleaner 👍 

I was not aware of this short-syntax for filter where you don't need to write the parameter 💯

fwilhe (approved):

packages/connectivity/src/scp-cf/destination/destination-service.ts
FrankEssenberger: function handler passing like in C++ :-)

FrankEssenberger: [le](https://xkcd.com/138/)

jjtang1985 (approved):
Please consider adding a changelog, as this is a fix.

PR #2795, opened 2022-08-24
chore: add type tests for OData properties of schema

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#672.

I added some type tests for the schema properties for both v2 and v4. Each version contains five different properties, which should have enough coverage for the type test.


FrankEssenberger (approved):
LGTM - the more tests the merrier

PR #2796, opened 2022-08-24
Remove references to defunct npm registry and deprecate npmrc cli option

fwilhe:


fwilhe:
Should we remove the "generate npmrc option" completely?

FrankEssenberger (approved):
[q] Could we remove the method completely?

florian-richter:
I agree we should remove the option completely

fwilhe:
Can we really just delete the option? It is listed as 'deprecated' in the `--help` output, but imo this would technically be a breaking change..

```
      --generateNpmrc                Deprecated. If set to true the generator
                                     will generate an .npmrc file specifying a
                                     registry for @sap scoped dependencies. This
                                     is not necessary anymore and will be
                                     skipped by default.
```

fwilhe:
> I think it was internal so removing is possible.

"internal" in what sense? The `npmrc` function has an `@internal` annotation, is this what you mean? What's the (semantic) meaning of this? It does not refer to the cli option in any way, does it?

FrankEssenberger:
> "internal" in what sense? The `npmrc` function has an `@internal` annotation, is this what you mean? What's the (semantic) meaning of this? It does not refer to the cli option in any way, does it?

This means that the method is not exported on root level and is not part of our public API. This means we can refactor, rename or delete methods like this as we please. Sometime we add a compatibility note but this is not needed in this case.

FrankEssenberger (approved):

PR #2802, opened 2022-08-25
fix: Remove vcap

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#824.

We still had an issue with a reference to `VCAP_SERVICES` in the code, which would cause problems on k8s.


deekshas8:
Since its a bug fix, please update the changelog

PR #2814, opened 2022-08-29
chore: trigger canary release nightly or on demand

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#836.

I changed the trigger event for the canary release to `schedule` and `workflow_dispatch`.

`workflow_dispatch`: The input `canary-release` variable will switch on/off the canary release.

Please double-check the `schedule` time to make it work with the nightly e2e test.

As required, the canary release (for both `schedule` and `workflow_dispatch`) will only happen if the `canary-release-pre-check` was successful, i.e., there were new commits during the last 24 hrs. Otherwise, after running all checks and tests, the actual `canary-release` workflow will be skipped. See [passing job outputs to another job](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-defining-outputs-for-a-job).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,6 +128,19 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 14
+      - name: Check if latest commit is within 24 hrs
+        run: |
+          lastCommitDate=$(git --no-pager log -n 1 main --pretty=format:"%at")
+          curDate=$(date +%s)
+          dateDiff=$(expr $curDate - $lastCommitDate)
+          echo $lastCommitDate, $curDate, $dateDiff
+
+          if [[ $dateDiff -gt 86400 ]]
+          then
+              echo 'No new commit found on ${{ github.ref }} within the last 24 hrs.'
+              echo 'Canceling the canary release...'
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -111,12 +119,25 @@ jobs:
...
+          lastCommitDate=$(git --no-pager log -n 1 main --pretty=format:"%at")
+          curDate=$(date +%s)
+          dateDiff=$(expr $curDate - $lastCommitDate)
+          echo $lastCommitDate, $curDate, $dateDiff
+
+          if [[ $dateDiff -gt 86400 ]]
+          then
+              echo 'No new commit found on ${{ github.ref }} within the last 24 hrs.'
+              echo 'Canceling the canary release...'
+              curl -X POST -H "Accept: application/vnd.github+json" -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/cancel
```
FrankEssenberger, line 139: [q] isn't there a more elegant way to cancel the current run instead of calling the api from a running job?

FrankEssenberger (approved):
LGTM

Issue #2244, opened 2022-03-04
Make logging less verbose when an error happens in @sap-cloud-sdk/http-client


brodo:
When an HTTP query fails in @sap-cloud-sdk/http-client I get h huge error message with a lot of information I don't care about. 

**Describe the solution you'd like**
Make logging less verbose. E.g. I'd like to see the response code and the response payload and that's it. Most of the space in the log is taken up by a binary buffer which contains the session cache. I'm pretty sure moste people don't want to see it and don't go to the hassle of decoding it.

**Describe alternatives you've considered**
If making it less verbose is not possible, make it configurable.

**Impact / Priority**

Affected development phase: Development, Release, Production

Impact:  Inconvenience
Log JSON here:
```json
  {
    "date": "Fri Mar 04 2022 10:14:29 GMT+0100 (Central European Standard Time)",
    "error": {
      "config": {
        "baseURL": "redacted",
        "data": "redacted",
        "headers": {
          "Accept": "application/json, text/plain, */*",
          "Content-Length": 379,
          "Content-Type": "application/json",
          "User-Agent": "axios/0.26.0",
          "authorization": "Basic redacted"
        },
        "httpAgent": {
          "_events": {},
          "_eventsCount": 2,
          "defaultPort": 80,
          "freeSockets": {},
          "keepAlive": false,
          "keepAliveMsecs": 1000,
          "maxFreeSockets": 256,
          "maxSockets": null,
          "maxTotalSockets": null,
          "options": {
            "path": null
          },
          "protocol": "http:",
          "requests": {},
          "scheduling": "lifo",
          "sockets": {},
          "totalSocketCount": 0
        },
        "httpsAgent": {
          "_events": {},
          "_eventsCount": 2,
          "_sessionCache": {
            "list": [
              "redacted::::::::true:::::::::::::"
            ],
            "map": {
              "redacted::::::::true:::::::::::::": {
                "data": ["gigantic array"],
                "type": "Buffer"
              }
            }
          },
          "defaultPort": 443,
          "freeSockets": {},
          "keepAlive": false,
          "keepAliveMsecs": 1000,
          "maxCachedSessions": 100,
          "maxFreeSockets": 256,
          "maxSockets": null,
          "maxTotalSockets": null,
          "options": {
            "path": null,
            "rejectUnauthorized": true
          },
          "protocol": "https:",
          "requests": {},
          "scheduling": "lifo",
          "sockets": {
[...]

PR #2823, opened 2022-08-31
fix: reduce default logs for failing requests

shojisoma:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/676
Original issue [Make logging less verbose when an error happens in @sap-cloud-sdk/http-client #2244](https://github.com/SAP/cloud-sdk-js/issues/2244)

This reduces error logs that are generated when an HTTP request fails in @sap-cloud-sdk/http-client. The error comes from `xssec`. Currently, our package takes whole error logs that contain too much unimportant info. This extracts a message section from the error and ignores the others.

#### Changes

- throw just a message property in the error, not the whole error
- add a test to check the error doesn't contain unimportant info


packages/connectivity/src/scp-cf/token-accessor.spec.ts
```diff
@@ -267,6 +267,19 @@ describe('token accessor', () => {
...
     });
 
+    it('throws an error having the cause, not the config property', async () => {
+      mockClientCredentialsGrantCall(
+        providerXsuaaUrl,
+        { access_token: signedJwt({ dummy: 'content' }) },
+        401,
+        destinationBindingClientSecretMock.credentials
+      );
+      await serviceToken('destination').catch(error => {
```
ZhongpinWang, line 277: `.catch(expect(...))` will give a false positive if there is no error thrown at all since `.catch` is never executed. For example, if `200` is returned instead of `401`, the test will still pass.

Try to use `rejects` instead:
```suggestion
      const promise = serviceToken('destination');
      await expect(promise).rejects.toHaveProperty('cause');
      await expect(promise).rejects.not.toHaveProperty('config');
```

packages/connectivity/src/scp-cf/token-accessor.spec.ts
```diff
@@ -267,6 +267,19 @@ describe('token accessor', () => {
       expect(retrieveFromCacheSpy).toHaveBeenCalledTimes(0);
     });
 
+    it('throws an error having the cause, not the config property', async () => {
```
ZhongpinWang, line 270:
```suggestion
    it('throws an error with only the cause property', async () => {
```

ZhongpinWang (commented):
A minor issue for the test case.

packages/connectivity/src/scp-cf/token-accessor.spec.ts
shojisoma: thanks. It must be tested with `.rejects`. I removed the try-catch block.

FrankEssenberger (approved):
LGTM

packages/connectivity/src/scp-cf/xsuaa-service.ts
```diff
@@ -116,7 +116,7 @@ export async function getClientCredentialsToken(
         null,
         subdomainAndZoneId.zoneId,
         (err: Error, token: string, tokenResponse: ClientCredentialsResponse) =>
-          err ? reject(err) : resolve(tokenResponse)
+          err ? reject(err.message) : resolve(tokenResponse)
```
FrankEssenberger, line 119: [req] Please add a string in front of the message like `Error in fetching the token for service ${serviceName}: ${error.message}`

shojisoma: I chenged the meessage. I added a type predicated and checker that is needed for generating the message as well.

FrankEssenberger (commented):
Thanks for adding the more detailed error message. Please go ahead and merge if the tests are green.

PR #2824, opened 2022-08-31
feat: support sending emails in parallel and in sequential

jjtang1985:
As discussed, we would like to let the users decide, whether parallel or sequential is better for their use cases.
By default, "parallel" is chosen.


packages/mail-client/src/mail-client.ts
```diff
@@ -232,21 +285,25 @@ function teardown(transport: Transporter<SentMessageInfo>, socket?: Socket) {
...
 ): Promise<MailResponse[]>;
 export async function sendMail<T extends MailOptions>(
   destination: DestinationOrFetchOptions,
+  mailClientOptions: MailClientOptions,
   mailOptions: T[]
 ): Promise<MailResponse[]>;
 /* eslint-disable jsdoc/require-jsdoc */
 export async function sendMail<T extends MailOptions>(
   destination: DestinationOrFetchOptions,
+  mailClientOptions: MailClientOptions,
```
FrankEssenberger, line 306: [pp] I would prefer the order sendMail(destination:Destiantion,mails:MailOptions[],mailClientOptions?:Options)). Doing so you can skip the options for the client which are options. You have then to pass an array for the mails but this would be fine I think.

FrankEssenberger: If more and more people are using typescript the people see directly in the compile step if it is an array or a single object.

FrankEssenberger (approved):
LGTM only on small personal preference in the API of send mail

PR #2828, opened 2022-09-01
draft: API idea for content-id in batch changeset

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#813.

I did a small draft how the content-id for the header and url could be set. Here an example:
The first request creates a entity with content-id `myContentId`
Then this created entity is used in the funciton which is the next changeset.

I created to setter for the header and url part see in the changed files. The names could be changed easily. Just let me know if you like the per request builder option or if you would like to do it differently.

```
--batch_test-boundary
Content-Type: multipart/mixed; boundary=changeset_test-boundary

--changeset_test-boundary
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-Id: myContentId

POST /sap/opu/odata/sap/API_TEST_SRV/A_TestEntity HTTP/1.1
Content-Type: application/json //request Custom header go here
Accept: application/json

{}

--changeset_test-boundary
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-Id: test-boundary

POST $myContentId/TestFunctionImportPOST?SimpleParam='someValue' HTTP/1.1
Content-Type: application/json //request Custom header go here
Accept: application/json



--changeset_test-boundary--
--batch_test-boundary--
--batch_test-boundary
```


packages/odata-common/src/request/odata-request-config.ts
```diff
@@ -38,6 +38,8 @@ export abstract class ODataRequestConfig {
   private _appendedPaths: string[] = [];
   private _fetchCsrfToken = true;
   private _timeout: number | undefined = undefined;
+  private _contentIdChangesetHeader: string | undefined;
+  private _contentIdChangesetUrl: string | undefined;
```
jjtang1985, line 42: [pp] I feel `Changeset` is not necessary, as you can use the header/url for both batch operations (retrieve/changeset). Maybe use `_contentIdHeader` and `_contentIdChangesetUrl` ?

[fyi]
1. As we want to support content id in the request, we might also want to think about the [response](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_MultipartBatchResponse).

2. In addition, the `contentId` can also be used with etag.
[reference](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ReferencinganETag)

[pp] Given the batch content id 3 use cases, I would do the following:
```
interface BatchContentId{
  header?: string | undefined;
  url?: string | undefined;
  etag?: string | undefined;
}
OdataRequestConfig{
  private _batchContentId: BatchContentId | undefined;
}
```

packages/odata-v2/src/request-builder/batch-request-builder.spec.ts
```diff
@@ -92,6 +92,66 @@ HTTP/1.1 200 OK
...
+      'Accept: application/json',
+      '',
+      '{}',
+      '',
+      '--changeset_test-boundary',
+      'Content-Type: application/http',
+      'Content-Transfer-Encoding: binary',
+      'Content-Id: test-boundary',
+      '',
+      "POST $myContentId/TestFunctionImportPOST?SimpleParam='someValue' HTTP/1.1",
```
jjtang1985, line 116: As discussed, this example is an unsupported odata feature, bound function.
Maybe add more examples by e.g., searching "content-id" [here](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html)

packages/odata-common/src/request/odata-request-config.ts
FrankEssenberger: oh, I missed the `eTag` thing. You are right with three facets a object makes total sense. I adjust this draft.

packages/odata-v2/src/request-builder/batch-request-builder.spec.ts
FrankEssenberger: sure I adjust the example. I ping you once the draft is updated.

packages/odata-common/src/request-builder/batch/batch-request-serializer.ts
```diff
@@ -65,10 +65,19 @@ export function serializeRequest(
...
+            request.requestConfig.contentIdBatch?.header || uuid()
+          }`
+        ]
+      : []), // make function with logging nicer - is this needed here?.
     '',
     `${method} ${getUrl(odataRequest, options.subRequestPathType)} HTTP/1.1`,
     ...(requestHeaders.length ? requestHeaders : ['']),
+    ...(request.requestConfig.contentIdBatch?.etag
+      ? [`If-Match: ${request.requestConfig.contentIdBatch?.etag}`]
+      : []),
```
jjtang1985, line 80: [pp] This function needs to be polished in the real implementation for better readability.

jjtang1985 (approved):
LGTM as API ideas, 3 use cases of the content id are covered.

packages/odata-common/src/request-builder/batch/batch-request-serializer.ts
FrankEssenberger: yes, I leave a comment. Also there should be a log statement if the `eTag` was also given in the header already.

marikaner:
Closing as we have discussed a new appraoch.

PR #2833, opened 2022-09-05
chore: Get All Destinations API ADR

tomfrenken:
This ticket concludes our ADR discussion on the Get All Destinations API ADR.

Part of SAP/cloud-sdk-backlog#738.


deekshas8 (commented):
LGTM.

knowledge-base/adr/0034-get-destinations.md
```diff
@@ -0,0 +1,106 @@
...
+export async function getAllDestinationsFromDestinationService(
+  options: DestinationOptions
+): Promise<Destination[] | null>
+```
+
+This function should throw a warning if one of the destinations is incomplete/compromised.
+
+If our users have feature requests for the other look-up functions, e.g. `searchEnvVariablesForDestination` etc., we shall implement them on-demand.
+
+If we end up implementing every look-up function to get all destinations, we are to implement the parent function and deprecate and remove the predecessors functions.
```
FrankEssenberger, line 43: [req] Please add also the options for fetching and the Return type to the ADR.

.changeset/unlucky-cooks-work.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Support fetching all destinations from the destination service simultaneously.
```
FrankEssenberger, line 5: [req] Be a bit more precise like: All destinations of a subacction and service instance simultaneously.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -296,3 +331,63 @@ describe('jwtType x selection strategy combinations. Possible values are {subscr
...
+      messageContext: 'destination-accessor'
+    });
+    const warnSpy = jest.spyOn(logger, 'warn');
+
+    const allDestinations = await getAllDestinationsFromDestinationService();
+    expect(allDestinations).toEqual(null);
+    expect(warnSpy).toHaveBeenCalledWith(
+      'Could not retrieve destinations from destination service.'
+    );
+  });
```
FrankEssenberger, line 392: [req] Please add a test where the response of the destination service is service empty. See the specs (empty array) 
https://api.sap.com/api/SAP_CP_CF_Connectivity_Destination/resource. 

Also an error case like 404 or 500 would be nice to mock to ensure we give a resonable exception.

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -85,3 +103,89 @@ export async function getDestination(
...
+      token.decoded.iss!
+    )}" from destination service.`
+  );
+
+  const [instance, subaccount] = await Promise.all([
+    fetchInstanceDestinations(destinationServiceUri, token.encoded, options),
+    fetchSubaccountDestinations(destinationServiceUri, token.encoded, options)
+  ]);
+
+  let loggerMessage = '';
```
FrankEssenberger, line 160: [pp] create helper methods for the logging

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -85,3 +103,89 @@ export async function getDestination(
...
+  const allDestinations = instance.concat(subaccount);
+
+  if (allDestinations && allDestinations.length !== 0) {
+    logger.debug(
+      `Successfully retrieved all destinations for account: "${parseSubdomain(
+        token.decoded.iss!
+      )}" from destination service.`
+    );
+  } else {
+    logger.warn('Could not retrieve destinations from destination service.');
```
FrankEssenberger, line 180: [req] If there are no destination this could be - I would not warn

FrankEssenberger: [pp] I would prefer an empty array in a non error case an not `null`

FrankEssenberger (changes requested):
Nothing big a few things requested. Overall looks good.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
tomfrenken: I think with the `mockThingsForCombinations(true, true);` I already test the empty case, however, I guess I can add some 404/500 case.

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -85,3 +102,79 @@ export async function getDestination(
...
   return destination;
 }
+
+/**
+ * Creates comprehensive log messages from a destinations array and their origin.
+ * @param origin - Origin of the destination.
+ * @param destinations - Array of destinations.
+ * @returns Logs of the retrival of destinations.
+ */
+function destinationFetchLogHelper(
```
FrankEssenberger, line 112: [pp] the `helper`feels a bit strange and our style guide also says: user verbs (https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md).

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -296,3 +350,83 @@ describe('jwtType x selection strategy combinations. Possible values are {subscr
...
+
+    expect(debugSpy).toHaveBeenCalledWith(
+      'Retrieving all destinations for account: "subscriber" from destination service.'
+    );
+    expect(debugSpy).toHaveBeenCalledWith(
+      'Retrieving subaccount destination: DESTINATION.\n'
+    );
+  });
+
+  it('should fetch all provider destinations when called without anything', async () => {
```
FrankEssenberger, line 390: [pp] without anything is perhaps a misleading doupble negation `without passing a JWT` perhaps?

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -296,3 +350,83 @@ describe('jwtType x selection strategy combinations. Possible values are {subscr
...
+    const logger = createLogger({
+      package: 'connectivity',
+      messageContext: 'destination-accessor'
+    });
+    const debugSpy = jest.spyOn(logger, 'debug');
+
+    const allDestinations = await getAllDestinationsFromDestinationService();
+    expect(allDestinations).toEqual([]);
+    expect(debugSpy).toHaveBeenCalledWith(
+      'Could not retrieve destinations from destination service.'
```
FrankEssenberger, line 409: [pp] This log statement is perhaps missleading if thre are simply no destinaitons.

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -264,6 +271,20 @@ class DestinationFromServiceRetriever {
     return { encoded, decoded: decodeJwt(encoded) };
   }
 
+  private static checkDestinationForCustomJwt(destination: Destination): void {
```
FrankEssenberger, line 274: [q] This is marked as a change here, but I guess nothing changed for the two private static methods?

FrankEssenberger (approved):
LGTM- only a few minor questions on naming of the tests and content of log statements

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
tomfrenken: Yes, they were just moved because private has to come before public

PR #2841, opened 2022-09-05
feat: set global transport

shojisoma:
Closes [#367](https://github.com/SAP/cloud-sdk-backlog/issues/367)

This added a new method that enables users to set transport globally. When the user sets the global transport, all transports in all loggers including default loggers are replaced with the global transport.

example
```
const fileTransport = new winston.transports.File({
  filename: "info.log",
  level: "info",
});
const httpTransport = new winston.transports.Http({
  host: "localhost",
  port: 8080,
});
setGlobalTransports([fileTransport, streamTransport]);
```

#### Changes

- made a new method `setGlobalTransports()` that sets transports globally
- added tests for the new function
- added dependency for the new function

### DIFF
--- a/.changeset/slow-emus-drum.md
+++ b/.changeset/slow-emus-drum.md
@@ -2,4 +2,4 @@
 '@sap-cloud-sdk/util': minor
 ---
 
-[New Functionality] Add method `setGlobalTransports` to support implementation of setting custom transport globally.
+[New Functionality] Add method `setGlobalTransports` to support setting custom transport globally.
--- a/.changeset/slow-emus-drum.md
+++ b/.changeset/slow-emus-drum.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/util': minor
+---
+
+[New Functionality] Add method `setGlobalTransports` to support implementation of setting custom transport globally.
--- a/packages/util/src/logger/cloud-sdk-logger.spec.ts
+++ b/packages/util/src/logger/cloud-sdk-logger.spec.ts
@@ -253,7 +253,7 @@ describe('Cloud SDK Logger', () => {
   describe('set global transport', () => {
     beforeEach(() => {
       logger = createLogger(messageContext);
-    })
+    });
     it('should replace all transpots in all active loggers with the global transport', async () => {
       const consoleSpy = jest.spyOn(process.stdout, 'write');
       const rootNodeModules = path.resolve(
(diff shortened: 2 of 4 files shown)


packages/util/src/logger/cloud-sdk-logger.ts
```diff
@@ -203,6 +204,28 @@ export function getGlobalLogLevel(): string | undefined {
   return container.options.level;
 }
 
+/**
+ * Change the global transport of the container which will set default transport for all active loggers.
+ * e.g., to set the global transport call `setGlobalTransports(httpTransport)`.
+ * @param customTransports - The transport to set the global transport to. Both array and varargs are supported.
+ */
+export function setGlobalTransports(
```
FrankEssenberger, line 212: [pp] I know we did this method overloading in many places to allow for options:
 f(a)
f([a,b])
 and f(a,b). 

However, I find this overload hard and difficult to read. So for me a `TransportStream | TransportStream[]` to allow for a single or array object should be fine.

.changeset/slow-emus-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/util': minor
+---
+
+[New Functionality] Add method `setGlobalTransports` to support implementation of setting custom transport globally.
```
FrankEssenberger, line 5:
```suggestion
[New Functionality] Add method `setGlobalTransports` to support setting custom transport globally.
```

FrankEssenberger (approved):
LGTM - one one minor [pp] regarding the method API. Also note that you should create a documentation PR where you mention this new feature under the logging section.

packages/util/src/logger/cloud-sdk-logger.ts
shojisoma: I changed the function to just accept a single transport object or an array with multiple transports for readability.

packages/util/src/logger/cloud-sdk-logger.spec.ts
```diff
@@ -245,6 +250,67 @@ describe('Cloud SDK Logger', () => {
     });
   });
 
+  describe('set global transport', () => {
+    beforeEach(() => {
+      logger = createLogger(messageContext);
+    });
+    it('should replace all transpots in all active loggers with the global transport', async () => {
```
deekshas8, line 257:
```suggestion
    it('should replace all transports in all active loggers with the global transport', async () => {
```

packages/util/src/logger/cloud-sdk-logger.ts
```diff
@@ -203,6 +203,29 @@ export function getGlobalLogLevel(): string | undefined {
   return container.options.level;
 }
 
+/**
+ * Change the global transport of the container which will set default transport for all active loggers.
+ * e.g., to set the global transport call `setGlobalTransports(httpTransport)`.
+ * @param customTransports - The transport to set the global transport to. Both single transport and an array with multiple transpots are supported.
```
deekshas8, line 209:
```suggestion
 * @param customTransports - The transport to set the global transport to. Both single transport and an array with multiple transports are supported.
```

packages/util/src/logger/cloud-sdk-logger.ts
```diff
@@ -203,6 +203,29 @@ export function getGlobalLogLevel(): string | undefined {
...
+ */
+export function setGlobalTransports(
+  customTransports: TransportStream | TransportStream[]
+): void {
+  container.options.transports = customTransports;
+  const isArray = (
```
deekshas8, line 215: [req] I don't think this method is needed. In the if statement you can already check with the `Array.isArray(customTransports)`.
[pp] I would also just use `?:` instead of the if else in that case.

deekshas8 (changes requested):

packages/util/src/logger/cloud-sdk-logger.ts
shojisoma: thanks. I removed the type predicates and just use the condition with a ternary operator.

deekshas8 (approved):
LGTM.

PR #2845, opened 2022-09-06
fix: next sdk version

jjtang1985:
The release build [failed](https://github.com/SAP/cloud-sdk-js/actions/runs/2994046205), as the next (minor) version was `2.8.1`.

This can be reproduced by the steps below:
1. make a `patch` release
2. make a `minor` release

This PR fixes the computation of the next version and adds some tests.


scripts/util.ts
```diff
@@ -19,13 +20,20 @@ export function transformFile(
...
+  const currentVersion = getPackageVersion();
   const releasePlan = await getReleasePlan(process.cwd());
-  const versionIncrease = releasePlan.releases
+
+  const versionIncreases = releasePlan.releases
     .map(({ type }) => versionOrder.indexOf(type))
     .sort((a, b) => b - a);
-  currentVersion[Math.min(...versionIncrease)] += 1;
-  return currentVersion.join('.');
+  const release = versionOrder[Math.min(...versionIncreases)];
```
tomfrenken, line 29: Kind of unintuitive that the "bigger" release (e.g. major > minor > patch) has the lowest index IMO, but I guess that is [pp].

tomfrenken (approved):
LGTM

PR #2846, opened 2022-09-06
Remove executable bits from files

fwilhe:
We had > 1000 executable files in our repo, most if not all of them should not be executable. This PR removes the executable bits.


florian-richter (approved):
👍

PR #2852, opened 2022-09-08
chore: Move `mock.restore` to `afterEach`

florian-richter:
I struggled with flakey unit tests locally. After investigation, it seems that we should always call `mock.restore` in `afterEach` as calling it in the test function or in `afterAll` may obfuscate or introduce errors. See [here](https://stackoverflow.com/questions/68686034/jest-tests-failing-with-enoent-no-such-file-or-directory-referring-to-the-spec). This change made the tests WAY more consistent for me.


marikaner (approved):
👍

FrankEssenberger (approved):

PR #2857, opened 2022-09-08
chore: clean up sdk metadata code

deekshas8:
Closes SAP/cloud-sdk-backlog#620.


packages/generator-common/package.json
```diff
@@ -37,15 +37,13 @@
     "readme": "ts-node ../../scripts/replace-common-readme.ts"
   },
   "dependencies": {
-    "@sap-cloud-sdk/util": "^2.8.0",
```
jjtang1985, line 40: [req] Perhaps this is not intended after the release?

test-packages/e2e-tests/package.json
```diff
@@ -57,7 +55,9 @@
     "requires": {
       "db": {
         "kind": "sqlite",
-        "model": "*"
```
jjtang1985, line 60: [q] could you please share the reason?

jjtang1985 (changes requested):
LGTM.
Please check the old SDK version used.

test-packages/e2e-tests/package.json
deekshas8: This change was unintended, will revert it.

packages/generator-common/package.json
deekshas8: You're right, I reverted it.

jjtang1985 (approved):

PR #2861, opened 2022-09-09
Mail sec2

jjtang1985:

### DIFF
--- a/packages/mail-client/src/mail-client-types.ts
+++ b/packages/mail-client/src/mail-client-types.ts
@@ -5,6 +5,8 @@ import {
   DestinationProxyType,
   ProxyConfiguration
 } from '@sap-cloud-sdk/connectivity';
+import { ConnectionOptions } from "tls";
+
 /**
  * Represents an e-mail address.
  * This interface is compatible with `Mail.Address` of `nodemailer`.
@@ -266,4 +268,16 @@ export interface MailClientOptions {
    * Option to define the strategy of sending emails. The emails will be sent in parallel when setting to true, otherwise in sequential. The default value is true.
    */
   parallel?: boolean;
+}
+
+/**
+ * Represents options for setting up the SMTP connection.
+ * This interface is compatible with `SMTPConnection.Options` of `nodemailer`.
+ * @experimental This API is experimental and might change in newer versions. Use with caution.
+ */
+export interface SmtpConnectionOptions {
+  /**
+   * Defines additional options to be passed to the socket constructor, e.g. { rejectUnauthorized: true }
+   */
+  tls?: ConnectionOptions | undefined
 }
--- a/packages/mail-client/src/mail-client-types.ts
+++ b/packages/mail-client/src/mail-client-types.ts
@@ -6,6 +6,9 @@ import {
   ProxyConfiguration
 } from '@sap-cloud-sdk/connectivity';
 import { ConnectionOptions } from "tls";
+import * as shared from "nodemailer/lib/shared";
+import net from "net";
+import { CustomAuthenticationHandlers } from "nodemailer/lib/smtp-connection";
 
 /**
  * Represents an e-mail address.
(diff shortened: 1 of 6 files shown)


test-packages/type-tests/test/mail-client/mail-client.spec.ts
```diff
@@ -6,9 +6,16 @@ const mailConfig = { from: 'from', to: 'to' };
...
+  }
 });
 // $ExpectType Promise<MailResponse[]>
-sendMail({ destinationName: 'dest' }, [mailConfig, mailConfig], {
-  parallel: false
-});
+sendMail(
+  { destinationName: 'dest' },
+  [mailConfig, mailConfig],
+  {},
```
jjtang1985, line 17: This might be a concern, as we now have a new optional option.

We can also merge the following into one optional parameter:
- MailClientOptions
- SmtpTransportOptions

However, as the SmtpTransportOptions has to sync with the lib (out of our control), we might have key conflicts theoretically. Therefore, I would then rename the `parallel` key and make it unique like { sdk: { parallel: true }}.

packages/mail-client/src/mail-client.ts
```diff
@@ -283,13 +271,15 @@ function teardown(transport: Transporter<SentMessageInfo>, socket?: Socket) {
...
  * @param mailConfigs - A single object or an array of {@link MailConfig}.
+ * @param smtpTransportOptions - A {@link SmtpTransportOptions} that defines how to set up an SMTP transport, including SSL and tls configurations.
  * @param mailClientOptions - A {@link MailClientOptions} that defines the configurations of the mail client, e.g., whether the mails are sent in parallel.
  * @returns A promise resolving to an array of {@link MailResponse}.
  * @see https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destination#referencing-destinations-by-name
  */
 export async function sendMail<T extends MailConfig>(
   destination: DestinationOrFetchOptions,
   mailConfigs: T | T[],
+  smtpTransportOptions?: SmtpTransportOptions,
```
FrankEssenberger, line 282: [pp] I am not a big fan of two optional arguments. If you just want to set the last one you need to add a dummy {}. What about using a union type or a larger object with two key?

test-packages/type-tests/test/mail-client/mail-client.spec.ts
FrankEssenberger: ah, I was leaving a comment before I read this. I think having a `sapCloudSdkOptions` or so is a good idea to avoud conflicts.

FrankEssenberger (approved):
LGTM - I would merge the two optional parameters and add one key holding the SDK options to avoid conflicts with the naitive node modules.

packages/mail-client/src/mail-client-types.ts
```diff
@@ -259,11 +262,75 @@ export interface MailDestination {
...
   /**
    * Option to define the strategy of sending emails. The emails will be sent in parallel when setting to true, otherwise in sequential. The default value is true.
    */
   parallel?: boolean;
 }
+
+/**
+ * Represents options of the mail client.
+ */
+export interface MailClientOptions extends SmtpTransportOptions {
```
FrankEssenberger, line 277: [pp] I prefer using types in such cases but interfaces work as well.

packages/mail-client/src/mail-client-types.ts
```diff
@@ -259,11 +262,75 @@ export interface MailDestination {
 }
 
 /**
- * Represents options of the mail client.
+ * Represents options for SDK behaviours, for example whether the mails are sent in parallel.
```
FrankEssenberger, line 265:
```suggestion
 * Represents options for sending mails provided by the SDK. For example whether the mails are sent in parallel.
```

FrankEssenberger (approved):
LGTM only minor suggestions

PR #2875, opened 2022-09-15
Support odata bound functions and actions

fwilhe:
This PR adds support for odata bound functions and actions for odata v4 only


packages/generator/src/entity/class.ts
```diff
@@ -102,6 +120,40 @@ function property(prop: VdmProperty): PropertyDeclarationStructure {
   };
 }
 
+function boundFunction(fn: any): MethodDeclarationStructure {
```
fwilhe, line 123: q: what is the correct type here?

packages/generator/src/edmx-parser/v4/edm-types.ts
```diff
@@ -33,7 +33,14 @@ export interface EdmxNavigationPropertyBinding {
  */
 export type EdmxEntityTypeV4 = EdmxEntityTypeBase<EdmxNavigationPropertyV4> &
   EdmxDerivedType &
-  EdmxNamespaced;
+  EdmxNamespaced &
+  EdmxEntityWithBound;
+
+interface EdmxEntityWithBound {
```
fwilhe, line 39: q: is there a better way to express this? is it ok to do it like that?

packages/generator/src/edmx-parser/common/edmx-parser.ts
```diff
@@ -27,9 +27,36 @@ export function parseEntityTypesBase(root: any): EdmxEntityTypeBase<any>[] {
...
     Property: forceArray(e.Property),
-    Namespace: e.Namespace
+    Namespace: e.Namespace,
+    BoundFunction: extractBoundFunctions(root),
+    BoundAction: extractBoundActions(root)
   }));
 }
+
+function extractBoundFunctions(root: any) {
+  const x = root[0];
```
fwilhe, line 37: q: how many items does this list have? What does that mean semantically? is there a proper way to access the "correct" object?

test-packages/e2e-tests/test/mail/mail.spec.ts
```diff
@@ -32,22 +32,22 @@ describe('Mail', () => {
...
-  it('should send 100 mails', async () => {
-    const mailOptions = buildArrayWithNatualNums(100).map(
-      mailIndex =>
-        ({
-          ...defaultMailOptions,
-          subject: `mail ${mailIndex}`
-        } as MailConfig)
-    );
-    const responses = await sendTestMail(mailOptions);
+  // it('should send 100 mails', async () => {
```
fwilhe, line 35: fixme: re-enable. commented out because the test does not run reliable locally, this has nothing to do with the change in this PR.

packages/generator/src/edmx-to-vdm/common/entity.ts
```diff
@@ -125,6 +130,46 @@ function properties(
...
+    // Remove first parameter which per spec always is the entity the function is bound to
+    // cf https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/os/odata-csdl-xml-v4.01-os.html#sec_Parameter
+    parameters: forceArray(f.Parameter)
+      .slice(1)
+      .map(p => ({
+        parameterName: p.Name,
+        jsType: edmToTsType(p.Type),
+        edmType: p.Type
+      })),
+      returnType: f.ReturnType.Type // edmToTsType(f.ReturnType.Type)
```
fwilhe, line 151: take from unbound functions

marikaner (changes requested):
I left some first comments. I am far away from looking through everything.

test-packages/test-services-odata-v2/test-service/TestEntityLvl2SingleLink.ts
```diff
@@ -9,6 +9,7 @@ import {
   DeSerializers,
   DeserializedType
 } from '@sap-cloud-sdk/odata-v2';
+import { camelCase } from '@sap-cloud-sdk/util';
```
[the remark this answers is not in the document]
fwilhe: Yes, this is not on purpose. The change is in `packages/generator/src/entity/file.ts`, but I think after our discussions on how multi-key entities should be handled this will go anyway from the generated code.

packages/odata-v4/src/request/odata-bound-function-request-config.ts
```diff
@@ -0,0 +1,66 @@
...
+  FunctionImportParameter
+} from '@sap-cloud-sdk/odata-common/internal';
+import { DeSerializers } from '../de-serializers';
+
+/**
+ * Function import request configuration for an entity type.
+ * @typeParam DeSerializersT - Type of the deserializer use on the request
+ * @typeParam ParametersT - Type of the parameter to setup a request with
+ */
+export class OdataBoundFunctionRequestConfig<
```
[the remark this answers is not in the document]
fwilhe: this seems to be fixed already

packages/generator/src/entity/class.ts
```diff
@@ -114,6 +142,122 @@ function property(prop: VdmProperty): PropertyDeclarationStructure {
...
+}
+
+function boundFunctionsParameterStatements(fn: VdmFunctionImport): string[] {
+  return ['const params = {'].concat(
+    fn.parameters?.map(
+      p =>
+        `${p.parameterName}: new FunctionImportParameter('${p.parameterName}', '${p.edmType}', ${p.parameterName}),`
+    ),
+    ['};']
+  );
```
marikaner, line 200: [req] I find this very odd to read. I think it would be better if this was either one string instead of an array of strings or if this function would just return the list of parameter statements.
Also, please use `array.join(,)` to concatenate the elements of the array with a comma.

packages/generator/src/entity/class.ts
```diff
@@ -98,6 +106,26 @@ function properties(entity: VdmEntity): PropertyDeclarationStructure[] {
...
 }
 
+function boundFunctions(
+  entity: VdmEntity,
+  service: VdmServiceMetadata
+): MethodDeclarationStructure[] {
+  if (entity.boundFunctions === undefined) {
+    return [];
+  }
+  return entity.boundFunctions.map(f => boundFunction(f, entity, service));
```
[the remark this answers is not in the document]
fwilhe: This is now not relevant anymore, but honestly I am not sure if I agree the latter is better. As this is a question of style there is probably no correct way and consistency is key, so I'm okay with doing it the short way, but I suspect the explicit handling of the undefined case might make it easier to read and understand.

packages/generator/src/edmx-to-vdm/common/entity.ts
```diff
@@ -125,6 +136,124 @@ function properties(
     };
   });
 }
+
+function emptyIfUndefined(input: string | undefined): string {
+  return input ? input : '';
```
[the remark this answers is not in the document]
fwilhe: code is deleted, resolving the comment

packages/generator/src/edmx-to-vdm/common/entity.ts
```diff
@@ -125,6 +136,124 @@ function properties(
...
+  if (input) {
+    return input.toLowerCase() === 'true';
+  }
+  return false;
+}
+
+/**
+ * @internal
+ */
+export function transformBoundFunctions(
```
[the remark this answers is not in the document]
fwilhe: updated

packages/generator/src/edmx-to-vdm/v4/entity.ts
```diff
@@ -89,6 +108,44 @@ function navigationProperties(
...
+  const entities: VdmEntityInConstruction[] = Object.keys(classNames).map(c => ({
+    className: c,
+    entityTypeName: c,
+    entityTypeNamespace: entityType.Namespace
+  }));
+
+  const enumTypes: VdmEnumType[] = generateEnumTypesV4(serviceMetadata, formatter);
+  const complexTypes: VdmComplexType[] = generateComplexTypesV4(serviceMetadata, enumTypes, formatter);
+
+  return generateActionImportsV4(serviceMetadata, entityType.Namespace, entities as VdmEntity[], complexTypes, formatter, true);
```
fwilhe, line 146: fixme: remove `as`

packages/generator/src/entity/class.ts
```diff
@@ -114,6 +135,128 @@ function property(prop: VdmProperty): PropertyDeclarationStructure {
...
 
+function boundFunction(
+  fn: VdmFunctionImport,
+  entity: VdmEntity,
+  service: VdmServiceMetadata
+): MethodDeclarationStructure {
+  return {
+    kind: StructureKind.Method,
+    name: fn.name,
+    returnType: `BoundFunctionRequestBuilder<${entity.className}<DeSerializersT>, DeSerializersT, any, ${fn.returnType.returnType} | null>`,
```
marikaner, line 146: [req] We should get rid of the `as any` parts.

marikaner (dismissed):

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -40,14 +41,14 @@ export function parseFunctionImportReturnTypes(
...
-  entities: VdmEntity[],
+  entities: VdmEntityInConstruction[],
   complexTypes: VdmComplexType[],
   extractResponse: ExtractResponse,
   serviceName: string
 ): VdmActionImportReturnType {
   return parseReturnTypes(
     returnType,
-    entities,
+    entities as VdmEntity[],
```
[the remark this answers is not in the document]
fwilhe: looks like this was leftover by accident as the function being called here does thake the in construction type

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
```diff
@@ -152,7 +153,7 @@ function getEdmReturnType(
 function getEntityReturnType(
   isCollection: boolean,
   isNullable: boolean,
-  entities: VdmEntity[],
+  entities: VdmEntityInConstruction[],
```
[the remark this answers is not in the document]
fwilhe: Not sure what a good name would be. It is a subset of an entity that is enough for here, but `VdmEntityLite` or `VdmEntityMini` are silly names.. Can someone describe what kind of an entity that is?

packages/generator/src/edmx-to-vdm/v4/action-import.ts
```diff
@@ -68,26 +68,47 @@ ${actionImportsWithoutActions
...
+      .filter(
+        ({ action: edmxAction }) =>
+          !hasUnsupportedParameterTypes(edmxAction, bindingEntitySetName)
+      );
+  }
+      // TODO 1571 remove when supporting entity type as parameter
+  return joinedFunctionData.filter(
+    ({ action: edmxAction }) =>
+      !hasUnsupportedParameterTypes(edmxAction, bindingEntitySetName)
+  );
```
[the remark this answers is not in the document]
fwilhe: all comments valid, thanks :)

I'd prefer to keep the if for simplicity.

packages/generator/src/edmx-to-vdm/common/entity.ts
```diff
@@ -62,7 +71,9 @@ export function transformEntityBase(
       : true,
     deletable: entityMetadata.entitySet
       ? isDeletable(entityMetadata.entitySet)
-      : true
+      : true,
+    boundFunctions: transformBoundFunctions(boundFunctions),
+    boundActions: transformBoundActions(boundActions)
```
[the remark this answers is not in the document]
FrankEssenberger: operation is new name

packages/generator/src/edmx-to-vdm/common/action-function-parameters.ts
```diff
@@ -34,30 +36,35 @@ export function getFunctionImportParameters(
...
     edmxFunctionImport,
     edmxParameters,
     swaggerDefinition,
-    formatter
+    formatter,
+    bindingEntitySetName
   );
 }
 
 function getParameter<T extends EdmxNamed>(
```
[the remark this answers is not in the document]
FrankEssenberger: done in marikas operations PR

packages/generator/src/entity/class.ts
```diff
@@ -98,6 +105,20 @@ function properties(entity: VdmEntity): PropertyDeclarationStructure[] {
   return entity.properties.map(prop => property(prop));
 }
 
+function boundFunctions(
+  entity: VdmEntity,
+  service: VdmServiceMetadata
+): MethodDeclarationStructure[] {
+  return entity.functions.map(f => boundFunction(f, entity, service));
```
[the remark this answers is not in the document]
FrankEssenberger: `operationFunctionBase` is used here as well.

packages/generator/src/edmx-to-vdm/v4/entity.ts
```diff
@@ -89,6 +111,46 @@ function navigationProperties(
...
 
+function transformBoundFunctions(
+  serviceMetadata: ServiceMetadata,
+  boundEntityType: EdmxEntityTypeV4,
+  boundEntitySet: EdmxEntitySet,
+  classNames: { [originalName: string]: string },
+  formatter: ServiceNameFormatter,
+  entitiesMetadata: JoinedEntityMetadata<EdmxEntitySet, EdmxEntityTypeV4>[]
+): VdmFunctionImport[] {
+  const entities: VdmEntityInConstruction[] = entitiesMetadata.map(({ entityType, entitySet }) => ({
```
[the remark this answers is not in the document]
FrankEssenberger: done

packages/generator/src/edmx-to-vdm/v4/function-import.ts
```diff
@@ -68,27 +68,46 @@ function joinFunctionImportData(
   return joinedFunctionImportData;
 }
 
+/**
+ * @internal
+ */
+export function filterFunctions(joinedFunctionData: JoinedFunctionImportData[], bindingEntitySetName?: string): JoinedFunctionImportData[] {
```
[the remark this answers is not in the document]
FrankEssenberger: operation was introduced

packages/generator/src/edmx-to-vdm/v4/entity.ts
```diff
@@ -89,6 +111,46 @@ function navigationProperties(
...
+    entityTypeName: entityType.Name,
+    className: classNames[entitySet.Name],
+    entityTypeNamespace: entityType.Namespace
+  }));
+
+  const enumTypes: VdmEnumType[] = generateEnumTypesV4(serviceMetadata, formatter);
+  const complexTypes: VdmComplexType[] = generateComplexTypesV4(serviceMetadata, enumTypes, formatter);
```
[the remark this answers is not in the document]
FrankEssenberger: removed

packages/generator/src/edmx-to-vdm/common/action-import.ts
```diff
@@ -14,9 +14,10 @@ export function transformActionImportBase(
   edmxActionImport: EdmxActionImport,
   edmxParameters: EdmxParameter[],
   swaggerDefinition: SwaggerPath | undefined,
-  formatter: ServiceNameFormatter
+  formatter: ServiceNameFormatter,
+  bindingEntitySetName?: string
 ): VdmActionImportBase {
-  const name = formatter.originalToActionImportName(edmxActionImport.Name);
+  const name = bindingEntitySetName ? formatter.originalToBoundActionImportName(bindingEntitySetName, edmxActionImport.Name) : formatter.originalToActionImportName(edmxActionImport.Name);
```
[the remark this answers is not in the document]
FrankEssenberger: done

packages/generator/src/edmx-to-vdm/common/action-function-return-types.ts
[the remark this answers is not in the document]
FrankEssenberger: I went for `Partial` entity to indicate that it is a subset.

packages/generator/src/edmx-to-vdm/edmx-to-vdm-util.ts
```diff
@@ -193,9 +193,11 @@ export const propertyJsType = (type: string): string | undefined =>
  * @internal
  */
 export function hasUnsupportedParameterTypes(
-  functionOrAction: EdmxAction | EdmxFunction | EdmxFunctionImportV2
+  functionOrAction: EdmxAction | EdmxFunction | EdmxFunctionImportV2, bindingEntitySetName?: string
 ): boolean {
-  const unsupportedParameters = functionOrAction.Parameter.filter(
+  const parameters = bindingEntitySetName ? functionOrAction.Parameter.slice(1) : functionOrAction.Parameter;
+
+  const unsupportedParameters = parameters.filter(
```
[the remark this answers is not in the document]
FrankEssenberger: done

packages/generator/src/edmx-to-vdm/common/action-function-parameters.ts
```diff
@@ -34,30 +36,35 @@ export function getFunctionImportParameters(
...
+  const parameters = bindingEntitySetName ? edmxParameters.slice(1) : edmxParameters;
+  return parameters.map(p => {
     const swaggerParameter = swaggerDefinition
       ? swaggerDefinition.parameters.find(param => param.name === p.Name)
       : undefined;
     const typeMapping = getTypeMappingActionFunction(p.Type);
     return {
       originalName: p.Name,
-      parameterName: formatter.originalToParameterName(
+      parameterName: bindingEntitySetName ? formatter.originalToBoundParameterName(bindingEntitySetName, edmxActionFunctionImport.Name, p.Name) :
```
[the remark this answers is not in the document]
FrankEssenberger: done

FrankEssenberger:
Once this is merged I will do a test run on all service specs - our E2E tests only cover S/4HANA which is a good portion but in this cases the extra numer of services could help. Current status:

  "numberOfServices": 1226,
  "numberFailedServices": 53

packages/odata-v4/package.json
```diff
@@ -47,6 +47,7 @@
   },
   "devDependencies": {
     "@sap-cloud-sdk/test-services-odata-v4": "^2.10.0",
+    "@sap-cloud-sdk/test-services-e2e": "^2.10.0",
```
jjtang1985, line 50: [req] I guess we have `test-services-odata-v4` and `@sap-cloud-sdk/test-services-e2e` for unit tests and e2e tests respectively.
- I would adjust/use the `test-services-odata-v4`, for the unit test `bound-function-import-request-builder.spec.ts`
- I would not use `@sap-cloud-sdk/test-services-e2e` for unit tests because the spec/client of the e2e tests should be updated after the cds data model updates, which is kind of passive, while our `test-services-odata-v4` is fully under our control.

test-resources/odata-service-specs-e2e/v4/API_TEST_SRV/API-TEST_SRV.edmx
```diff
@@ -6,14 +6,21 @@
                 <EntitySet Name="TestEntity" EntityType="TestService.TestEntity">
                     <NavigationPropertyBinding Path="ToMultiLink" Target="TestEntityLink"/>
                 </EntitySet>
+                <EntitySet Name="TestEntityWithMultipleKeys" EntityType="TestService.TestEntityWithMultipleKeys"/>
```
jjtang1985, line 9: [q] I assume this is not a manual update and is retrieved from the cap app.

test-packages/e2e-tests/test/bound-action.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+      const entity: TestEntityWithMultipleKeys =
+        await testEntityWithMultipleKeysApi
+          .requestBuilder()
+          .getByKey(101, 'a', true)
+          .execute(destination);
+
+      const actionResult = await entity
+        .boundActionWithoutArgumentsWithMultipleKeys({})
+        .execute(destination);
+      expect(actionResult).toEqual('abc');
```
jjtang1985, line 36: [req] This returns the same return value as the line 22.
Please use a different value.

test-packages/e2e-tests/test/bound-function.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      const entity: TestEntityWithMultipleKeys =
+        await testEntityWithMultipleKeysApi
+          .requestBuilder()
+          .getByKey(101, 'a', true)
+          .execute(destination);
+
+      const functionResult = await entity
+        .boundFunctionWithoutArgumentsWithMultipleKeys({})
+        .execute(destination);
+      expect(functionResult).toEqual('xyz');
```
jjtang1985, line 38: [req] This returns the same value as the line 23, please use a different value.

packages/odata-v4/src/request-builder/bound-function-import-request-builder.spec.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  TestEntity,
+  testService
+} from '@sap-cloud-sdk/test-services-e2e/v4/test-service';
```
jjtang1985, line 4: This is the case, which uses e2e service in unit tests.

packages/generator/src/service-name-formatter.ts
```diff
@@ -106,6 +106,20 @@ export class ServiceNameFormatter {
     return applyPrefixOnJsConflictFunctionImports(newName);
   }
 
+  originalToBoundOperationName(
```
jjtang1985, line 109: [req] I guess the unit test is missing also for line 165.

packages/generator/src/operations/return-type.ts
```diff
@@ -31,11 +36,24 @@ export function operationReturnType({
...
     type = `${type} | null`;
   }
-  type = wrapRequestBuilderAroundType(
-    requestBuilderName,
-    parametersTypeName,
-    type
-  );
+
+  if (isBound && !boundEntitySetName) {
+    throw new Error(
```
jjtang1985, line 41: [q]/[req] I assume this will stop the generation?
For people who's using our generator and their specs do not meet our expectation, this will be a breaking change.
Before, the bound functions are ignored, and now the generation will be broken.
We should ignore this case to avoid the breaking changes.

packages/generator/src/operations/parameters-interface.spec.ts
```diff
@@ -7,6 +7,10 @@ import {
 import { parametersInterface } from './parameters-interface';
 
 describe('parameters-interface', () => {
+  it('function parameter interface for bound function', () => {
```
jjtang1985, line 10: [req] left over?

packages/generator/src/entity/class.spec.ts
```diff
@@ -43,4 +53,23 @@ describe('entity class generator', () => {
...
+  xit('has expected parameters for bound function', () => {
+    const fn = classDeclaration.methods?.find(
+      ({ name }) => name === 'getPrice'
+    );
+
+    expect(fn?.parameters?.map(({ name }) => name)).toEqual(['meal']);
+  });
+
+  // TODO decide if list of parameters func(p1,p2,...) or like for unbound and currently implemented func({paramsObj},
+  xit('has expected parameters for bound action', () => {
```
jjtang1985, line 71: [req] left over?

packages/generator/src/entity/imports.spec.ts
```diff
@@ -14,6 +14,15 @@ describe('imports', () => {
     ).toEqual(expectedEntityImports);
   });
 
+  // todo
```
jjtang1985, line 17: [req] left over?

packages/generator/src/entity/imports.ts
```diff
@@ -12,8 +14,28 @@ import { VdmEntity, VdmServiceMetadata } from '../vdm-types';
...
+        oDataVersion
+      ),
+      ...complexTypeImportDeclarations(entity.properties),
+      {
+        namedImports: [`${entity.className}Api`],
+        moduleSpecifier: `./${entity.className}Api`,
+        kind: StructureKind.ImportDeclaration,
+        isTypeOnly: true
+      },
+      ...operationImportDeclarations(service, 'action', entity.actions), // fixme(fwilhe) need to filter here
```
jjtang1985, line 33: [req] left over?

jjtang1985 (changes requested):
Looks good.
Please consider my comments and add the changelog.

packages/generator/src/edmx-parser/v4/edm-types.ts
```diff
@@ -106,7 +106,7 @@ export interface EdmxOperation extends EdmxNamed, EdmxNamespaced {
   /**
    * @internal
    */
-  IsBound: boolean;
+  IsBound: string;
```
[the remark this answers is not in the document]
fwilhe: See the second point in the original PR description where Frank described the issue. In the edmx "parser" we don't do much parsing, and both keeping it as string or as boolean here has issues. I'd argue for keeping this in sync with 'isComposed' and other boolean properties which currently all are strings afaik.

packages/generator/src/edmx-parser/v4/edmx-parser.spec.ts
```diff
@@ -12,6 +12,37 @@ import {
...
+    expect(
+      parseOperations(metadataEdmx.root, 'function').map(
+        operation => operation.IsBound
+      )
+    ).toContain('false');
+    expect(
+      parseOperations(metadataEdmx.root, 'function').map(
+        operation => operation.IsBound
+      )
+    ).toContain('true');
```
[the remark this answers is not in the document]
fwilhe: I did not write this test and I have to admit that it was not obvious to me what it does test. I guess the idea was to make it a bit more resilient to changes of the edmx file. I'll try to improve it.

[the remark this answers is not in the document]
fwilhe: Rewrote the test, I hope it is both more clear and more strict now.

test-resources/odata-service-specs-e2e/v4/API_TEST_SRV/API-TEST_SRV.edmx
fwilhe: I'm not aware that we have a workflow to automatically re-generate those files. I've added the command to regenerate the edmx to the cap app [here](https://github.com/SAP/cloud-sdk-js/blob/bfa038ea5f907d0494afb49393bdecd0e19e9001/test-packages/e2e-tests/package.json#L24), but applying those changes here is a manual effort. I'd be happy to see this improved, but my understanding is that fully automating this is not trivial.

test-packages/e2e-tests/test/bound-action.spec.ts
fwilhe: improved with a unique value

test-packages/e2e-tests/test/bound-function.spec.ts
fwilhe: improved with a unique value

packages/generator/src/operations/return-type.ts
fwilhe: I'm not sure if I agree here. Yes this is breaking for invalid edmx files, but is our policy to always avoid this? I think that this could be handled with a compatibility note and allow us to ship better software if we clearly define what we expect, rather than having more warnings.

packages/generator/src/operations/parameters-interface.spec.ts
fwilhe: looks like, @FrankEssenberger can you comment on that?

packages/generator/src/entity/class.spec.ts
fwilhe: Yes. @FrankEssenberger I've tried to update the test but I'm not sure how I can actually get the parameters here.

FrankEssenberger: I think we aligned to keep the object for consistency in the call yesterday. In this case you have to check there is a paramter with name `paramaters` and the the type is `GetPriceParameterType` or however it is called. Then to make it 100% you would check the definition of the `GetPriceParameterType` that the object contains two properties. For this you have to check the `parametersInterfac()` result. Perhaps it is enough to check here the type and the two properties in the internal `statements` of the class and check if the `parametersInterface()` has a sparate test.

packages/generator/src/operations/parameters-interface.spec.ts
FrankEssenberger: ah, look at my comment below. I wanted to add a test that the parameter object contains the expected values but forgot about it. Feel free to add a test.

fwilhe: I've tried adding a test, but the function seems to do nothing different for bound/unbound.. I don't see additional value in that test.

packages/generator/src/entity/imports.ts
fwilhe: yes, thanks!

packages/generator/src/operations/return-type.ts
jjtang1985: As a general generator CLI design (which is done only in the openapi generator), we provide an option `--skipValidation`, so the user can still finish the generation and use the generated client.
In the long term, the odata generator should also have this option ([ticket](https://github.com/SAP/cloud-sdk-backlog/issues/506)) and it should not block the users.

Currently, the policy is, we allows limited breaking changes for bug fix or improvement and mark them in the compatibility notes.

Therefore, both of the following options are valid:
1. show warning instead of error and wait for v3
2. show error and stop the generation and mark it in the compatibility notes + support potential issues

As we are going to work on the generator [milestone](https://github.com/SAP/cloud-sdk-backlog/milestone/19) soon, I would prefer 1, because we will then have the `--skipValidation` option, and the major version bump better fits breaking changes.

test-packages/e2e-tests/package.json
```diff
@@ -55,7 +57,9 @@
     "requires": {
       "db": {
         "kind": "sqlite",
-        "model": "*"
+        "credentials": {
```
jjtang1985, line 60: [req] I guess this is not intended?

test-packages/e2e-tests/package.json
```diff
@@ -43,6 +44,7 @@
     "json-schema-faker": "^0.5.0-rcv.46",
     "mock-fs": "^5.2.0",
     "moment": "^2.29.4",
+    "nock": "^13.2.9",
```
jjtang1985, line 47: [req] In the e2e test package, `nock` as a mock package should not be used?

test-packages/e2e-tests/test/bound-function-import-request-builder.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  const destination: Destination = {
+    url: host,
+    username: 'username',
+    password: 'password',
+    sapClient: '123',
+    authTokens: [],
+    originalProperties: {}
+  };
+
+  it('executes a bound action', async () => {
```
jjtang1985, line 22: [req] I guess this is the location where nock is used.
Not sure whether this is the test that came from the previous unit test?
- if this is an e2e test, then we should use the real e2e app
- if this is a unit test, it should be located in the odata-* package with unit test client

packages/generator/src/operations/return-type.ts
jjtang1985: @FrankEssenberger , please leave your opinion about this.

test-packages/e2e-tests/package.json
fwilhe: true. sorry.

test-packages/e2e-tests/test/bound-function-import-request-builder.spec.ts
fwilhe: Yes I moved the test here thinking this is ok.. I'll look into how to improve it

test-packages/e2e-tests/package.json
fwilhe: removed

test-packages/e2e-tests/test/bound-function-import-request-builder.spec.ts
fwilhe: I think this test is a duplicate of the existing e2e tests for bound actions. removed.

jjtang1985:
@fwilhe , please add the changelog and there is a comment in `imports.spec.ts`

packages/generator/src/edmx-to-vdm/common/operation.ts
```diff
@@ -1,29 +1,35 @@
...
  */
 export function transformOperationBase(
-  edmxOperation: EdmxFunctionImportV2 | EdmxOperationImport,
+  edmxOperation: EdmxFunctionImportV2 | EdmxJoinedOperation,
   edmxParameters: EdmxParameter[],
   type: 'function' | 'action',
   swaggerDefinition: SwaggerPath | undefined,
-  formatter: ServiceNameFormatter
+  formatter: ServiceNameFormatter,
+  edmxBindingEntitySetName?: string
```
FrankEssenberger, line 21: ywa good name now

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -20,95 +29,280 @@ const logger = createLogger({
...
 const extractResponse = (response: string) => `${response}.value`;
 
-interface JoinedOperationData {
-  operationImport: EdmxOperationImport;
-  operation: EdmxOperation;
+function splitMissingOperation(
+  operationImports: EdmxOperationImport[],
+  operations: EdmxOperation[]
+): [EdmxOperationImport[], EdmxJoinedOperation[]] {
+  return operationImports.reduce<
```
FrankEssenberger, line 36: [req] didn't you introduce a `partition` or `divide` method on arrays? If so please use this here or create a follow up to use this for the two methods here.

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -20,95 +29,280 @@ const logger = createLogger({
...
+    [[], []]
+  );
+}
+
+function extractEntitySetName(type: string): string {
+  const components = type.split('.');
+  return components.pop()!;
+}
+
+function splitMissingParameter( // fixme name
```
FrankEssenberger: [req] here the name it is more than a split. Transformation of data.

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -20,95 +29,280 @@ const logger = createLogger({
...
+
+      if (!curr.Parameter.length) {
+        return [validOperations, [...withoutParameter, curr]];
+      }
+
+      const entitySetName = extractEntitySetName(curr.Parameter[0].Type);
+      if (entitySetName) {
+        const bound = {
+          ...curr,
+          entitySetName,
```
FrankEssenberger, line 84: [req] Also put here something like `edmxEntityName` or so to indicate that it is the raw name with potential `A_`

packages/generator/src/entity/file.ts
```diff
@@ -14,10 +15,13 @@ export function entitySourceFile(
   return {
     kind: StructureKind.SourceFile,
     statements: [
-      ...entityImportDeclarations(entity, service.oDataVersion),
+      ...entityImportDeclarations(entity, service, service.oDataVersion),
       ...otherEntityImports(entity, service),
       entityClass(entity, service),
-      entityTypeInterface(entity, service)
+      entityTypeInterface(entity, service),
+      ...[...entity.functions, ...entity.actions].map(operation =>
```
FrankEssenberger, line 22: spread city @marikaner

packages/generator/src/edmx-to-vdm/v4/operation.ts
fwilhe: `partition` already existed in `array.ts` but it can't be easily used here because we have 2 type parameters for the 2 result lists with transformation happening inside the reduce.. maybe we can split the transformation and the partition in 2 steps?

packages/generator/src/edmx-to-vdm/v4/operation.ts
fwilhe: not sure what a good name would be.

packages/generator/src/edmx-to-vdm/v4/operation.ts
FrankEssenberger: no I missed the point of the two data points.

packages/generator/src/edmx-to-vdm/v4/operation.ts
FrankEssenberger: Perhaps you can use the `partition`of the array inside the method and add the setting of the parameter in a separate method with a good name.

packages/generator/src/operations/return-type.ts
```diff
@@ -44,5 +60,16 @@ function wrapRequestBuilderAroundType(
   parameterName: string,
   type: string
 ) {
+  // BoundActionImportRequestBuilder<TestEntity<T>,T,BoundActionWithoutArgumentsParameters<T>,string>
```
FrankEssenberger, line 63: [q] leftover?

packages/generator/src/operations/return-type.ts
```diff
@@ -44,5 +60,16 @@ function wrapRequestBuilderAroundType(
...
   return `${requestBuilderName}<DeSerializersT, ${parameterName}<DeSerializersT>, ${type}>`;
 }
+
+function wrapRequestBuilderAroundTypeBound(
+  entityName: string,
+  requestBuilderName: string,
+  parameterName: string,
+  type: string
+) {
+  // BoundActionImportRequestBuilder<TestEntity<T>,T,BoundActionWithoutArgumentsParameters<T>,string>
```
FrankEssenberger, line 73: [q] leftover?

packages/generator/src/operations/return-type.ts
FrankEssenberger: I think this case is just a indicator that we messed something up. If the method isBound there must be an entityClassName since it is bound. So this should never happen at runtime. However, since the property is optional for the unbound case we need a check.

FrankEssenberger (approved):
LGTM - please clean away the few old comments Junjie and I spotted. Also consider the renaming suggestions besides that I think we can merge it.

packages/generator/src/operations/return-type.ts
jjtang1985: Discussed with Frank.
This check turns out to be related to the ts implementation, as `boundEntitySetName` is optional.
An invalid spec will not trigger this error.

jjtang1985 (approved):
LGTM, please go ahead.

PR #2877, opened 2022-09-15
feat: `orderBy` set ascending order by default

shojisoma:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/169

#### Changes
This makes the oderBy method set in ascending order by default:
```
peopleApi
  .requestBuilder()
  .getAll()
  .orderBy(peopleApi.schema.GENDER)
```
The URL for the query will be:
```
/People?$orderby=Gender asc
```


packages/odata-common/src/request-builder/count-request-builder.spec.ts
```diff
@@ -3,55 +3,47 @@ import { getAllRequestBuilder } from '@sap-cloud-sdk/test-services-odata-common/
 import { defaultDestination } from '../../../../test-resources/test/test-util/request-mocker';
 import { Filter } from '../filter/filter';
 
-describe('CountRequestBuilder', () => {
+describe('orderBy', () => {
```
FrankEssenberger, line 6: [q] Why did you change the tests in this file?

shojisoma: this is wrong. I edited this instead of `get-all-request-builder.spec.ts`. I reset all.

packages/odata-common/src/request-builder/get-all-request-builder-base.ts
```diff
@@ -61,13 +61,48 @@ export abstract class GetAllRequestBuilderBase<
...
+      EntityT,
+      DeSerializersT,
+      EntityApi<EntityBase, DeSerializersT>
+    >[]
   ): this {
-    this.requestConfig.orderBy = transformVariadicArgumentToArray(first, rest);
+    this.requestConfig.orderBy = transformVariadicArgumentToArray(
+      first,
+      rest
+    ).map(entity => {
```
FrankEssenberger, line 100: [pp]I would not use the term `entity` since we use it also for other things. Perhaps use `element` or `orderabelOrOrderableInput`

packages/odata-v2/src/request-builder/get-all-request-builder.spec.ts
```diff
@@ -47,6 +48,41 @@ describe('GetAllRequestBuilder', () => {
...
+        .requestBuilder()
+        .getAll()
+        .orderBy(
+          desc(testEntityApi.schema.COMPLEX_TYPE_PROPERTY.stringProperty)
+        )
+        .url(defaultDestination);
+      expect(request).toBe(expected);
+    });
+
+    it('should set ascending order when any order not specified', async () => {
```
FrankEssenberger, line 75: [req] Could you add one more test where two properties are given where one has nothing (default should be asc) and the other one has explicitly `desc` set.

orderBy(
          desc(prop1,desc(prop2))
        )

to see that the looping over the elemts wokrs if the types are mixed.

packages/odata-v4/src/request-builder/get-all-request-builder.spec.ts
```diff
@@ -31,6 +32,41 @@ describe('GetAllRequestBuilder', () => {
...
+        .requestBuilder()
+        .getAll()
+        .orderBy(
+          desc(testEntityApi.schema.COMPLEX_TYPE_PROPERTY.stringProperty)
+        )
+        .url(defaultDestination);
+      expect(request).toBe(expected);
+    });
+
+    it('should set ascending order when any order not specified', async () => {
```
FrankEssenberger, line 59:
```suggestion
    it('should set ascending order as default if no order is specified', async () => {
```

packages/odata-v2/src/request-builder/get-all-request-builder.spec.ts
FrankEssenberger, line 75:
```suggestion
    it('should set ascending order as default if no order is specified', async () => {
```

FrankEssenberger (changes requested):
LGTM only asked for one more test. Also I think it would be ok to remove the test from the v4 package, because the logic comes from common so it is ok to test it there.

packages/odata-common/src/request-builder/get-all-request-builder-base.ts
shojisoma: I changed the parameter name entity.

packages/odata-v2/src/request-builder/get-all-request-builder.spec.ts
shojisoma: Is the additonal test will be `orderBy(prop1, desc(prop2))`?
I think the `prop1,desc(prop2)` cannot be wrapped by `desc` because `asc` and `desc` takes a single argument.

FrankEssenberger: oh, yes typo from my side. Only one desc

shojisoma: I added the test in V2 package

FrankEssenberger (approved):
LGTM

PR #2883, opened 2022-09-16
improve: expose proxy key of the `transport`

jjtang1985:
Related to: https://github.com/SAP/cloud-sdk-backlog/issues/840

This PR exposes the `proxy` key of the `nodemailer` when setting up the `transport`, so users can decide whether they would like to pass the value, e.g., from the environment variables.

Also, added a verification call.


test-packages/e2e-tests/test/on-prem/odata.spec.ts
```diff
@@ -54,3 +76,13 @@ describe('OData OnPrem', () => {
       });
   }, 60000);
 });
+
+describe('OData Cloud', () => {
```
FrankEssenberger, line 80: [pp] perhaps move this test to a file not in the `on-prem` folder?

test-packages/e2e-tests/test/on-prem/odata.spec.ts
```diff
@@ -54,3 +76,13 @@ describe('OData OnPrem', () => {
...
+  it('should go through proxy if proxy authorization header matches', async () => {
```
FrankEssenberger, line 81: Very nice that we have a E2E test now, could you perhaps also add one more test here where you set the proxu in the `http_proxy` env variable as string and use the `registerDestination` or `mockDestinationService` to put only the 

```ts
Destination = {
  url: e2eDestination?.url,
 name: 'myDest',
  authentication: 'NoAuthentication'
}
```

and the proxy with headers should be added by the fetching methods. There are units tests inthe `proxy-util` but no E2E test I think.

packages/mail-client/src/mail-client-types.ts
```diff
@@ -333,4 +333,8 @@ export interface SmtpTransportOptions {
    * Connected socket to use instead of creating and connecting a new one. If secure option is true, then socket is upgraded from plaintext to ciphertext.
    */
   connection?: net.Socket | undefined;
+  /**
+   * A proxy URL used for connecting the SMTP server.
```
FrankEssenberger, line 337: [pp] Perhaps be explicit that this string is handed to the nodemailer implementation which does the proxy handling for us.

test-packages/e2e-tests/test/on-prem/odata.spec.ts
jjtang1985: I renamed the directory to `proxy`.

PR #2891, opened 2022-09-20
fix: Add force option for `clearOutputDir`

FrankEssenberger:
Add the force option so it does not fail in case the directory is not there.


packages/openapi-generator/src/generator.ts
```diff
@@ -71,7 +71,9 @@ export async function generateWithParsedOptions(
 
   if (options.clearOutputDir) {
     const rm = promisesFs.rm || promisesFs.rmdir;
-    await rm(options.outputDir, { recursive: true });
+    const forceOption =
```
jjtang1985, line 74: [req] please add a comment, so we know the `rm` was introduced in `v14.14.0`.

jjtang1985 (approved):
LGTM. I left a suggestion.

PR #2893, opened 2022-09-21
chore: Get improved script from other branch

florian-richter:
This improves the error message when the changelogs are incorrect.

Closes SAP/cloud-sdk-backlog#834.


tomfrenken (approved):
Overall LGTM, however if there would be an issue with the RegEx I would probably not catch it

PR #2898, opened 2022-09-22
Fix: Fix type in delete request builder

marikaner:
I found this accidentally and believe this is a typo.

Changelog is still missing.


jjtang1985 (approved):
LGTM.
[pp] The PR title is "Fix: Fix xxx", which seems to be strange.

PR #2899, opened 2022-09-22
adr: destination ts type

jjtang1985:
When support the mail destination, our `Destination` cannot be used due to the mandatory key `url`.
Here is an ADR for this topic.

### DIFF
--- a/knowledge-base/adr/0035-destination-ts-type.md
+++ b/knowledge-base/adr/0035-destination-ts-type.md
@@ -78,7 +78,7 @@ const md4: MailDestination = { originalProperties: {'mail.smtp.host': ''}}
 - [-] The code `executeHttpReqeust(getDestination(...), ...)` is not possible, as `getDestination()` does not know the type (`HTTP`/`MAIL`) in compile time.
 - [-] At least medium complexity, as all the `Destination`s in the `http-client` are affected.
 
-### C:
+### C: Only create a new `MailDestination`
 We can also keep the `Destination` for both `connectivity` and `http-client` packages and only create a new `MailDestination`.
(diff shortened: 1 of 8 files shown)


knowledge-base/adr/0035-destination-ts-type.md
```diff
@@ -0,0 +1,108 @@
...
+  type?: 'HTTP' | 'LDAP' | 'MAIL' | 'RFC';
+  ...
+}
+```
+
+However, when supporting destinations with type `MAIL`, the mandatory `url` key becomes a problem, as mail destination saves the host url in the properties of the destination.
+
+## Alternatives
+
+### A: Make `url` optional
```
FrankEssenberger, line 27: [pp] I would go for option A, since it seems the most comfortable and return the `Destination` type in the getDestination() method and allow also the `Destination` in the `executeHttpRequest`. I would create the types `HttpDestination` and `MailDestination` anyhow and provided some typeguards `isHttpDestination` and `isMailDestination` which are used in our code then to narrow the type within the `executeHttpRequest` so the internal interface so we good there and throw a exception if the wrong type is provided.

Since the response of `getDestination` depends on a service at runtime we do not get much from a static type check and need the runtime exception anywat. Therefore I would keep it simple in the user facing interface.

FrankEssenberger: But option `B` is also fine for me.

jjtang1985: `B` is a breaking change, unfortunately.

marikaner (approved):
I agree with your decision. But this is a breaking change and should only be introduced in version 3. Let's keep mail support experimental until then. I will create a ticket for this.

PR #2901, opened 2022-09-23
chore: Improve destination cache tests

FrankEssenberger:
We have a customer issue, where destination are chaced forever, I just double checked the tests we have and I guess this is not the case - perhaps the token lifetime on the customer side is too long. Since I went over the test I made some small improvements.


deekshas8 (approved):
LGTM

PR #2917, opened 2022-09-29
fix: API docs module section issues

shojisoma:
Closes [SAP/cloud-sdk-backlog#883](https://github.com/SAP/cloud-sdk-backlog/issues/883)

This fixes 2 issues in our API docs module section.
#### issue 1: The connectivity package has an internal layer that contains unnecessary list item [`Protocol`](https://sap.github.io/cloud-sdk/api/2.9.0/)
this comes from that typedoc wrongly recognizes the internal namespace [`Protocol`](https://github.com/SAP/cloud-sdk-js/blob/b7344eb5c737984de5dff3f0948e6e343e2ab15c/packages/connectivity/src/scp-cf/protocol.ts#L14) is exported instead it has `@internal` tag. That's because the same name enum [`Protocol`](https://github.com/SAP/cloud-sdk-js/blob/b7344eb5c737984de5dff3f0948e6e343e2ab15c/packages/connectivity/src/scp-cf/protocol.ts#L6) is exported in the same file then typedoc see the internal namespace as exported one. the namespace `Protocol` and the enum `Protocol` must have different names. ~~this changed the namespace name to the identifiable name `ProtocolNameSpace`.~~
**update**: Since the namespace is only used for the `of` method inside it, this extracts the function as a new function `getProtocol()` outside of the namespace and deletes it.
#### issue 2: The mail package does not follow the common naming scheme and instead shows a file path.
Currently, the mail package link is a file path [`mail-client/src`](https://sap.github.io/cloud-sdk/api/2.9.0/modules/mailclient_src) but this should be `@sap-cloud-sdk/mail-client`.
This comes from missing JSDoc annotations in the `index.ts`. It needs the annotations:
```
/**
 * [[include:mail-client/README.md]]
 * @packageDocumentation
 * @module @sap-cloud-sdk/mail-client
 */
```
**[Actually, this has already been added.](https://github.com/SAP/cloud-sdk-js/commit/838aaef7d280eb002d9abb563f483104c0a86da4)** The incorrect module name will be updated correctly.

### DIFF
--- a/packages/connectivity/src/scp-cf/get-protocol.ts
+++ b/packages/connectivity/src/scp-cf/get-protocol.ts
@@ -1,6 +1,6 @@
 import { createLogger } from '@sap-cloud-sdk/util';
 import { Destination } from './destination/destination-service-types';
-import { Protocol, ProtocolNamespace } from './protocol';
+import { Protocol, getProtocol } from './protocol';
 
 const logger = createLogger({
   package: 'connectivity',
@@ -23,7 +23,7 @@ export function getProtocolOrDefault(destination: Destination): Protocol {
     );
     return Protocol.HTTPS;
   }
-  const casted = ProtocolNamespace.of(protocol[0]);
+  const casted = getProtocol(protocol[0]);
   if (casted) {
     return casted;
   }
--- a/packages/connectivity/src/scp-cf/get-protocol.ts
+++ b/packages/connectivity/src/scp-cf/get-protocol.ts
@@ -1,6 +1,6 @@
 import { createLogger } from '@sap-cloud-sdk/util';
 import { Destination } from './destination/destination-service-types';
-import { Protocol } from './protocol';
+import { Protocol, ProtocolNamespace } from './protocol';
 
 const logger = createLogger({
   package: 'connectivity',
@@ -23,7 +23,7 @@ export function getProtocolOrDefault(destination: Destination): Protocol {
     );
     return Protocol.HTTPS;
   }
-  const casted = Protocol.of(protocol[0]);
+  const casted = ProtocolNamespace.of(protocol[0]);
   if (casted) {
     return casted;
   }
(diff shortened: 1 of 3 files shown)


packages/connectivity/src/scp-cf/get-protocol.ts
```diff
@@ -23,7 +23,7 @@ export function getProtocolOrDefault(destination: Destination): Protocol {
     );
     return Protocol.HTTPS;
   }
-  const casted = Protocol.of(protocol[0]);
+  const casted = ProtocolNamespace.of(protocol[0]);
```
FrankEssenberger, line 26: [req] I guess the idea behind the namespace was to give the `of`method some context so that it is clear what it creates. I guess if this leads to some name collision with the `Protocol` object I would just create a method `createProtocol(str:string)` or `getProtocal` and get rid of the namespace.

FrankEssenberger: [q] There was also the issue with the mail client package name. Is this solved?

shojisoma: > [q] There was also the issue with the mail client package name. Is this solved?

yes. it comes from missing JSDoc annotations for typedoc in the index.ts. It needs the annotations:
```
/**
 * [[include:mail-client/README.md]]
 * @packageDocumentation
 * @module @sap-cloud-sdk/mail-client
 */
```
[Actually, this has already been added](https://github.com/SAP/cloud-sdk-js/commit/838aaef7d280eb002d9abb563f483104c0a86da4) but it is not on docs repo. The incorrect module name will be updated correctly.

shojisoma: > [req] I guess the idea behind the namespace was to give the `of`method some context so that it is clear what it creates. I guess if this leads to some name collision with the `Protocol` object I would just create a method `createProtocol(str:string)` or `getProtocal` and get rid of the namespace.

I made a `getProtocol()` to get rid of the namespace. It also works fine.

FrankEssenberger (approved):

PR #2918, opened 2022-09-29
Improve error messages in ts compiler

fwilhe:
In some cases our compiler created error messages such as:

```
Caused by:
Error: Compilation Errors:
.../Foo.ts:2:772 - error TS7053: [object Object]
```

This commit fixes that by selecting the string property if the object containing the error is not a string.


jjtang1985 (approved):
LGTM

PR #2942, opened 2022-10-10
Add entity api to entity base

fwilhe:
Change `EntityBase` so it has an `EntityApi` field. This change is a prerequisite for implementing OData bound actions and functions in PR #2875.


test-resources/odata-service-specs/v2/service-mapping.json
```diff
@@ -1,7 +1,7 @@
 {
   "API_MULTIPLE_SCHEMAS_SRV": {
     "directoryName": "multiple-schemas-service",
-    "servicePath": "VALUE_IS_UNDEFINED",
```
fwilhe, line 4: not sure if this makes sense or is required @marikaner

packages/odata-common/src/entity-base.ts
```diff
@@ -72,13 +72,17 @@ export abstract class EntityBase {
    */
   protected _customFields: Record<string, any>;
 
+  private schema: Record<string, any>;
+
   abstract readonly _oDataVersion: any;
 
-  constructor(private schema: Record<string, any>) {
+  constructor(readonly _entityApi: any) {
```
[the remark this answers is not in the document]
FrankEssenberger: I am always in favor of keeping only one object - hence we need the  bigger object `_entityApi`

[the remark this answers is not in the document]
fwilhe: Addressed in [26d142f](https://github.com/SAP/cloud-sdk-js/pull/2942/commits/26d142f9556a63dcfb7757375bab9c813996bf8a).

It seems like this is really just a small local change. I have not found any other places where I need to change something.

fwilhe:
@FrankEssenberger @marikaner @jjtang1985 I think this PR is ready for review now.

Please have a close look at those commits while reviewing

make entity api not private f7f768c
explicit type annotation 01b6c6c
snapshot 06a3e9a

packages/generator/src/entity/class.ts
```diff
@@ -27,6 +27,18 @@ export function entityClass(
   return {
     kind: StructureKind.Class,
     name: `${entity.className}<T extends DeSerializers = DefaultDeSerializers>`,
+    ctors: [
+      {
+        parameters: [
+          {
+            name: '_entityApi',
```
FrankEssenberger, line 34: [q] why `_` prefix for the name? Is this our convention for private things?

fwilhe: This seems to be the convention for properties on entities to avoid name clashes.

packages/odata-common/src/entity-base.ts
```diff
@@ -74,10 +74,10 @@ export abstract class EntityBase {
 
   abstract readonly _oDataVersion: any;
 
-  constructor(private schema: Record<string, any>) {
-    nonEnumerable(this, 'schema');
+  constructor(readonly _entityApi: any) {
```
FrankEssenberger, line 77: [q] I know we show in all the tutorials to use the `entityBuilder` to create entities. However, if someone used the constructor for some obscure reason directly this would break now. I would add a compatibility note. Also I would add the @internal annotation on the constructor.

test-packages/e2e-tests/test/mail/mail.spec.ts
```diff
@@ -37,22 +37,23 @@ describe('Mail', () => {
...
-    expect(responses[0].accepted?.length).toBe(2);
+  //   expect(responses.length).toBeGreaterThan(9);
+  //   expect(responses[0].accepted?.length).toBe(2);
 
-    const mails = fs.readdirSync(join(resolve('test'), 'mail', 'test-output'));
-    expect(mails.length).toBeGreaterThan(9);
-  }, 60000);
+  //   const mails = fs.readdirSync(join(resolve('test'), 'mail', 'test-output'));
+  //   expect(mails.length).toBeGreaterThan(9);
+  // }, 60000);
```
FrankEssenberger, line 56: [req] Thish should get back in before merging.

packages/odata-common/src/entity-base.ts
fwilhe: You're right, I was also wondering about backwards compatibility, my understanding was that typically users don't call the constructor. I'll add an annotation and a note.

test-packages/e2e-tests/test/mail/mail.spec.ts
fwilhe: Oh I forgot about this, you're right. Thanks.

FrankEssenberger (changes requested):
Please add the compatibility note and put the test back in. Then I can approve. Content is fine.

fwilhe:
> Please add the compatibility note and put the test back in. Then I can approve. Content is fine.

Done, please have another look.

Thanks!

.changeset/flat-socks-listen.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-cloud-sdk/odata-common': patch
+'@sap-cloud-sdk/odata-v2': patch
+'@sap-cloud-sdk/odata-v4': patch
```
jjtang1985, line 4: For changes with compatibility notes, please use minor.

packages/odata-v2/src/uri-conversion/get-expand.spec.ts
```diff
@@ -54,7 +54,8 @@ const testExpandMultiLink = {
 const testNestedExpandLink = {
   expand: testEntityApi.schema.TO_SINGLE_LINK.select(
     testEntitySingleLinkApi.schema.TO_MULTI_LINK.select(
-      testEntityMultiLinkApi.schema.TO_MULTI_LINK
+      // Incorrect type, therefore type assertion to "any", level 2 api does not have navigation properties
```
jjtang1985, line 57: [q] Is this a todo or follow up?

packages/odata-v4/src/uri-conversion/get-expand.spec.ts
```diff
@@ -64,7 +64,8 @@ const testExpandMultiLink = {
 const testNestedExpandLink = {
   expand: testEntityApi.schema.TO_SINGLE_LINK.expand(
     testEntitySingleLinkApi.schema.TO_SINGLE_LINK.expand(
-      testEntityMultiLinkApi.schema.TO_MULTI_LINK_1
+      // Incorrect type, therefore type assertion to "any", level 2 api does not have navigation properties
```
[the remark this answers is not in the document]
fwilhe: No, this is documenting. We need to use type any here to make the ts compiler happy.

packages/odata-v4/src/entity.spec.ts
```diff
@@ -36,7 +36,7 @@ describe('entity', () => {
         testEntitySingleLinkApi
           .entityBuilder()
           .toSingleLink(
-            testEntityLvl2MultiLinkApi
+            testEntityLvl2SingleLinkApi
```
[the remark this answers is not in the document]
fwilhe: Yes, as far I I understand the test was wrong before. Does not compile if we leave it as is.

packages/odata-v2/src/uri-conversion/get-expand.spec.ts
fwilhe: No, this is documenting. We need to use type any here to make the ts compiler happy.

FrankEssenberger (approved):
LGTM

PR #2951, opened 2022-10-11
add workflow

FrankEssenberger:


FrankEssenberger:
Add file once so that we can dispatch it from a different branch from now on.

PR #2964, opened 2022-10-17
chore: delete other v1 api docs

jjtang1985:
This PR deletes other v1 api docs and only keep the latest v1 doc.

Another reason to delete the v1 docs:
> your delete old api docs PR was to large in the browser it crashed and in the IDE I was at least able to approve but the load changes also crashed


FrankEssenberger (approved):
LGTM - I checked only 1.x.y got deleted. However on the repo I see releases up to v1.54.0 (https://github.com/SAP/cloud-sdk-js/releases?q=v1&expanded=true) do we did not release api doc for the version?

jjtang1985:
> However on the repo I see releases up to v1.54.0 (https://github.com/SAP/cloud-sdk-js/releases?q=v1&expanded=true) do we did not release api doc for the version?

The release pipeline has been changed after the v2 release (v1.54.0 was the latest at that time) and only v2 docs are considered now.

PR #2966, opened 2022-10-17
fix: update broken API doc links

shojisoma:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/889

this updates broken API doc links

### DIFF
--- a/packages/connectivity/README.md
+++ b/packages/connectivity/README.md
@@ -41,7 +41,7 @@ If you would like to contribute to the SAP Cloud SDK, please make yourself famil
 
 <br>
 
-- [SAP Cloud SDK Documentation portal](https://sap.github.io/cloud-sdk/)
+- [SAP Cloud SDK Documentation portal](https://sap.github.io/cloud-sdk/api/latest)
 - [SAP Cloud SDK Documentation portal - Getting started guide](https://sap.github.io/cloud-sdk/docs/js/getting-started)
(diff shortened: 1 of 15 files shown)


packages/connectivity/README.md
```diff
@@ -41,7 +41,7 @@ If you would like to contribute to the SAP Cloud SDK, please make yourself famil
 
 <br>
 
-- [SAP Cloud SDK Documentation portal](https://sap.github.io/cloud-sdk/)
+- [SAP Cloud SDK Documentation portal](https://sap.github.io/cloud-sdk/api/latest)
```
FrankEssenberger, line 44: [req] when I follow the link `https://sap.github.io/cloud-sdk/api/latest` in my browser I am redirected to `https://sap.github.io/cloud-sdk/api/2.9.0/` which gives me a page not found. So something is not working.

shojisoma: It works locally. Let me check it.

FrankEssenberger (approved):
As discussed, once the docs PR is merged adjusting the redirect this should work. Please check before mergingn.

packages/connectivity/README.md
shojisoma: The Problem was the API docs are not under baseURL `sap.github.io/cloud-sdk/` because it is static content. I removed `useBaseUrl()` then just navigate to `/cloud-sdk/api/LATEST-VERSION`.

PR #2978, opened 2022-10-19
[chore] Use fs to write files

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#736.

As a first step to call the prettier on the source files programatically we need to write the sources not via the `project.save` of ts-morph. In the remove ts-morph tickets this would have happened anyhow. O moved the shared file write stuff to the generator common package.


deekshas8 (approved):
LGTM. So I assume now we can just run the prettier on the `content` in `createFile` method?

FrankEssenberger:
> LGTM. So I assume now we can just run the prettier on the `content` in `createFile` method?

Yes, this was my plan. I just wanted to make the replacement first.

PR #2996, opened 2022-10-25
chore: update api docs link

deekshas8:
This PR fixes the links for docs portal and api docs in README files.


packages/connectivity/README.md
```diff
@@ -41,9 +41,9 @@ If you would like to contribute to the SAP Cloud SDK, please make yourself famil
 
 <br>
 
-- [SAP Cloud SDK Documentation portal](https://sap.github.io/cloud-sdk/api/latest)
+- [SAP Cloud SDK Documentation portal](https://sap.github.io/cloud-sdk)
 - [SAP Cloud SDK Documentation portal - Getting started guide](https://sap.github.io/cloud-sdk/docs/js/getting-started)
-- [SAP Cloud SDK Documentation portal - API documentation](https://sap.github.io/cloud-sdk/docs/js/api-reference-js-ts)
+- [SAP Cloud SDK Documentation portal - API documentation](https://sap.github.io/cloud-sdk/api/latest)
```
FrankEssenberger, line 46: [q] What if the java sdk wanted to host their JavaDoc somewhere? The URL does not contain the `js` part? @marikaner should we change this in the future?

marikaner: If we change this in the future, we should change it here as well, but for now we can leave it. I guess if we change this we will also add another redirect.

marikaner (approved):
LGTM

FrankEssenberger (approved):
LGTM - just a general question

PR #3003, opened 2022-10-27
chore: Introduce Operations instead of Actions/Functions

marikaner:
This removes lots of code duplication, where we have no difference between actions and functions. I found it very confusing because I would always crosscheck both to see if there is a difference, but there never was. 
Even worse: some of the code would differ, but do the same!
The only actual difference I found that the original emdx file has an import that has either an `Action` or `Function` property. After that, everything is the same, except for naming of request builders and co.

I did not change any logic (just some places with minor refactoring).
The generated code is the same, except for some ordering of imports, some whitespace and some wording in comments.
I adjusted the tests for this, but did not change any other logic in the tests.

Further, I found a test with `.only` which excluded the other test in this file. I reenabled the test.


packages/generator/src/action-function-import/operation.ts
```diff
@@ -0,0 +1,77 @@
...
+import voca from 'voca';
+import { VdmOperation, VdmServiceMetadata } from '../vdm-types';
+import { isEntityNotDeserializable } from '../edmx-to-vdm/common';
+import { getRequestBuilderArgumentsBase } from './request-builder-arguments';
+import { operationReturnType } from './return-type';
+
+const parameterName = 'parameters';
+
+/**
+ * @internal
```
FrankEssenberger, line 11: [pp] Add some doc to explain what this object represents: I guess function and action.

packages/generator/src/vdm-types.ts
```diff
@@ -270,7 +270,7 @@ export interface VdmComplexType {
 /**
  * @internal
  */
-export interface VdmFunctionImportBase {
+export interface VdmOperationBase {
```
FrankEssenberger, line 273: [pp] Since the name is now not function or action consider adding a small JS doc to explain what this represents.

packages/generator/src/vdm-types.ts
```diff
@@ -64,11 +64,11 @@ export interface VdmServiceEntities {
   /**
    * @internal
    */
-  functionImports: VdmFunctionImport[];
+  functionImports: VdmOperation[];
   /**
    * @internal
    */
-  actionImports?: VdmActionImport[];
+  actionImports?: VdmOperation[];
```
FrankEssenberger, line 71: [q] Should we create a follow up to merge the actions and functions also on a Vdm level? We end up with a action-imports.ts and function-imports.ts in the generated clients and customers need to know if it is a action or function they want to import. We could also create a operations.ts which contains actions and functions. I think this would need some discussion - but we could get rid of the two request builders and many types.

FrankEssenberger: Perhaps create a follow up ticket.

packages/generator/src/batch/type.ts
```diff
@@ -60,7 +57,7 @@ function getReadRequestType(service: VdmServiceMetadata): string {
   );
   const functionImportsReturnTypes = service.functionImports
     .filter(f => f.httpMethod.toLowerCase() === 'get')
-    .map(f => functionImportReturnType(f));
+    .map(fn => operationReturnType(fn));
```
FrankEssenberger, line 60: [pp] in the line above it is called `f` perhaps align it.

packages/generator/src/edmx-to-vdm/common/operation.ts
```diff
@@ -1,42 +1,44 @@
...
   edmxParameters: EdmxParameter[],
+  type: 'function' | 'action',
   swaggerDefinition: SwaggerPath | undefined,
   formatter: ServiceNameFormatter
-): VdmFunctionImportBase {
-  const name = formatter.originalToFunctionImportName(edmxFunctionImport.Name);
-  const functionImport = {
-    originalName: edmxFunctionImport.Name,
+): VdmOperationBase {
+  const name = formatter.originalToFunctionImportName(edmxOperation.Name);
```
FrankEssenberger, line 21: [pp] Perhaps rename the method on the formatter to `originalToOperationName

packages/generator/src/edmx-to-vdm/common/operation.ts
```diff
@@ -1,42 +1,44 @@
...
-import { getFunctionImportParameters } from './action-function-parameters';
+import { EdmxActionImport, EdmxFunctionImportV4 } from '../../edmx-parser/v4';
+import { getOperationParameters } from './operation-parameter';
 
 /**
  * @internal
  */
-export function transformFunctionImportBase(
-  edmxFunctionImport: EdmxFunctionImportV2 | EdmxFunctionImportV4,
+export function transformOperationBase(
```
FrankEssenberger, line 14: [pp] Also here a JSdoc to explain that the operations represent funciton and actikns could help.

packages/generator/src/edmx-to-vdm/common/operation.ts
```diff
@@ -1,42 +1,44 @@
...
+  edmxOperation: EdmxFunctionImportV2 | EdmxFunctionImportV4 | EdmxActionImport,
```
FrankEssenberger, line 15: [req] I guess you can use the `EdmxOperation` here as well?

packages/generator/src/edmx-to-vdm/common/operation-parameter.ts
```diff
@@ -14,38 +14,11 @@ import { getTypeMappingActionFunction } from '../edmx-to-vdm-util';
...
-    swaggerDefinition,
-    formatter
-  );
-}
-
-function getParameter<T extends EdmxNamed>(
-  edmxActionFunctionImport: T,
+export function getOperationParameters(
+  edmxActionFunctionImport:
+    | EdmxFunctionImportV2
```
FrankEssenberger, line 19: [req] I guess you can use the `EdmxOperation` here as well?

packages/generator/src/edmx-to-vdm/common/operation-parameter.ts
```diff
@@ -14,38 +14,11 @@ import { getTypeMappingActionFunction } from '../edmx-to-vdm-util';
...
-    edmxParameters,
-    swaggerDefinition,
-    formatter
-  );
-}
-
-function getParameter<T extends EdmxNamed>(
-  edmxActionFunctionImport: T,
+export function getOperationParameters(
+  edmxActionFunctionImport:
```
FrankEssenberger, line 18: [pp] rename the parameter to edmxOperation

packages/generator/src/edmx-parser/v4/edm-types.ts
```diff
@@ -94,7 +94,7 @@ export interface EdmxActionImport extends EdmxNamed, EdmxNamespaced {
 /**
  * @internal
  */
-export interface EdmxFunction extends EdmxNamed, EdmxNamespaced {
```
FrankEssenberger, line 97: [q] I got confused. What is the different between the `EdmOperation` (was EdmxFunction` before and the `EdmxFunctionImport`? I guess there are two things in the Edmx?

FrankEssenberger: Perhaps it would make sense to create a `EdmxOperationImport`?

packages/generator/src/vdm-types.ts
marikaner: I like this idea. Just creating one file with all of them should be good enough.

packages/generator/src/edmx-parser/v4/edm-types.ts
FrankEssenberger: [req] I think Florian also stumbled over this. So I guess in v4 you have the Function and FunctionImport and we filter out cases where there is no import to the function right? I think it would make sense to add some doc here to explain that.

packages/generator/src/edmx-to-vdm/common/operation-parameter.ts
marikaner: I guess you mean `EdmxOperationImport`? If so, done.

packages/generator/src/edmx-parser/v4/edm-types.ts
marikaner: I added quite extensive documentation on how we now handle `EdmxOperationImport` and also added a note to the `EdmxOperation` that it is linked to `EdmxOperationImport`. Please let me know if this is good enough or if you would like it to be more detailed.

FrankEssenberger (approved):
LGTM - If we have now a `operations` folder in the generator I would in a follow up potential rename the functionImportV2 in the VDM also to operation to reduce the appearance of the function and align the two version of OData even more.

PR #3004, opened 2022-10-28
chore: Add automatically generated CDS config

marikaner:
It seems that the latest version of CAP needs this configuration and creates these files when running, which leads to untracked and changed files locally.


jjtang1985 (approved):
LGTM.
Could you please share the reference of this change?
I could not find it in the doc.

PR #3005, opened 2022-10-28
feature: Use prettier api

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#736.

Use the prettier API to format .ts and other files. Source files like `.js` are excluded because this could break the source maps. In the past we created the map files before formating `.ts` which was wrong. Now it is correct.

Hence there are many changes in the generated clients files.


packages/generator-common/src/compiler.ts
```diff
@@ -50,8 +53,39 @@ export async function transpileDirectory(
...
+    fileName,
+    text,
+    writeByteOrderMark,
+    onError?,
+    sourceFiles?,
+    data?
+  ) => {
+    const parsed = parse(fileName);
+    const promise = mkdir(parsed.dir, { recursive: true }).then(() => {
+      // All not emitted files like .md or .json should be already prettified.
```
marikaner, line 69: [nth] General: In the section above we mention prettying, linting and formatting - I would prefer to reuse one word.

packages/generator-common/src/compiler.ts
```diff
@@ -50,8 +53,39 @@ export async function transpileDirectory(
...
+      // Formatting emitted .js file could break source map -> skip these.
+      // Formatting the .map files also is not necessary increases only file size - only .d.ts files are nicer if linted
+      const usePrettier = getFileExtension(fileName) === 'd.ts' ? true : false;
+
+      // Prettier config should be cached from previous steps already.
+      // Emit also overwrites and copyright is either already there.
```
marikaner, line 75: [req] I do not understand this sentence ;) Can you rephrase it?

packages/generator-common/src/file-writer/create-file.ts
```diff
@@ -1,9 +1,138 @@
...
+
+/**
+ * This method considers also double dots like `.map.js`
+ * @param fileName
+ * @returns The complete file extension containing multiple dots
+ * @internal
+ */
+export function getFileExtension(fileName: string): string {
+  const fileType = parse(fileName).base.split('.').slice(1).join('.');
+  return fileType;
```
marikaner, line 104: [pp] Maybe return directly?
```suggestion
  return parse(fileName).base.split('.').slice(1).join('.');
```

packages/generator-common/src/file-writer/create-file.ts
```diff
@@ -1,9 +1,138 @@
...
+}
+
+async function formatWithPrettier(
+  fileName: string,
+  content: string,
+  prettierConfigPath: string | undefined
+) {
+  const prettierConfig = await readPrettierConfig(prettierConfigPath);
```
marikaner, line 112: [q] We read the config for every file (I know, that we have a cache). But wouldn't it be simpler to just read it once and add it to the options? That would also make this function synchronous.

packages/generator-common/src/file-writer/create-file.ts
```diff
@@ -1,9 +1,138 @@
...
+
+function addCopyrightHeader(content: string, withCopyright: boolean): string {
+  if (!withCopyright) {
+    return content;
+  }
+  return (
+    codeBlock`
+${getCopyrightHeader()}
+${content}
+` + unixEOL
```
marikaner, line 133: [req] I would put the `unixEOL` into the codeblock.
```suggestion
${content}
${unixEOL}
`
```
or:

```suggestion
${content}\n
`
```
[pp] After using it for a while, I see that I am not a fan of the `unixEOL`.  `\n` is just shorter, does not require escaping in strings and you don't need to import it. Maybe we could reconsider this.

packages/generator/package.json
```diff
@@ -50,6 +50,7 @@
     "fast-xml-parser": "^4.0.11",
     "fs-extra": "^10.0.0",
     "glob": "^8.0.1",
+    "prettier": "^2.7.1",
```
marikaner, line 53: [req] This seems unnecessary? I didn't find any use in the generator package, only generator-common.

packages/generator-common/src/file-writer/create-file.ts
```diff
@@ -1,9 +1,138 @@
...
+  /**
+   * Flag to indicate if copy write header is added.
+   */
+  withCopyright?: boolean;
+  /**
+   * Path to the prettier config with respect to the process.cwd().
+   */
+  prettierConfigPath?: string;
+  /**
+   * Flag to indicate if the file is prettified before.
```
marikaner, line 36: [req] Maybe we should avoid "prettier" as a verb and instead say "formatted with prettier".

marikaner (approved):
I like the new create options and I don't see any major blockers.
I wasn't sure if I got it right: When do we run the prettier? before transpilation or after? or both?

FrankEssenberger:
> I like the new create options and I don't see any major blockers. I wasn't sure if I got it right: When do we run the prettier? before transpilation or after? or both?

Now we run it before the `.ts` files are written so before the transpilation to `.js`. Which is correct. Before we did it after the transpilation simply on all files which would break the source maps.

packages/generator-common/src/file-writer/create-file.ts
FrankEssenberger: I also considered this. Since we have the options object now, it should not hurt I thing. Let me try that.

packages/generator/package.json
FrankEssenberger: ups, yes

marikaner:
> > I like the new create options and I don't see any major blockers. I wasn't sure if I got it right: When do we run the prettier? before transpilation or after? or both?
> 
> Now we run it before the `.ts` files are written so before the transpilation to `.js`. Which is correct. Before we did it after the transpilation simply on all files which would break the source maps.

But we also run it on the .d.ts files, so we run it twice?

packages/generator-common/src/compiler.ts
FrankEssenberger: removes because nonsense.

PR #3029, opened 2022-11-07
feat: setting batch ids and retrieving batch references from requests

shojisoma:
Closes SAP/cloud-sdk-backlog#814

This adds a new feature for setting and getting an identifier for a batch request.

#### Changes
as mentioned in the [ADR](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/adr/0036-batch-references.md), I added 2 the functions and also changed the place where the initial batch id is generated.
- add `getBatchReference()` for getting the ID in requests `getByKey`, `delete`,  `create`, `update`, and `bound/unbound functions/actions`(Note: adding `getBatchReference()` into `delete-request-builder` is not mentioned in ADR but there is a test case that needs the function [here](https://github.com/SAP/cloud-sdk-js/blob/09787ddd50c22bccb732f8abbd710ccbe544fe52/packages/odata-common/src/request-builder/batch/batch-request-serializer.spec.ts#L129))
- add `setBatchId()` for setting the ID in requests  `create`, `delete`, and `update` 
- change the place where batch id is generated from on the serialization to on the request creation


packages/odata-common/src/request-builder/action-function-import-request-builder-base.ts
```diff
@@ -15,14 +17,26 @@ export abstract class ActionFunctionImportRequestBuilderBase<
    * Base class for function  and actions imports.
    * @param responseTransformer - Transformation function for the response.
    * @param requestConfig - Request config for a action or function import.
+   * @param batchReference - Identifier for the batch request.
    */
   protected constructor(
     readonly responseTransformer: (data: any) => ReturnT,
-    requestConfig: RequestConfigT
+    requestConfig: RequestConfigT,
+    private batchReference: BatchReference = { id: uuid() }
```
FrankEssenberger, line 25: [q] Could you add the `batchReference` property to the `MethodRequestBuilder` class which is the parent for all the request builders? Same is true for the `getBatchReference` and `setBatchId` so that you do not have to duplicate the code multiple times?

packages/odata-common/src/request-builder/batch/batch-request-serializer.ts
```diff
@@ -75,6 +88,60 @@ export function serializeRequest(
...
 }
 
+type AllBuilderTypes<
+  EntityT extends EntityBase,
+  DeSerializersT extends DeSerializers,
+  ReturnT,
+  ParametersT
+> =
+  | Omit<MethodRequestBuilder, 'execute'>
+  | ActionFunctionImportRequestBuilderBase<
```
FrankEssenberger, line 98: [pp] You could create a type `RequestBuilderWithBatchReference` containing the Create,Delte and so on. Then you define the `AllBuilderType` as `RequestBuilderWithBatchReference | Omit<MethodRequestBuilder, 'execute'>` Then you have the long type only once.

FrankEssenberger (changes requested):
I have a question about the position of the getter and setter.

packages/odata-common/src/request-builder/action-function-import-request-builder-base.ts
shojisoma: I got the same idea and Marika planned that too. When that is placed in `MethodRequestBuilder`, we can reduce the duplicated codes across the `RequestBuilderBase`s, and also the complicated union type and type guard in the serializer can be removed because just `MethodRequestBuilder` is enough in the case.It's simple and looks better. 
However, Junjie suggested that all  `RequestBuilderBase`s don't need the `batchReference`, the getter, and the setter. `MethodRequestBuilder ` may be too broad for that.
@marikaner has already aware of these pros and cons, she will add comments.

FrankEssenberger: ok, sure. No matter what you decide I would like to define a interface `ReferencableInBatch` or something which defines the getter,setter methods and is added to the request builders under implements so that the methods are ways the same. Perhaps @marikaner can also comment if the init should be done in constructor or lazy in the get method.

jjtang1985:
Discussed with Shoji and we'll take over the open question and continue working on this.

packages/odata-common/src/request-builder/action-function-import-request-builder-base.ts
jjtang1985: Yes, I discussed with Shoji regarding the abstraction of `batchReference`.
There is a tradeoff as not all the request builders need it.
Unfortunately, Marika cannot provide feedback for now.
@FrankEssenberger , I checked the changes of this PR, and feel we can wait for Marika.
Even we create a v3 branch and this amount of files can still be handled and merged to two branches.
Let me know, if you want to merge before the v3 branch.

jjtang1985: I agree with Frank

FrankEssenberger: perhaps TS supports default implementation for interface finally, then we could safe even the coding.

packages/odata-common/src/request-builder/action-function-import-request-builder-base.ts
```diff
@@ -10,7 +12,9 @@ import { MethodRequestBuilder } from './request-builder-base';
 export abstract class ActionFunctionImportRequestBuilderBase<
   ReturnT,
   RequestConfigT extends ODataRequestConfig
-> extends MethodRequestBuilder<RequestConfigT> {
+> extends MethodRequestBuilder<RequestConfigT>  implements WithBatchReference{
+
+  private _batchReference: BatchReference = { id: uuid() };
```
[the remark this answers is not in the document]
FrankEssenberger: true, I wanted to keep the implementation of the get as short as possible to avoid code duplication. I had in each method then `if(!batchReference){this.setBatchReference(uuid())}` which I could not put in a helper method. Hence I decided to initialize on creation, which is I think acceptable because it is really a small object.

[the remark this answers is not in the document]
FrankEssenberger: was a leftover - removed

packages/odata-v4/src/request-builder/bound-function-import-request-builder.ts
```diff
@@ -2,8 +2,10 @@ import {
   ActionFunctionImportRequestBuilderBase,
   EntityApi,
   EntityBase,
-  FunctionImportParameters
+  FunctionImportParameters,
+  BatchReference
 } from '@sap-cloud-sdk/odata-common/internal';
+import { v4 as uuid } from 'uuid';
```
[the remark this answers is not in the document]
FrankEssenberger: was a leftover - removed

packages/odata-common/src/request-builder/action-function-import-request-builder-base.ts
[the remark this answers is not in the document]
jjtang1985: Understood.
The cost is about the number of calls to `uuid`, which is related to performance and memory.
I guess for normal cases, it might be fine.
For big number of big entities, it might be a problem, which happened before when the memory issue was raised.

I tried the following:
- The [mixin](https://codeburst.io/multiple-inheritance-with-typescript-mixins-d92d01198907) does partial work, and the methods of the parent class cannot be reused.
- This package [ts-mixer](https://www.npmjs.com/package/ts-mixer) seems to be helpful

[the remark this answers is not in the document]
jjtang1985: Here is the playground [test](https://www.typescriptlang.org/play?target=5&ts=4.1.5#code/MYGwhgzhAEBKCmBHArvCAXAQsgliAJvAE4DCA9gLYVkB2A3gLABQ008AHvMMuvABQBKRi2gBfZuKbNQkGAHUc6ABaYw6YEoQAzYawD6ReFoD8ALmgYiOGgHNoAXmjIahLdfj4A3M1Y346bUFdVmgcLT4AQmUcCAA6AyMhHxDWaLiErQdoAHJkXHxs5NZJYolmZmteIi0wYHg4JFQMTDxCUkpqGgVlVXVNIzZ2XhcYBBQ0LFwCYnIqWmEyqSYAemXoChx2a2h8MjRoGjJ0aAB3MiIAa2ZV6BkoaBJDNXgxpsnW4lCKAAcQeAp4DR0KNGhMWtN2nMuooVGoNNoFis1pJrmtkBB6sCALQbThEaBac6DYb4ax2CjIEDoPD1O4YiDScD3AAi8D+vFeYKmbWJgPw8hhvXhRkRKKYwFoGGgx0cNHgJ2grPZL1BzW5xEE3iRThcRnc+GkkrIf1iIDINj46Hihi0Ai1Vr8ASMmtR0AAbmAQKhQjRFDhPTgAF4eQ00CDG+Cm82W62JTxAA)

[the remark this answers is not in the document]
FrankEssenberger: ok then I make it lazy if we have the experience that there are many request build und UUID is called often.

FrankEssenberger (approved):

jjtang1985 (approved):
LTGM
Please merge after fixing the pipeline

fwilhe:
@FrankEssenberger @jjtang1985 

Hi, it looks like this PR introduced unused type parameters. Pipeline is green, but it generates warnings. Can you check if we can remove them without breaking anything?

Screenshot 2022-11-28 at 15 58 54

FrankEssenberger:
The method `serializeRequest` is marked as internal and the type `AllRequestBuilders` is not exported at all. Feel free to removed the unused types and open a PR.

PR #3038, opened 2022-11-08
chore: use node 16 for `get-changelog` action

devenes:
🚨 Node 12 has an end of life on April 30, 2022.

The GitHub Actions workflow gives the following warning while running the action:

> Node.js 12 actions are deprecated. For more information see: github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12.

This PR updates the default runtime to [node16](https://github.blog/changelog/2021-12-10-github-actions-github-hosted-runners-now-run-node-js-16-by-default/) rather then node12.

This is supported on all Actions Runners v2.285.0 or later.

Updated `actions/core` dependency:

- `"@actions/core":` `"^1.9.1"` --> `"^1.10.0"`

Signed-off-by: Enes <ahmedenesturan@gmail.com>


jjtang1985:
Hi @devenes ,

Thanks for raising this up.

The "get changelog" GH action is created for internal usage.
Are you using it somewhere?

Sure, we will update the version from 12 to e.g., 16.
My question is, why you touched 100+ files, which are not related.

devenes:
Hello @jjtang1985, thanks for your quick review! 

> The "get changelog" GH action is created for internal usage. Are you using it somewhere?

I don't use it regularly I just tested it for a few times. 

> Sure, we will update the version from 12 to e.g., 16. My question is, why you touched 100+ files, which are not related.

This PR actually aims to resolve the 2 different workflow warnings that sometimes broke pipelines. 

1. `Node.js 12 actions are deprecated`. 
		- And it can be solved by updating `action.yaml` 

2. `The set-output command is deprecated`.
		- This need to be solved by updating  dependency `@actions/core`  --> `"^1.10.0"`.
		- And updating the `@actions/core` creating required files in `node_modules` folder like uuid, http_client etc. 

I can revert the changes about packages and go to Node update solution only.

devenes:
> 100 irrelevant files are touched.

Reverted `core` package updates. 

Please let me know if you think of any method to get a better solution. 🚀

jjtang1985 (approved):
LGTM.

PR #3046, opened 2022-11-14
Convert code generation of request-builder to template-based approach

fwilhe:
Closes SAP/cloud-sdk-backlog#724.

Convert code generation of request-builder to template-based approach

Note: This PR introduces a few duplicate methods for import handling with changed types. This will be resolved once we get rid of the ts-morph dependency. Writing those methods in a generic way seems like unreasonable effort for an ongoing migration and would only save a handful of lines of code.


FrankEssenberger:
Danger I see the sqlite file checked in.

fwilhe:
> Danger I see the sqlite file checked in.

I think this is okay now. Marika recently made the PR to add the file to the repo and the tests don't seem to fail. I don't exactly know why the file has changed, but I looked at the diff [as described here](https://github.com/SAP/cloud-sdk-js/blob/v2-main/knowledge-base/how-tos/0008-diff-sqlite-db.md) and it seems some uninteresting schema change. That's why I committed the file as the e2e tests produced it.

packages/generator/src/generator-without-ts-morph/request-builder/class.ts
```diff
@@ -0,0 +1,247 @@
...
+      }
+    )}` +
+    unixEOL +
+    codeBlock`getAll(): GetAllRequestBuilder<${entity.className}<T>, T> {
+    return new GetAllRequestBuilder<${entity.className}<T>, T>(this.entityApi);
+  }`
+  );
+}
+
+function createRequestBuilder(entity: VdmEntity): string {
```
FrankEssenberger, line 90: [pp] you can deconstruct the className from the entity in the argument or directly only pass the class name in. Also applies to other methods.

FrankEssenberger: Perhaps destructuring in the argument is even nicer, because you have a short `className` var name but you see where it came from in the method declaration.

packages/generator/src/generator-without-ts-morph/request-builder/class.ts
```diff
@@ -0,0 +1,247 @@
...
+    delete(
+      ${deleteRequestBuilderParameters(entity)
+        .map(p => `${p.name}${p.hasQuestionToken ? '?' : ''}: ${p.type}`)
+        .join(', ')}
+    ): DeleteRequestBuilder<${entity.className}<T>, T> {
+      ${deleteRequestBuilderStatements(entity)}
+    }`;
+}
+
+interface deleteRequestBuilderParameter {
```
FrankEssenberger, line 189: [req] Types should start Capital

packages/generator/src/internal.ts
```diff
@@ -1,30 +1,29 @@
...
-export * from './request-builder';
-export * from './sdk-metadata';
-export * from './service';
-export * from './swagger-parser';
 export * from './file-generator';
+export * from './generator';
 export * from './generator-options';
 export * from './generator-utils';
-export * from './generator';
+// Had to be added for public API check
```
FrankEssenberger, line 12: [pp] perhaps also add a comment that the `generator-without-ts-morph` will go away after complete migration.

FrankEssenberger (approved):
LGTM I like the class.ts with a method for each builder. Easy to read and follow. Left some minor comments.

packages/generator/src/internal.ts
fwilhe: I don't remember touching this file, but it looks like this is done by the ide or linter.. anyway, I'll add a comment.

FrankEssenberger:
> > LGTM I like the class.ts with a method for each builder. Easy to read and follow. Left some minor comments.
> 
> Thanks for the review. I had to look up the destructuring syntax but I agree its better. I've used it in places now where we only use the className property. Not sure if we could do that even more aggressively in other places too?

I think this can be done in the boy scout way. It is not essential. Something is understandable either way. It is just a bit nicer some times.

PR #3047, opened 2022-11-14
fix: logging in registered destination look-up

tomfrenken:
Previously, the logging in the registerDestination lookup was always logging success, now it only logs a success if a destination was actually retrieved.


FrankEssenberger (approved):
LGTM as in meeting

PR #3055, opened 2022-11-16
Configure and document how to diff sqlite schema changes

fwilhe:
By default, git won't show how *binary* files have changed.
Since we have a sqlite file as part of the end to end tests, it might be useful to see how the schema of that database has changed.
This document describes how to configure git to show schema changes.


jjtang1985 (approved):
LGTM, by checking the original answer.
Question: have you tried?

PR #3056, opened 2022-11-16
chore: oauth2refreshtoken auth flow

deekshas8:
Relates SAP/cloud-sdk-backlog#884


FrankEssenberger (approved):
LGTM - I am very happy with the implementation. Also a mocked test for positive and negative are there and a sample in the autho-flow tests. Very good. I changed the PR form `closes` issue to `relates` because please also add a real E2E test in our nightly job and a follow up ticket to investigate and optimize for the potentially unused call to fetch the service token again.

PR #3058, opened 2022-11-16
fix: fix formatting bug

YuanZhuang107:
Closes SAP/cloud-sdk-backlog#953

Add "Service" in the reserved words list.


FrankEssenberger (approved):
LGTM

packages/generator/src/service-name-formatter.spec.ts
```diff
@@ -112,6 +112,7 @@ describe('name-formatter', () => {
       expect(formatter.originalToComplexTypeName('BigNumber')).toBe(
         'BigNumber_1'
       );
+      expect(formatter.originalToEntityClassName('Service')).toBe('Service_1');
```
FrankEssenberger, line 115: Just a general comment. It is always nice to have tests not to test different things so that when a test fails you directly see what was broken. In this case it is still fine but often it is good to create a second dedicated `it` step and not put the expect in an existing test.

YuanZhuang107: I see. Thanks for sharing :)

jjtang1985 (commented):
Sorry for late review.
Here are my comments.
- please also make a cherry pick to `v2-main` branch, as this is a bug, requested by a user
- you said the changelog was missing
  - please just create another PR for adding it to `main`
  - then cherry pick the commit to `v2-main`

PR #3082, opened 2022-11-21
Bootstrap version 3 upgrade guide

fwilhe:
Closes SAP/cloud-sdk-backlog#934.


V3-Upgrade-Guide.md
```diff
@@ -0,0 +1,35 @@
...
+
+## How to upgrade SAP Cloud SDK for JavaScript to Version 3
+
+This document collects information on how to upgrade Cloud SDK version 2 to version 3.
+If you encounter issues with the documentation or while upgrading, please [open an issue in our GitHub repository](https://github.com/SAP/cloud-sdk-js/issues/new/choose).
+
+## Update your project dependencies
+
+Search for occurrences of `@sap-cloud-sdk/[some module]` in your `package.json` files.
+Replace the `2.x.y` version numbers with the [most recent `3.x.y` version number](https://github.com/SAP/cloud-sdk-js/releases).
```
florian-richter, line 22:
```suggestion
Replace the version numbers with `^3`.
```
[req] https://github.com/SAP/cloud-sdk/blob/main/STYLEGUIDE.md#do-not-use-formats-xy-or-xyz

florian-richter: [pp] Can we supply a `sed` (and windows equivalent) command to replace this with a single command?

V3-Upgrade-Guide.md
```diff
@@ -0,0 +1,35 @@
...
+We recommend updating your applications in one commit or pull request and making sure everything still works using your existing test suite.
+
+## Generator CLI
+
+<!-- TODO: This is only meant as an example for sections in the upgrade guide. Improve this section and add new sections as you see fit. -->
+
+The SAP Cloud SDK includes two "generator" cli applications for OData and for OpenAPI clients.
+For historic reasons the command-line arguments of both applications were different in cases where this does not make sense.
+In version 3, the arguments are aligned and deprecated arguments have been removed.
+Please see (insert link here) for the current documentation on the cli arguments.
```
florian-richter, line 35:
```suggestion
<!-- TODO: This is only meant as an example for sections in the upgrade guide. Improve this section and add new sections as you see fit.

## Generator CLI

The SAP Cloud SDK includes two "generator" cli applications for OData and for OpenAPI clients.
For historic reasons the command-line arguments of both applications were different in cases where this does not make sense.
In version 3, the arguments are aligned and deprecated arguments have been removed.
Please see (insert link here) for the current documentation on the cli arguments.
-->
```
[pp] This avoids a problem if we forget to remove this later.

florian-richter (changes requested):

florian-richter (approved):
Approved but I would appreciate an answer to my [pp] comment to supply a bash one-liner to edit the package.json according to the upgrade guide.

V3-Upgrade-Guide.md
```diff
@@ -0,0 +1,35 @@
...
+It will eventually end up in the SDK docs portal and release notes for version 3.
+
+Please add you items below when creating a change which will involve manual tasks for the user when performing the upgrade.
+Add sections to the document as you see fit.
+
+<!-- Everything below this line should be written in the style of enduser documentation. If you need to add hints for SDK developers, to that above. -->
+
+## How to upgrade SAP Cloud SDK for JavaScript to Version 3
+
+This document collects information on how to upgrade Cloud SDK version 2 to version 3.
```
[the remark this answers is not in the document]
fwilhe: Oh it took me a while to understand what you meant, but now I got it. The metaphor of a 'todo-list' didn't seem too obvious to me, maybe we can find a better way to express and structure that, but I've taken it to v3 for now for consistencies sake. Thanks.

fwilhe:
> Approved but I would appreciate an answer to my [pp] comment to supply a bash one-liner to edit the package.json according to the upgrade guide.

Thanks. I did not forget about your comment, but I wanted to do some research on how to do this properly.

For example this is what you get by docusaurus:
Screenshot 2022-11-22 at 11 18 05
 
I guess we could also tell users to do `npm i @sap-cloud-sdk/foobar@latest`, but then they would need to know which sdk artifacts they use. I know this can be inspected with `npm ls` but that sounds a bit complex to describe.

I'm not sure if providing an `sed(1)` snippet will be of such big help. Is there a proper npm/yarn solution that works cross-platform?

florian-richter:
I mean this would be a great use-case for a CLI... sadly we discontinued our CLI long ago. Running the check on `npm install` is a bad practice and we don't have access to a `serve` command or similar.

Maybe we could provide a mini-package to be executed via `npx @sap-cloud-sdk/upgrade` or similar...

PR #3083, opened 2022-11-21
chore: make request config parameter optional in executeHttpRequest

deekshas8:
Relates https://github.com/SAP/cloud-sdk-backlog/issues/958

In `executeHttpRequest` function, the `url` property of `requestConfig` is optional. That means a user can make the following call (skipping the requestConfig url):

`await executeHttpRequest(destination!, { method: 'post', data: {} });`

But, will get the error when making Csrf token fetch request`:

TypeError: Cannot read properties of undefined (reading 'endsWith')

```
    at appendSlash (node_modules/sap-cloud-sdk-http-client-canary/src/csrf-token-header.ts:129:27)
    at makeCsrfRequest (node_modules/sap-cloud-sdk-http-client-canary/src/csrf-token-header.ts:85:42)
    at buildCsrfHeaders (node_modules/sap-cloud-sdk-http-client-canary/src/csrf-token-header.ts:35:29)
    at getCsrfHeaders (node_modules/sap-cloud-sdk-http-client-canary/src/http-client.ts:531:23)
    at addCsrfTokenToHeader (node_modules/sap-cloud-sdk-http-client-canary/src/http-client.ts:549:13)
    at node_modules/sap-cloud-sdk-http-client-canary/src/http-client.ts:123:29
```
~~One way to fix this would be add a check in [appendSlash method](https://github.com/SAP/cloud-sdk-js/blob/637a8ce4a4e70f4c5bc64ebeda73e9c0223557e9/packages/http-client/src/csrf-token-header.ts#L129) where if the requestConfig url is empty, we use the destination url instead. 
However, this might not be a good practice as the destination url should ideally not include the service url. Even in the odata/openapi clients, the url (relativeUrl) is [passed in the requestConfig to the executeHttpRequest method](https://github.com/SAP/cloud-sdk-js/blob/637a8ce4a4e70f4c5bc64ebeda73e9c0223557e9/packages/odata-common/src/request/odata-request.ts#L231).~~

~~This PR makes the `url` property in the `HttpRequestConfigBase` mandatory. **This will be a breaking change in v3**. This hasn't been reported yet by users in v2, so the assumption is that they invoke the `executeHttpRequest` function with destination url and serviceUrl correctly.~~

This PR makes the `requestConfig` property in the `executeHttpRequest` and `executeHttpRequestWithOrigin` optional and fixes the TypeError by handling the empty `url` property in the fetch csrf token function.


FrankEssenberger (commented):
LGTM - however perhaps you wait also for a feedback from a colleague. Also I changed the tag from closes to relates because I see two subtaks:
- adjust the docs - here I am not sure if we have the version 3 tag?
- check if the E2E tests need to be adjusted - here I am not sure if we have the next version tests?

packages/http-client/src/http-client-types.ts
```diff
@@ -78,7 +78,7 @@ export interface HttpRequestConfigBase {
    * Server URL that will be used for the request.
    * Relative `url` can be used together with {@link DestinationHttpRequestConfig#baseURL}.
    */
-  url?: string;
+  url: string;
```
jjtang1985, line 81: [q] This is a breaking change, let's discuss.

jjtang1985 (commented):
Please add a changelog

jjtang1985:
I would just touch this [line](https://github.com/SAP/cloud-sdk-js/blob/main/packages/http-client/src/csrf-token-header.ts#L129) and return '/', if the `url` is missing.

jjtang1985 (changes requested):
Let's have a call.

deekshas8:
> I would just touch this [line](https://github.com/SAP/cloud-sdk-js/blob/main/packages/http-client/src/csrf-token-header.ts#L129) and return '/', if the url is missing.

This would mean that the call is made against 'only' the destination url (as no service url is available). I tested this case with CCT (fetching csrf token), and got a `405 Method Not Allowed` error.

So we will avoid the TypeError but the overall request will still fail

jjtang1985 (approved):
Discussed.
This is a breaking change that will only be applied to v3.
Please consider the changeset.

marikaner (changes requested):
As discussed, I think the better solution would still be to keep the URL optional.

This is why:

1. What is the case for not having a URL? I assume in most cases if you go against an SAP System (OData or OpenAPI), where you will need a URL anyways if you want to make a modifying request (e.g., to the business partner you want to modify). In those cases it is irrelevant whether the URL is optional or mandatory, because it is needed anyways.

```ts
const destination = { url: 'https://some-system.com' };
// not a problem because we have a URL anyways
executeHttpRequest(destination, { method: 'post', url: '/sap/opu/bupa' } );
```

2. Let's assume you don't need the additional URL: In this case, whatever the user entered in the destination URL must accept POST requests. If the destination URL accepts POST and requires a CSRF token, I assume it also accepts CSRF fetch requests.

```ts
const destination = { url: 'https://some-system.com/sap/opu/bupa' };
// If POST does not need additional URL, why would CSRF fetch need an additional URL?
executeHttpRequest(destination, { method: 'post' } );
```
The example before might be bad practice, but there might be other non-SAP system examples that work exactly like this.
I argue, that the usability would suffer from a mandatory URL in this case, because users would have to add a useless url parameter with either "" or "/". This seems like a technical workaround for something that the SDK did not manage to solve internally.

2.b. Let's assume you don't need the URL and you actually just want to execute a GET request. In this case you don't even need the URL technically - otherwise same case as 2.


(3. There is no way (currently) to set a different URL for the actual POST request and the CSRF fetch request (with HEAD):

| POST URL? | CSRF Fetch URL? |
|---|---|
|true|true|
|false|false|
| The following CANNOT happen |
|true|false| 
|false|true|

So only the two cases I described exist.
)

FrankEssenberger:
@deekshas8 we had a discussion and I did not have the `GET` case on the radar which has the same options, but there is no error is the URL is left. We decided to make the full options object optional i.e. we have a default `GET` for method and handle the missing URL gracefully. Either with `/` as default value or using the baseURL is URL is not set.

FrankEssenberger (commented):
The PR description is outdated. Please update the description so that we have a clear history.

packages/http-client/src/csrf-token-header.spec.ts
```diff
@@ -172,6 +172,31 @@ describe('buildCsrfHeaders', () => {
...
+      'x-csrf-token': 'mocked-x-csrf-token',
+      'set-cookie': ['mocked-cookie-0;mocked-cookie-1', 'mocked-cookie-2']
+    };
+
+    nock(defaultHost)
+      .head(request.serviceUrl() + '/')
+      .reply(200, undefined, mockedHeaders);
+
+    const expected = {
+      cookie: 'mocked-cookie-0;mocked-cookie-2',
```
FrankEssenberger, line 190: [q] why is the `mocked-cookie-1` not set?

packages/http-client/src/csrf-token-header.ts
```diff
@@ -126,7 +126,9 @@ function getResponseHeadersFromError(error: any): Record<string, any> {
 }
 
 function appendSlash<T extends HttpRequestConfigBase>(requestConfig: T): T {
-  if (!requestConfig.url!.endsWith('/')) {
+  if (!requestConfig.url) {
+    requestConfig.url = '/';
+  } else if (!requestConfig.url!.endsWith('/')) {
```
FrankEssenberger, line 131: [q] is the second `!` needed. You would expect that we know `requestConfig.url` is defined because of the first `if`statement.

packages/http-client/src/http-client.ts
```diff
@@ -348,7 +351,7 @@ function logRequestInformation(request: HttpRequestConfig) {
  */
 export function executeHttpRequest<T extends HttpRequestConfig>(
   destination: DestinationOrFetchOptions,
-  requestConfig: T,
+  requestConfig?: T,
```
FrankEssenberger, line 354: [pp] I do not like two optional parameters in a row. I know the background that we do not want to mix our options with the ones from the http client, but now I would need a dummy object if I want to disable csrfTokenFetching which is I guess not possible because you would need to change the default method from `get` to `post`. So I guess it is fine.

packages/http-client/src/http-client.ts
```diff
@@ -401,10 +404,16 @@ export function executeHttpRequestWithOrigin<
   T extends HttpRequestConfigWithOrigin
 >(
   destination: DestinationOrFetchOptions,
-  requestConfig: T,
+  requestConfig?: T,
   options?: HttpRequestOptions
 ): Promise<HttpResponse> {
-  return execute(executeWithAxios)(destination, requestConfig, options);
+  const requestConfigWithOrigin =
```
FrankEssenberger, line 410: [req] This name is not good. Perhaps `requestConfigWithDefaults` - perhaps you can also do a `{...getDefault(),...requestConfig} This would mean the config get all the default values.

packages/http-client/src/http-client.ts
```diff
@@ -487,6 +496,15 @@ export function getAxiosConfigWithDefaultsWithoutMethod(): Omit<
   };
 }
 
+/**
+ * @internal
+ */
+export function getDefaultHttpRequestConfigOptions(): HttpRequestConfigWithOrigin {
+  return {
+    method: 'get'
```
FrankEssenberger, line 504: [q] Could one perhaps add `url:'/'` here so that no if statement is needed in the csrf-token-header?

packages/http-client/src/http-client.ts
FrankEssenberger: I guess since the method is mandatory this does not make a difference.

packages/http-client/src/http-client.ts
FrankEssenberger: I think to make it 100% correct we would have to make a conditional type for the RequestConfig: If the method is post the URL is mandatory. If the method is not POST the URL is optional in the config.

FrankEssenberger (approved):
LGTM - I thought about it longer and I guess it is best the way it is. If people do post with a destination containing a full path it will work - no URL needed which is convenient.

packages/http-client/src/csrf-token-header.spec.ts
deekshas8: The logic [here](https://github.com/SAP/cloud-sdk-js/blob/c370baa7b7c919969b895be13b1f822db4991569/packages/http-client/src/csrf-token-header.ts#L158) only picks the first string from each item in the array.

packages/http-client/src/csrf-token-header.ts
deekshas8: Yes, correct. I fixed it now.

packages/http-client/src/http-client.ts
deekshas8: I adjusted the variable name to `requestConfigWithDefaults` .
The idea for `requestConfig ?? getDefaultHttpRequestConfigOptions();` was to avoid the getDefault function altogether since the method property is mandatory.

PR #3102, opened 2022-11-24
fix: Fix warnings for deprecated options

marikaner:
I noticed that currently we ALWAYS show a warning that `generateNpmrc` is deprecated. This PR fixes this to only happen when the option is used.


fwilhe (approved):

marikaner:
@FrankEssenberger I'll go ahead and enable auto-merge. There will be a second PR for duplicates, in case you have further feedback, I can add it then.

PR #3103, opened 2022-11-24
chore!: Remove deprecated api

fwilhe:
Closes SAP/cloud-sdk-backlog#938.

### DIFF
--- a/V3-Upgrade-Guide.md
+++ b/V3-Upgrade-Guide.md
@@ -18,6 +18,7 @@ Add sections to the document as you see fit.
 This document will guide you through the steps necessary to upgrade to version 3 of the SAP Cloud SDK. Depending on your project, some steps might not be applicable. The To-Do list is:
 
 - [Update your project dependencies](#update-your-project-dependencies)
+- [Check for removed deprecated functions and replace them if required](#check-for-removed-deprecated-functions-and-replace-them-if-required)
 
 ### Update your project dependencies
 
@@ -27,6 +28,24 @@ Depending on if you're using `npm` or `yarn`, run `npm install` or `yarn` in the
 
 Running your tests or deploying your application might fail at this point in time if you need to adapt to any breaking changes.
 We recommend updating your applications in one commit or pull request and making sure everything still works using your existing test suite.
+
+### Check for removed deprecated functions and replace them if required
+
+While the SAP Cloud SDK maintains backwards compatibility within a major version where possible, a new major release breaks compatibility where required to simplify the programming interface.
+Most of the removed functions had been deprecated before, so ideally they are not used anymore.
+The following sub-sections describe affected modules, functions and interfaces with instructions on how to replace them.
+
+#### HTTP Client
+
+The overload, that accepted `HttpRequestConfigWithOrigin` as a parameter, is removed and replaced by the function `executeHttpRequestWithOrigin`.
+
+#### Logger
+
+The field `logger` on the interface `LoggerOptions` was not used and is removed from the interface.
+
+#### Array Utils
+
+The function `variadicArgumentToArray` is replaced by the function `transformVariadicArgumentToArray`.
 
 <!-- TODO: This is only meant as an example for sections in the upgrade guide. Improve this section and add new sections as you see fit.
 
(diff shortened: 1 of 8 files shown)


scripts/current-sdk-version.ts
```diff
@@ -1,5 +1,7 @@
 import { readFileSync } from 'fs';
 
+//todo(fwilhe): should this be removed as part of v3? Seems like non-user facing anyway.
```
FrankEssenberger, line 3: yes from the text it looks like it. Do we use this somewhere?

fwilhe:
Remaining occurrences of the term 'deprecated' are either related to a BTP authentication thing (which I guess we can't just remove), or relate to open api / npm packages (which is not related to this PR as it is not deprecated code).

Also I've left out the generator options, not sure if this is part of the scope of this PR.

Apart from that I've quite aggressively removed deprecated stuff, in-depth review by people who have better context-knowledge on the individual code pieces are required.

fyi @marikaner @FrankEssenberger @florian-richter

FrankEssenberger (approved):

packages/http-client/src/http-client.ts
```diff
@@ -351,20 +350,6 @@ export function executeHttpRequest<T extends HttpRequestConfig>(
...
-// eslint-disable-next-line jsdoc/require-returns-check
-/**
- * Builds a {@link DestinationHttpRequestConfig} for the given destination, merges it into the given `requestConfig` and executes it (using Axios).
- * @deprecated This overload is replaced by the function {@link executeHttpRequestWithOrigin}.
- * @param destination - A destination or a destination name and a JWT.
- * @param requestConfig - Any object representing an HTTP request.
- * @param options - An {@link HttpRequestOptions} of the HTTP request for configuring e.g., CSRF token delegation. By default, the SDK will fetch the CSRF token.
- * @returns A promise resolving to an {@link HttpResponse}.
- */
-export function executeHttpRequest<T extends HttpRequestConfigWithOrigin>(
```
deekshas8, line 363: Since this overload is removed, we should also remove the other declaration and only have the function where `T extends HttpRequestConfig`.

fwilhe: Yes, you are right, thanks for the hint.

packages/odata-common/src/entity-builder.spec.ts
```diff
@@ -154,18 +154,16 @@ describe('EntityBuilder', () => {
       expect(entity).toStrictEqual(expectedEntity);
     });
 
-    it('should build an entity from json with legacy _customFields', () => {
+    it('should build an entity from json with custom fields', () => {
```
deekshas8, line 157: Possible duplicate of test in line 105

fwilhe: thanks, fixed

V3-Upgrade-Guide.md
```diff
@@ -28,6 +29,57 @@ Depending on if you're using `npm` or `yarn`, run `npm install` or `yarn` in the
...
+
+The field `logger` on the interface `LoggerOptions` was not used and is removed from the interface.
+
+#### Array Utils
+
+The function `variadicArgumentToArray` is replaced by the function `transformVariadicArgumentToArray`.
+
+#### JWT
+
+The generic types of `JwtKeyMapping` is simplified so the second type argument `JwtKeysT` are always strings.
```
jjtang1985, line 52: [req] The function name is listed, which looks good.
However, the section title is called "JWT", I as a user would not know whether this is related to a package that I'm using.
Instead of using "JWT", I would prefer using the npm package name. (@sap-cloud-sdk/connectivity?)

jjtang1985 (changes requested):
Looks good.
Please consider my comments about the structure of the upgrade guide.

V3-Upgrade-Guide.md
fwilhe: Hi @jjtang1985 

thanks for your feedback, I think having npm package names as the first structure element makes sense. For the odata specific things I'm not sure what the second structure element should be, but I feel like we need a structure there because the text is quite long. I'm happy to take any suggestion you might have to improve it.

Please have another look.

jjtang1985 (approved):
LGTM

jjtang1985:
Maybe another minor comment: is it intended to use `chore!` as the PR prefix?

FrankEssenberger:
> Maybe another minor comment: is it intended to use `chore!` as the PR prefix?

I guess on @deekshas8 keyboard the `:` and `!` key are close?

fwilhe:
> > Maybe another minor comment: is it intended to use `chore!` as the PR prefix?
> 
> I guess on @deekshas8 keyboard the `:` and `!` key are close?

no, use en-US keyboard, the colon is on shift-O Umlaut and the bang is on shift-1

that was intentional as @deekshas8 pointed out :)

PR #3125, opened 2022-11-29
chore!: Make constructor private

FrankEssenberger:
In a [support issuse](https://answers.sap.com/questions/13764308/sf-api-select-using-navigation-property.html) a customer used the constructor to get a API object. This is not good, since the init step is not invoked. 

This PR makes the constructor private to protect users (at least if they use TypeScript or a IDE which understands the private keyword).

The static method I introduced is called `_privateFactory` - here I am up for any good suggestion.


fwilhe (approved):

.changeset/loud-radios-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/odata-common': patch
```
jjtang1985, line 2:
```suggestion
'@sap-cloud-sdk/odata-common': major
```

.changeset/loud-radios-arrive.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+[Compatibility Note] The constructor of the entity API is private to avoid accidental usage. Use the service object to get a API instance.
```
jjtang1985, line 5: [q] Should we update the "upgrade guide"?

packages/generator/src/generator-without-ts-morph/entity-api/class.ts
```diff
@@ -25,10 +25,18 @@ export function classContent(
...
   public deSerializers: DeSerializersT;
 
-  constructor(
+  private constructor(
     deSerializers: DeSerializersT = defaultDeSerializers as any) {
     this.deSerializers = deSerializers;
   }
+  
+  ${documentationBlock`Do not use this method or the constructor directly.
+Use the service object to get a API instance.`}
```
jjtang1985, line 34:
```suggestion
Use the service object to get an API instance.`}  
```

jjtang1985 (commented):
Looks good.
As this is not refined, maybe ask for more review.
I left some comments.

Other alternatives:
- lint rule for checking the constructor use case (lint and the rules are required)
- runtime error that shows the wrong usage (over engineering?)

PR #3126, opened 2022-11-29
chore: Configure dependabot

fwilhe:
Ignore major dependabot updates for chalk, execa and axios because new versions require ESM which we don't use yet.


FrankEssenberger (approved):

FrankEssenberger:
Question: How can we remember to remove this when we migrate to ESM? If there is a ticket for this task perhaps we can add this PR so that we revert this when possible.

PR #3137, opened 2022-11-30
chore: Align generator options

YuanZhuang107:
Deprecate and add options with compatibility notes, reflect the updated names

Closes SAP/cloud-sdk-backlog#506.


florian-richter (commented):
It is unclear to me how the new and the deprecated flags are merged? From what I see, the deprecated flags are just ignored in the code. Is that correct? Is there any handling that would throw an error if both (new and old) flags are set at the same time (especially if they have different values)?

.changeset/shaggy-yaks-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/generator': minor
+---
+
+[Compatibility Notes] Deprecate `writeReadme` option. You can use the new option `readme` to set if the generator will write README.md file instead.
```
FrankEssenberger, line 5: I am not sure if you need separate files for each note. @florian-richter I guess the changeset will handle multiple entries here?

florian-richter: Not multiple entries per file but you can use a markdown list as you see here: https://github.com/SAP/cloud-sdk-backlog/wiki/Changesets#best-practices

.changeset/beige-rocks-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/generator': minor
+---
+
+[Compatibility Notes] Deprecate `processesJsGeneration` option. Use the new hidden option `transpilationProcesses` to set number of processes for generation instead.
```
marikaner, line 5: [req] This is hidden and therefore not documented behavior.
```suggestion
[Compatibility Notes] Deprecate the `processesJsGeneration` option.
```

.changeset/clever-apes-know.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/generator': minor
+---
+
+[Compatibility Notes] Deprecate `generatePackageJson` option. You can use the new option `packageJson` to set if the generator will generate a package.json file instead.
```
marikaner, line 5:
```suggestion
[Compatibility Notes] Deprecate the `generatePackageJson` option. Use `packageJson` instead.
```

.changeset/shaggy-yaks-shop.md
marikaner, line 5:
```suggestion
[Compatibility Notes] Deprecate the `writeReadme` option. Use `readme` instead.
```

.changeset/strange-schools-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/generator': minor
+---
+
+[Compatibility Notes] Deprecate `licenseInPackageJson` option. Use the `include` option to add your own package.json file instead.
```
marikaner, line 5:
```suggestion
[Compatibility Notes] Deprecate the `licenseInPackageJson` option. Use `include` instead.
```

.changeset/witty-snails-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/generator': minor
+---
+
+[Compatibility Notes] Deprecate `forceOverwrite` option. Use the new option `overwrite` to set if the generator will overwrite existing files instead.
```
marikaner, line 5:
```suggestion
[Compatibility Notes] Deprecate the `forceOverwrite` option. Use `overwrite` instead.
```

packages/generator/src/generator-options.ts
```diff
@@ -32,20 +32,32 @@ export interface GeneratorOptions {
...
   /**
    * Exit when encountering a file that already exists.
    * When set to true, it will be overwritten instead.
    * Please note that compared to the `clearOutputDir` option, this will not delete outdated files.
    */
   forceOverwrite: boolean;
+  /**
+   * Exit when encountering a file that already exists.
+   * When set to true, it will be overwritten instead.
+   * Please note that compared to the `clearOutputDir` option, this will not delete outdated files.
```
marikaner, line 58:
```suggestion
   * Note, that compared to the `clearOutputDir` option, this will not delete outdated files.
```

florian-richter (approved):
LGTM, but I think another review would be good.

marikaner:
> LGTM, but I think another review would be good.

I reviewed it, triggered through your addition of `packageJson` I found that there was no link between `generatePackageJson` and the new `packageJson`. The replaceBy was missing, I added it for all the options.

PR #3140, opened 2022-12-01
fix: Fix parsing when using deprecated options

marikaner:
This PR fixes and improves the following cases:
- If an option was deprecated replaced by another one and a user sets the old option, set the value to the new option instead.
- If both options were set ignore the old value.
- If both options were set, show a warning.
- Handle setting default values ourselves as this can be reused for CLI and programmatic use.
- Introduce a `ParserOptions` type where all options are required, and depreceted, replaced options are excluded.

Closes SAP/cloud-sdk-backlog#986.

- [ ] If my change will be merged into the `main` branch (for v3), I've updated [V3-Upgrade-Guide.md](./V3-Upgrade-Guide.md) in case my change has any implications for users updating to SDK v3


packages/generator/src/generator-cli.ts
```diff
@@ -46,9 +48,15 @@ export function parseCmdArgs(): GeneratorOptions {
...
     .strict(true)
-    .recommendCommands().argv as unknown as GeneratorOptions;
+    .parserConfiguration({
+      'strip-aliased': true,
+      'strip-dashed': true
+    })
+    .recommendCommands();
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { _, $0, ...userOptions } = parsedOptions.argv as any;
```
FrankEssenberger, line 59: [q] is this part of the stable yargs command - test is definatly needed to see if this breaks.

packages/generator/src/generator-options.ts
```diff
@@ -176,6 +195,7 @@ export const generatorOptionsCli: KeysToOptions = {
     describe:
       'By default, the generator will exit when encountering a file that already exists. When set to true, it will be overwritten instead. Please note that compared to the --clearOutputDir option, this will not delete outdated files.',
     type: 'boolean',
+    deprecated: 'Since now. Use overwrite instead.',
```
FrankEssenberger, line 198: [q] Should we put the next version like 2.11.1 ior so?

packages/generator/src/generator-options.ts
```diff
@@ -200,7 +221,8 @@ export const generatorOptionsCli: KeysToOptions = {
       'By default, when generating package.json file, the generator will set a version by using the generator version. It can also be set to a specific version.',
     type: 'string',
     deprecated:
-      "Since v2.6.0. Use the 'include' option to add your own package.json file instead."
+      "Since v2.6.0. Use the 'include' option to add your own package.json file instead.",
+    replacedBy: 'packageVersion'
```
FrankEssenberger, line 225: [q] This I do not understang. The replace is `packageVersion` and in the sentence above it is mentioned to use `include`.

packages/generator/src/generator-options.ts
```diff
@@ -93,34 +89,58 @@ export interface GeneratorOptions {
...
+ */
+export type ParsedGeneratorOptions = ParsedOptions<
+  GeneratorOptions,
+  typeof generatorOptionsCli
+>;
+
+/**
+ * @internal
+ */
+export const generatorOptionsCli = {
```
FrankEssenberger, line 139: Is `inputDir`not also deprecated?

packages/generator/src/generator-options.ts
FrankEssenberger: [q] Should there be a `replacedBy` also for this one?

FrankEssenberger: In general it seems that a lot of deprecation changes are missing.

packages/generator/src/generator.ts
```diff
@@ -76,33 +78,38 @@ export async function generate(options: GeneratorOptions): Promise<void> {
  * TODO: Actually split the parsing of options between CLI and programmatic options. Currently users have to provide everything.
  */
 export async function generateWithParsedOptions(
-  options: GeneratorOptions
+  options: ParsedGeneratorOptions
```
FrankEssenberger, line 81: [q] If I got it right are the parsedOptions: All options required and the deprecated omitted. So I guess the TODO could go to the `generate` function - or could you make all things beside input/output optional anyhow?

packages/generator/src/generator-cli.ts
marikaner: The two parameters `_` and `$0` are part of the yargs API and I think this is very stable. However, those two parameters are not used. What I am doing here is basically remove them from the object - if they are not there, they are undefined, but we don't care.

I just investigated a bit: the problem is that `.argv` could be a promise. I would suggest to take care of this in the follow up, when we [improve the API with `GeneratorOptions`](https://github.com/SAP/cloud-sdk-backlog/issues/980). I partially started it here, but don't want to do everything in one PR.

packages/generator/src/generator-options.ts
marikaner: Sorry, this was just for a manual test and was not supposed to be here. There is another PR for the deprecation [here](#3137).

packages/generator/src/generator-options.ts
marikaner: Yes, this is confusing. Technically, the replacement for this is `packageVersion`. We moved this to a hidden parameter `packageVersion` because it was useful to us for the pregenerated clients, but most likely not relevant for users.

However, I now see the issue that if you do batch processing, `include` is not an adequate replacement, because you would always add the same package.json for all services...

packages/generator/src/generator.ts
marikaner: I'll just remove it, this was added before the other ticket was opened.

packages/generator/src/generator-options.ts
```diff
@@ -243,14 +264,14 @@ export const generatorOptionsCli: KeysToOptions = {
...
+    // default: false,
     hidden: true
   },
   generateCSN: {
     describe:
       'When set to true a CSN file will be generated for each service definition in the output directory.',
-    type: 'boolean',
-    default: false
+    type: 'boolean'
+    // default: false
```
marikaner, line 274:
```suggestion
    default: false
```

packages/generator/src/generator-options.ts
```diff
@@ -243,13 +264,13 @@ export const generatorOptionsCli: KeysToOptions = {
     describe:
       'When set to true, the description of the generated packages will be specific to S/4HANA Cloud.',
     type: 'boolean',
-    default: false,
+    // default: false,
```
marikaner, line 267:
```suggestion
    default: false,
```

packages/generator/src/generator-options.ts
```diff
@@ -249,7 +270,7 @@ export const generatorOptionsCli: KeysToOptions = {
   generateCSN: {
     describe:
       'When set to true a CSN file will be generated for each service definition in the output directory.',
-    type: 'boolean',
+    type: 'boolean'
```
marikaner, line 273:
```suggestion
    type: 'boolean',
```

test-packages/e2e-tests/test/generator-cli.spec.ts
```diff
@@ -122,7 +122,10 @@ describe('generator-cli', () => {
         '--generateNpmrc'
       ])
     ).rejects.toThrow(
-      /Deprecated options used.*\n\t--generateNpmrc:.*\n\t--versionInPackageJson:/
+      expect(err.stdout).toContain(
+        "The option 'generateNpmrc' is deprecated since v2.8.0."
+      );
+    }
```
marikaner, line 128:
```suggestion
      /Deprecated options used.*\n\t--generateNpmrc:.*\n\t--versionInPackageJson:/
```

packages/generator/src/options-parser.ts
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * @internal
+ * Remove defaults from CLI options. This is necessary to handle default setting on our own.
+ * @param options - CLI options, record mapping option name to option config
+ * @returns CLI options without default values
+ */
+export function getOptionsWithoutDefaults<
+  CliOptionsT extends Record<string, Option>
+>(options: CliOptionsT): CliOptionsT {
+  return Object.entries(options).reduce((opts, [name, option]) => {
```
FrankEssenberger, line 20: [pp] perhaps use a longer name than `ops`

packages/generator/src/options-parser.ts
```diff
@@ -0,0 +1,174 @@
...
+    const { default: def, ...optionWithoutDefault } = option;
```
FrankEssenberger, line 21: [q] why don't you do `const { default, ...optionWithoutDefault } = option;`

packages/generator/src/options-parser.ts
```diff
@@ -0,0 +1,174 @@
...
+  }, {} as CliOptionsT);
+}
+
+/**
+ * @internal
+ * Parse options for programmatic and CLI use.
+ * Warn, if deprecated options are used or duplicate.
+ * Adds defaults if unset.
+ * @param options - Available CLI options along with their configuration.
+ * @param userOptions - Options as set by user.
```
FrankEssenberger, line 42: [nth] This can be either from the CLI input or the programmatic call right? Perhaps you mention this.

packages/generator/src/options-parser.ts
```diff
@@ -0,0 +1,174 @@
...
+ * @returns Parsed options with default values.
+ */
+export function parseOptionsWithDefaults<
```
FrankEssenberger, line 45: [pp] I would remove the `withDefaults` from the method name and include it in the return type. The method you called is gone after the invocation. However, the types remains and would be `ParseOptionsWithDefault` you could perhaps even argue the this is clear since all properties are not optional on the parse type.

packages/generator/src/options-parser.ts
```diff
@@ -0,0 +1,174 @@
...
+  constructor(
+    private options: CliOptionsT,
+    private userOptions: GeneratorOptionsT
+  ) {}
+
+  parseOptionsWithDefaults(): ParsedOptions<GeneratorOptionsT, CliOptionsT> {
+    this.warnIfDeprecatedOptionsUsed();
+    this.warnIfDuplicateOptionsUsed();
+    const parsedOptions = this.sanitizeIfReplacedOptionsUsed();
+    return this.setDefaults(parsedOptions);
```
FrankEssenberger, line 68: [pp] When I see a `this.setXXX` I assume that the state of the class is changed. In this case it is not so I would perhaps use `addDefaults()`

packages/generator/src/options-parser.ts
```diff
@@ -0,0 +1,174 @@
...
+  CliOptionsT extends Record<string, Option>,
+  GeneratorOptionsT extends Record<string, any>
+>(
+  options: CliOptionsT,
+  userOptions: GeneratorOptionsT
+): ParsedOptions<GeneratorOptionsT, CliOptionsT> {
+  return new OptionsParser(options, userOptions).parseOptionsWithDefaults();
+}
+
+class OptionsParser<
```
FrankEssenberger, line 55: [pp] There are two arguments maximum: CLI and user options. There you could create even a type if they are alwyas both needed. I think you could also pass the two around and use functions instead of the class. You moved the `getOptionsWithoutDefaults` already out of the class for testing.

FrankEssenberger (approved):
LGTM - a lot of PP but I guess it is a good starting point for @YuanZhuang107

packages/generator/src/options-parser.ts
```diff
@@ -0,0 +1,174 @@
...
+ * Remove defaults from CLI options. This is necessary to handle default setting on our own.
+ * @param options - CLI options, record mapping option name to option config
+ * @returns CLI options without default values
+ */
+export function getOptionsWithoutDefaults<
+  CliOptionsT extends Record<string, Option>
+>(options: CliOptionsT): CliOptionsT {
+  return Object.entries(options).reduce((opts, [name, option]) => {
+    const { default: def, ...optionWithoutDefault } = option;
+    const describe =
```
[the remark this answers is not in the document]
marikaner: I should have mentioned this in the description of the PR. As we do the default value handling now, yargs only has values without defaults in the beginning, therefore I am adding the default value to the description if available.

packages/generator/src/options-parser.ts
marikaner: I think `ParsedOptions` already implies that the defaults are in there, because this is the result of the parsing. I can remove this from the name and mention it only in the comment. I think that is legit.

packages/generator/src/options-parser.ts
marikaner: I changed a lot of things over time, I'll try this out 😊

PR #3157, opened 2022-12-05
chore: Add PR checks

marikaner:
Add checks on PRs. The checks run on changes to the PR metadata and on external PRs.
It checks:
- cc is used
- title is upper case
- there is a body and it is not the template
- there is a changelog that fits the cc preamble

Closes SAP/cloud-sdk-backlog#926.


YuanZhuang107 (commented):
I have a question regarding `.github/actions/check-pr/node_modules/`. Should this directory be ignored, or is it intended to include them?

marikaner:
> I have a question regarding `.github/actions/check-pr/node_modules/`. Should this directory be ignored, or is it intended to include them?

This is intended. For actions this is required unfortunately.

fwilhe:
> This is intended. For actions this is required unfortunately.

I've never worked with actions inline in the repo, but would not be building the action using [ncc](https://github.com/actions/setup-node/blob/main/package.json#L8) be a way to avoid that? this produces one big inlined js artifact which can be executed without node modules..

fwilhe:
Do we need the test files in the changeset dir?

.github/actions/check-pr/index.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+function parseTitle(title: string): ParsedTitle {
+  const groups = title.match(
+    /(?<preamble>\w+)(\((?<topic>\w+)\))?(?<isBreaking>!)?: (?<description>.*)/
+  ).groups;
+  if (groups) {
+    return { ...groups, isBreaking: !!groups.isBreaking };
+  } else {
+    throw new Error('Could not parse PR title');
```
fwilhe, line 21: not an regex expert, but could we print more information to help understand what's wrong?

.github/actions/check-pr/index.ts
```diff
@@ -0,0 +1,120 @@
...
+function validateTitle(title: string | undefined): void {
+  if (!title) {
+    setFailed(
+      `PR title does not have a title (after conventional commit preamble).`
+    );
+  }
+
+  if (title[0] === title[0].toLowerCase()) {
+    setFailed(
+      `PR title title should be capitalized (after conventional commit preamble).`
```
fwilhe, line 46:
```suggestion
      `PR title title must be capitalized (after conventional commit preamble).`
```

if we throw an error it's not 'should'

.github/actions/check-pr/index.ts
```diff
@@ -0,0 +1,120 @@
...
+
+async function validateBody() {
+  const body = context.payload.pull_request.body.replace(/\r\n/g, '\n');
+  const prTemplate = await readFile(
+    resolve('.github', 'PULL_REQUEST_TEMPLATE.md'),
+    'utf-8'
+  );
+
+  if (!body || body === prTemplate) {
+    setFailed('PR should have a description');
```
fwilhe, line 103:
```suggestion
    setFailed('PR must have a description');
```

fwilhe (commented):
minor comments

marikaner:
> ncc

I think this is a nice improvement. But I would prefer to separate this from this PR, given that we have another action, that I would probably have to touch as well. I will create a ticket.

.github/actions/check-pr/package-lock.json
```diff
@@ -0,0 +1,431 @@
+{
+  "name": "check-pr",
```
florian-richter, line 2: [q] Why is this a `package-lock.json` and not `yarn.lock`?

.github/actions/check-pr/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "check-pr",
```
florian-richter, line 2: [q] Could this be in a separate repo and just be released to NPM (and then called with `npx` or as a github action step) to avoid cluttering the SDK repo?

.changeset/other-test.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/generator': minor
+---
+
+[Compatibility Notes] Deprecate the `generatePackageJson` option. Use `packageJson` instead.
```
florian-richter, line 5: I assume these test changeset files will be removed before the PR is merged. Otherwise I think they will pollute our release notes...

florian-richter (commented):
I am not a fan of putting all these files into our repo... hopefully there are alternative solutions.

marikaner:
> Do we need the test files in the changeset dir?

Nope, I removed them now. Good catch 👍

.github/actions/check-pr/index.ts
marikaner: I changed this quite a bit, I hope it is more readable now.

.github/actions/check-pr/package-lock.json
marikaner: I just did the same thing as with the other action. I think yarn might not be supported in GH actions, but I didn't want to try on this PR...

.github/actions/check-pr/package.json
marikaner: It could, but in my opinion this would be major overkill...

.changeset/other-test.md
marikaner: Yes, I forgot to remove them before.

.github/actions/check-pr/package-lock.json
fwilhe: I'm fine with npm, just wanted to mention that its possible to use yarn for actions, I've done so on my [personal open source project](https://github.com/fwilhe2/setup-kotlin)

marikaner: I guess, if we are using yarn it makes sense to do this here as well. Still, I prefer to do this in a separate PR to adjust the other action that we have as well. I created a ticket to checkout ncc, I will add this as well.

marikaner: => https://github.com/SAP/cloud-sdk-backlog/issues/995

.github/actions/check-pr/package.json
marikaner: Yes, they are, example => https://github.com/SAP/cloud-sdk-js/pull/2776. I ❤️ GH.

marikaner: I can add this publishing idea to the ticket. But I really would like to avoid additional complexity. Commiting node_modules is ugly, but not complex. Making changes in a separate repo and publishing is nicer, but much more complex.

fwilhe (approved):
LGTM

YuanZhuang107 (approved):
LGTM. Learned more about the rules of PR btw :)

PR #3159, opened 2022-12-05
Configure ci for sdk v3

fwilhe:
Configure CI runs for `main` and `v2-main` branches separately. This allows us to only run nodejs 18 for main where we don't need to care for older versions and allows us to do other tweaks to the build as long as v3 is not yet released.

### DIFF
--- a/.github/workflows/build-v3.yml
+++ b/.github/workflows/build-v3.yml
@@ -51,7 +51,7 @@ jobs:
             node_modules
             */*/node_modules
           key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
-      - run: yarn install --frozen-lockfile
+      - run: yarn install --ignore-engines --frozen-lockfile
       - run: yarn lint
         name: Static Code Check
       - run: yarn check:test-service
@@ -83,5 +83,5 @@ jobs:
             node_modules
             */*/node_modules
           key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
-      - run: yarn install --frozen-lockfile
+      - run: yarn install --ignore-engines --frozen-lockfile
       - run: yarn test:e2e
--- a/.github/workflows/build-v3.yml
+++ b/.github/workflows/build-v3.yml
@@ -1,0 +1,87 @@
+name: build v3
+
+on:
+  pull_request:
+    branches: ['main']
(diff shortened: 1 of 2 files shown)


FrankEssenberger:
I am confused. Let's have a call tomorrow. Why is the `v3` script merged in the `v2-main` branch?

.github/workflows/build-v3.yml
```diff
@@ -0,0 +1,87 @@
...
+  pull_request:
+    branches: ['main']
+  push:
+    branches: ['main']
+    paths-ignore:
+      - 'docs/**'
+  schedule:
+    - cron: '0 1 * * *'
+
+jobs:
```
jjtang1985, line 13: I assume you copied from the build.yml.
[req] Please also do so for dependabot step [here](https://github.com/SAP/cloud-sdk-js/blob/v2-main/.github/workflows/build.yml#L98).

.github/workflows/build-v3.yml
```diff
@@ -0,0 +1,87 @@
...
+
+on:
+  pull_request:
+    branches: ['main']
+  push:
+    branches: ['main']
+    paths-ignore:
+      - 'docs/**'
+  schedule:
+    - cron: '0 1 * * *'
```
jjtang1985, line 11:
```suggestion
```
[req] The nightly job is not be needed.

.github/workflows/build-v3.yml
```diff
@@ -0,0 +1,87 @@
+name: build v3
```
jjtang1985, line 1: The v3 pipeline is [here](https://github.com/SAP/cloud-sdk-js/blob/main/.github/workflows/build.yml), which was created already.
I was not able to double check, when I was outside today.
Sorry for confusing.
You can just touch the node versions there.

fwilhe: I know this builds v3 but it also builds the matrix of node 14,16,18 but we only want node 18

fwilhe:
> I am confused. Let's have a call tomorrow. Why is the `v3` script merged in the `v2-main` branch?

because our default branch `v2-main` is relevant for what github builds in ci. the goal of this pr is to have a separate pipeline definition for v3 as long as it is not released so we only build it with node 18 and can do other tweaks.

PR #3164, opened 2022-12-06
chore: use node 18

jjtang1985:
Note: the tsconfig uses the following:
```
"target": "es2021",
```


fwilhe (changes requested):
this means we won't test v2 on non-node18 versions. is that intended? why not do [this](https://github.com/SAP/cloud-sdk-js/pull/3159) instead?

test-packages/test-services-e2e/v4/test-service/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2019",
+    "target": "es2021",
```
FrankEssenberger, line 3: [q] If I look at the [specs](https://www.typescriptlang.org/tsconfig#target) there are `es2022`and `esnext` as even newer standards. Is there a reason why we use 2021?

FrankEssenberger (approved):
LGTM only renaming and warning fixes as far as I can see. One question.

test-packages/test-services-e2e/v4/test-service/tsconfig.json
jjtang1985: With es2022, it introduces lots of compile time/runtime changes and the clients are generated differently.
I switched back to es2021, and now everything are solved right away.
Therefore, I would make this as a first step and merge it.
Based on this, I'll continue my investigation.

PR #3173, opened 2022-12-07
chore: Don't run dependency check as part of compile

fwilhe:
This speeds up the build process and dependency check is still run as part of the pipeline.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -10,7 +10,9 @@
   ],
   "pipeline": {
     "compile": {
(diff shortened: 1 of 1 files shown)


florian-richter:
To my understanding this would cause the dep check to not be run at all anymore, right?

FrankEssenberger:
I would also like to keep it. Sometimes I get an error which is blocking my flow but often I realize mistakes by this.

PR #3198, opened 2022-12-09
chore: Use @types/node 18

jjtang1985:
Related to: https://github.com/SAP/cloud-sdk-backlog/issues/939


V3-Upgrade-Guide.md
```diff
@@ -46,6 +46,9 @@ Therefore, all SAP Cloud SDK for JavaScript libraries will switch to node 18 as
 If you run an older (<18) node version, please update to a newer version.
 You can find a list of breaking changes in the news section of the node.js [website](https://nodejs.org/en/blog/).
 
+Please not, in the `tsconfig.json`, we updated the configuration, so it looks like `"module": "Node16"`.
```
fwilhe, line 49:
```suggestion
Please note, in the `tsconfig.json`, we updated the configuration, so it looks like `"module": "Node16"`.
```

V3-Upgrade-Guide.md
```diff
@@ -46,6 +46,9 @@ Therefore, all SAP Cloud SDK for JavaScript libraries will switch to node 18 as
...
+As `Node18` is not a valid value of `module` for the time being, we might update it later.
```
fwilhe, line 50: I don't know if they'll do that.. is this sentence important?

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2021",
-    "module": "commonjs",
+    "module": "Node16",
```
fwilhe, line 4: are you sure this is what we want need? why? cf [docs](https://www.typescriptlang.org/tsconfig#node16nodenext-nightly-builds)

yarn.lock
```diff
@@ -1626,11 +1626,16 @@
   resolved "https://registry.npmjs.org/@types/node/-/node-12.20.55.tgz#c329cbd434c42164f846b909bd6f85b5537f6240"
   integrity sha512-J8xLz7q2OFulZ2cyGTLE1TbbZcjpno7FaN6zdJNrgAdrJ+DZzh/uFR6YrTb4C+nXakvud8Q4+rbhoIWlYQbUFQ==
 
-"@types/node@^14.14.35", "@types/node@^14.18.34":
+"@types/node@^14.14.35":
```
fwilhe, line 1629: does this line hint at leftovers of node 14 types?

fwilhe (changes requested):
I have some questions that need to be addressed before merging.

yarn.lock
jjtang1985: This comes from transitive dependency.

V3-Upgrade-Guide.md
jjtang1985: This is a note for ourselves and the users, as you might be confused by the "16".
@marikaner , what's your opinion?

marikaner: [pp] Also - I agree it is not releavant for the upgrade guide. Maybe rather a compatibility note.

tsconfig.json
marikaner: [q] Ah, is it the same? According to documentation it should be, but I thought I saw it was different. That's great then 😊

marikaner (approved):

.changeset/node-18.md
```diff
@@ -0,0 +1,20 @@
...
+'@sap-cloud-sdk/test-util': major
+'@sap-cloud-sdk/util': major
+'@sap-cloud-sdk/openapi-generator': major
+'@sap-cloud-sdk/generator-common': major
+'@sap-cloud-sdk/temporal-de-serializers': major
+'@sap-cloud-sdk/resilience': major
+---
+
+
+- [Compatibility Note] Switch the compilerOptions.module to `Node16` instead of `CommonJS` for better interoperability with ES modules.([handbook](https://www.typescriptlang.org/docs/handbook/esm-node.html))
```
fwilhe, line 20:
```suggestion
- [Compatibility Note] Switch the compilerOptions.module to `Node16` instead of `CommonJS` for better interoperability with ES modules. See the [TypeScript handbook](https://www.typescriptlang.org/docs/handbook/esm-node.html) for technical details if needed.
```

yarn.lock
fwilhe: do you know which dependency? can we upgrade it? I guess it's not ideal to have types from both node versions, not sure if it will bite us but it seems unclean to me.

fwilhe (approved):

PR #3203, opened 2022-12-12
chore: Circuit Breaker via Middleware

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#667.

Still in draft status. @deekshas8 one test was failing due to refreshtoken not beeing JSON. Is it possible to extract the tenant from the refresh token? I just encoded it and hoped for the best :-). Perhaps you can find a realistic content of the refresh token we can use in the test.

This [test](https://github.com/SAP/cloud-sdk-js/blob/main/packages/connectivity/src/scp-cf/circuit-breaker.spec.ts) could provide valuable information what happens if the xsuaa fails on http errors.


deekshas8:
> Is it possible to extract the tenant from the refresh token
A realistic value is of the format `9548314dc4004d8e836817b7dcc15a4e-r`. From what I read about refresh tokens, the payload is also encrypted and is only meant to be decrypted by the OAuth server, not the client. Why do we need to decode this ?

> This test could provide valuable information what happens if the xsuaa fails on http errors.
I debugged this for two cases: `client_secret` is wrong or it is missing.
In the first case, xssec returns an `AxiosError` with expected `401` in `error.response.status`. I added the `httpErrorFilter` to the `circuitbreakerXSUAA` and noticed that this still opened the breaker. This is because the promise is rejected with this [message](https://github.com/SAP/cloud-sdk-js/blob/9630f6afb53970591826f37ebead143401abac27/packages/connectivity/src/scp-cf/xsuaa-service.ts#L110) so the filter does not work. This will need to be adjusted.

For the second case (missing), the error is not of type `AxiosError` but something like:
```
statuscode: 500
message: 'Invalid config: Missing clientsecret.'
stack: 'Error: Invalid config: Missing clientsecret.\n    at buildOptions (C:\\repo\\cloud-sdk-js\\node_modules\\@sap\\xssec\\lib\\requests.js:149:19)\n    at 
```
This statuscode is [added by the library](https://github.wdf.sap.corp/CPSecurity/node-xs2sec/blob/77135e3c8ce57271fb88941af0537721372e2ded/lib/requests.js#L338).

packages/connectivity/src/scp-cf/destination/destination-accessor-loading-precendence.spec.ts
```diff
@@ -103,5 +103,5 @@ describe('destination loading precedence', () => {
     ).rejects.toThrowErrorMatchingInlineSnapshot(
       '"Could not fetch client credentials token for service of type "destination"."'
     );
-  });
+  }, 99999);
```
FrankEssenberger, line 106: [req] remove long timeout - I guess debugging

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -251,6 +246,30 @@ export async function fetchCertificate(
...
+function getTenantFromTokens(token: AuthAndExchangeTokens | string): string {
+  if (typeof token === 'string') {
+    const decoded = decodeJwt(token);
+    if (decoded.zid) {
+      return decoded.zid;
+    }
+    if (decoded.iss) {
+      return parseSubdomain(decoded.iss);
+    }
+    throw new Error('Could not obtain tenant identifier from jwt.');
```
FrankEssenberger, line 258: [req] I think the `exchangeHeaderJwt` wukk also be used in the bring your own JWT approach. In this case the JWT could not contain the `zid` or `iss` property. So perhaps we restructure this a bit that we do not call the method itself but a helper method which checks the zid and iss and return this or undefinend. Then we do all the cases and see if we find something. In the very end we throw if non of the ways worked.

packages/connectivity/src/scp-cf/token-accessor.spec.ts
```diff
@@ -240,16 +237,15 @@ describe('token accessor', () => {
       expect(retrieveFromCacheSpy).toHaveBeenCalledTimes(0);
     });
 
-    it('throws an error with only the cause property', async () => {
+    it('throws an error with the cause property', async () => {
```
FrankEssenberger, line 240: [req] Here the text was not clear. The PR from shoji was that it has the cause, but not the cause.config which contained the full axios config. Perhaps make this more clear in the test description and adjust the assertion so that it checks for that.

packages/connectivity/src/scp-cf/xsuaa-service.ts
```diff
@@ -92,32 +63,19 @@ export async function getClientCredentialsToken(
...
-            : service;
-
-          return err
-            ? reject(
-                `Error in fetching the token for service ${serviceName}: ${err.message}`
-              )
-            : resolve(tokenResponse);
-        }
+        (err: Error, token: string, tokenResponse: ClientCredentialsResponse) =>
+          err ? reject(err) : resolve(tokenResponse)
```
FrankEssenberger, line 67: [req] here we need to put back some part that only the message is included otherwise the whole axios config could end up in the logs again.

packages/connectivity/src/scp-cf/token-accessor.spec.ts
FrankEssenberger: The `getUserToken` method should behave the same so perhaps add a second test.

packages/resilience/src/circuit-breaker.ts
```diff
@@ -0,0 +1,147 @@
...
+
+function xsuaaErrorFilter(error: AxiosError | Error): boolean {
+  if (isAxiosError(error)) {
+    if (
+      error.response?.status &&
+      error.response.status.toString().startsWith('4')
+    ) {
+      return true;
+    }
+  } else if (error['statuscode'] && error['statuscode'] === 500) {
```
FrankEssenberger, line 36: [q] I assume this means some setup error right? It is a bit missleading why 500 are not counted. Perhaps add a comment.

packages/resilience/src/circuit-breaker.ts
```diff
@@ -0,0 +1,147 @@
...
+  if (
+    error.response?.status &&
+    error.response.status.toString().startsWith('4')
+  ) {
+    return true;
+  }
+  return false;
+}
+
+function isAxiosError(err: AxiosError | Error): err is AxiosError {
```
FrankEssenberger, line 24: [pp] I think we could make our lives easier and just have one CB for HTTP and XSUAA and filter for both just the 400 axios errors out. A wrong config with respect to client credentials is unlikely in a real scenario where this comes from the service binding.
If you want to keep them both also fine.

packages/resilience/src/circuit-breaker.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+      request
+    );
+
+    expect(Object.keys(circuitBreakers)).toEqual([
+      `${host}::tenant1`,
+      `${host}::tenant2`
+    ]);
+  });
+
+  it('creates circuit breaker for each service', async () => {
```
FrankEssenberger, line 129: [req] I guess the description is outdated.

packages/connectivity/src/scp-cf/destination/destination-service.ts
deekshas8: Tried to restructure this, have a look.

packages/resilience/src/circuit-breaker.ts
deekshas8: I simplified this, so we now have a common function for sb for xsuaa and http

packages/connectivity/src/scp-cf/circuit-breaker.spec.ts
```diff
@@ -89,20 +87,17 @@ describe('circuit breaker', () => {
...
-        if (e.message === 'Breaker is open') {
+        if (e.message.match(/Breaker is open/)) {
           keepCalling = false;
         }
       }
     }
     // Since we exit the loop breaker opened.
     expect(failedCalls).toBeGreaterThan(0);
-  }, 99999);
+  }, 999999);
```
FrankEssenberger, line 100: [req] reduce timeout

packages/connectivity/src/scp-cf/token-accessor.spec.ts
```diff
@@ -240,15 +238,30 @@ describe('token accessor', () => {
...
         { access_token: signedJwt({ dummy: 'content' }) },
         401,
         destinationBindingClientSecretMock.credentials
       );
       const promise = serviceToken('destination');
-      await expect(promise).rejects.toHaveProperty('cause');
+      await expect(promise).rejects.not.toHaveProperty('cause.config');
+    });
+
+    it('jwtBearerToken should throw an error without  cause.config property', async () => {
```
FrankEssenberger, line 252: good that we have a test now also for the other method.

packages/resilience/src/circuit-breaker.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+          context,
+          request
+        )
+      ).rejects.toThrowError(/Request failed with status code 401/);
+      keepCalling = !mock.isDone();
+    }
+    expect(circuitBreakers[`${host}::myTestTenant`].opened).toBe(false);
+  });
+
+  it('works together with a timeout', async () => {
```
FrankEssenberger, line 201: [nth] You could move this test to the resilience.spec.ts @fwilhe introduced. This contains all the combined tests.

packages/connectivity/src/scp-cf/circuit-breaker.spec.ts
deekshas8: I changed it for testing and forgot. Set it to 15 sec

packages/resilience/src/circuit-breaker.ts
```diff
@@ -0,0 +1,114 @@
...
+      ) as CircuitBreaker<any, ReturnT>
+    ).fire(options.fn);
+}
+
+/**
+ * Map of all existing circuit breakers.
+ * Entries are added in a lazy way.
+ * @internal
+ */
+export const circuitBreakers: Record<string, CircuitBreaker> = {};
```
FrankEssenberger, line 56: [pp] Could you move this variable to the top of the file. I like to have such maps directly visible when you open a file.

FrankEssenberger (commented):
LGTM - I can not approve cause I opened it. Feel free to merge the PR.

packages/resilience/src/circuit-breaker.spec.ts
deekshas8: Done

deekshas8 (approved):

PR #3212, opened 2022-12-13
chore: Add verbose option and change unimportant info logs to verbose logs

YuanZhuang107:
Closes SAP/cloud-sdk-backlog#977

For this PR, I have added `verbose` option and change unimportant info logs to verbose logs.


.changeset/clean-olives-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/generator': major
+---
+
+[Compatibility Note] Add `verbose` generator option. By default, only error and warning logs will be displayed. If `verbose` is set to true, all logs will be displayed.
```
marikaner, line 5:
```suggestion
[Compatibility Note] Add `verbose` generator option. By default, only error and warning logs will be displayed. If `verbose` is set to `true`, all logs will be displayed.
```

marikaner (changes requested):
@YuanZhuang107 You suggested to add tests and I very much agree. My proposal on the strategy:
Use the generator programmatically (once with the default `verbose` option and once set to `true`) and check output on `stdout`. [Here is a similar test](https://github.com/SAP/cloud-sdk-js/blob/90942b886350b312f36b7798f7b82440f3c8aae2/packages/util/src/logger/cloud-sdk-logger.spec.ts#L257).

marikaner (approved):

PR #3213, opened 2022-12-13
chore: Add default retry middleware

fwilhe:
Closes SAP/cloud-sdk-backlog#668.


packages/resilience/src/resilience.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+describe('combined resilience features', () => {
+  const HTTP_STATUS = {
+    OK: 200,
+    UNAUTHORIZED: 401,
+    FORBIDDEN: 403,
+    SERVICE_UNAVAILABLE: 504
+  };
+
+  it('Needs to retry with delay below timeout', async () => {
```
FrankEssenberger, line 15: [nth] start with lower case in the sentence so the `it` - applies also for the other tests.

packages/resilience/src/resilience.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    await expect(
+      executeWithMiddleware(
+        [timeout(600), retry(2)],
+        { uri: 'https://example.com', args: [] },
+        request
+      )
+    ).resolves.not.toThrow();
+  });
+  it('Needs to retry with delay above timeout', async () => {
+    nock('https://example.com', {})
```
FrankEssenberger, line 40: [pp] you could change the test slightly so that both mocks return some okish status code but different data e.g. `killded by timeout` and `response ok`. Then you make the call and assert that the data is `response ok`. Doing that you test that the timeout actually killed the first request because return was 200 and the retry was executed getting the data. In the current test you do not really know if the retry stopped the first request or the bad status.

packages/resilience/src/resilience.spec.ts
```diff
@@ -0,0 +1,63 @@
+import axios from 'axios';
+import nock from 'nock';
+import { executeWithMiddleware } from './middleware';
+import { retry } from './retry';
+import { timeout } from './timeout';
+
+describe('combined resilience features', () => {
```
FrankEssenberger, line 7: [q] I like the separate file for the combined features. Could you have a look and move a combinded test from timeout.spec.ts if there is any.

packages/resilience/src/retry.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    await expect(
+      executeWithMiddleware(
+        [retry(2)],
+        { uri: 'https://example.com', args: [] },
+        request
+      )
+    ).resolves.not.toThrow();
+  });
+
+  it('Does not retry on HTTP Status 401', async () => {
```
FrankEssenberger, line 58: [req] Add a second nock with status 200 after the unauthorized to ensure that no retry is done.

packages/resilience/src/retry.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      executeWithMiddleware(
+        [retry(2)],
+        { uri: 'https://example.com', args: [] },
+        request
+      )
+    ).rejects.toThrowError('Request failed with status code 401');
+  });
+
+  it('Does not retry on HTTP Status 403', async () => {
+    nock('https://example.com', {}).get('/retry').reply(HTTP_STATUS.FORBIDDEN);
```
FrankEssenberger, line 80: same here - would like to be more explicit on the no retry is done. Either with a spy or the the nock is not done.

packages/resilience/src/retry.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const defaultRetryCount = 3;
+
+/**
+ * Helper method to build a retry middleware.
+ * @param retryCount - Number of retry attempts. Default value is 3.
+ * @returns The middleware adding a retry to the function.
+ */
+export function retry<ReturnType, ContextType extends Context>(
+  retryCount: number = defaultRetryCount
```
FrankEssenberger, line 12: [pp] I would prefer to put the aysnc-retry options here with a default setup. So that users can also adjust more then the count. On the other hand we are more simple doing that. @jjtang1985 do you have an opinion on that?

packages/resilience/src/retry.ts
```diff
@@ -0,0 +1,35 @@
...
+    options: MiddlewareIn<ReturnType, ContextType>
+  ): MiddlewareOut<ReturnType> {
+    return () =>
+      asyncRetry.default(
+        async bail => {
+          try {
+            return await options.fn();
+          } catch (error) {
+            // Don't retry on error statuses where a second attempt won't help
+            const status = error.response.status;
```
FrankEssenberger, line 24: [req] make the acces on `error.response.status` optional with `?` - you may add a low level log statement informing that the property is not present.

packages/resilience/src/timeout.ts
```diff
@@ -3,7 +3,7 @@ import { Context, Middleware, MiddlewareIn, MiddlewareOut } from './middleware';
 const defaultTimeout = 10000;
 
 /**
- * Helper method to build a timout middleware.
+ * Helper method to build a timeout middleware.
```
FrankEssenberger, line 6: thanks

FrankEssenberger (approved):
LGTM - some minor questions and suggestions for improving the tests a bit.

packages/resilience/src/resilience.spec.ts
fwilhe: yeah this makes sense, thanks for the hint

PR #3223, opened 2022-12-15
chore: Fix `next` version release

jjtang1985:
This git command does not work even locally: `git --no-pager log -n 1 main --pretty=format:"%at"`
It might be related to non-default branch.
As it does not bring much value anyways, I'll remove this check.


deekshas8 (approved):
LGTM

PR #3244, opened 2022-12-19
feat: Add Retry to Destination by Name call

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#1010.
 
Add a option to retry the call to the destination service  by name call. This is the one where the token is fetched by the destination service and the one which was a bit flaky sometimes.


packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -375,8 +422,20 @@ async function callDestinationService(
     headers
   };
 
+  let resilience = [
+    timeout<fpp, HttpMiddlewareContext>(),
+    circuitBreakerHttp<fpp, HttpMiddlewareContext>()
+  ];
+  // Do a retry only for the get destination by Name calls which are somethimes flaky for the token retrieval
+  if (options?.destinationName && options.retry) {
+    resilience = [
```
deekshas8, line 431: We can replace this condition part when we have `resilience()` method.

packages/connectivity/src/scp-cf/destination/destination-service-types.ts
```diff
@@ -230,6 +230,12 @@ export type DestinationRetrievalOptions = CachingOptions & {
    * By default, IsolationStrategy.Tenant_User is set.
    */
   isolationStrategy?: IsolationStrategy;
+  /**
+   * Sometimes the `destination/{name}` performing a token exchange are flaky due to an unstable token exchange endpoint.
```
deekshas8, line 234:
```suggestion
   *  The call to `destination/{name}` sometimes also performs a token exchange. Such calls can be flaky due to an unstable token exchange endpoint.
```

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -323,6 +328,42 @@ function errorMessageFromResponse(
...
+        } catch (error) {
+          const status = error?.response?.status;
+          if (status.toString().startsWith('4')) {
+            bail(new Error(`Request failed with status code ${status}`));
+            // We need to return something here but the actual value does not matter
+            return undefined as any;
+          }
+          throw error;
+        }
+        retryCount++;
```
deekshas8, line 354: [req] unreachable code

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -323,6 +328,42 @@ function errorMessageFromResponse(
...
+function retryDestination(
+  destinationName: string
+): Middleware<AxiosResponse<DestinationJson>, HttpMiddlewareContext> {
+  return options => () => {
+    let retryCount = 0;
+    return asyncRetry.default(
+      async bail => {
+        try {
+          const destination = await options.fn();
+          if (retryCount < 3) {
```
deekshas8, line 340: [q] Is this if necessary? the retry count will never be more than 3 right?

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -9,12 +9,16 @@ import axios, { AxiosRequestConfig, AxiosResponse, AxiosError } from 'axios';
 import {
   executeWithMiddleware,
   HttpMiddlewareContext,
-  circuitBreakerHttp
+  circuitBreakerHttp,
+  Context,
+  timeout
```
deekshas8, line 14: [req] All middlewares are exported, so should not be imported from internal path.

packages/connectivity/src/scp-cf/destination/destination-service.ts
FrankEssenberger: The reason for this is that we return the destination with the errors in the auth tokens. I am not sure if this is a good idea because in most cases the destintaion will be useless. However, this is the way it was before and I wanted to keep this. Hence the `buildAuthorizationHeaders` is not executed on the last retry to make the promise resolve.

packages/connectivity/src/scp-cf/destination/destination-service.ts
FrankEssenberger: moved to right place - also added a test.

packages/connectivity/src/scp-cf/destination/destination-service.ts
FrankEssenberger: yes nice.

.changeset/red-snakes-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Add a `retry` option in the `DestinationFetchOption`. The retries apply to the destination by name call and not the full destination retrieval process.
```
fwilhe, line 5: > apply to the destination by name call and not the full destination retrieval process

can we rephrase this to make more clear what it means? Not sure if our users understand this sentence, I certainly have issues understanding what it means

packages/connectivity/src/scp-cf/destination/destination-service-types.ts
```diff
@@ -230,6 +230,12 @@ export type DestinationRetrievalOptions = CachingOptions & {
    * By default, IsolationStrategy.Tenant_User is set.
    */
   isolationStrategy?: IsolationStrategy;
+  /**
+   * The call to `destination/{name}` sometimes also performs a token exchange. Such calls can be flaky due to an unstable token exchange endpoint.
+   * This parameter enables 3 retries on the destination by name calls.
+   * By default, the retry is disabled.
```
fwilhe, line 236: why is it disabled by default? would not enabled be a good default to have more resilience without requiring user action?

FrankEssenberger: hm, I was afraid to add load to the destination service @marikaner what do you think?

.changeset/red-snakes-shout.md
FrankEssenberger: changed it to not mention technical details but when to enable it.

packages/connectivity/src/scp-cf/destination/destination-service-types.ts
florian-richter: I like disabled by default. Behavior stays as is unless configured otherwise.

packages/connectivity/src/scp-cf/destination/destination-service.spec.ts
```diff
@@ -631,6 +632,123 @@ describe('destination service', () => {
       delete process.env.no_proxy;
     });
 
+    it('does a retry if request fails wit 500 error', async () => {
```
florian-richter, line 635:
```suggestion
    it('does a retry if request fails with 500 error', async () => {
```

packages/connectivity/src/scp-cf/destination/destination-service.spec.ts
```diff
@@ -631,6 +632,123 @@ describe('destination service', () => {
...
+        authTokens: [
+          {
+            type: 'Bearer',
+            value: 'some.token',
+            expires_in: '3600',
+            error: '',
+            http_header: {
+              key: 'Authorization',
+              value: 'Bearer some.token'
+            } as any // TODO fix the authTokens type in DestinationJson
```
florian-richter, line 704: [q] Is there a ticket for this?

packages/connectivity/src/scp-cf/destination/destination-service-types.ts
fwilhe: > Behavior stays as is unless configured otherwise

v3 is our chance to do breaking changes where we see fit. Sure we need to consider if it is worth it, but in this case I'd say being more resilient by default is added usability. Most users don't change options. Is it good to have resilience for most users? I'd say: yes.

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -323,6 +328,42 @@ function errorMessageFromResponse(
...
+          const destination = await options.fn();
+          if (retryCount < 3) {
+            retryCount++;
+            // this will throw if the destination does not contain valid auth headers and a second try is done to get a destination with valid tokens.
+            await buildAuthorizationHeaders(parseDestination(destination.data));
+          }
+          return destination;
+        } catch (error) {
+          const status = error?.response?.status;
+          if (status.toString().startsWith('4')) {
```
florian-richter, line 348:
```suggestion
          if (status >= 400 && status < 500) {
```
[pp] It's slightly shorter and avoids coercing to string.

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -323,6 +328,42 @@ function errorMessageFromResponse(
     : '';
 }
 
+function retryDestination(
+  destinationName: string
+): Middleware<AxiosResponse<DestinationJson>, HttpMiddlewareContext> {
+  return options => () => {
+    let retryCount = 1;
+    return asyncRetry.default(
```
florian-richter, line 336: [q] Is there any waiting time between retries? I would assume ~100ms between request could be helpful...

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -375,8 +422,20 @@ async function callDestinationService(
     headers
   };
 
+  let resilience = [
+    timeout<fpp, HttpMiddlewareContext>(),
+    circuitBreakerHttp<fpp, HttpMiddlewareContext>()
+  ];
+  // Do a retry only for the get destination by Name calls which are sometimes flaky for the token retrieval
```
florian-richter, line 429:
```suggestion
  // Do a retry only for `getDestination({destinationName})` calls which are sometimes flaky for the token retrieval
```

packages/connectivity/src/scp-cf/destination/destination-service.ts
deekshas8: There is a 1000ms wait before the first retry (default). Thereafter, it uses an exponential backoff strategy. As explained [here](https://github.com/tim-kos/node-retry#tutorial): With the default settings (10 in async-retry), this means the last attempt is made after 17 minutes and 3 seconds.

packages/connectivity/src/scp-cf/destination/destination-service.spec.ts
deekshas8: None that I could find, but I am also not entirely aware of the issue. I have created a note in the backlog for now.

packages/connectivity/src/scp-cf/destination/destination-service.spec.ts
```diff
@@ -631,6 +632,123 @@ describe('destination service', () => {
...
+        .reply(200, response);
+
+      const actual = await fetchDestination(destinationServiceUri, jwt, {
+        destinationName: 'HTTP-BASIC',
+        retry: true
+      });
+      expect(actual).toEqual(parseDestination(response));
+    });
+
+    it('does a no retry if request fails with 401 error', async () => {
```
tomfrenken, line 656:
```suggestion
    it('does no retry if request fails with 401 error', async () => {
```

packages/connectivity/src/scp-cf/destination/destination-service.ts
florian-richter: That seems pretty long... maybe factor 3 and starting with 100ms would be better? Does not need to be changed in this PR though.

florian-richter (approved):
All my comments were addressed. I did not do another in-depth review since the first time it seemed pretty good already. (So please let me know if something bigger changed or if an opinion on any topic is necessary)

packages/connectivity/src/scp-cf/destination/destination-service.ts
deekshas8: I agree, I also had the same thought working on tests combining retry and cb. Will discuss with @FrankEssenberger since we also don't expose other options to the user.

FrankEssenberger: I thought we do 3 retries, let's discuss why it takes so long.

deekshas8: Yes the default is 3 in SDK. The explanation in my first comment was about the default behavior in async-retry (where retry count is 10 and it takes 17 mins or so to finish). Even in the SDK, it's possible to set a higher value. I wanted to test a scenario where > 10 retries opens cb for 5xx error, and the test did take around 17 mins to finish.

FrankEssenberger: As discussed directly we will keep the options simple first. If there is feedback that the retry should have more options we will most likely add the union type `number | fullOptions` so that users can provide both.

PR #3245, opened 2022-12-19
feat: Add resilience method

deekshas8:
Closes SAP/cloud-sdk-backlog#1009.

Add `resilience()` method.


packages/resilience/src/resilience.spec.ts
```diff
@@ -64,7 +65,7 @@ describe('combined resilience features', () => {
 
     expect(response.status).toBe(HTTP_STATUS.OK);
   }, 10000);
-  it('circuit breaker works together with a timeout', async () => {
+  it('uses circuit breaker and timeout with resilience()', async () => {
```
FrankEssenberger, line 68: [req] Perhaps you can add a few more tests with the retry being added in this file.

packages/resilience/src/resilience.ts
```diff
@@ -0,0 +1,63 @@
...
+    middlewares.push(timeout(resilienceOption.timeout));
+  } else if (resilienceOption.timeout) {
+    middlewares.push(timeout());
+  }
+
+  if (resilienceOption.circuitBreaker) {
+    middlewares.push(circuitBreakerHttp());
+  }
+
+  if (typeof resilienceOption.retry === 'number') {
```
FrankEssenberger, line 57: [nth] add a check if the number if negative and throw.

packages/resilience/src/resilience.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @param options - Resilience Options.
+ * @returns Array of middleware functions.
+ */
+export function resilience<ReturnT, ContextT extends Context>(
+  options?: ResilienceOptions
+): Middleware<ReturnT, ContextT>[] {
+  const resilienceOption = { ...defaultResilienceOptions, ...options };
+  const middlewares: Middleware<ReturnT, ContextT>[] = [];
+  if (typeof resilienceOption.timeout === 'number') {
+    middlewares.push(timeout(resilienceOption.timeout));
```
FrankEssenberger, line 48: [nth] add a check for number greater 0. Also a check is nice when the number is low indicating that someone made an input assuming the value is seconds instead of milliseconds. In this case a warning is enough like values below 10 are most likely an error.

FrankEssenberger (approved):
LGTM - just a few nice to have requests. Implemented as suggested in the ADR.

PR #3251, opened 2022-12-20
chore: Improve parsed options handling

marikaner:
- Remove default value for processes
- Fix types to align between cli, user options and parsed options
- Replace yargs coercion with coercion applied by Options parser => so it can be reused by programmatic approach. yargs only parses the input, no defaults, no coercion.
- Strenghen parsed types based on inferred types.

Relates to SAP/cloud-sdk-backlog#980.


packages/generator/src/options/options.ts
```diff
@@ -200,16 +183,22 @@ export const generatorOptionsCli = {
...
       'Configuration file to ensure consistent names between multiple generation runs with updated / changed metadata files. Will be generated if not existent. By default it will be saved to/read from the input directory as "service-mapping.json".',
     type: 'string',
-    coerce: coercePathArg,
-    normalize: true
+    coerce: (
+      arg: string | undefined,
+      options: GeneratorOptions & { config?: string }
+    ) =>
+      coerceRequiredPathArg(
+        arg ? arg : join(options.inputDir.toString(), 'service-mapping.json'),
```
FrankEssenberger, line 191: [q] I would have expected the `join(options.inputDir.toString(), 'service-mapping.json')` as default - or is thisnot possible?

packages/generator/src/options/options-parser.ts
```diff
@@ -1,12 +1,62 @@
...
+  /**
+   * Coerce function with additional parameter for all options.
+   */
+  coerce?: (arg: any, options: any) => any;
+} & Required<Pick<Options, 'describe' | 'type'>>;
+
+/**
+ * @internal
+ * Helper to represent parsed options based on a public generator options type and a CLI options configuration.
+ * - Makes all properties required.
```
FrankEssenberger, line 30: [pp] Perhaps you can add to the list that it also sets the defaults.

packages/generator/src/generator.ts
```diff
@@ -85,41 +85,28 @@ export async function generateWithParsedOptions(
...
-    options as ParsedGeneratorOptions
-  );
+  await generateFilesWithoutTsMorph(services, options);
 
   if (options.transpile) {
     const directories = services
       .filter(async service => {
         const files = await readdir(
-          resolvePath(service.directoryName, options as ParsedGeneratorOptions)
+          resolveOutDirPath(service.directoryName, options)
```
FrankEssenberger, line 100: [pp] As I understood are at this point the paths already resolved. I would therefore use `join` and name the method accordingly.

packages/generator/src/options/options-parser.spec.ts
```diff
@@ -68,6 +76,27 @@ describe('options parser', () => {
       });
     });
 
+    it('coerces value', () => {
```
FrankEssenberger, line 79: [req] Add a test to ensure the desired behaviour for coerce with a property having a default value.

packages/generator/src/options/options.ts
```diff
@@ -220,14 +209,14 @@ export const generatorOptionsCli = {
...
-      'When set to true, the generator will write a README.md file into the root folder of every package. This option does not make that much sense without also set useSwagger to "true".',
+      "When set to true, the generator will write a README.md file into the root folder of every package. The information in the readme are mostly derived from accompanying Swagger or OpenAPI files. Therefore it is recommended to use the 'readme' option in combination with 'useSwagger'.",
     type: 'boolean',
     default: false,
     hidden: true
   },
   writeReadme: {
     describe:
-      'When set to true, the generator will write a README.md file into the root folder of every package. This option does not make that much sense without also set useSwagger to "true".',
+      "When set to true, the generator will write a README.md file into the root folder of every package. The information in the readme are mostly derived from accompanying Swagger or OpenAPI files. Therefore it is recommended to use the 'writeReadme' option in combination with 'useSwagger'.",
```
FrankEssenberger, line 219: [q] If this is merged in main - should this option be removed?

FrankEssenberger: If so also look for other deprecated options in the list.

scripts/check-public-api.ts
```diff
@@ -10,7 +10,7 @@ import {
   readCompilerOptions,
   transpileDirectory
 } from '@sap-cloud-sdk/generator-common/internal';
-import {defaultPrettierConfig} from "@sap-cloud-sdk/generator-common/dist/file-writer";
+import { defaultPrettierConfig } from '@sap-cloud-sdk/generator-common/dist/file-writer';
```
FrankEssenberger, line 13: [q] Could we avoid importing from dist and using the ts file?

packages/generator/src/options/options-parser.ts
```diff
@@ -1,12 +1,62 @@
...
+ * @typeParam GeneratorOptionsOptionsT - Public generator options.
+ * @typeParam CliOptionsT - Configuration of CLI options.
+ */
+export type ParsedOptions<CliOptionsT extends Record<string, Option>> = Omit<
+  { [K in keyof CliOptionsT]: ParsedOptionType<CliOptionsT[K]> },
+  OptionsWith<'deprecated' | 'replacedBy', CliOptionsT>
+>;
+
+type ParsedOptionType<OptionT extends Option> = OptionT extends {
+  coerce: (...arg: any) => infer ReturnT;
```
FrankEssenberger, line 42: [pp] I think our `coerceRequiredPathArg` and `coercePathArg` have the same signature like (arg,options)=>ReturnT which is always string. Could we type it stronger here?

packages/generator/src/options/options-parser.ts
```diff
@@ -1,12 +1,62 @@
...
+   */
+  coerce?: (arg: any, options: any) => any;
+} & Required<Pick<Options, 'describe' | 'type'>>;
+
+/**
+ * @internal
+ * Helper to represent parsed options based on a public generator options type and a CLI options configuration.
+ * - Makes all properties required.
+ * - Removes deprecated and replaced options.
+ * - Replaces input types with coerced types.
```
FrankEssenberger, line 32: [pp] Mention that this means that for example path are resolved.

packages/generator/src/options/options-parser.ts
```diff
@@ -47,37 +100,43 @@ export function getOptionsWithoutDefaults<
...
-  GeneratorOptionsT extends Record<string, any>
+  GeneratorOptionsT extends Record<keyof CliOptionsT, any>
 > {
   constructor(
     private options: CliOptionsT,
     private userOptions: GeneratorOptionsT
   ) {}
 
-  parseOptions(): ParsedOptions<GeneratorOptionsT, CliOptionsT> {
+  parseOptions(): ParsedOptions<CliOptionsT> {
```
FrankEssenberger, line 120: [req] I think it would be good to have a way to get the options used by the `generateWithParsedOptions` so I would perhaps add here a log statement showing the final option being passed. Doing so we can ask for that output on support issues. This should be a low log level `debug`  programmatic use case OR if the `verbose` option is activated in the cli use case. - If we have something like this already please ignore this comment.

FrankEssenberger (changes requested):
I like the changes a lot - all parsing and default setting adjusting is in one place. I left a few questions and requests for changes.

packages/generator/src/options/options.ts
FrankEssenberger: Or is this done in this ticket: https://github.com/SAP/cloud-sdk-backlog/issues/978?

marikaner: > Or is this done in this ticket: [SAP/cloud-sdk-backlog#978](https://github.com/SAP/cloud-sdk-backlog/issues/978)?

Yes, exactly

scripts/check-public-api.ts
marikaner: I guess so, I don't know why this was done this way originally (just formatted the file).

packages/generator/src/options/options.ts
marikaner: This is not possible, because the option relies on another option (`inputDir`). I agree that maybe `coerce` should not handle defaults. However, these two activities are quite similar. Maybe, now that we have the `default` option in our hands, we could actually do more, like using functions for the default.

This can get more complicated though, because from a function it will be difficult to get a description of the default, so we probably would need ad *another* additional property.

For simplicity, I would probably rather go with the slightly dirty coerce function. If you disagree, I would be happy to discuss this and work on this in a separate PR.

packages/generator/src/options/options-parser.ts
marikaner: Yes!

packages/generator/src/options/options-parser.spec.ts
marikaner: Done, let me know what you think.

packages/generator/src/options/options-parser.ts
marikaner: But it doesn't, it depends what the `coerce` function does. I will add it to the path arg coersion functions instead.

packages/generator/src/options/options-parser.ts
marikaner: Not sure I understand what you mean.

I could change `(...arg: any) => infer ReturnT` to something like `(arg: any, options: XY) => infer ReturnT;`, but what is the benefit? (and I don't think that is what you meant)

`coerceRequiredPathArg` and `coercePathArg` return `string` and `string | undefined` which is a not always `string`. The coercion functions could however return an arbitrary type (e.g. `string | boolean`), so restricting this to `string` would be wrong in my opinion.

scripts/check-public-api.ts
marikaner: `git blame`s you ;) I believe this was just overlooked in the review. Fixed it.

marikaner: Aaaaand, reverted it. When I remove `dist` it does not work. I didn't really check why, but I would suggest to separate this from this PR. I only changed this file, because apparently it is not covered by the formatter.

packages/generator/src/generator.ts
marikaner: I now deleted the old function and everywhere it was called I am using `join(outputDir, path)`

packages/generator/src/options/options-parser.ts
marikaner: Done.

FrankEssenberger (approved):
LGTM As discussed with @marikaner  the compiler already complains if the types do not match

packages/generator/src/options/options-parser.ts
FrankEssenberger: discussed in meeting

PR #3252, opened 2022-12-20
chore: Release Major Canary

FrankEssenberger:
Tackel prerelease problem introducing a 3.0.0 prerelease version. Currently the `next` tag points to a version 2.11.0-20221220 which fine in principle: https://www.npmjs.com/package/@sap-cloud-sdk/util?activeTab=versions

However, in the dependencies the changeset bumps to the next version which is 2.12.0. This is however a valid version two release now. Since it is missleading to use version 2.X.Y for version 3 prereleases this change will release a version 3.0.0-20221220 with the tag next.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -153,7 +153,7 @@ jobs:
           echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' >> .npmrc
           date=`date +%Y%m%d%H%M%S`
           rm -f .changeset/*.md
-          cp canary-release-changeset.md .changeset
+          cp major-release-changeset.md .changeset
           yarn changeset pre enter ${date}
           yarn changeset version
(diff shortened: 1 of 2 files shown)


marikaner (approved):
I think this should probably do what you want. We should however make absolutely sure ;)
@florian-richter maybe you can also take a look

marikaner:
@FrankEssenberger I changed the base to `main`.

FrankEssenberger:
I am a bit afraid that we forget to change it back to the patch changeset after the 3.0 is released and the version in the package.json is 3.X.Y and we would create a 4.X.Y if we change something.

tomfrenken (approved):
LGTM, as I mentioned in slack, we should probably collect these "after-release-TODOs" in a ticket or add them as TODOs to the release v3 ticket.

PR #3253, opened 2022-12-21
chore: Remodel Actions to follow best practices

fwilhe:
This PR involves the following changes:

- Delete node_modules folder
- Switch build tool / lockfile to yarn to be in sync with other components
- Move code to a yarn package to integrate with the overall build
- Use ncc to build the executable js code with dependencies
- Configure eslint
- Follow structure from [GitHub's ts action template](https://github.com/actions/typescript-action)

Closes SAP/cloud-sdk-backlog#995.
Closes SAP/cloud-sdk-backlog#1012.


.github/actions/check-pr/dist/sourcemap-register.js
```diff
@@ -0,0 +1 @@
...pe._compile=function(e,r){p[r]=e;f[r]=undefined;return t.call(this,e,r)};n.prototype._compile.__sourceMapSupport=true}}if(!l){l="emptyCacheBetweenOperations"in r?r.emptyCacheBetweenOperations:false}if(!u){u=true;Error.prepareStackTrace=prepareStackTrace}if(!s){var o="handleUncaughtExceptions"in r?r.handleUncaughtExceptions:true;try{var i=dynamicRequire(e,"worker_threads");if(i.isMainThread===false){o=false}}catch(e){}if(o&&hasGlobalProcessEventEmitter()){s=true;shimEmitUncaughtException()}}};r.resetRetrieveHandlers=function(){h.length=0;d.length=0;h=S.slice(0);d=_.slice(0);v=handlerExec(d);m=handlerExec(h)}},147:e=>{"use strict";e.exports=require("fs")},17:e=>{"use strict";e.exports=require("path")}};var r={};function __webpack_require__(n){var t=r[n];if(t!==undefined){return t.exports}var o=r[n]={id:n,loaded:false,exports:{}};var i=true;try{e[n](o,o.exports,__webpack_require__);i=false}finally{if(i)delete r[n]}o.loaded=true;return o.exports}(()=>{__webpack_require__.nmd=e=>{e.paths=[];if(!e.children)e.children=[];return e}})();if(typeof __webpack_require__!=="undefined")__webpack_require__.ab=__dirname+"/";var n={};(()=>{__webpack_require__(284).install()})();module.exports=n})();
```
FrankEssenberger, line 1: [q] why this file is generated?

.github/actions/check-pr/dist/licenses.txt
```diff
@@ -0,0 +1,695 @@
+@actions/core
```
FrankEssenberger, line 1: [q] where do these `licenceses.txt` files in the dist come from? If they are auto generated mention it perhaps in the markdown? Are they mandatory?

.github/actions/README.md
```diff
@@ -0,0 +1,18 @@
+# Actions
+
+This directory contains code for GitHub Actions that are used in this repo's workflow.
+
+The actions are built with yarn like the rest of this repo.
+
+To build / update the actions, run this inside the action's directory:
+
+```
+yarn
```
FrankEssenberger, line 10: [q] Is this line neded?

.github/actions/README.md
```diff
@@ -0,0 +1,18 @@
...
+yarn run all
+```
+
+This will compile, lint and package the actions.
```
FrankEssenberger, line 14: [q] As it looks this only compiles and lints the action which is your CWD right? There are two the `get-changelog` and `check-pr` right?

FrankEssenberger (commented):
In general I would like to create a followup ticket so that we create a `action-packages` which is managed from yarn as the normal and test packages on every compile install lint etc. and the dist is then simply generated in the actions folder.

.github/actions/check-pr/dist/sourcemap-register.js
fwilhe: This file is generated in also in github's action template, cf https://github.com/actions/typescript-action/tree/main/dist

I think that's controlled by the `--source-map` flag provided to ncc (which is the tool that bundles all the dependencies into one js file that can be run without node_modules installed)

Honestly I don't know if github needs that but I wanted to stay close to github's template.

.github/actions/check-pr/dist/licenses.txt
fwilhe: That's generated by ncc via the `--license licenses.txt` flag. Technically this is not mandatory, and maybe we don't need it here but it feels correct to me to mention the licenses of dependencies we vendor into the js file we have in our repo.

.github/actions/README.md
fwilhe: Good question. Currently if you don't run it the dependencies are not installed. Is there a better way in yarn to automat this than to add that call to the `run all` script?

.github/actions/README.md
fwilhe: yes, I did not want to duplicate the readme so I've put it to the top level directory. will add a hint on the cwd.

fwilhe:
> In general I would like to create a followup ticket so that we create a `action-packages` which is managed from yarn as the normal and test packages on every compile install lint etc. and the dist is then simply generated in the actions folder.

hm maybe I can just do this as part of this pr? I did not think about this as I'm not yet too familar with the details of yarn, but I'll look into it. Makes sense to ensure dist is up-to-date I assume.

.github/actions/README.md
FrankEssenberger: yes

FrankEssenberger:
For me it is also fine to merge this frist and create a follow up. I guess the setup is not that difficult. Just add the new packages everywhere you see the `test-packages` in the yarn config.

.github/actions/get-changelog/.eslintignore
```diff
@@ -0,0 +1,4 @@
+dist/
+lib/
+node_modules/
+jest.config.js
```
florian-richter, line 4: [q] If this ignores basically all files, why is eslint (and prettier) even included in this package?

florian-richter (commented):
I think this does not resolve the problem of outdated dependencies being used. The action would need to be recompiled which is not happening automatically...

.github/actions/get-changelog/.eslintignore
fwilhe: it applies to the ts code of the action. since we don't have a lot of code we can discuss if we need eslint/prettier here, this is the config from github's template and I think all exclusions make sense.

florian-richter: I didn't realize that `dist/` and `lib/` does not include the code. My assumption was `lib` is code and `dist` is compile output, but dist was just renamed lib... my bad!

fwilhe:
> I think this does not resolve the problem of outdated dependencies being used. The action would need to be recompiled which is not happening automatically...

Okay I see. Yes the action is not compiled automatically, I'm not sure if we want to do that (on what event?) or if we want to include it into the main yarn project as frank suggested. Then this could happen automatically when you run `yarn whatever` to lint/build the whole project.

I think from a complexity/severity tradeoff perspective that would be okay. Getting updated dependencies for the action is nice, but this won't be 0-day-level of urgency so getting a new update once someone re-runs `yarn lint:fix` would be good enough imo.

florian-richter:
I dislike to mix lint and build. But in general I would be fine if it happened periodically. I would assume we could create an action that rebuilds on change (trigger through dependabot updating the deps) and maybe also every x days?

build-packages/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+To build / update the actions, run this inside the action's directory:
+
+```
+yarn run all
+```
+
+This will compile, lint and package the actions.
+
+Changes in the `dist` folder need to be committed to the git repository as this is the code which is being run by GitHub.
```
florian-richter, line 15: [q] Is there still a `dist` folder? Seems like `dist` was replaced by `lib` and doesn't actually need to be committed anymore (see `.gitignore`).

build-packages/check-pr/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "Checks the current PR for semantical correctness and appropriate changelog.",
+  "license": "Apache-2.0",
+  "private": true,
+  "scripts": {
+    "compile": "tsc -p tsconfig.json",
+    "lint": "eslint --ignore-pattern '!index.ts' && prettier --check '**/*.ts'",
+    "lint:fix": "eslint --ignore-pattern '!index.ts' --fix --quiet && prettier --write '**/*.ts'",
+    "check:dependencies": "depcheck --skip-missing=true .",
+    "package": "ncc build lib/index.js --out ../../.github/actions/check-pr/",
+    "all": "yarn && yarn run compile && yarn run lint && yarn run package"
```
florian-richter, line 13:
```suggestion
    "compile": "tsc -p tsconfig.json",
    "lint": "eslint --ignore-pattern '!index.ts' && prettier --check '**/*.ts'",
    "lint:fix": "eslint --ignore-pattern '!index.ts' --fix --quiet && prettier --write '**/*.ts'",
    "check:dependencies": "depcheck --skip-missing=true .",
    "postcompile": "ncc build lib/index.js --out ../../.github/actions/check-pr/",
    "all": "yarn && yarn run lint && yarn run compile"
```
[nth] `lint` can and should run before `compile`
[pp] If `package` is a follow up of `compile` (either through renaming to `postcompile` or by adding it to `turbo.json` accordingly), then we build and package regularly... would be nice 😊

.gitignore
```diff
@@ -1,10 +1,10 @@
 node_modules
-!.github/actions/**/node_modules
 .vscode
 build/**
 dist
 .next/**
-!.github/actions/**/dist
+!build-packages/**/dist
```
florian-richter, line 6:
```suggestion
```

I think this can be removed?

build-packages/README.md
fwilhe: Forgot to update. The compiled action (which was in dist before is now in the 'root' of the action in `.github/actions/action-name`. Will update the readme, thanks for the hint.

florian-richter (approved):
Looks very good! One more personal preference comment though 😉

PR #3254, opened 2022-12-21
chore: Allow Manual Canary Release

FrankEssenberger:
I wanted to get my release after the #3252 was merged. However the release was skipped. I remembered that we do it only every 24hours so I did a manual trigger which should overwrite this.

However I think the `boolean` os not taken correctly from the event intputs see [here](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) and [here](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/)

This PR changes this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,6 +25,8 @@ jobs:
       matrix:
         node-version: [18]
     steps:
+      - name: Show input from user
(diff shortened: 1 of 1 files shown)


tomfrenken (approved):
LGTM

marikaner (approved):

PR #3260, opened 2022-12-23
chore!: Url Property Optional

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#892.

- Makes the `url` property optional
- Remove all checks on this on the parsing of a destination
- Introduces a type `HttpDestination` for points in the code where it is clear a destination with URL is needed (HttpDestination)
- Uses type-guard `isHttpDestionation`  or type assertion `assertHttpDestination` to check if needed.
- Add a generic to the `destinationOrFetchOption` so that for if an `executeHttprequest` is done and a destination is passed the `url` is mandatory to help users with typing.


packages/connectivity/src/scp-cf/destination/destination.ts
```diff
@@ -469,7 +432,7 @@ export function noDestinationErrorMessage(
 /**
  * Type that is either a {@link Destination} or (XOR) {@link DestinationFetchOptions & DestinationForServiceBindingOptions}.
  */
-export type DestinationOrFetchOptions = Xor<
-  Destination,
+export type DestinationOrFetchOptions<T extends Destination> = Xor<
```
marikaner, line 435: [pp] No strong opinion, but I have 2 thoughts:

1. `DestinationOrFetchOptions<HttpDestination>` is longer than `HttpDestinationOrFetchOptions`, therefore I am not sure whether the generic is better than just two types.
2. Could we find a better (shorter) name than `Destination....<Destination>`? My ideas do not convince me: `FetchOptionsOr<Destination>`,

packages/connectivity/src/scp-cf/destination/destination-service-types.ts
```diff
@@ -239,13 +239,44 @@ export type DestinationRetrievalOptions = CachingOptions & {
...
-export function isDestination(destination: any): destination is Destination {
-  return destination.url !== undefined;
+export function isHttpDestination(
+  destination: HttpDestination | Destination | null
+): destination is HttpDestination {
+  if (destination?.type && destination.type !== 'HTTP') {
+    return false;
+  }
+
+  return !!destination?.url;
```
marikaner, line 279: [pp] Shorter and better readable in my opinion:
```suggestion
return destination?.type === 'HTTP' && destination?.url;
```

marikaner (dismissed):
I did not review every single file, as many are just replacements.

test-packages/integration-tests/test/auth-flows/auth-flow.spec.ts
```diff
@@ -96,7 +101,7 @@ describe('OAuth flows', () => {
       destinationName: systems.s4.providerBasic,
       jwt: accessToken.provider
     });
-
+    assertHttpDestination(destination!);
```
deekshas8, line 104: [q] So now when we use `getDestination`, we'll always have to call `assertHttpDestination` for http destinations, before calling execute?

packages/connectivity/src/scp-cf/destination/destination.ts
```diff
@@ -181,55 +179,20 @@ export function getDestinationConfig(
...
  * @returns String containing information on the destination.
  */
 export function toDestinationNameUrl(
-  destination: DestinationOrFetchOptions
+  destination: DestinationOrFetchOptions<Destination>
 ): string {
   return isDestinationFetchOptions(destination)
     ? `name: ${destination.destinationName}`
-    : `name: ${destination.name}, url: ${destination.url}`;
+    : `name: ${destination.name} ${destination.url ? ', url: ' : ''}${
```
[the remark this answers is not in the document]
FrankEssenberger: I guess if you give a full destination from the service the name is defined. If you build it on your own, maybe not. I make it more forgiving - have a look.

packages/connectivity/src/scp-cf/destination/destination.ts
FrankEssenberger: In principle I like the idea to use generic as part of the name as in `FetchOptionsOr<Destination>` however this is not used anywhere else.

FrankEssenberger: I use two types now

fwilhe:
I assume this implies breaking changes for users who have variables of type Destination. Please describe how to upgrade in the `V3-Upgrade-Guide.md`

fwilhe:
looks good to me so far, please take care of the upgrade guide then it should be good from my perspective

packages/connectivity/src/scp-cf/destination/destination-from-registration.spec.ts
```diff
@@ -38,14 +43,22 @@ describe('register-destination', () => {
     unmockDestinationsEnv();
   });
 
-  it('registers destination and retrieves it', async () => {
+  it('registers http destination and retrieves it', async () => {
```
florian-richter, line 46:
```suggestion
  it('registers HTTP destination and retrieves it', async () => {
```
[pp]

packages/connectivity/src/scp-cf/destination/destination-service-types.ts
```diff
@@ -238,14 +238,44 @@ export type DestinationRetrievalOptions = CachingOptions & {
...
+): asserts destination is HttpDestination {
+  if (!isHttpDestination(destination)) {
+    if (!destination.url) {
+      throw new Error(
+        `The 'url' property is not set for destination ${destination.name} which is mandatory if you use it as an 'HTTP destination`
+      );
+    }
+    throw new Error(
+      `The 'type' property is  ${destination.type} instead of  HTTP for destination '${destination.name}' which is mandatory if you use it as an 'HTTP destination`
+    );
```
florian-richter, line 262:
```suggestion
    throw new Error(destination.url ? 
      `The 'type' property is  ${destination.type} instead of  HTTP for destination '${destination.name}' which is mandatory if you use it as an 'HTTP destination` : `The 'url' property is not set for destination ${destination.name} which is mandatory if you use it as an 'HTTP destination`
    );
```
[pp]

packages/connectivity/src/scp-cf/destination/destination-service-types.ts
```diff
@@ -238,14 +238,44 @@ export type DestinationRetrievalOptions = CachingOptions & {
...
+ * @returns Boolean.
  */
-export function isDestination(destination: any): destination is Destination {
-  return destination.url !== undefined;
+export function isHttpDestination(
+  destination: HttpDestination | Destination | null
+): destination is HttpDestination {
+  return (
+    !!destination &&
+    !!destination.url &&
```
florian-richter, line 276: [q] Could this return an incorrect result, e.g. if the `destination.url` is an empty string?
```
{
  url: "",
  type: "HTTP"
}
```

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -123,9 +106,9 @@ async function fetchDestinations(
     headers
   )
     .then(response => {
-      const destinations: Destination[] = response.data
-        .filter(isParsable)
-        .map(parsableDestination => parseDestination(parsableDestination));
+      const destinations: Destination[] = response.data.map(
+        parsableDestination => parseDestination(parsableDestination)
```
florian-richter, line 110:
```suggestion
        destination => parseDestination(destination)
```
[nth] `isParsable` is not used to filter so this should be named accordingly.

packages/connectivity/src/scp-cf/destination/destination-service-types.ts
florian-richter: [q] Previously it looked like we also supported `Type` (with capital T). If this changed, it should be mentioned in the compatibility note.

FrankEssenberger: This should be the same. I think in the past we did already an assertion before the transformation of the destination. Not it is done later on when the `type` property is already in lower case.

packages/odata-common/src/request-builder/request-builder-base.ts
```diff
@@ -56,6 +59,7 @@ export abstract class MethodRequestBuilder<
    * @param middlewares - Middlewares to be applied to the executeHttpRequest().
    * @returns The request builder itself, to facilitate method chaining.
    */
+
```
florian-richter, line 62:
```suggestion
```
[nth]

test-resources/test/test-util/request-mocker.ts
```diff
@@ -11,7 +11,7 @@ import {
 } from '@sap-cloud-sdk/odata-common/internal';
 import { createODataUri as createODataUriV2 } from '@sap-cloud-sdk/odata-v2/internal';
 import { createODataUri as createODataUriV4 } from '@sap-cloud-sdk/odata-v4/internal';
-import { Destination } from '@sap-cloud-sdk/connectivity';
+import { HttpDestination,Destination} from '@sap-cloud-sdk/connectivity/internal';
```
florian-richter, line 14:
```suggestion
import { HttpDestination, Destination } from '@sap-cloud-sdk/connectivity/internal';
```
[req] formatting

test-resources/test/test-util/request-mocker.ts
```diff
@@ -33,7 +33,7 @@ export const defaultRequestHeaders = {
 
 export const defaultDestinationName = 'Testination';
 
-export const defaultDestination: Destination = {
+export const defaultDestination:HttpDestination = {
```
florian-richter, line 36:
```suggestion
export const defaultDestination: HttpDestination = {
```
[req] formatting

florian-richter (changes requested):
Some formatting otherwise LGTM

V3-Upgrade-Guide.md
```diff
@@ -60,6 +60,8 @@ The following sub-sections describe affected modules, functions and interfaces w
 ### Package `@sap-cloud-sdk/connectivity` <!-- omit from toc -->
 
 - The generic type parameter `JwtKeysT` in `JwtKeyMapping` is now narrowed to extend `string`.
+- The property `url` on the `Destination` interface becomes optional.
+It is only present for HTTP destinations and not for Mail destinations.
```
florian-richter, line 64:
```suggestion
  It is only present for HTTP destinations and not for Mail destinations.
```
[q] I assume this is meant to be part of the bullet point. Without the leading spaces it would render as a line below the list.

florian-richter (approved):
LGTM (with one minor point in the upgrade guide to be changed)

PR #3276, opened 2022-12-29
chore: Fix typos

marikaner:
Fix some typos + minor lint fixes

### DIFF
--- a/packages/mail-client/src/mail-client-types.ts
+++ b/packages/mail-client/src/mail-client-types.ts
@@ -267,7 +267,7 @@ export interface MailDestination {
  */
 export interface SDKOptions {
   /**
-   * Option to define the strategy of sending emails. The emails will be sent in parallel when setting to `true`, otherwise in sequential. The default value is `true`.
+   * Option to define the strategy of sending emails. The emails will be sent in parallel when set to `true`, otherwise in sequential. The default value is `true`.
    */
   parallel?: boolean;
 }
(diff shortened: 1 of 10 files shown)


packages/mail-client/src/mail-client-types.ts
```diff
@@ -267,7 +267,7 @@ export interface MailDestination {
  */
 export interface SDKOptions {
   /**
-   * Option to define the strategy of sending emails. The emails will be sent in parallel when setting to true, otherwise in sequential. The default value is true.
+   * Option to define the strategy of sending emails. The emails will be sent in parallel when setting to `true`, otherwise in sequential. The default value is `true`.
```
deekshas8, line 270:
```suggestion
   * Option to define the strategy of sending emails. The emails will be sent in parallel when set to `true`, otherwise in sequential. The default value is `true`.
```

packages/mail-client/src/mail-client-types.ts
```diff
@@ -277,7 +277,7 @@ export interface SDKOptions {
  */
 export interface MailClientOptions extends SmtpTransportOptions {
   /**
-   * Defines the SDK behaviours, for example whether the mails are sent in parallel.
+   * Defines, whether mails are sent in parallel.
```
deekshas8, line 280:
```suggestion
   * Defines whether mails are sent in parallel.
```

scripts/check-public-api.spec.ts
```diff
@@ -138,7 +138,7 @@ describe('check-public-api', () => {
       ]);
     });
 
-    it('parses barrelfile file without matching', () => {
+    it('parses barrel file file without matching', () => {
```
deekshas8, line 141:
```suggestion
    it('parses barrel file without matching', () => {
```

deekshas8 (approved):
Noticed few minor typos.

PR #3278, opened 2022-12-29
chore!: Replace enum `IsolationStrategy` with string literal union type

marikaner:
Replacing enum with strings.

Relates to SAP/cloud-sdk-backlog#1014.


deekshas8 (approved):
LGTM.

PR #3280, opened 2022-12-29
chore: Remove internal enum representations

marikaner:
The internal enums brought more complexity than value. In this PR I replaced them with simple literal string values.

Relates to [SAP/cloud-sdk-backlog#1014 ](https://github.com/SAP/cloud-sdk-backlog/issues/1014)


fwilhe:
Nice thing getting rid of the enums.

Not wanting to bike shed, but I'm wondering if [this alternative pattern](https://youtu.be/0fTdCSH_QEU?t=288) could make sense for us? I think this would preserve the nice aspects of enums without their downsides (as far as I understand with my limited typescript knowledge)

cf [playground here](https://www.typescriptlang.org/play?#code/KYOwrgtgBACgTgewB4E8DKAXOBDDwDmKUA3gFBRQByA8gPowBK1AGgJoA05U1l9DAogFkAkmn58WHLsMoAVfg0r9ZEtpwqNhANQCC8gDIyA0qtakAvqVIBjBCADOGWIlSYceQpWAB3KAF4SLhpTAC4oAHIQBABaAAcXFHD1bl5GIVFxRkkw8Ls4uGAIAEt7YCTpOQUlFSy2HKKQPDgQYAxyjQZtPX5DShNa1hz4ooA3XGBogBsGgGtwiyhseyhbBwwrDBRY4GdkdCxxzx9-KE3thAAzXdcDjxQvbwBtGeAUS9Ot4Hf4PbdD+58AF0rKQLmAQNYMEU7FAACYIDAACwa+AAjAAKJBhH43dwEFAASkCFAoq3sCEmwAAdJMEPhMQSLFYwRCoTD4UiUQAmTHYhJ-O4PIlkEkrOzkyk0ukMpmkDnIkBo9E4-Z4whUmTyRTKUyM+Xc5X8274h5U4IDRlAA)

packages/connectivity/src/scp-cf/destination/http-proxy-util.ts
```diff
@@ -19,33 +19,35 @@ const logger = createLogger({
...
  * @internal
+ * Determines the proxy strategy. If the 'no_proxy' env variable is set, the `ProxyConfiguration` in the destination is omitted.
+ * For the 'on-premise' and 'internet' proxy strategies the connectivity service or environment variables are checked to fill the `ProxyConfiguration`.
+ * @param destination - Destination to derive the proxy strategy from.
+ * @returns The proxy strategy for the given destination.
  */
-export function proxyStrategy(destination: Destination): ProxyStrategy {
+export function proxyStrategy(
+  destination: Destination
+): 'no-proxy' | 'on-premise' | 'internet' | 'private-link' {
```
deekshas8, line 30: [q] Should we not create these as types?

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -105,7 +94,7 @@ function isParsable(
 async function fetchDestinations(
   destinationServiceUri: string,
   jwt: string,
-  type: DestinationType,
+  type: 'instance' | 'subaccount',
```
FrankEssenberger, line 97: [pp] For the `DestinationType` and `ProxyStrategy` you could introduce a dedicated type even if they are not used multiple time. No strong opinion.

FrankEssenberger (approved):
LGTM only minor PP

packages/connectivity/src/scp-cf/destination/http-proxy-util.ts
marikaner: I added a `ProxyStrategy` type.

packages/connectivity/src/scp-cf/destination/destination-service.ts
marikaner: I added a type for `ProxyStrategy` because it is used as a return type there, but I didn't do it for the `DestinationType` because it really is more of a fancy boolean here. If needed this can be added at any time.

PR #3295, opened 2023-01-02
chore: Update tj-actions version

deekshas8:
This PR updates version of tj-actions/changes-files action.


fwilhe:
Question: Do we need to update this manually? Should be handled by [dependabot via this config](https://github.com/SAP/cloud-sdk-js/blob/v2-main/.github/dependabot.yml#L42)

deekshas8:
> Question: Do we need to update this manually? Should be handled by [dependabot via this config](https://github.com/SAP/cloud-sdk-js/blob/v2-main/.github/dependabot.yml#L42)

 v35 was released 2 weeks ago and it wasn't updated automatically by the config. So I am not sure if the config has an issue or it's not used for updating this. Will take a look.

fwilhe (approved):

PR #3302, opened 2023-01-03
fix: Update Request Builder filters out updates to null

fwilhe:
Relates to SAP/cloud-sdk-backlog#1003.


FrankEssenberger (commented):
I think if you use `closes`in the PR description it will close the ticket. But I see you also want to add this to `main` - so better to use `relates` or so.

packages/odata-v2/src/request-builder/update-request-builder.spec.ts
```diff
@@ -140,6 +143,46 @@ describe('UpdateRequestBuilder', () => {
     expect(actual).toEqual(entity.setOrInitializeRemoteState());
   });
 
+  it('update request is executed when nullable string is set to null', async () => {
```
FrankEssenberger, line 146: [pp] I like to have the `it` to be part of the sentence.

packages/odata-v2/src/request-builder/update-request-builder.spec.ts
```diff
@@ -208,7 +258,11 @@ describe('UpdateRequestBuilder', () => {
 
   it('update request should contain version identifier when set on entity', async () => {
     const entity = createTestEntity().setVersionIdentifier('not-a-star');
-    const requestBody = { Int32Property: entity.int32Property };
+    const requestBody = {
+      Int32Property: entity.int32Property,
+      StringProperty: null,
```
FrankEssenberger, line 263: [q] Why are the two options set to null here?

FrankEssenberger (changes requested):
Currently I am afraid that we send null values if a user does not touch the `nullable` values. I would assume that this should not be done per default, only if you explicitly set the value to null. So you need the state: set to value (includes null) and not set. @marikaner is my understanding correct? I would like to see also a E2E test using our CDS local app. This would increase my trust. The E2E test should cover:
- create an entity with property A=123
- update property B do not mention property A -> should remain 
- update a property A=null -> should become null

packages/odata-v2/src/request-builder/update-request-builder.spec.ts
fwilhe: Would it be bad?

Yes this changes that all null values are being sent where before none has been sent.

My understanding was that this is nock specific because it limits what the 'server' reacts to. My assumption is that real servers don't care about additional null fields. If the value was null before it does not change anything, if the value was not null before you wanted to set it to null.

Not sure how we could avoid that if we don't add another parameter for 'null fields I want to update'.

packages/odata-v2/src/request-builder/update-request-builder.spec.ts
```diff
@@ -140,6 +143,86 @@ describe('UpdateRequestBuilder', () => {
     expect(actual).toEqual(entity.setOrInitializeRemoteState());
   });
 
+  it('executes the update when nullable string is set to null', async () => {
+    // Given: Entity with only key properties and one nullable string property which has non-null value
+    const keyPropGuid = uuid();
```
FrankEssenberger, line 148: [nth] consider a local helper function to build the entity and do the mock perhaps.

FrankEssenberger (approved):
LGTM only a minor comment.

packages/odata-v2/src/request-builder/update-request-builder.spec.ts
fwilhe: I did think about this but I'm not sure. We do already have the create test entity helper, adding just a second version of that probably would not be helpful as we would need a new function for every variation of the test entity. Maybe there would be a good way to make that configurable, but this sounds like too complex code for simple test object creation imo.

fwilhe: Chose to extract it into a method

PR #3308, opened 2023-01-04
feat!: Add skip validation

FrankEssenberger:
Relates  SAP/cloud-sdk-backlog#976.

Adds a `skipValidation` method to the OData generator. This is disabled per default and generations will fail if a name needs to be changed due to a conflict with a reserved keyword or non uniqness. If the user enables `skipValidation` generation works and the the name is changed.


.changeset/loud-starfishes-type.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-cloud-sdk/generator': major
+---
+
+[Compatibility Note] Per default the OData generator does not change object names anymore to resolve conflicts.
+Instead, the generation process will fail.
+Switch on the `skipValidation` flag if you want to generate anyway with adjusted names.
```
fwilhe, line 7:
```suggestion
[Compatibility Note] The OData generator won't automatically rename identifiers to avoid name conflicts.
The generation process will fail if identifiers have conflicting names.
Switch on the `skipValidation` flag if you want to generate despite name conflicts and are okay with changed identifier names to avoid conflicts.
```

fwilhe: hope this is better to understand

fwilhe (commented):
looks good to me so far, I have a suggestion for the changelog to make it more readable

fwilhe (approved):
minor comment, looks good to me

PR #3316, opened 2023-01-05
chore: Configure lint to not allow enums

fwilhe:
We got rid of enums in 0bc120593d66f6dc198d27f58e0ceb4547625cfa because we learned that they behave in unexpected ways in TypeScript and don't add much value. This lint rule will make us aware of this in the future in case someone accidentally wants to introduce new enums.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,6 +57,13 @@ module.exports = {
     }
   ],
   rules: {
-    radix: 'off'
+    radix: 'off',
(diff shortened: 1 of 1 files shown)


marikaner (approved):
"weird" lol 😊

FrankEssenberger (approved):

fwilhe:
> "weird" lol 😊

want to suggest a more 'enterprise' phrasing? :)

PR #3309, opened 2023-01-04
chore!: Remove csn generation

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#991.
 
Remove the `generateCSN` option.

PR #3319, opened 2023-01-05
chore!: Remove Remaining Deprecated Options

FrankEssenberger:
Relates  SAP/cloud-sdk-backlog#978.

Removed the remaining deprecated options after in #3312 and #3309 the first bunch was removed.

### DIFF
--- a/V3-Upgrade-Guide.md
+++ b/V3-Upgrade-Guide.md
@@ -70,9 +70,9 @@ The options `versionInPackageJson` and `licenseInPackageJson`, that allowed sett
 When the `packageJson` option is enabled, a `package.json` file with version `1.0.0` and license `UNLICENSED` is created.
 If you want to set custom values, use the `include` option to add a custom `package.json`.
 
-The following deprecated options were removed with a clear replacement:
+The following deprecated options were renamed:
 - `forceOverwrite` becomes `overwrite`
-- `generatePackageJson` becomes `overwrite`
+- `generatePackageJson` becomes `packageJson`
 - `writeReadme` becomes `readme`
 - `processesJsGeneration` becomes `transpilationProcesses`
 
--- a/V3-Upgrade-Guide.md
+++ b/V3-Upgrade-Guide.md
@@ -77,8 +77,7 @@ The following deprecated options were renamed:
 - `processesJsGeneration` becomes `transpilationProcesses`
 
 The deprecated `generateNpmrc` option of the generator is removed. 
-In most cases the `.npmrc` was present in the project root anyhow. 
-If this is not the case, use the `include` option to add a `.npmrc` to the generated code if needed.
+Use the `include` option to add a `.npmrc` to the generated code if needed.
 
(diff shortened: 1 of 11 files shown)


V3-Upgrade-Guide.md
```diff
@@ -70,6 +70,19 @@ The options `versionInPackageJson` and `licenseInPackageJson`, that allowed sett
 When the `packageJson` option is enabled, a `package.json` file with version `1.0.0` and license `UNLICENSED` is created.
 If you want to set custom values, use the `include` option to add a custom `package.json`.
 
+The following deprecated options were removed with a clear replacement:
```
fwilhe, line 73:
```suggestion
The following deprecated options were renamed:
```

V3-Upgrade-Guide.md
```diff
@@ -70,6 +70,19 @@ The options `versionInPackageJson` and `licenseInPackageJson`, that allowed sett
...
+- `forceOverwrite` becomes `overwrite`
+- `generatePackageJson` becomes `overwrite`
+- `writeReadme` becomes `readme`
+- `processesJsGeneration` becomes `transpilationProcesses`
+
+The deprecated `generateNpmrc` option of the generator is removed. 
+In most cases the `.npmrc` was present in the project root anyhow.
```
fwilhe, line 80: why? is this something the user needs to do? not clear to me.

FrankEssenberger: We used the .npmrc to set the repository url for all `@sap/` dependencies to the internal NPM this is not needed anymore. In general I think most of the options of the `npmrc` apply to the whole project so it is better to put it in the project root, instead of having configs in subsfolders which only apply if you execute the install in this folder. I will remive the sentence and just say -> use include now.

fwilhe:
from viewing the diff its not clear to me how the rename of option works, like `writeReadme` to `readme` because I see the first removed but the second not added. I guess that's some missing context which I can't see in the diff and the second option already existed.

In general the PR looks good to me, but I have not tested functional correctness in any way. Do we have integration tests for this?

fwilhe (approved):
feel free to merge when you trust its good @FrankEssenberger 

note my comment above on context and functional correctness

FrankEssenberger:
> from viewing the diff its not clear to me how the rename of option works, like `writeReadme` to `readme` because I see the first removed but the second not added. I guess that's some missing context which I can't see in the diff and the second option already existed.
> 
> In general the PR looks good to me, but I have not tested functional correctness in any way. Do we have integration tests for this?

@marikaner added a [nice class](https://github.com/SAP/cloud-sdk-js/blob/7340703e051f20670d1d4283a622d209b852c1b4/packages/generator/src/options-parser.ts#L98) which maps the old options to the new ones. For each property there was the [replacedBy](https://github.com/SAP/cloud-sdk-js/blob/7340703e051f20670d1d4283a622d209b852c1b4/packages/generator/src/generator-options.ts#L219) which took care of this.

Issue #3349, opened 2023-01-12
Cloud SDK fails to take subscriber subdomain when requesting token for connectivity service


ww917352:
**Describe the bug**

When we use Cloud SDK to execute a request with a destination configured with a cloud connector, due to lack of an original access token, we have to set the `tokenIssueUrl` in the request, and the provided URL is the token URL for the subscriber sub account, not the provider sub account. However, Cloud SDK fetches a token for the provider sub account.

We found the problematic code to be here

https://github.com/SAP/cloud-sdk-js/blob/a78f4a91a6ff06d7763e615df5e365f3ae0293e6/packages/connectivity/src/scp-cf/connectivity-service.ts#L120-L133

**To Reproduce**

Use `setTokenIssuelUrl` instead of passing a token to the execution with a OnPrem destination configured with a cloud connector

**Expected behavior**

It should execute the request.

**Used Versions:**

The latest version of Cloud SDK `"@sap-cloud-sdk/connectivity": "2.13.0"`

**Impact / Priority**

We are facing high customer incidents due to this issue.

ww917352:
@FrankEssenberger 

Via remote debugging, I can say for sure that `jwt` is undefined when `proxyAuthorizationHeader` is called.

Can you help further? Thanks.

FrankEssenberger:
The idea is to set the `iss` property directly `getDestination` or `execute` call:
```ts
.execute({
   destinationName:'yourDest',
   iss:'http://yourSubscriberDomain.localhost:8080/uaa/oauth/token'})
```
The SDK then build a dummy JWT from this value and this should be passed to the `proxyAuthorizationHeader` method. 

Best
Frank

FrankEssenberger:
Hi @ww917352 the version [2.14.0](https://github.com/SAP/cloud-sdk-js/releases/tag/v2.14.0) containing the fix is release. Please test and let me know if this fixes the issue on your side.

PR #3351, opened 2023-01-12
fix: Fix ISS for OnPremise Basic

FrankEssenberger:
relates https://github.com/SAP/cloud-sdk-js/issues/3349

Adds the subscriber sevice token containing the proper subdomain in case of `iss`property with onPrem destination.


packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -532,7 +532,9 @@ Possible alternatives for such technical user authentication are BasicAuthentica
       case 'on-premise':
         return addProxyConfigurationOnPrem(
           destination,
-          this.subscriberToken?.userJwt
+          this.subscriberToken?.userJwt || this.subscriberToken?.type === 'iss'
```
florian-richter, line 535:
```suggestion
          this.subscriberToken?.userJwt ?? this.subscriberToken?.type === 'iss'
```
[nth]

florian-richter (approved):

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
FrankEssenberger: somehow the tenary operator was behaving strange so I refactored.

florian-richter: `??` is nullish coalescing (like `||` but only returns the right hand value on `null` or `undefined`). Ternary is `?`

PR #3382, opened 2023-01-17
chore: Fix `getPackageVersion` Function

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#1028.

While investigating the issue I saw that we replaced the `currentVersion` constant with a method - however the method `getPackageVersion` was not invoked. This PR fixes this and also execute the prettier on the scripts folder.


tomfrenken:
All these changes LGTM, but do you think this will fix the issue mentioned in https://github.com/SAP/cloud-sdk-backlog/issues/1028 ?

scripts/check-public-api.ts
```diff
@@ -118,15 +118,16 @@ export async function checkApiOfPackage(pathToPackage: string): Promise<void> {
...
-        {
-          compilerOptions: opts,
-          //We have things in our sources like  #!/usr/bin/env node in CLI .js files which is not working with parser of prettier.
-          createFileOptions:{overwrite:true,prettierOptions:defaultPrettierConfig,usePrettier:false}
-        }
-    );
+    console.log(JSON.stringify(opts));
+    await transpileDirectory(pathToSource, {
+      compilerOptions: opts,
+      //We have things in our sources like  #!/usr/bin/env node in CLI .js files which is not working with parser of prettier.
```
florian-richter, line 124:
```suggestion
      // We have things in our sources like  `#!/usr/bin/env node` in CLI `.js` files which is not working with parser of prettier.
```
[nth]

florian-richter (approved):
LGTM

FrankEssenberger:
> All these changes LGTM, but do you think this will fix the issue mentioned in [SAP/cloud-sdk-backlog#1028](https://github.com/SAP/cloud-sdk-backlog/issues/1028), or is this a separate PR?

No we use the `v2-main` branch there. I could not reproduce the issue to be honest.

PR #3383, opened 2023-01-17
chore: Run yarn install with 'ignore engine' flag in ci

fwilhe:
Not all dependencies specify yet that they run on node 18.

This change should allow us to merge this dependabot pr: https://github.com/SAP/cloud-sdk-js/actions/runs/3934981468/jobs/6730242461


florian-richter:
We had this in the past, but removed it to make sure we are aware if problems like this exist. Should we have an additional check (that is optional in the PR checks) to validate that our dependencies correctly specify their engine support?

PR #3391, opened 2023-01-18
chore: Csrf Middleware

FrankEssenberger:
Relates to: https://github.com/SAP/cloud-sdk-backlog/issues/1043

This PR contains the ADR and the implementation.


florian-richter:
The two solutions at the bottom seem like bad alternatives to me. The middleware (+ disabled CSRF by default) that is otherwise proposed is a very clean solution. I would prefer to follow the middleware approach and delay it to a later major version if we can't add that in version 3. Maybe version 4 can already be released in 6 months.

FrankEssenberger:
> The two solutions at the bottom seem like bad alternatives to me. The middleware (+ disabled CSRF by default) that is otherwise proposed is a very clean solution. I would prefer to follow the middleware approach and delay it to a later major version if we can't add that in version 3. Maybe version 4 can already be released in 6 months.

I left the idea in the end just for the discussion - I think @marikaner and I also liked the `csrf` middleware best.

knowledge-base/adr/0038-csrf-middleware.md
```diff
@@ -0,0 +1,181 @@
...
+Note that the `fn` may contain already middleware layers from before.
+Same is true for the expiring access tokens from a destination cache.
+
+An idea could be, to add the argument type and a reference to the `fn`:
+
+```ts
+interface Function<ReturnType,ArguemntType> {
+
+    (): Promise<ReturnType>,
+    arg: ArguemntType
```
marikaner, line 48: I am wondering why this has to be a separate property. Could it be in the fn signature?

`(arg: ArgumentType): Promise<ReturnType>` ?

FrankEssenberger: We make it a argument of the function and add a reference to the context so that it can be changed.

knowledge-base/adr/0038-csrf-middleware.md
```diff
@@ -0,0 +1,106 @@
+## Status
+
+accepted
+
+### Structure of the Code Before Cahnge
```
fwilhe, line 5:
```suggestion
### Structure of the Code Before Change
```

knowledge-base/adr/0038-csrf-middleware.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Currently, the middleware is only added to the `executeFn()` to the target system.
+For the fetch request it is not considered.
+Note that any kind of resilience entering via the middleware is not considered.
+
+### General Improvement - Disable Fetch CSRF by Default
+
+Not all systems require a CSRF token.
+We follow also the trend of optional functionality not being enabled e.g. cache.
```
fwilhe, line 29:
```suggestion
We follow also the trend of optional functionality not being enabled by default, like cache.
```

knowledge-base/adr/0038-csrf-middleware.md
```diff
@@ -0,0 +1,106 @@
...
+Currently, the CSRF token fails gracefully, and we continue this means we may have unnecessary calls.
```
fwilhe, line 30:
```suggestion
Currently, the CSRF token fails gracefully, and the SDK retries to get one, this means we may have unnecessary HTTP requests.
```

knowledge-base/adr/0038-csrf-middleware.md
```diff
@@ -0,0 +1,106 @@
...
+
+Currently, the middleware is only added to the `executeFn()` to the target system.
+For the fetch request it is not considered.
+Note that any kind of resilience entering via the middleware is not considered.
+
+### General Improvement - Disable Fetch CSRF by Default
+
+Not all systems require a CSRF token.
+We follow also the trend of optional functionality not being enabled by default, like cache.
+Currently, the CSRF token fails gracefully, and the SDK retries to get one, this means we may have unnecessary HTTP requests.
```
marikaner, line 30: [req] "CSRF token fails" => "CSRF token fetching fails"
```suggestion
Currently, when the CSRF token fetching fails, the SDK retries the fetch with an adjusted URL.
This means we may have unnecessary HTTP requests.
```

knowledge-base/adr/0038-csrf-middleware.md
```diff
@@ -0,0 +1,106 @@
...
+If the default is off, one should really fail.
```
marikaner, line 31: [q] What is "one"? Is it "the fetching"?

knowledge-base/adr/0038-csrf-middleware.md
```diff
@@ -0,0 +1,106 @@
...
+- dedicated timeout for the actual request and csrf fetching
+- circuit breaker instance based on destination URL will record actual request and csrf fetching
+- retry is not enabled per default
+
+In case a user need something different they can pass their own `csrf()` middleware to adjust the method or `URL`.
+
+## Migration Discussion
+
+- Too late in version 3 development to change the CSRF default behavior.
+- Deprecate `skipCsrfTokenFetchion()` in version 3 and remove it in version 4.
```
marikaner, line 104: [req] I assume this is meant: https://sap.github.io/cloud-sdk/api/v2/interfaces/sap_cloud_sdk_http_client.HttpRequestOptions.html#fetchCsrfToken
```suggestion
- Deprecate `fetchCsrfToken` in version 3 and remove it in version 4.
```

packages/http-client/src/csrf-token-middleware.ts
```diff
@@ -0,0 +1,199 @@
...
+  HttpRequestConfig,
+  HttpRequestConfigBase,
+  HttpRequestConfigWithOrigin,
+  Method
+} from './http-client-types';
+
+const logger = createLogger('csrf-middleware');
+
+/**
+ * Options for middleware to fetch csrf tokens.
```
marikaner, line 25:
```suggestion
 * Options for middleware to fetch CSRF tokens.
```

packages/http-client/src/csrf-token-middleware.ts
```diff
@@ -0,0 +1,199 @@
...
+ */
+export interface CsrfMiddlewareOptions {
+  /**
+   * Method used for the token fetching. Default is `head`.
+   */
+  method?: Method;
+  /**
+   * Url used for the token fetching. Default is the resource path without parameters.
```
marikaner, line 33:
```suggestion
   * URL used for the token fetching. Default is the resource path without parameters.
```

packages/http-client/src/csrf-token-middleware.ts
```diff
@@ -0,0 +1,199 @@
...
+      csrfToken.cookie = requestConfig.headers?.cookie
+        ? [requestConfig.headers?.cookie, csrfToken?.cookie].join(';')
+        : csrfToken?.cookie;
+    }
+    requestConfig.headers = { ...requestConfig.headers, ...csrfToken };
+    return middlewareOptions.fn(requestConfig);
+  };
+}
+
+function noActionNeeded(requestConfig: HttpRequestConfig): boolean {
```
marikaner, line 67: [req] According to our style guide a function should start with a verb. So, this would probably be: `isNoActionNeeded`.
[pp] Action is very general and the negation in the name adds unnecessary potential for confusion in my opinion. How about returning the opposite boolean and calling it `needsCsrfToken`?

packages/http-client/src/csrf-token-middleware.ts
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Middleware for fetching a CSRF token. This middleware is added to all request per default.
+ * Use the `fetchCsrfToken` option to disable it.
+ * @param options - Options like URL or method to configure the token fetching.
+ * @returns The middleware for fetching CSRF tokens.
+ */
+export function csrf(options?: CsrfMiddlewareOptions): HttpMiddleware {
+  return (middlewareOptions: HttpMiddlewareOptions) => async requestConfig => {
+    if (noActionNeeded(requestConfig)) {
```
marikaner, line 50: [pp] I prefer to test the positive case to avoid confusion through potential double negation. See my comments below, I would find it more clear like this:
```suggestion
    if (!needsCsrfToken(requestConfig)) {
```

marikaner (approved):
I left some comments, but nothing major at all. Looks really good!

knowledge-base/adr/0038-csrf-middleware.md
FrankEssenberger: There are two optins: `skipCsrfTokenFetching` on the typed clients and `fetchCsrfToken` on the http-client.

FrankEssenberger: I removed the typo and included them both.

PR #3384, opened 2023-01-17
chore: Upgrade yarn lock file to update dependencies

fwilhe:
This bumps `@sap/hdi-deploy` to version 4.5.1 which is compatible with node 18

Pin turbo to 1.6.3 because turbo 1.7 causes issues with the build as of now.

This should allow us to merge [this pr](https://github.com/SAP/cloud-sdk-js/pull/3372)

PR #3400, opened 2023-01-19
chore: Update compiled code of github action

fwilhe:
Not sure why this was not done in https://github.com/SAP/cloud-sdk-js/pull/3384


FrankEssenberger:
I think I execute the `prettier` on the `scripts folder which seems to also change the index.js I guess. Then you re-compile it which leads a not prettified version. This could be the reason - perhaps we can include the prettier in the compile script or exclude it in the `.prettierignore`

FrankEssenberger (approved):

PR #3410, opened 2023-01-20
chore: Generate one API doc per major version

florian-richter:
Part of SAP/cloud-sdk-backlog#948.
Related to SAP/cloud-sdk#1226.

API docs are now moved to `knowledge-base/api-reference` instead of `docs/api`. This new folder contains the generated API docs for the code on this commit. So on `v2-main` this will be the v2 docs, on `main` this will currently be `main`.

The GIthub Actions script should now run on both `main` and `v*-main` (so not `1.0-main` but that is probably fine as there will be no further v1 releases). Once a new API docs is generated it will overwrite the corresponding version in the docs repo (other versions in the docs repo will be left untouched).

I also removed the part of the `generate-docs` script that removed the `.html` endings from links since that broke local deployments and it should work with them on Github Pages (at least in my manual tests). It also simplifies the script 😊 

The same changes will need to be applied to `main`, but I will open that once this PR is ready to merge.


FrankEssenberger (approved):
LGTM

PR #3414, opened 2023-01-23
feat: Add Function Arguments to Middleware

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#1042.

Add the function arguments to the middleware context so that users can change the value of function pparameters in the middleware easily.

I also moved the `HttpMiddlewareContext` to the HttpClient where it belonged in my opinion because it is introduced there and renamed `Context` to `MiddlewareContext` to give it more context :-).


packages/resilience/src/middleware.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  it('adds middlewares in the expected order', async () => {
+    const infoSpy = jest.spyOn(logger, 'info');
+    const actual = await executeWithMiddleware(
+      [middleWareAppend1, middleWareAppend2],
+      { uri: 'dummyUri', fnArgument: 'initial Input', tenantId: 'dummyTenant' },
+      (s: string) => Promise.resolve(s)
+    );
+    expect(infoSpy).toHaveBeenNthCalledWith(2, 'append1');
+    expect(infoSpy).toHaveBeenNthCalledWith(1, 'append2');
+    expect(actual).toBe('initial Input append2 append1');
```
deekshas8, line 53: [q] I am not sure I understand, but shouldn't `append1` be added first?

FrankEssenberger: The functions are like a stack so the one added first is executed last.

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -411,14 +413,22 @@ async function callDestinationService(
   const resilienceMiddleware =
     destinationName && retry
       ? [
-          ...resilience<AxiosResponse, HttpMiddlewareContext>(),
+          ...resilience<
+            RawAxiosRequestConfig,
+            AxiosResponse,
+            MiddlewareContext<RawAxiosRequestConfig>
+          >(),
```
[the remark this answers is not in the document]
FrankEssenberger: We use it also for the XSUAA requests which have different arguments, I tried to make the generic not so often.

packages/connectivity/src/scp-cf/xsuaa-service.ts
```diff
@@ -54,22 +54,33 @@ export async function getClientCredentialsToken(
   const serviceCredentials = resolvedService.credentials;
   const subdomainAndZoneId = getSubdomainAndZoneId(userJwt);
 
-  const xssecPromise: () => Promise<ClientCredentialsResponse> = () =>
+  const fnArgs = [
```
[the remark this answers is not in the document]
FrankEssenberger: It had a type like `null| string|serviceCredenital []`  so I was not sure what to get from such a broad type -but I did not like it so I will add a object to have the properties  - array is not nice anyway... have a look

packages/connectivity/src/scp-cf/xsuaa-service.ts
```diff
@@ -91,31 +102,37 @@ export function getUserToken(
 ): Promise<string> {
   const subdomainAndZoneId = getSubdomainAndZoneId(userJwt);
 
-  const xssecPromise: () => Promise<string> = () =>
+  const fnArgs = [
```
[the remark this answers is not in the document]
FrankEssenberger: added new function const with different name

.changeset/moody-kids-double.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-cloud-sdk/http-client': minor
+'@sap-cloud-sdk/resilience': minor
+---
+
+[New Functionality] The HTTP request config is not part of the middleware context so that middlewares can change the request config easily.
```
florian-richter, line 6: [nth] I don't think that I understand this sentence. Maybe this could be rephrased to be clearer.

packages/connectivity/src/scp-cf/xsuaa-service.ts
```diff
@@ -50,31 +57,41 @@ export async function getClientCredentialsToken(
...
-  const subdomainAndZoneId = getSubdomainAndZoneId(userJwt);
+  const fnArgument: XsuaaParameters = {
+    ...getSubdomainAndZoneId(userJwt),
+    serviceCredentials: resolveService(service).credentials
+  };
 
-  const xssecPromise: () => Promise<ClientCredentialsResponse> = () =>
+  const xssecPromise: (
+    arg: XsuaaParameters
+  ) => Promise<ClientCredentialsResponse> = arg =>
```
florian-richter, line 67:
```suggestion
  type foo = (arg: XsuaaParameters) => Promise<ClientCredentialsResponse>;
  const xssecPromise: foo = arg =>
```
[nth] I find that definition really hard to read because I overlooked the `=` (assumed it was a `=>`).
`foo` is only a placeholder of course... 😉

florian-richter: [q] Maybe this could even use the `function` keyword instead of assigning an arrow function?
Same applies on line 115

florian-richter (approved):
looks good but it is pretty complicated. We definitely need a user guide and probably also some developer documentation (e.g. updating the ADR with implementation specifics). So I would recommend to create a follow-up ticket for this.

packages/connectivity/src/scp-cf/xsuaa-service.ts
FrankEssenberger: I agree made the same mistake - added the function

FrankEssenberger:
> looks good but it is pretty complicated. We definitely need a user guide and probably also some developer documentation (e.g. updating the ADR with implementation specifics). So I would recommend to create a follow-up ticket for this.

Added a TODO in the parent ticket.

PR #3416, opened 2023-01-23
chore: Use Parser also for OpenApi

FrankEssenberger:
Relates SAP/cloud-sdk-backlog#979.

Just the first step in aligning the two generators. Use the `options-parser.ts` in OData and now also OpenApi.
I also aligned the `Readme.md` files - however there was some new line added when I executed the script.


packages/generator/test/test-util/create-generator-options.ts
```diff
@@ -3,7 +3,7 @@ import {
   cliOptions,
   ParsedGeneratorOptions
 } from '../../src/options';
-import { parseOptions } from '../../src/options/options-parser';
+import { parseOptions } from '../../../generator-common/src/options-parser';
```
florian-richter, line 6: [q] Should this not be imported from `@sap-cloud-sdk/generator-common`?

packages/openapi-generator/src/cli-parser.ts
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env node
```
florian-richter, line 1: [q] I doubt that this will execute with node because of the type annotations. So why do we have this shebang?

packages/openapi-generator/src/options/options-per-service.spec.ts
```diff
@@ -1,3 +1,5 @@
+import {ParsedGeneratorOptions} from "./options";
```
florian-richter, line 1: [req] This needs to be formatted
```suggestion
import { ParsedGeneratorOptions } from "./options";
```

packages/openapi-generator/src/options/tsconfig-json.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { defaultTsConfig } from '@sap-cloud-sdk/generator-common/internal';
 import mock from 'mock-fs';
-import { ParsedGeneratorOptions } from './generator-options';
 import { tsconfigJson } from './tsconfig-json';
+import {ParsedGeneratorOptions} from "./options";
```
florian-richter, line 4: format

packages/openapi-generator/test/test-config-wrong-key.json
```diff
@@ -0,0 +1,5 @@
+{
+  "input": "some-repository",
```
florian-richter, line 2: [q] Should this and the other test config be actual files or could we use mock-fs?

packages/generator/test/test-util/create-generator-options.ts
FrankEssenberger: for things in `src` for sure. For the test it would mean that you need to compile so a new version is in `dist`. Since I execute the tests in the IDE directly in many cases I would like to avoid this.

packages/openapi-generator/src/cli-parser.ts
FrankEssenberger: This was a copy paste from the [OData](https://github.com/SAP/cloud-sdk-js/blob/main/packages/generator/src/cli-parser.ts) parser where a code reuse was difficult - but I will try to add something in the generator-commons.

packages/openapi-generator/test/test-config-wrong-key.json
FrankEssenberger: Since we use `yargs` now the mock did not work. Seems yargs uses not the `fs` module to lead the file or spawns a different process to load the file which also block the mock.

packages/openapi-generator/src/cli-parser.ts
FrankEssenberger: removed it - in the `cli.ts` which is the bin in the package.json it makes sense. Here not I think.

packages/openapi-generator/src/options/tsconfig-json.spec.ts
FrankEssenberger: linted

packages/generator/test/test-util/create-generator-options.ts
florian-richter: Understood. Turborepo does take care of this (and probably compiles this anyways if you run your tests through `yarn test ...` instead of `jest ...`), but I understand the reason.

florian-richter (approved):
LGTM

PR #3425, opened 2023-01-24
chore: Generate one API doc per major version

florian-richter:
Related to #3410 
Related to SAP/cloud-sdk#1226
Related to SAP/cloud-sdk-backlog#948.


packages/connectivity/src/scp-cf/connectivity-service.ts
```diff
@@ -111,7 +111,7 @@ async function proxyHeaders(
 }
 
 async function proxyAuthorizationHeader(
-  jwt?
+  jwt?: string
```
FrankEssenberger, line 114: Good change, but why here?

FrankEssenberger (approved):

packages/connectivity/src/scp-cf/connectivity-service.ts
florian-richter: I don't remember 😅 I think I had that file still open when I started working on this ticket

PR #3431, opened 2023-01-25
chore!: Align Coerece Methods

FrankEssenberger:
Realtes SAP/cloud-sdk-backlog#979.

PR #3436, opened 2023-01-26
chore!: Align service per options

deekshas8:
Closes SAP/cloud-sdk-backlog#974.

- rename `serviceMapping` to `optionsPerService`
- Align option behaviour in both generators 


packages/generator/src/generator.spec.ts
```diff
@@ -226,6 +234,109 @@ describe('generator', () => {
     });
   });
 
+  describe('optionsPerService', () => {
+    beforeEach(async () => {
+      mock({
+        common: {},
+          // '/prettier/config': JSON.stringify({ printWidth: 66 }),
```
FrankEssenberger, line 241: [req] remove comment

packages/generator/src/generator.spec.ts
```diff
@@ -226,6 +234,109 @@ describe('generator', () => {
...
+              npmPackageName: 'test-service'
+            }
+          },
+          null,
+          2
+        )
+      );
+    });
+
+    xit('overwrites writes options per service', async () => {
```
FrankEssenberger, line 282: [q] these are the not working thing right which work in OpenApi - add a TODO and mention the lines in the ticket - TDD :-)

packages/generator/src/options/options.ts
```diff
@@ -139,19 +140,24 @@ export const cliOptions = {
...
     type: 'string',
     coerce: (
       arg: string | undefined,
       options: GeneratorOptions & { config?: string }
-    ) =>
-      resolveRequiredPath(
-        arg ? arg : join(options.inputDir.toString(), 'service-mapping.json'),
-        options
-      )
+    ) => {
```
FrankEssenberger, line 151: add a named method with some small doc explaining what is done.

FrankEssenberger: Also in the OpenApi client you could use this method then.

packages/generator/src/generator.ts
```diff
@@ -190,16 +190,19 @@ export async function generateProject(
...
+  if (options.optionsPerService) {
+    logger.verbose('Generating options per service ...');
 
-  project.createSourceFile(
-    resolve(options.serviceMapping.toString()),
-    serviceMappingFile(services),
-    { overwrite: true }
-  );
+    const dir = dirname(options.optionsPerService);
+    await mkdir(dir, { recursive: true });
```
FrankEssenberger, line 197: [q] What if the folder already exists? I guess covered by tests alrady.

FrankEssenberger: Made a quick test and got an error so please double check with a test 
image

FrankEssenberger (changes requested):
Check that it also works for existing folders and files.

packages/generator/src/generator.ts
deekshas8: I added few more tests, have a look

packages/generator/src/options/options.ts
deekshas8: I'll do this after this [PR](https://github.com/SAP/cloud-sdk-js/pull/3431) is merged, since `resolveRequiredPath` was moved to common.

packages/generator/src/service-mapping.ts
```diff
@@ -1,5 +1,6 @@
 import { parse } from 'path';
-import { unixEOL, createLogger, readJSON } from '@sap-cloud-sdk/util';
```
FrankEssenberger, line 2: [pp] Rename the full file `service-mapping`should go away completely as a term.

packages/generator/src/generator.ts
FrankEssenberger: the `recursive: true` makes the difference. No error.

FrankEssenberger (approved):
LGTM - Please consider my last comment regarding the name of the file `service-mapping.ts` should also be called `options-per-service.ts` or so. Same applies to the `spec` file.

packages/generator/src/service-mapping.ts
deekshas8: Removed all occurrences of the term.

PR #3437, opened 2023-01-26
chore: Rename Middleware types

FrankEssenberger:
While adjusting the middleware documentation I found that our names are not optimal:
- The `MiddleWareIn` and `MiddlewareOut` was not super nice -> Options and Function are better names
- For the public API we should use the `HTTPRequest` type instead of the axios one.


packages/resilience/src/middleware.ts
```diff
@@ -75,8 +75,8 @@ export type Middleware<
   ReturnT,
   ContextT extends MiddlewareContext<ArgumentT>
 > = (
-  options: MiddlewareIn<ArgumentT, ReturnT, ContextT>
-) => MiddlewareOut<ArgumentT, ReturnT>;
+  options: MiddlewareOptions<ArgumentT, ReturnT, ContextT>
```
deekshas8, line 78: [req] Update the jsdoc. It still mentions `MiddlewareIn`

packages/resilience/src/middleware.ts
```diff
@@ -130,7 +130,7 @@ function addMiddlewaresToInitialFunction<
   ContextT extends MiddlewareContext<ArgumentT>
 >(
   middlewares: Middleware<ArgumentT, ReturnT, ContextT>[],
-  initial: MiddlewareIn<ArgumentT, ReturnT, ContextT>
+  initial: MiddlewareOptions<ArgumentT, ReturnT, ContextT>
 ): MiddlewareOut<ArgumentT, ReturnT> {
```
deekshas8, line 134: [q] the return type here can also be changed to `MiddlewareFunction` right?

packages/resilience/src/retry.ts
```diff
@@ -31,7 +31,7 @@ export function retry<
   }
 
   return function (
-    options: MiddlewareIn<ArgumentType, ReturnType, ContextType>
+    options: MiddlewareOptions<ArgumentType, ReturnType, ContextType>
   ): MiddlewareOut<ArgumentType, ReturnType> {
```
deekshas8, line 35: [q] The return type here as well?

packages/resilience/src/middleware.ts
FrankEssenberger: see here: https://github.com/SAP/cloud-sdk/pull/1234

packages/resilience/src/middleware.ts
FrankEssenberger: yes removed them all

packages/resilience/src/retry.ts
FrankEssenberger: see above

packages/resilience/src/middleware.ts
deekshas8: No, I meant the comment in line 62. 
 

> The input is the MiddlewareIn containing the initial function and some context information e.g. axios request and the request context.

packages/http-client/src/http-client-types.ts
```diff
@@ -1,13 +1,12 @@
 import * as http from 'http';
 import type { MiddlewareContext, Middleware } from '@sap-cloud-sdk/resilience';
 // eslint-disable-next-line import/named
```
deekshas8, line 3:
```suggestion
```

packages/http-client/src/http-client.spec.ts
```diff
@@ -239,7 +239,7 @@ describe('generic http client', () => {
       // Doing the dummy object the name of the key is taken as function name.
       const dummy = {
         [appendedText](
-          options: MiddlewareIn<
+          options: MiddlewareOptions<
             RawAxiosRequestConfig,
```
deekshas8, line 243: [req] change to `HttpRequestConfig`

deekshas8 (approved):
Left one suggestion, but otherwise, LGTM.

PR #3438, opened 2023-01-26
chore: Generate v3 docs

florian-richter:
Generate docs with dummy version `3.0.0-alpha.0` until we trigger a release which would automatically regenerate this.


tomfrenken (approved):
LGTM

PR #3442, opened 2023-01-29
chore: Merge actions and functions into operations

YuanZhuang107:
Closes SAP/cloud-sdk-backlog#941.


packages/generator/src/operations/import.ts
```diff
@@ -117,29 +124,41 @@ export function operationImportDeclarations(
...
       ];
 
+  const namedImports = [
+    ...edmRelatedImports(returnTypes),
+    ...complexTypeRelatedImports(returnTypes),
+    ...responseTransformerImports(returnTypes),
+    'DeSerializers',
+    'DefaultDeSerializers',
+    'defaultDeSerializers',
+    ...propertyTypeImportNames(parameters)
```
FrankEssenberger, line 134: [req] The lines below are a clear use case of the tenary operator:
```ts
includesUnbound && hasFunction ? ['FunctionImportRequestBuilder'] : []
```

or which I would find even better, create a little helper function `operationRelatedImport(operation)` which returns a list of string, and you move the code above and below in this helper function and use it like the other methods above `...responseTransformerImports`

packages/generator/src/service/index-file.ts
```diff
@@ -21,11 +21,9 @@ export function indexFile(service: VdmServiceMetadata): SourceFileStructure {
         exportStatement(complexType.typeName)
       ),
       ...service.enumTypes.map(enumType => exportStatement(enumType.typeName)),
-      ...(service.functionImports && service.functionImports.length
-        ? [exportStatement('function-imports')]
-        : []),
-      ...(service.actionImports && service.actionImports.length
-        ? [exportStatement('action-imports')]
+      ...((service.functionImports && service.functionImports.length) ||
```
FrankEssenberger, line 24: [pp] you could use conditional access `service.functionImports?.length` to shorten it.

FrankEssenberger (approved):
Looks good. I think marika had the same comment, but I would like to also align the request and request builder to use a OperationRequestBuilder. For version 3 we could still re-export the other request builders with the old names but deprecate them. Please create a follow up ticket for this.

packages/generator/src/operations/operation.ts
```diff
@@ -45,6 +45,8 @@ export function operationFunctionBase(
       [
         getDocDescription(operation),
         `@param parameters - Object containing all parameters for the ${operation.type}.`,
+        // TODO: is this still necessary, considering that the line above indicates the type of operation?
+        `@type: ${operation.type}`,
```
marikaner, line 49: [req] This is not only not necessary, but even wrong, because in TSDoc this tag (`@type`) does not exist and in JSDoc it represents an actual type => https://jsdoc.app/tags-type.html

packages/generator/src/service/index-file.ts
```diff
@@ -21,11 +21,9 @@ export function indexFile(service: VdmServiceMetadata): SourceFileStructure {
...
+      (service.actionImports && service.actionImports.length)
+        ? [exportStatement('operations')]
```
marikaner, line 26: [req] You can simplify this statement as below. (But I would again expect `service.operations?.length` here.)
```suggestion
      ...(service.functionImports?.length || service.actionImports?.length)
        ? [exportStatement('operations')]
```

marikaner (dismissed):
Looks quite good to me, but I left a lot of questions and some requests to consider.

packages/generator/src/entity/imports.spec.ts
```diff
@@ -63,10 +63,10 @@ const expectedEntityImportsV4 = [
         'DeSerializers',
         'DefaultDeSerializers',
         'defaultDeSerializers',
-        'ActionImportParameter',
-        'ActionImportRequestBuilder',
         'FunctionImportParameter',
-        'FunctionImportRequestBuilder'
+        'ActionImportParameter',
+        'FunctionImportRequestBuilder',
+        'ActionImportRequestBuilder'
```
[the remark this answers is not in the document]
YuanZhuang107: I didn't look into if they could be merged as well, but yes I agree that we could do a follow-up to merge them if possible.

packages/generator/src/entity/imports.ts
```diff
@@ -30,8 +30,10 @@ export function entityImportDeclarations(
         kind: StructureKind.ImportDeclaration,
         isTypeOnly: true
       },
-      ...operationImportDeclarations(service, 'action', entity.actions),
-      ...operationImportDeclarations(service, 'function', entity.functions),
+      ...operationImportDeclarations(
+        service,
+        entity.actions.concat(entity.functions)
```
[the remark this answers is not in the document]
YuanZhuang107: I think `VdmEntity` doesn't have the field `operetaions`, only `functions` and `actions` for now. Should we do a follow-up to merge these in `VdmEntity` as well?

packages/generator/src/generator-without-ts-morph/service/class.spec.ts
```diff
@@ -24,27 +24,20 @@ describe('class', () => {
...
 
-  it('contains functionImports if in VDM', () => {
+  it('contains operations if in VDM', () => {
     const result = serviceClass({
       ...service,
       functionImports: [
         { name: 'myFunction', parametersTypeName: 'paraName' } as any
+      ],
+      actionImports: [
+        { name: 'myAction', parametersTypeName: 'paraName' } as any
```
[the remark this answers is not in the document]
YuanZhuang107: I think this is because `VdmServiceMetadata` separates `actionImports` and `functionImports` now. Should these be merged as well, possibly in a follow-up?

packages/generator/src/generator-without-ts-morph/service/class.spec.ts
```diff
@@ -24,27 +24,20 @@ describe('class', () => {
...
       ]
     });
-    expect(result).toContain('functionImport');
+    expect(result).toContain('operations');
```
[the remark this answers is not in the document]
YuanZhuang107: I think this is because `serviceClass(...)` returns a string that contains all the `service` code (I just found out this now), and this `expect...` line basically tried to see if the generated `service` code contained the substring `functionImport`, which it did.

packages/generator/src/generator-without-ts-morph/service/class.ts
```diff
@@ -54,32 +54,30 @@ export function serviceClass(service: VdmServiceMetadata): string {
...
 
-function getActionFunctionImports(
-  service: VdmServiceMetadata,
-  type: 'functionImports' | 'actionImports'
-): string {
-  if (service[type] === undefined || service[type]!.length === 0) {
+function getOperations(service: VdmServiceMetadata): string {
+  const operations = service.functionImports.concat(
+    service.actionImports || []
+  );
```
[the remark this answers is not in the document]
YuanZhuang107: For the first question, this is also because `VdmServiceMetadata` doesn't have `operations`, only `actionImports` and `functionImports`. Maybe I didn't understand the scope of the issue correctly -- should we also merge these in to `operations` in `VdmServiceMetadata` and the likes?

For the second question, the definitions of these fields in `VdmServiceEntities`:
```
  /**
   * @internal
   */
  functionImports: VdmOperation[];
  /**
   * @internal
   */
  actionImports?: VdmOperation[];
```
show that actionImports is optional. Maybe this is not intended? Then we could change that in a follow-up as well.

packages/generator/src/generator-without-ts-morph/service/file.ts
```diff
@@ -17,25 +17,23 @@ ${serviceBuilder(service.className, service.oDataVersion)}
...
 
-function getImports(
-  actionFunctionImports:
-    | { name: string; parametersTypeName: string }[]
-    | undefined,
-  moduleIdentifier: 'function-imports' | 'action-imports'
-): Import[] {
-  if (actionFunctionImports === undefined) {
+function getImports(service: VdmServiceMetadata): Import[] {
+  const operationImports = service.functionImports.concat(
```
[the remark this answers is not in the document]
YuanZhuang107: I'll change this to destructuring as well, and I'll leave the discussion on `VdmServiceMetadata` to the thread(s) above.

packages/generator/src/generator.ts
```diff
@@ -392,29 +392,16 @@ export async function generateSourcesForService(
     );
   });
 
-  if (service.functionImports?.length) {
+  // Merge generated function-imports.ts and action-imports.ts into one operations.ts.
+  if (service.functionImports.length || service.actionImports?.length) {
     logger.verbose(
-      `[${service.originalFileName}] Generating function imports ...`
+      `[${service.originalFileName}] Generating operation imports ...`
```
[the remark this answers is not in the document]
YuanZhuang107: Thanks!

packages/generator/src/operations/import.ts
YuanZhuang107: Thanks for the suggestion! I will move these to a helper function as suggested.

packages/generator/src/operations/operation.ts
YuanZhuang107: Thanks, I'll remove this then.

packages/generator/src/service/index-file.ts
YuanZhuang107: Good idea, and I think this part of the code could be updated further because by definition `service.functionImports` can't be undefined or null. I'll use the conditional access on `service.actionImports`.

packages/generator/src/service/index-file.ts
YuanZhuang107: Yes, I'll shorten this as well, although we don't need the conditional access on `service.functionImports` because its TS definition shows it always exists.

packages/generator/src/generator-without-ts-morph/service/class.spec.ts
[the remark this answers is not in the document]
marikaner: I agree. I'll create a ticket for this.

.changeset/beige-garlics-call.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-cloud-sdk/odata-common': major
+'@sap-cloud-sdk/generator': major
+'@sap-cloud-sdk/util': major
+---
+
+[Compatability Note] By default, generation of OData client will fail if a service path cannot be determined. Either provide `servicePath` value in the `options-per-service.json` or set `skipValidation` to true, in which case, `/` will be used.
```
marikaner, line 7:
```suggestion
[Compatability Note] By default, generation of OData clients fails if a service path cannot be determined. Either provide `servicePath` value in the `options-per-service.json` or set `skipValidation` to true, in which case, `/` will be used.
```

packages/generator/src/operations/import.ts
```diff
@@ -98,10 +97,17 @@ export function operationImportDeclarations(
     .length;
   const includesUnbound = !!operations.filter(operation => !operation.isBound)
     .length;
+  const hasFunctionWithParameters = operations.some(
+    operation => operation.parameters.length && operation.type === 'function'
+  );
+  const hasActionWithParameters = operations.some(
+    operation => operation.parameters.length > 0 && operation.type === 'action'
```
marikaner, line 104:
```suggestion
    operation => operation.parameters.length && operation.type === 'action'
```

marikaner (approved):
LGTM, I created the requested follow up

PR #3455, opened 2023-01-30
chore: Create Common Interface

FrankEssenberger:
Closes SAP/cloud-sdk-backlog#979.

Finally create common interface for the two generators after many prerequisites.


packages/generator-common/src/options.ts
```diff
@@ -0,0 +1,171 @@
...
+      alias: 'p',
+      describe:
+        'Specify the path to the prettier config. If not given a default config will be used for the generated sources.',
+      coerce: resolvePath,
+      requiresArg: true
+    },
+    skipValidation: {
+      type: 'boolean',
+      describe:
+        'By default, the generation fails, when there are duplicate or invalid names for operations and/or path parameters after transforming them to camel case. Set this to true to enable unique and valid name generation. The names will then be generated by appending numbers and prepending prefixes.',
```
deekshas8, line 72: I feel path parameters are specific to OpenAPI. Maybe we can adjust it so it works for both

deekshas8 (approved):
LGTM
As discussed, changes for resolving `inputDir` and `input` separately from `output` will be done in a follow-up PR.

PR #3483, opened 2023-02-02
chore!: Remove VALUE_IS_UNDEFINED in OData generator

deekshas8:
Closes SAP/cloud-sdk-backlog#1033.

- Remove `VALUE_IS_UNDEFINED` and its occurences
- Fail generation if service path cannot be determined (`skipValidation` is false)
- Dont fail generation if service path cannot be determined (`skipValidation` is true) and set it to `/` in `options-per-service.json`


test-packages/test-services-e2e/generate-e2e-services.ts
```diff
@@ -49,7 +49,7 @@ async function generateE2E() {
       '..',
       'test-resources',
       'odata-service-specs-e2e',
-      'v4'
+      'TripPin'
```
FrankEssenberger, line 52: [q] I do not get why this was changed - TripIn is also the service above?

FrankEssenberger: never mind got it - was the wrong options per service before.

FrankEssenberger (approved):
LGTM - I see tests fro the order, changelog and migration guide. Nothing to comment from my side :-).

PR #3485, opened 2023-02-02
chore: Middleware Cleanup

FrankEssenberger:
Relates [SAP/cloud-sdk-backlog#ISSUENUMBER.](https://github.com/SAP/cloud-sdk-backlog/issues/1042)

While writing the documentation we realized a few things. This adjusts them:
- We add the middlewares from right to left which is in line with the compose of function offered in other languages or `rambda` lib in JS
- The reference to the function argument is not needed.


deekshas8 (approved):
LGTM.

PR #3498, opened 2023-02-06
chore: `input` coerce function returns paths matching glob pattern in openapi generator

deekshas8:
The coerce function for `input` option in openapi-generator now returns array of paths matching glob pattern.


packages/generator-common/src/options-parser.ts
```diff
@@ -110,6 +110,45 @@ export function resolveRequiredPath<GeneratorOptionsT>(
...
+}
+
+/**
+ * Recursively searches through a given input path and returns all file paths as a string array.
+ * @param input - the path to the input directory.
+ * @returns all file paths as a string array.
+ * @internal
+ */
+export function getInputFilePaths(input: string): string[] {
+  if (glob.hasMagic(input)) {
```
FrankEssenberger, line 137: [q]I assume the `hasMagic` means that it is a directory? Please add a comment.

packages/generator-common/src/options-parser.ts
```diff
@@ -110,6 +110,45 @@ export function resolveRequiredPath<GeneratorOptionsT>(
...
+ * @returns Absolute path as a `string`.
+ */
+export function resolveInputGlob<GeneratorOptionsT>(
+  arg: string,
+  options: GeneratorOptionsT & { config?: string }
+): string[] {
+  const inputPath = options.config
+    ? resolve(dirname(options.config), arg.toString())
+    : resolve(arg.toString());
+  return getInputFilePaths(inputPath.split(sep).join(posix.sep));
```
FrankEssenberger, line 127: [pp] When I workd on the [glob](https://github.com/SAP/cloud-sdk-js/blob/2650859415f7bc54de39caa7bf9156700dc8a628/packages/generator-common/src/options-parser.ts#L87) resolving I used the `cwd` property of the glob. To merge before has the same effect - We should align the two ways. Either remove the `cwd` in the mentioned method or use it also here to have the code as consistent as possible.

FrankEssenberger (approved):
Please consider my comment on aligning the code.

packages/generator-common/src/options-parser.ts
deekshas8: Removed the `cwd`. Also fixed the tests that were affected as a result

packages/generator-common/src/options-parser.ts
deekshas8: Added a comment. `hasMagic` checks if pattern has special characters.

PR #3565, opened 2023-02-23
chore: Adjust upgrade guide

tomfrenken:
This PR adjusts the upgrade to give clearer instructions and improves some grammar.

Closes SAP/cloud-sdk-backlog#1065


V3-Upgrade-Guide.md
```diff
@@ -81,23 +81,20 @@ The following sub-sections describe affected modules, functions and interfaces w
...
   .execute({
-    destinationName:
-    'DESTINATION',
+    destinationName: 'DESTINATION',
     useCache: true,
     isolationStrategy: 'tenant'
   })
   ```
 
-- The generic type parameter `JwtKeysT` in `JwtKeyMapping` is now narrowed to extend `string`.
```
fwilhe, line 91: why delete this line?

tomfrenken: Because it is already there a few lines above

fwilhe:
My original motivation which moved into this ticket was that I was missing guidance on how to fix issues with renamed identifiers such as `Destination` to `HttpDestination`. I still don't see any mention of this in the document. We have a line saying that the `Destination` type as changed, but imo it does not really properly tell me what I need to do.

We don't need too much handholding, but something like "Renaming instances of Destination to HttpDestination` is something we should have in the guide, imo.

Not sure if we have more cases where this would apply. My hope would be that with the guide a developer who is unaware of sdk internals has enough information to fix issues caused by the upgrade without needing to google or guess.

FrankEssenberger:
> My original motivation which moved into this ticket was that I was missing guidance on how to fix issues with renamed identifiers such as `Destination` to `HttpDestination`. I still don't see any mention of this in the document. We have a line saying that the `Destination` type as changed, but imo it does not really properly tell me what I need to do.
> 
> We don't need too much handholding, but something like "Renaming instances of Destination to HttpDestination` is something we should have in the guide, imo.
> 
> Not sure if we have more cases where this would apply. My hope would be that with the guide a developer who is unaware of sdk internals has enough information to fix issues caused by the upgrade without needing to google or guess.

I think it would be best to add things you find missing also to this PR.

FrankEssenberger (approved):
The changes are fine for me - I agree with Florian that in many places the guide could be a bit more elaborate what to do.

tomfrenken:
Yeah, I wasn't sure at what point I explain too much but I guess I can add more instructions

PR #3585, opened 2023-02-28
chore: support varargs in middleware func

deekshas8:
Closes SAP/cloud-sdk-backlog#1073.

Middleware should allow to pass varargs


packages/openapi/src/openapi-request-builder.spec.ts
```diff
@@ -125,7 +125,7 @@ describe('openapi-request-builder', () => {
 
     const timeoutAboveDelay = () =>
       new OpenApiRequestBuilder('get', '/with-delay')
-        .middleware([timeout(delayInResponse * 10)])
+        .middleware(timeout(delayInResponse * 10))
```
marikaner, line 128: [pp] Instead of testing this implicitly, you could have a separate test to see whether both variants work. I am not sure if we now have a test that checks `[]` as well. Then again we already have tests for `transformVariadicArgumentToArray()`. So maybe this is not necessary.

marikaner (approved):
Looks good

packages/openapi/src/openapi-request-builder.spec.ts
deekshas8: Added two tests for both overloads.

FrankEssenberger:
LGTM

Issue #3544, opened 2023-02-20
Calls to destination with certificates lead to "unsupported" issue with Nodejs version > 17


lgunko:
**Describe the bug**
When destination has certificates set up - cloud sdk adds this certificate to the httpsAgent despite the fact that destination is not using the ClientCertificateAuthentication. This leads to [Axios Response] ERROR MESSAGE: "unsupported" when the legacy ciphers are used (for example SHA-1)

**To Reproduce**
Steps to reproduce the behaviour:
1. Set up the destination with SHA-1 SAML certificate 
2. Execute cloudSdkConnectivity.getDestinationFromDestinationService and executeHttpRequest to this destination
3. See error:
 'unsupported', stack: 'Error: unsupported\\n    at configSecureContext (node:internal/tls/secure-context:277:15)\\n    at Object.createSecureContext (node:_tls_common:117:3)\\n    at Object.connect (node:_tls_wrap:1629:48)\\n    at Agent.createConnection (node:https:150:22)\\n    at Agent.createSocket (node:_http_agent:350:26)\\n    at Agent.addRequest (node:_http_agent:297:10)\\n    at new ClientRequest (node:_http_client:335:16)\\n    at Object.request (node:https:360:10)\\n    at RedirectableRequest._performRequest (/app/node_modules/follow-redirects/index.js:284:24)\\n    at new RedirectableRequest (/app/node_modules/follow-redirects/index.js:66:8)'

**Expected behavior**
destination should be called without adding the certificate to the httpsAgent

**Used Versions:**

- node version via `18.12.1`
- npm version via `9.5.0`
- SAP Cloud SDK version you used as dependency: 2.12.1

**Impact / Priority**
we can not connect to some destinations

Affected development phase: e.g. Production

Impact: e.g. Inconvenience

Timeline: e.g. Go-Live is in 11 weeks.

Issue on Axis side:
https://github.com/axios/axios/issues/4778

marikaner:
Hey @lgunko thank you for reporting this. What authentication type does your destination have?
I wonder why there is a certificate if it not supposed to be used and whether this would occur with other users as well.

ptesny:
Hello, 

Indeed, it seems that sap-cloud-sdk **always** appends anything in the certificates array from the find destination payload  to the axios httpsAgent config; and that despite the fact that my destination is `OAuth2SAMLBearerAssertion` type and not `ClientCertificateAuthentication` type; 
[...]

PR #3588, opened 2023-02-28
fix: send certs only for MTLS authentication

marikaner:
Relates to #3544 

### DIFF
--- a/packages/connectivity/src/http-agent/http-agent.ts
+++ b/packages/connectivity/src/http-agent/http-agent.ts
@@ -122,6 +122,7 @@ function getKeyStoreOption(destination: Destination): Record<string, any> {
   if (
     destination.keyStoreName &&
     destination.keyStorePassword &&
+    // Only add certificates, when using MTLS (https://github.com/SAP/cloud-sdk-js/issues/3544)
     destination.authentication === 'ClientCertificateAuthentication'
   ) {
     const certificate = selectCertificate(destination);
--- a/packages/connectivity/src/http-agent/http-agent.ts
+++ b/packages/connectivity/src/http-agent/http-agent.ts
@@ -119,7 +119,11 @@ function getTrustStoreOptions(destination: Destination): Record<string, any> {
  * @returns Options, which can be used later by tls.createSecureContext() e.g. pfx and passphrase or an empty object, if the protocol is not 'https:' or no client information are in the definition.
  */
 function getKeyStoreOption(destination: Destination): Record<string, any> {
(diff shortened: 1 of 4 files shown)


packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -119,7 +119,11 @@ function getTrustStoreOptions(destination: Destination): Record<string, any> {
  * @returns Options, which can be used later by tls.createSecureContext() e.g. pfx and passphrase or an empty object, if the protocol is not 'https:' or no client information are in the definition.
  */
 function getKeyStoreOption(destination: Destination): Record<string, any> {
-  if (destination.keyStoreName && destination.keyStorePassword) {
+  if (
+    destination.keyStoreName &&
+    destination.keyStorePassword &&
+    destination.authentication === 'ClientCertificateAuthentication'
```
FrankEssenberger, line 125: [nth] Perhaps add a link to the original issue or one or two sentences to explain that the keyStore property are present also for the desintaiton service to sign SamlAssertion but only for the `ClientCertificateAuthentication` we as the SDK need to consider it.

FrankEssenberger (approved):
LGTM - would like to see a comment on the condition so that we remember why it is there.

packages/connectivity/src/http-agent/http-agent.ts
marikaner, line 125:
```suggestion
    // Only add certificates, when using MTLS (https://github.com/SAP/cloud-sdk-js/issues/3544)
    destination.authentication === 'ClientCertificateAuthentication'
```

ptesny: not only; please note that from now on all the destinations authentication types where a token is fetched via a token issuance endpoint do support explicit mTLS keystore dedicated to token fetching only...we have agreed this will be managed as a feature request going forward...

marikaner: @ptesny Yes, we talked about this, but I later realized, that this is nothing that we need to do. I believe the destination service is already doing that for us: https://help.sap.com/docs/CP_CONNECTIVITY/cca91383641e40ffbe03bdc78f00f681/2c162aaa9e464480b2307879d1b865f5.html?q=mtls

In the client secret case, we are not doing anything either.

ptesny: @marikaner , indeed, that's correct; thus no further action on your side; many thanks

PR #3590, opened 2023-02-28
chore: cherry-pick #3588

marikaner:
* fix: send certs only for MTLS authentication

* Update packages/connectivity/src/http-agent/http-agent.ts


packages/connectivity/src/http-agent/http-agent.spec.ts
```diff
@@ -159,6 +159,30 @@ describe('createAgent', () => {
     );
   });
 
+  it("does not return an agent for destinations with authentication types that have a certificate but don't use MTLS", () => {
+    const destination: Destination = {
```
marikaner, line 163:
```suggestion
    const destination: HttpDestination = {
```

FrankEssenberger (approved):

PR #3596, opened 2023-03-01
chore: Improve logging for package name generation

marikaner:
I noticed that we missed this minor issue when changing the options per service behavior.

### DIFF
--- a/packages/generator-common/src/util.ts
+++ b/packages/generator-common/src/util.ts
@@ -20,49 +20,60 @@ const npmMaxLength = 214;
 const npmRegex = /^(?:@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/;
 
 /**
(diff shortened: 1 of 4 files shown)


packages/generator-common/src/util.ts
```diff
@@ -20,49 +20,60 @@ const npmMaxLength = 214;
 const npmRegex = /^(?:@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/;
 
 /**
- * Takes a name and returns a transformation that is guaranteed to be compliant with npm naming rules.
- * @param name - The name to be transformed if necessary.
- * @returns Name that is guaranteed to be compliant.
+ * Checks whether a name is compliant with npm naming rules. Logs a warning if not.
```
marikaner, line 23:
```suggestion
 * Checks whether a name is compliant with NPM naming rules. Logs a warning if not.
```

PR #3627, opened 2023-03-07
chore: Create daily pre-release with 'canary' tag instead of 'next'

fwilhe:
Related to  SAP/cloud-sdk-backlog#1068.

This PR reverts changes made during the development of v3. It contains the following changes:

- We will not create a 'next' on npm for daily pre-released but a 'canary' tag instead
- Remove v2/v3 conditional for canary release which is not needed anymore
- Re-enable check for new commits within the last day and only release canary if this is the case

Minor chore changes:

- Use node 18 in a ci job which fails now when running with node 14 due to newer runtime requirements of a dependency
- Update to new syntax for github actions state output, cf https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


deekshas8 (approved):
LGTM.

PR #3647, opened 2023-03-13
fix: Mail-client proxy undefined error

deekshas8:
Closes SAP/cloud-sdk-backlog#1085

Fixes the following issues:
- Using `MAIL` destination with proxytype `Onpremise` throws error: `The proxy configuration is undefined`
- Sending mails to OnPrem server failes with `Greeting never received error`


packages/mail-client/src/mail-client.spec.ts
```diff
@@ -139,6 +196,51 @@ describe('mail client', () => {
     ).resolves.not.toThrow();
   });
 
+  xit('should work with registered destination- proxy-type OnPremise', async () => {
```
tomfrenken, line 199: Why only `xit` ?

tomfrenken (approved):
Overall LGTM, I guess you still have to lint the code & consider my one comment

packages/mail-client/src/mail-client.ts
```diff
@@ -128,13 +130,18 @@ async function createSocket(mailDestination: MailDestination): Promise<Socket> {
   const socketConnection = await SocksClient.createConnection(
     connectionOptions
   );
-  return socketConnection.socket;
+
+  const socksSocket = socketConnection.socket as SocksSocket;
+  // Setting `_readableListening` to true in the next line makes the socket readable.
+  // Otherwise nodemailer is not able to receive the SMTP greeting message and send emails.
```
jjtang1985, line 136: 👍

.changeset/angry-timers-press.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-cloud-sdk/connectivity': patch
+'@sap-cloud-sdk/mail-client': patch
+---
+
+[Fixed Issues] Set `_readableListening` property of Socket to `true` to allow nodemailer to receive SMTP greeting messages. Fixes `Greeting never received` error when sending emails to onPrem mail servers.
```
jjtang1985, line 6: [req] I would also mention one item for reflecting the error `The proxy configuration is undefined`

jjtang1985 (changes requested):
LGTM.
I left a comment regarding the release notes.

.changeset/angry-timers-press.md
marikaner, line 5: Maybe mention what was fixed and then how it was done. Starting with the technical details might not be relevant for most users.
```suggestion
[Fixed Issues] Set `_readableListening` property of socket to `true` to allow nodemailer to receive SMTP greeting messages. Fixes `Greeting never received` error when sending emails to On-Premise mail servers.
```

packages/connectivity/src/scp-cf/destination/http-proxy-util.ts
```diff
@@ -42,28 +46,30 @@ export function proxyStrategy(destination: HttpDestination): ProxyStrategy {
...
   }
 
-  const destinationProtocol = getProtocolOrDefault(destination);
-  if (!getProxyEnvValue(destinationProtocol)) {
-    logger.debug(
-      `Could not find proxy settings for ${destinationProtocol} in the environment variables - no proxy used.`
-    );
-    return 'no-proxy';
-  }
+  if (isHttpDestination(destination)) {
```
marikaner, line 49: [q] Is this only relevant to HTTP destinations? If I understand the code correctly this is about reading the env variables, shouldn't those apply to the MAIL destinations too? (at least partially, except for the part with the protocol)

packages/connectivity/src/scp-cf/destination/http-proxy-util.ts
```diff
@@ -42,28 +46,30 @@ export function proxyStrategy(destination: HttpDestination): ProxyStrategy {
...
-    logger.debug(
-      `Proxy settings for ${destinationProtocol} are found in environment variables.`
-    );
-    return 'internet';
+    if (getProxyEnvValue(destinationProtocol)) {
+      logger.debug(
+        `Proxy settings for ${destinationProtocol} are found in environment variables.`
+      );
+      return 'internet';
+    }
```
marikaner, line 72: [pp] It seems to me that it would be very helpful to extract this code into a separate function with a speaking name, you have to read too much code to understand the flow of this function in my opinion.

packages/mail-client/src/mail-client-types.ts
```diff
@@ -330,3 +330,21 @@ export interface SmtpTransportOptions {
...
+
+/**
+ * @internal
+ */
+interface ReadableState {
+  readableListening: boolean;
+}
+
+/**
+ * Represent a Socket object used fon OnPrem proxy.
```
marikaner, line 342:
```suggestion
 * Represent a socket object used for On-Premise proxy.
```

jjtang1985 (approved):
LGTM.

packages/connectivity/src/scp-cf/destination/http-proxy-util.ts
deekshas8: @marikaner  `getProtocolOrDefault` expects an `HttpDestination` and extracts the protocol from the `url` property (not relevant for the `MAIL` destinations). Based on this protocol, the proxy strategy is decided and if it's `Internet`, the env variables are read and the proxyConfiguration object is built for the destination. However, looking at the mail-client code, we only consider `proxyConfiguration` for `OnPremise` proxy type. So for `Internet`, I don't think mail has to be considered. @jjtang1985 please correct me if my understanding is wrong.

Also, in this [PR](https://github.com/SAP/cloud-sdk-js/pull/2883), we exposed the Transport proxy option so users could decide themselves if they want to pass this.

jjtang1985: @deekshas8 
Here is the [ADR](https://github.com/SAP/cloud-sdk-js/blob/main/knowledge-base/adr/0033-send-email.md#protocol-of-the-proxy) about the proxy, defined in the env.
Basically, the same logic as the http-client, which is more or less the same as the old implementation.
The changes in this file now specifically changed this.

.changeset/angry-timers-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/mail-client': patch
+---
+
+[Fixed Issues] Fix error `Greeting never received` when sending emails to On-Premise mail servers. The `_readableListening` property of socket is set to `true` to allow nodemailer to receive SMTP greeting messages.
```
marikaner, line 5:
```suggestion
[Fixed Issue] Fix error `Greeting never received` when sending emails to On-Premise mail servers. The `_readableListening` property of socket is set to `true` to allow nodemailer to receive SMTP greeting messages.
```

marikaner (approved):
👍

PR #3648, opened 2023-03-13
deserializer should load in batches with only function imports

fwilhe:

### DIFF
--- a/packages/odata-common/src/request-builder/batch/batch-request-builder.ts
+++ b/packages/odata-common/src/request-builder/batch/batch-request-builder.ts
@@ -4,7 +4,11 @@ import { first } from '@sap-cloud-sdk/util';
 import { MethodRequestBuilder } from '../request-builder-base';
 import { ODataRequest } from '../../request/odata-request';
 import { ODataBatchRequestConfig } from '../../request/odata-batch-request-config';
-import { DefaultDeSerializers, DeSerializers } from '../../de-serializers';
+import {
+  defaultDeSerializers,
+  DefaultDeSerializers,
+  DeSerializers
+} from '../../de-serializers';
 import { EntityBase } from '../../entity-base';
 import { GetAllRequestBuilderBase } from '../get-all-request-builder-base';
 import { GetByKeyRequestBuilderBase } from '../get-by-key-request-builder-base';
@@ -44,9 +48,12 @@ export class BatchRequestBuilder<
     )[]
   ) {
     super(new ODataBatchRequestConfig(defaultBasePath));
-    this.deSerializers = first(
-      Object.values(this.getEntityToApiMap())
-    )?.deSerializers;
+
+    const entityApi = first(Object.values(this.getEntityToApiMap()));
+
+    this.deSerializers = entityApi
+      ? entityApi?.deSerializers
+      : (defaultDeSerializers as DeSerializersT);
   }
 
   withSubRequestPathType(subRequestPathType: BatchSubRequestPathType): this {
--- a/test-packages/e2e-tests/test/batch.spec.ts
+++ b/test-packages/e2e-tests/test/batch.spec.ts
@@ -3,6 +3,7 @@ import { DefaultDeSerializers } from '@sap-cloud-sdk/odata-v4';
 import {
   batch,
   changeset,
+  createTestEntityById,
   createTestEntityByIdReturnId,
   returnInt,
   returnSapCloudSdk
@@ -109,26 +110,24 @@ describe('batch', () => {
   });
 
   it('deserializer should load in batches with only function imports', async () => {
-    const myFunctionImport = returnSapCloudSdk as any;
+    const myOperationUnderTest = createTestEntityById;
     //     ^?
-
-    const inputs = [1, 2, 3];
-    const functions = inputs.map(x => changeset(myFunctionImport({ foo: x })));
+    const inputs = [77];
+    const functions = inputs.map(x => changeset(myOperationUnderTest({ id: x })));
     //      ^?
     const results = await batch(...functions)
       //  ^?
       .withSubRequestPathType('relativeToEntity')
       .execute(destination);
 
-    // const r = (results[0] as unknown as WriteResponse<DefaultDeSerializers>);
-    const r = results[0] as any;
+    const r = results[0];
     //    ^?
-    if (r && r.responses) {
-      const deSerializedTestEntity = r.responses[0].as(testEntityApi);
+    if (r && r.isWriteResponses()) {
+      const deSerializedTestEntity = r.responses[0].as?.(testEntityApi);
       //      ^?
       expect(deSerializedTestEntity).toBeDefined();
     } else {
-      throw new Error('unexpected');
+      throw new Error('unexpected branch');
     }
   });
 });
--- a/test-packages/e2e-tests/test/batch.spec.ts
+++ b/test-packages/e2e-tests/test/batch.spec.ts
@@ -15,7 +15,7 @@ import {
 } from './test-utils/test-entity-operations';
 
 const entityKey = 456;
-const entityKeysUsedInTests = [456, 1000, 1001];
+const entityKeysUsedInTests = [77, 456, 1000, 1001];
 const requestBuilder = testEntityApi.requestBuilder();
 
 async function deleteAllCreatedEntities() {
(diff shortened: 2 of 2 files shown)


.changeset/witty-ducks-design.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sap-cloud-sdk/e2e-tests': patch
+'@sap-cloud-sdk/odata-common': patch
+'@sap-cloud-sdk/odata-v2': patch
+'@sap-cloud-sdk/odata-v4': patch
+---
+
+Deserializer load in batches with only function imports
```
marikaner, line 8: [req] Please rephrase this to a full sentence. I don't really get what is meant by this.

packages/odata-common/src/request-builder/action-function-import-request-builder-base.ts
```diff
@@ -3,13 +3,15 @@ import { HttpResponse } from '@sap-cloud-sdk/http-client';
...
 import { BatchReference, WithBatchReference } from '../request';
+import { DeSerializers } from '../de-serializers/de-serializers';
 import { MethodRequestBuilder } from './request-builder-base';
 
 /**
  * Create OData request to execute a action or function import.
  * @typeParam ReturnT - Type of the function import return value.
  */
 export abstract class ActionFunctionImportRequestBuilderBase<
+    DeSerializersT extends DeSerializers,
```
marikaner, line 14: [q] This is actually a breaking change. Maybe we should add this as the last type parameter and make it optional + add a todo to change this in the future and give it a compatibility note.
However, this will make things a bit more complicated, so I am wondering whether this is worth it. I doubt anyone is using this directly, therefore maybe it is fine with a compatibility note.

marikaner: [req] Please also add the typeParam to the API Doc

packages/odata-common/src/request-builder/batch/batch-request-builder.ts
```diff
@@ -38,15 +38,25 @@ export class BatchRequestBuilder<
...
-    )?.deSerializers;
+
+    const entityApi = first(Object.values(this.getEntityToApiMap()));
+    const entityDeserializer = entityApi?.deSerializers;
+
+    const operationDeserializer = this.getOperationDeserializer(requests);
+
+    this.deSerializers = entityDeserializer
+      ? entityDeserializer
+      : operationDeserializer;
```
marikaner, line 59: [pp] This can be shortened quite a bit and I think would be appropriate here as it makes it slightly more readable in my opinion.

```suggestion
    this.deSerializers = entityApi?.deSerializers || this.getOperationDeserializer(requests);
```

packages/odata-common/src/request-builder/batch/batch-request-builder.ts
```diff
@@ -103,6 +113,31 @@ export class BatchRequestBuilder<
     });
     return request;
   }
+
+  private getOperationDeserializer(
```
marikaner, line 117: [req/pp] As this returns (de-)**S**erialiser**S** I would suggest to name it that 😊 Note that those are serializers and deserializers and therefore `deSerializers` not `deserializers`.
```suggestion
  private getOperationDeSerializers(
```

packages/odata-common/src/request-builder/batch/batch-request-builder.ts
```diff
@@ -103,6 +113,31 @@ export class BatchRequestBuilder<
...
+    )[]
+  ) {
+    const outerRequest = first(requests);
+    if (outerRequest instanceof BatchChangeSet) {
+      const innerRequest = first(outerRequest.requests);
+      if (innerRequest && isActionOrFunctionImport(innerRequest)) {
+        return innerRequest._deSerializers;
+      }
+    }
+    return undefined;
```
marikaner, line 139: [pp] this line is unnecessary
```suggestion
```

test-packages/e2e-tests/test/batch.spec.ts
```diff
@@ -107,4 +108,43 @@ describe('batch', () => {
     expect(changesetResponse.isSuccess()).toBe(true);
     expect(deleteRes.isSuccess()).toBe(true);
   });
+
+  it('deserializer should load in batches with only function imports', async () => {
```
marikaner, line 112:
```suggestion
  it('(de-)serializers should be used in batch requests with only actions', async () => {
```

test-packages/e2e-tests/test/batch.spec.ts
```diff
@@ -107,4 +108,43 @@ describe('batch', () => {
...
+      const deSerializedTestEntity = r.responses[0].as?.(testEntityApi);
+      expect(deSerializedTestEntity).toBeDefined();
+    } else {
+      throw new Error(
+        'Expected response to be of type WriteResponse which is not the case.'
+      );
+    }
+  });
+
+  it('deserializer should load in batches with only bound function imports', async () => {
```
marikaner, line 129:
```suggestion
  it('(de-)serializers should be used in batch requests with only bound actions', async () => {
```

test-packages/e2e-tests/test/batch.spec.ts
```diff
@@ -107,4 +108,43 @@ describe('batch', () => {
...
   });
+
+  it('deserializer should load in batches with only function imports', async () => {
+    // See https://github.com/SAP/cloud-sdk-js/issues/3539 for the original issue
+    const results = await batch(changeset(createTestEntityById({ id: 77 })))
+      .withSubRequestPathType('relativeToEntity')
+      .execute(destination);
+
+    const r = results[0];
+    if (r && r.isWriteResponses()) {
```
marikaner, line 119: [pp] I am not a fan of variables called `r` (except maybe in arrow functions - which is not the case here)
Also, some of this code can be shortened, so naming gets easier:
```suggestion
    const changesetRes = first(await batch(changeset(createTestEntityById({ id: 77 })))
      .withSubRequestPathType('relativeToEntity')
      .execute(destination))

    if (changesetRes?.isWriteResponses()) {
```

(also applies to the test below)

marikaner (changes requested):
Mostly text and code style, otherwise the solution seems quite legit.

packages/odata-common/src/request-builder/batch/batch-request-builder.ts
fwilhe: applied

.changeset/witty-ducks-design.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sap-cloud-sdk/e2e-tests': patch
+'@sap-cloud-sdk/odata-common': patch
+'@sap-cloud-sdk/odata-v2': patch
+'@sap-cloud-sdk/odata-v4': patch
+---
+
+[Compatibility Note] Fix deserializer so it works in batch requests which only contain functions or actions. This changes the type parameters of `ActionFunctionImportRequestBuilderBase`.
```
marikaner, line 8: [req] We should have two changesets: one for the compatibility note and one for the fix.

marikaner, line 8: [req] Please note that we are talking about deserializers and serializers, therefore (de-)serializers. Also, as this is a collection of functions we use the plural throughout the docs.

marikaner, line 8: [req] And last point: We did not change anything in the (de-)serializers, therefore "fix deserializer" does not really describe what happened. My suggestion:

1. [Fixed Issue] Fix batch requests to accept only functions or actions.
2. [Compatibility Note] The `ActionFunctionImportRequestBuilderBase` has an additional mandatory type parameter to pass the according (de-)serializer type. This was missing previously due to a bug.

Of course feel free to chose better words. ;)

packages/odata-common/src/request-builder/action-function-import-request-builder-base.ts
```diff
@@ -3,13 +3,17 @@ import { HttpResponse } from '@sap-cloud-sdk/http-client';
 import { v4 as uuid } from 'uuid';
 import { ODataRequestConfig } from '../request/odata-request-config';
 import { BatchReference, WithBatchReference } from '../request';
+import { DeSerializers } from '../de-serializers/de-serializers';
 import { MethodRequestBuilder } from './request-builder-base';
 
 /**
  * Create OData request to execute a action or function import.
+ * @typeParam DeSerializersT - Type of the (De-)serializers.
```
marikaner, line 11:
```suggestion
 * @typeParam DeSerializersT - Type of the (de-)serializers.
```

marikaner (approved):
I had some nitpicky comments on the changeset - the devil lies in the details.
But looks very good otherwise. 👍

.changeset/witty-ducks-design.md
fwilhe: got it, thanks for the explanation.

fwilhe: I think that wording is very good. adapted.

PR #3661, opened 2023-03-14
chore: Ignore commits in git blame

fwilhe:
Configure git to ignore some commits in git blame.
cf https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
Inspired by https://devblogs.microsoft.com/typescript/typescripts-migration-to-modules/#wait-what-was-that-about-git


marikaner (approved):
I am still not sure that this in general is a good idea. But I don't think it should break anything, therefore, let's go!

fwilhe:
> I am still not sure that this in general is a good idea

Okay, I thought this was a quick win without significant drawbacks, but we should only do this if we all agree on it. Closed the PR.

PR #3689, opened 2023-03-20
chore: Add ignores for depcheck in e2e tests

fwilhe:
Fixes the following build issue:

@sap-cloud-sdk/e2e-tests:check:dependencies: Missing dependencies
@sap-cloud-sdk/e2e-tests:check:dependencies: * @sap-cloud-sdk/odata-v2: ./test/generator-test-output/API_TEST_SRV/BatchRequest.ts
@sap-cloud-sdk/e2e-tests:check:dependencies: * bignumber.js: ./test/generator-test-output/API_TEST_SRV/service.ts


marikaner:
I think sqlite is ignored because it is used by CAP and we have to install their dependencies. I don't know why the other libs are ignored, but generally, I think it is not a good idea to add those exception (except in cases where the check can't know - like sqlite).

PR #3692, opened 2023-03-20
chore: More readable github workflow names

fwilhe:
This change should make it easier to navigate the list of workflows in the github ui.

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -1,4 +1,4 @@
-name: api-docs
+name: Push API Docs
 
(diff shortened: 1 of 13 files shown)


marikaner (commented):
I personally prefer the shorter names, because they align with the file name and I find it easier to find (+ I am used to the names I guess). You could consider the `run-name` instead. => https://docs.github.com/en/actions/using-workflows/about-workflows#understanding-the-workflow-file

Please align with the other colleagues to decide whether this should be done. @deekshas8 @tomfrenken

PR #3717, opened 2023-03-22
test gh actions

fwilhe:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,6 +39,7 @@ jobs:
           SANDBOX_TOKEN: ${{ secrets.API_HUB_SANDBOX_TOKEN }}
       - run: yarn test:integration
       - run: yarn test:self
+      - run: yarn test:build-packages
       - run: yarn test:type
   checks:
     runs-on: ubuntu-latest
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "compile": "turbo run compile",
     "test": "turbo run test",
     "test:unit": "turbo run test --filter=./packages/*",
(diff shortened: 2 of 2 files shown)


fwilhe:
Thanks for taking care @YuanZhuang107 

I think there would also be an option to avoid duplicating the pr-template file by ensuring we use the correct working directory when running the unit tests, but I'm fine with doing it like this for now. It might be a potential follow-up to refactor the tests so they don't need that duplicated file.

Since I created the PR I can't approve it. @marikaner can you approve if you're fine with this change too?

build-packages/check-pr/.github/PULL_REQUEST_TEMPLATE.md
marikaner, line 1: [q/req] Is there no way to make the test search for the actual file?

marikaner (changes requested):
Overall looks good, but I think we should avoid the duplicating file, if somehow possible and us fs-mock instead.

marikaner:
@fwilhe I made some changes to reflect what I meant, please review again and feel free to make changes and/or to merge.

marikaner (approved):
👍

build-packages/check-pr/validators.spec.ts
```diff
@@ -1,12 +1,23 @@
-import { readFile } from 'node:fs/promises';
-import { resolve } from 'node:path';
+import mock from 'mock-fs';
 import { validateTitle, validateBody, validateChangesets } from './validators';
 
+const prTemplate = `unchanged PR template
+with multiple lines`;
```
marikaner, line 5:
```suggestion
with multiple lines`;

```

PR #3718, opened 2023-03-22
chore: set default validity time of cache

YuanZhuang107:
change validity time of cache from optional to required and set default validity time to 5min.

Closes SAP/cloud-sdk-backlog#706.


marikaner (changes requested):
Let's align on the requirements, I think this PR does something different than I expected.

.changeset/smart-garlics-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Improvement] Change the default validity time of the cache to 5 minutes. If you need to customize it, use the `expires_in` option.
```
marikaner, line 5: [req] This description is outdated after the changes.

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -86,7 +79,7 @@ export class Cache<T> implements CacheInterface<T> {
   set(key: string | undefined, item: CacheEntry<T>): void {
     if (key) {
       const expires =
-        item.expires ?? inferExpirationTime(this.defaultValidityTimeInMs);
+        item.expires ?? inferExpirationTime(this.defaultValidityTime);
```
marikaner, line 82: [req] Please adjust the signature of `inferExpirationTime` to not allow using `undefined`. Maybe consider adding it as a private method to the class, and rename it to something like `inferDefaultExpirationTime`. Then you don't need to pass `this.defaultValidityTime`. This is anyways the only place this is used, if I'm not mistaken.

packages/connectivity/src/scp-cf/destination/destination-from-registration.spec.ts
```diff
@@ -116,18 +116,6 @@ describe('register-destination', () => {
     ).resolves.toBe(true);
   });
 
-  it('caches with unlimited time', async () => {
```
marikaner, line 119: [req] I think we still need this test. The register destination cache should never expire by default, but as far as I can see there is no such option currently possible.

The ticket for this says "Align on a parameter that reflects the value "never" (ideas: "never", null, undefined)" => I think `undefined` is not a good choice because it is not explicit enough. I would maybe consider `0` because this value does not make sense as a default and could signify that it never expired. I think I saw this in other APIs as well.

marikaner (changes requested):
We still need some way to make the cache unlimited. It should just not be unlimited by default. This part looks great already. 😊

packages/connectivity/src/scp-cf/cache.ts
YuanZhuang107: Thanks for the suggestion! I updated this part as suggested.

.changeset/smart-garlics-study.md
YuanZhuang107: I updated this description to reflect newer changes.

packages/connectivity/src/scp-cf/destination/destination-from-registration.spec.ts
YuanZhuang107: I've now re-enabled this test. For unlimited expiration time, I went with your suggestion to use `0` to signify "never expired".

packages/connectivity/src/scp-cf/cache.spec.ts
```diff
@@ -27,8 +27,10 @@ const dummyToken = {
 // Cache with expiration time
 const cacheOne = new Cache<Destination>(300000);
 
-// Cache without expiration time
-const cacheTwo = new Cache<ClientCredentialsResponse>();
+// Cache with a long expiration time
+const cacheTwo = new Cache<ClientCredentialsResponse>(
+  365 * 24 * 60 * 1000 /* 1 year */
+);
```
marikaner, line 33: [pp] How about:
```suggestion
// Never expiring cache
const cacheTwo = new Cache<ClientCredentialsResponse>(0);
```

marikaner (commented):
Looks good, I would already approve, I am just wondering about the removed test.

.changeset/smart-garlics-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Improvement] Change the default validity time parameter of the Cache constructor to be required, rename it from `validityTimeInMs` to `defaultValidityTime`, and allow never-expiring cache with explicitly setting this parameter to `0`. Set the default validity time of `ClientCredentialsTokenCache` to 5 minutes.
```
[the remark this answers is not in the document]
YuanZhuang107: That makes sense. Thanks for the suggestion!

packages/connectivity/src/scp-cf/cache.spec.ts
YuanZhuang107: Thanks, yes the previous `1 year` was legacy code that I changed to simulate a very long cache, but in this case keeping the old behavior with our new "0 as never expiring" is definitely better.

packages/connectivity/src/scp-cf/cache.spec.ts
```diff
@@ -80,11 +82,6 @@ describe('Cache', () => {
     expect(cacheOne.hasKey('one')).toBeFalsy();
   });
 
-  it('should return the item when its expiration time is undefined', () => {
```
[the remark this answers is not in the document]
YuanZhuang107: This shouldn't get removed. It was removed because our the misunderstanding of the issue requirement. I will add this back.

packages/connectivity/src/scp-cf/destination/destination-cache.ts
```diff
@@ -51,7 +51,7 @@ export interface DestinationCacheInterface {
 export class DefaultDestinationCache implements DestinationCacheInterface {
   cache: Cache<Destination>;
 
-  constructor(validityTimeInMs?: number) {
+  constructor(validityTimeInMs = 0) {
```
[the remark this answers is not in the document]
YuanZhuang107: This is a good idea. I'll do that. Thanks!

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -41,14 +41,11 @@ export class Cache<T> implements CacheInterface<T> {
   private cache: Record<string, CacheEntry<T>>;
 
   /**
-   * Default validity period for each entry in cache.
-   * If `undefined`, all cached entries will be valid indefinitely.
+   * Creates an instance of Cache.
+   * @param defaultValidityTime - The default expiration time in milliseconds. Use 0 for unlimited cache duration.
```
marikaner, line 45:
```suggestion
   * @param defaultValidityTime - The default validity time in milliseconds. Use 0 for unlimited cache duration.
```

marikaner (approved):
LGTM

PR #3729, opened 2023-03-25
chore: Move dependencies to dev dependencies

marikaner:
I think these dependencies were added to the wrong category by accident. I don't see a reason to make them productive dependencies.


FrankEssenberger (approved):
I guess all types can go to dev since they are not needed in the shipped artifact.

deekshas8 (approved):

PR #3735, opened 2023-03-27
chore: Fix tests on windows runner

fwilhe:
The tests on windows have been failing consistently for quite a while. This pr aims to improve that by taking a few steps:

- Increase timeouts which are likely to cause failures
- run unit tests with `--concurrency=1` option

See latest run of [windows tests here](https://github.com/SAP/cloud-sdk-js/actions/workflows/tests-windows.yml).


.github/workflows/tests-windows.yml
```diff
@@ -10,23 +10,17 @@ on:
...
     steps:
       - uses: actions/checkout@v3
       - run: git fetch --depth=1
       - uses: actions/setup-node@v3
         with:
           node-version: 18
+          cache: 'yarn'
       - run: yarn install --frozen-lockfile
-      - run: yarn test:unit --continue -- -- -- --json --outputFile=unit-test-output.json
-      - run: yarn test:integration -- -- -- --json --outputFile=int-test-output.json
```
fwilhe, line 22: can we keep the `concurrency=1` and revert the output file things as before?

deekshas8: I feel we can do without archiving the results as the test output is in the console already. I generally just go through that instead of downloading from the archive. But I don't have a strong opinion on this.

deekshas8: We added this initially to run some analysis on flaky windows tests, but I don't think we ever did that.

fwilhe:
I can't approve because I opened the PR, but looks good to me so far. Definitely an improvement over what we had before.

deekshas8 (approved):

PR #3742, opened 2023-03-28
wip iss/jwt

fwilhe:


packages/connectivity/src/scp-cf/xsuaa-service.ts
```diff
@@ -47,20 +48,55 @@ interface XsuaaParameters {
...
  * Make a client credentials request against the XSUAA service.
  * @param service - Service as it is defined in the environment variable.
- * @param userJwt - User JWT.
+ * @param jwtOrOptions - User JWT or ServiceTokenOptions.
  * @returns Client credentials token.
  */
 export async function getClientCredentialsToken(
   service: string | Service,
-  userJwt?: string | JwtPayload
+  jwtOrOptions?: string | JwtPayload
```
fwilhe, line 83: I hope this is not considered a breaking change. The name of the parameter changed, but afaik this should not be breaking because typescript does not support named parameters.

I've added the overload to allow the old and the new way of calling the function and both are equally supported, neither is deprecated.

PR #3748, opened 2023-03-29
feat: Allow iss/jwt combination

fwilhe:
Allow to use the `iss` and `jwt` combination, where `iss` is used to fetch the service token for the destination service and the `jwt` is used for the `x-user-token` header for user related auth types.

Closes SAP/cloud-sdk-backlog#[1058](https://github.com/SAP/cloud-sdk-backlog/issues/1058).


packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -562,18 +567,14 @@ Possible alternatives for such technical user authentication are BasicAuthentica
     }
   }
 
-  private selectSubscriberJwt(): JwtPayload {
-    if (!this.subscriberToken) {
+  private selectSubscriberJwt(): JwtPair {
```
marikaner, line 570: @FrankEssenberger I have changed this function and I think logically my changes are the same for the old state. This is used for the cache key, and I guess this is not correct for the combination case (iss and jwt), because both tokens are relevant I guess. Do you agree?

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -471,11 +477,12 @@ Possible alternatives for such technical user authentication are BasicAuthentica
         )
       };
     }
-    // Case 2 Subscriber and provider account not the same OR custom jwt -> x-user-token  header passed to determine user and tenant in token service URL and service token to get the destination
+    // Case 2 Subscriber and provider account not the same OR custom JWT -> x-user-token header passed to determine user and tenant in token service URL and service token to get the destination
     const serviceJwt =
       origin === 'provider'
         ? this.providerServiceToken
-        : this.subscriberToken.serviceJwt!;
+        : // TODO: What is the meaning of this? Why do we assume this is defined. Technically, it might not be.
```
marikaner, line 484: @FrankEssenberger We are not ensuring that the service JWT is really not undefined. With `!` we are basically just assuming this, but it could still happen. Or not?

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -545,9 +552,7 @@ Possible alternatives for such technical user authentication are BasicAuthentica
       case 'on-premise':
         return addProxyConfigurationOnPrem(
           destination,
-          this.subscriberToken?.type === 'iss'
-            ? this.subscriberToken.serviceJwt
-            : this.subscriberToken?.userJwt
+          this.subscriberToken?.userJwt || this.subscriberToken?.serviceJwt
```
marikaner, line 555: @FrankEssenberger I lack a bit of knowledge here - what is this used for and is the new implementation correct?

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -68,7 +70,11 @@ interface DestinationSearchResult {
  * We name these tokens "subscriber tokens", because they are related to the subscriber account in contrast to the"provider account", where the application is running.
  * The tenant defined in the subscriber token is the provider tenant for single tenant applications.
  */
-type SubscriberToken = IssToken | XsuaaToken | CustomToken;
+interface SubscriberToken {
+  userJwt?: JwtPair;
+  serviceJwt?: JwtPair;
+}
+// IssToken | XsuaaToken | CustomToken;
```
fwilhe, line 77: leftover or does this have documentation value?

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
marikaner: Also: Why would we ever pass a service token here?

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
marikaner: If custom JWT, then origin has to be provider!

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -448,14 +418,14 @@ Possible alternatives for such technical user authentication are BasicAuthentica
     // This covers OAuth to user-dependent auth flows https://help.sap.com/viewer/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/39d42654093e4f8db20398a06f7eab2b.html and https://api.sap.com/api/SAP_CP_CF_Connectivity_Destination/resource
     // Which is the same for: OAuth2UserTokenExchange, OAuth2JWTBearer and OAuth2SAMLBearerAssertion
 
-    // If user JWT is not XSUAA enforce the JWKS properties are there - destination service would do that as well. https://help.sap.com/docs/CP_CONNECTIVITY/cca91383641e40ffbe03bdc78f00f681/d81e1683bd434823abf3ceefc4ff157f.html
-    if (this.subscriberToken.type === 'custom') {
+    // If subscriber token does not include service JWT (aka. originally passed JWT was not an XSUAA JWT) enforce the JWKS properties are there - destination service would do that as well. https://help.sap.com/docs/CP_CONNECTIVITY/cca91383641e40ffbe03bdc78f00f681/d81e1683bd434823abf3ceefc4ff157f.html
+    if (!this.subscriberToken.serviceJwt) {
```
deekshas8, line 422: [q] We previously had this check only for custom jwt because self-signed jwts should have the jwks set in the destination. Wouldn't this now check also for xsuaa jwts (where jwks is not really required as it has jku in the headers?) ?

PR #3753, opened 2023-03-30
fix: Fix `FromJsonType`

marikaner:
Fix the `FromJsonType` to allow nullable arrays. This should accomodate a breaking change from TypeScript.

### DIFF
--- a/packages/odata-common/src/entity-builder.ts
+++ b/packages/odata-common/src/entity-builder.ts
@@ -38,7 +38,7 @@ export type PureEntityType<T> = Omit<NonNullishType<T>, keyof EntityBase>;
 export type FromJsonType<JsonT> = {
   [key: string]: any; // custom properties
 } & {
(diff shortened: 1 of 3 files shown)


packages/odata-common/src/entity-builder.ts
```diff
@@ -38,7 +38,7 @@ export type PureEntityType<T> = Omit<NonNullishType<T>, keyof EntityBase>;
 export type FromJsonType<JsonT> = {
   [key: string]: any; // custom properties
 } & {
-  [P in keyof PureEntityType<JsonT>]?: PureEntityType<JsonT>[P] extends (infer U)[]
+  [P in keyof PureEntityType<JsonT>]?: PureEntityType<JsonT>[P] extends (infer U)[] | null | undefined
```
marikaner, line 41: This check previously failed for collection types.

fwilhe (approved):
awesome, thanks 💯

PR #3754, opened 2023-03-30
chore: Fix memory tests after v3 release

fwilhe:
Closes SAP/cloud-sdk-backlog#1096.

This PR fixes the failing memory tests due to changes caused by the v3 release of the SDK.

See green test run [here](https://github.com/SAP/cloud-sdk-js/actions/runs/4564894493/jobs/8055250254).

The new baseline comparison is v2 instead of v1.

v2 is the baseline because it is not updated anymore, so it is a stable reference point to judge how memory usage changes in v3/canary.

Note that the old 90% threshold does not seem to work anymore, updated to 105%. This should at least let us know if we are getting worse in memory usage. Threshold may be updated if we improve memory usage in the future.

### DIFF
--- a/test-packages/memory-tests/sdk-v2/package.json
+++ b/test-packages/memory-tests/sdk-v2/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@sap-cloud-sdk/memory-tests-canary",
+  "name": "@sap-cloud-sdk/memory-tests-v2",
   "version": "0.0.1",
   "description": "Memory usage tests of the SAP Cloud SDK for JavaScript",
   "homepage": "https://sap.github.io/cloud-sdk/docs/js/overview",


.github/workflows/memory-tests.yml
```diff
@@ -34,5 +34,5 @@ jobs:
         working-directory: test-packages/memory-tests/sdk-canary/
       - run: node --inspect test-script.js && cat result.txt
         working-directory: test-packages/memory-tests/sdk-canary/
-      - run: yarn ts-node scripts/compare-memory-usage.ts $(cat ./test-packages/memory-tests/sdk-v1/result.txt) $(cat ./test-packages/memory-tests/sdk-canary/result.txt)
-        name: compare v1 and canary
+      - run: yarn ts-node scripts/compare-memory-usage.ts $(cat ./test-packages/memory-tests/sdk-v2/result.txt) $(cat ./test-packages/memory-tests/sdk-canary/result.txt)
+        name: compare v2 and canary
```
deekshas8, line 38: [q] Instead of `v2`, should we use something like `previous` so we don't have to maintain this in the future? Same for variable names in `compare-memory-usage.ts`

fwilhe: I think we have to maintain this anyway because of breaking changes and I feel that 'compared to v2' gives us a better understanding than 'compared to previous'. This part is a simple search and replace, I'd like to keep it because I think the added clarity of what we compare is worth the effort of this search/replace.

tomfrenken (approved):
LGTM, I assume the heap limit was increased as the heap consumption went beyond the threshold.

scripts/compare-memory-usage.ts
```diff
@@ -1,17 +1,17 @@
-const threshold = 90;
+const threshold = 105;
```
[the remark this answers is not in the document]
fwilhe: As described in the PR description, the old threshold does not hold anymore. I assume in v2 we made optimisations which made v2 more memory efficient than v1 which is why a < 100% TH worked, but this is not the case in v2 -> v3.

PR #3759, opened 2023-04-03
chore: fix caching and on prem connectivity for `iss` and `jwt` combination

marikaner:
This fixes caching and  for `iss` and `jwt`.
Caching and on-prem connectivity take the following for identification:

|                        | user | tenant |
|---|---|---|
| xsuaa jwt       | user jwt | service token based on user jwt |
| custom jwt     | user jwt | user jwt |
| iss only           | service token based on iss (is most likely undefined) | service token based on iss |
| iss + user jwt | user jwt | service token based on iss|

Closes SAP/cloud-sdk-backlog#[1058](https://github.com/SAP/cloud-sdk-backlog/issues/1058).


fwilhe:
Thanks for writing down that table, that's definitely helpful. I'm wondering: What about our documentation, both for sdk users and for developers. Is it obvious enough how this behaves based on what parameters you pass? Is it obvious enough what the tradeoffs are with providing a custom issuer?

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
```diff
@@ -92,3 +101,63 @@ function getJwtForServiceToken(
...
+  token: SubscriberToken | undefined
+): Required<SubscriberToken> {
+  if (token) {
+    const { userJwt, serviceJwt } = token;
+
+    if (hasTokens(token)) {
+      // in case one of the tokens is missing use the other token for tenant and user identification
+      // at least one of the properties is defined, therefore ! is safe
+      return {
+        userJwt: userJwt || serviceJwt!,
```
fwilhe, line 120: our return type is required but with the `!` we do potentially allow the value to be undefined.. is this intended? should we rather throw an error if anything is undefined?

fwilhe: oh nevermind, I've only now read the comment..

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
```diff
@@ -92,3 +101,63 @@ function getJwtForServiceToken(
...
+
+/**
+ * @internal
+ * Get a subscriber token pair with required fields. Checks that at least one of the tokens exists and sets defaults if needed.
+ * @returns The decoded subscriber tokens.
+ */
+export function getRequiredSubscriberToken(
+  token: SubscriberToken | undefined
+): Required<SubscriberToken> {
+  if (token) {
```
fwilhe, line 113: style: any reason not to do an early return here which would allow the rest of the function not to be inside the scope of the if?

```
if (!token) {
 throw error
}

// rest of the function
```

marikaner:
> Thanks for writing down that table, that's definitely helpful. I'm wondering: What about our documentation, both for sdk users and for developers. Is it obvious enough how this behaves based on what parameters you pass? Is it obvious enough what the tradeoffs are with providing a custom issuer?

No, that's why I added this: https://github.com/SAP/cloud-sdk/pull/1336

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
marikaner: That is what I added the comment for. The problem is, that given the check before it cannot be undefined, but the code to ensure that TS also know that will be a lot more verbose. I agree that this is not a good practice in general, that is why I added the comment. Still, someone could change the code and that makes it potentially dangerous. 
I don't have a strong opinion - I can change the code to be more verbose to make the TS compiler get it too. Let me know what you think.

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
marikaner: I'd have to throw the error twice.

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
fwilhe: nice, I like that solution 💯

fwilhe (approved):
With my limited understanding of the subject, looks good to me. Definitely brings more clarity than we had before.

PR #3763, opened 2023-04-04
feat: Add NoAuthentication destination support back

tomfrenken:
This PR reverts a previous change where we removed support for `NoAuthentication` on-premise destinations, as we assumed it was deprecated.

In the current state of the BTP documentation, it seems the decision to remove/deprecate `NoAutentication` destinations was reverted, at least there is no mention of it being legacy anymore, see [here](https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/authentication-to-on-premise-system?q=no%20authentication).

Closes SAP/cloud-sdk-backlog#1097.


packages/connectivity/src/scp-cf/authorization-header.ts
```diff
@@ -147,6 +147,22 @@ function headerForProxy(
...
 }
 
+function noAuthOnPremiseProxy(
+  destination: Destination
+): Record<string, any> {
+  let principalPropagationHeader;
+  try {
+    principalPropagationHeader = headerForPrincipalPropagation(destination);
+  } catch (e) {
+    logger.warn('No principal propagation header found.');
```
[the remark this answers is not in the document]
tomfrenken: Yeah, I think you could argue that this is rather a debug message, I believe it was a warning because previously NoAuth was used as some principal propagation workaround, so if those headers were missing a warning was thrown, I will convert it into a debug message.

packages/connectivity/src/scp-cf/authorization-header.ts
```diff
@@ -169,9 +185,7 @@ function getProxyRelatedAuthHeaders(
     destination.proxyType === 'OnPremise' &&
     destination.authentication === 'NoAuthentication'
   ) {
-    throw Error(
-      'OnPremise connections are not possible with NoAuthentication. Please select a supported authentication method e.g. PrincipalPropagation or BasicAuthentication.'
-    );
+    return noAuthOnPremiseProxy(destination) as any;
```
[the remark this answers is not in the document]
tomfrenken: Here I just used the previous implementation, I've narrowed the type down a bit, let me know what you think.

.changeset/gentle-drinks-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Add support for NoAuthentication on-premise destinations back.
```
marikaner, line 5:
```suggestion
[New Functionality] Add support for authentication type `NoAuthentication` on-premise destinations.
```

.changeset/gentle-drinks-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Add support for authentication type `NoAuthentication` on-premise destinations.
```
marikaner, line 5: Do we need a compatibility note?

packages/connectivity/src/scp-cf/authorization-header.ts
```diff
@@ -156,6 +172,10 @@ interface AuthenticationHeaderOnPrem {
 interface AuthenticationHeaderProxy {
   'Proxy-Authorization': string;
 }
+interface AuthenticationNoAuthHeaders {
+  'Proxy-Authorization'?: string;
+  'SAP-Connectivity-Authentication'?: string;
+}
```
marikaner, line 178: [nth] Looking at the lines above and below, I have the impression that there is some potential for de-duplication.

marikaner: [req] I just noticed that some of the interfaces here are actually here twice! E.g. `AuthenticationHeaderOnPrem`. I know this was not part of the ticket, but please remove the duplicate types... 🤪

packages/connectivity/src/scp-cf/authorization-header.ts
```diff
@@ -164,14 +184,12 @@ interface AuthenticationHeaders {
 
 function getProxyRelatedAuthHeaders(
   destination: Destination
-): AuthenticationHeaderProxy | undefined {
+): AuthenticationHeaderProxy | AuthenticationNoAuthHeaders | undefined {
```
marikaner, line 187: [q] I think this defeats the purpose of this function. There are two functions for proxy related headers and auth related headers. I think this might be intended differently than you "hacked" it ;)

marikaner (changes requested):
Please align with the intended structure (or change the structure if needed).

PR #3766, opened 2023-04-05
feat: Allow registerDestination to infer mTLS certs on CF

fwilhe:


fwilhe:
Notes on open questions from Discussion with @marikaner:

- Where should `inferMtlsCertificate` be specified?
  - `RegisterDestinationOptions` seems like a good place, not yet clear how this can be accessed from `getAgentConfig`
- What properties should `Destination` have (does already have a lot)?
- What would actual usage look like in application code consuming the sdk?
  - Which information is available at which point/when calling `registerDestination`?

fwilhe:
replaced by #3781

PR #3768, opened 2023-04-06
chore: Refactor environment accessor

marikaner:
When trying to get rid of the SDK's dependency on the XSUAA services where not needed, I found that the environment accessor had a few issues. It grew quite big, some functionality behaves differently than expected and wasn't tested. Also,  some functionality was unused.

I separated, reordered, tested and removed some parts.


packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+function getCredentialsWithoutJwt(
+  xsuaaCredentials: XsuaaServiceCredentials[]
+): XsuaaServiceCredentials {
+  logResult(xsuaaCredentials, false);
+  return xsuaaCredentials[0];
+}
+
+function logResult(credentials: XsuaaServiceCredentials[], usedToken: boolean) {
```
tomfrenken, line 75: Perhaps a more speaking name could be used for this utility function

tomfrenken (approved):
LGTM, only two small comments

packages/connectivity/src/scp-cf/environment-accessor/env.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+export function getServiceByInstanceName(
+  serviceInstanceName: string
+): Service | undefined {
+  const service = xsenv.filterServices(serviceInstanceName);
+
+  if (!service.length) {
+    throw Error(`Could not find service with name: '${serviceInstanceName}'.`);
+  }
+  // This never happens as xsenv always takes the last matching binding
```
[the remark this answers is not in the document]
marikaner: I left it in to ensure that we control the behavior ultimately. I don't know why it was in in the first place (without comment) but maybe the behavior changed?

packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
marikaner: I'll do another refactoring PR and I will think about it in the next one than. Hope that's ok.

PR #3781, opened 2023-04-12
feat: Allow to use mTLS certificates on CloudFoundry for registered destinations

fwilhe:
This PR introduces a convenience feature for using mTLS on CloudFoundry.

Closes SAP/cloud-sdk-backlog#1054.

Related docs pr: https://github.com/SAP/cloud-sdk/pull/1343


test-packages/integration-tests/test/mtls.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    isTrustingAllCertificates: true
+  };
+
+  const options: RegisterDestinationOptions = {
+    inferMtls: true
+  };
+
+  // fixme: actually needs to assert on cert/key being part of the request
+  it('can perform http request with mTLS on registered destination', async () => {
+    nock('https://example.com').get('/mtls').reply(200);
```
deekshas8, line 44: In [this](https://github.com/SAP/cloud-sdk-js/blob/68470c5a9e208b85a06bd60af02c505339663937/packages/http-client/src/http-client.spec.ts#L716) test we check if the axios request is called with httpsAgent config, maybe you can try this?

packages/connectivity/src/http-agent/http-agent.spec.ts
```diff
@@ -258,4 +259,63 @@ describe('getAgentConfig', () => {
...
+
+        expect(actual.cert).toEqual('my-cert');
+        expect(actual.key).toEqual('my-key');
+        expect(actual.pfx).not.toBeDefined();
+        expect(actual.passphrase).not.toBeDefined();
+      });
+    });
+
+    it('returns an object with key "httpsAgent" and mTLS options for destinations with inferMtls option and environment variables are not present', async () => {
+      const destination: HttpDestination = {
```
deekshas8, line 299: [q] If `inferMtls` option is true but `CF_INSTANCE_CERT` and `CF_INSTANCE_KEY` are not present in the env variables, should this not lead to any warning or error?

test-packages/integration-tests/test/mtls.spec.ts
fwilhe: Yes, that works. Thanks for the suggestion, I'll update the test.

packages/connectivity/src/http-agent/http-agent.spec.ts
fwilhe: I don't have a strong opinion here, my thinking was that 'infer' does not guarantee it will do something so I'm not sure if we should really warn if it does not happen.

tomfrenken: i'm also in favor of throwing a warning in this case

fwilhe: added a log and adapted the unit test to assert it is there

tomfrenken (approved):
Only made some small adjustments, LGTM.

PR #3799, opened 2023-04-19
chore: Remove unused functions

deekshas8:
Removed functions like `getAuthHeaders` and `addDestinationToRequestConfig`


marikaner (approved):
Yes! More deletions than additions 😊

PR #3805, opened 2023-04-20
chore: Add function to retrieve service binding by JWT for destination binding

marikaner:
In this PR I refactored the environment accessor further and introduced a generic function to retrieve service credentials by JWT.
Now, if a JWT is passed to `getDestinationServiceCredentials()` it will also be matched by client id.

Related to SAP/cloud-sdk-backlog#865.

### DIFF
--- a/packages/connectivity/src/scp-cf/environment-accessor/service-credentials.spec.ts
+++ b/packages/connectivity/src/scp-cf/environment-accessor/service-credentials.spec.ts
@@ -3,52 +3,6 @@ import { getServiceCredentials } from './service-credentials';
 
 const logger = createLogger('environment-accessor');
 
-// const services = {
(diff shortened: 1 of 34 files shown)


packages/connectivity/src/scp-cf/environment-accessor/service-credentials.spec.ts
```diff
@@ -0,0 +1,192 @@
+import { createLogger } from '@sap-cloud-sdk/util';
+import { getServiceCredentials } from './service-credentials';
+
+const logger = createLogger('environment-accessor');
+
+// const services = {
```
deekshas8, line 6: Probably a leftover?

deekshas8 (approved):
LGTM.

PR #3823, opened 2023-04-25
chore: remove unnecessary duplication of functions and actions

YuanZhuang107:
Closes SAP/cloud-sdk-backlog#1072.


packages/generator/src/batch/imports.ts
```diff
@@ -19,10 +19,10 @@ export function importBatchDeclarations(
         'GetByKeyRequestBuilder',
         'ODataBatchRequestBuilder',
         'UpdateRequestBuilder',
-        ...(service.functionImports.length
+        ...(service.operationImports.some(op => op.type === 'function')
```
marikaner, line 22: [q] Can we also combine `ActionImportRequestBuilder` and `FunctionImportRequestBuilder` into an `OperationRequestBuilder`?

packages/generator/src/batch/type.ts
```diff
@@ -34,15 +34,12 @@ function getWriteRequestType(service: VdmServiceMetadata): string {
     e =>
       `CreateRequestBuilder<${e.className}<DeSerializersT>, DeSerializersT> | UpdateRequestBuilder<${e.className}<DeSerializersT>, DeSerializersT> | DeleteRequestBuilder<${e.className}<DeSerializersT>, DeSerializersT>`
   );
-  const functionImportsReturnTypes = service.functionImports
-    .filter(fn => fn.httpMethod.toLowerCase() !== 'get')
+  const operationImportsReturnTypes = service.operationImports
```
marikaner, line 37: [nth] Generally, let's try to avoid the wording "operationIMPORT" and use "operation" instead. This was a misunderstanding initially.

packages/generator/src/batch/type.ts
```diff
@@ -34,15 +34,12 @@ function getWriteRequestType(service: VdmServiceMetadata): string {
...
+    .filter(fn => fn.type === 'action' || fn.httpMethod.toLowerCase() !== 'get')
```
marikaner, line 38: [q] Is this differentiation necessary? Wouldn't it suffice to check for "get"?

packages/generator/src/batch/type.ts
```diff
@@ -55,8 +52,10 @@ function getReadRequestType(service: VdmServiceMetadata): string {
     e =>
       `GetByKeyRequestBuilder<${e.className}<DeSerializersT>, DeSerializersT>`
   );
-  const functionImportsReturnTypes = service.functionImports
-    .filter(fn => fn.httpMethod.toLowerCase() === 'get')
+  const functionImportsReturnTypes = service.operationImports
+    .filter(
+      fn => fn.type === 'function' && fn.httpMethod.toLowerCase() === 'get'
```
marikaner, line 57: [q] Same here. Aren't all actions not GET?

packages/generator/src/edmx-to-vdm/v2/service-entities.ts
```diff
@@ -29,6 +29,6 @@ export function getServiceEntitiesV2(
     complexTypes,
     enumTypes: [],
     entities,
-    functionImports
+    operationImports: functionImports
```
marikaner, line 32: [pp] I'd suggest to rename this as well.

packages/generator/src/edmx-to-vdm/v4/entity.spec.ts
```diff
@@ -182,27 +182,34 @@ describe('entity', () => {
...
+    ).toEqual(['fn1IsBound']);
+    expect(
+      entities[0].operations.filter(functionFilter)[0].parameters.length
+    ).toBe(1);
+    expect(
+      entities[0].operations.filter(actionFilter).map(({ name }) => name)
+    ).toEqual(['action1IsBound']);
+    expect(
+      entities[0].operations.filter(actionFilter)[0].parameters.length
+    ).toBe(0);
```
marikaner, line 200: [pp] I think this would be easier to read as follows (the order might need to be adjusted, same for `entities[1]`):
```suggestion
    const [function, action] = entities[0].operations;
    
    // function
    expect(function.name).toEqual('fn1IsBound');
    expect(function.parameters.length).toBe(1);
    
    // action
    expect(action.name).toEqual('action1IsBound');
    expect(action.parameters.length).toBe(0);
```

packages/generator/src/edmx-to-vdm/v4/entity.ts
```diff
@@ -65,26 +65,28 @@ export function generateEntitiesV4(
...
+      ...generateBoundOperations(
+        serviceMetadata,
+        entityMetadata.entityType.Namespace,
+        'action',
+        entities,
+        complexTypes,
+        formatter,
+        entityMetadata.entitySet.Name,
+        classNames[entityMetadata.entitySet.Name]
+      )
```
marikaner, line 88: [q] Is it still necessary to have to calls of `generateBoundOperations` here or can we merge them into one?

packages/generator/src/edmx-to-vdm/v4/service-entities.ts
```diff
@@ -49,7 +49,6 @@ export function getServiceEntitiesV4(
     complexTypes,
     enumTypes,
     entities,
-    functionImports,
-    actionImports
+    operationImports: [...actionImports, ...functionImports]
```
marikaner, line 52: [q] Same as above, can those be merged before?

packages/generator/src/generator-without-ts-morph/service/file.ts
```diff
@@ -18,9 +18,7 @@ ${serviceClass(service)}`;
 }
 
 function getImports(service: VdmServiceMetadata): Import[] {
-  const operationImports = service.functionImports.concat(
-    service.actionImports || []
-  );
+  const operationImports = service.operationImports;
   if (operationImports.length === 0) {
```
marikaner, line 22: [req] I know this was here before, but it does not match our style guide: https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-truthyfalsy-checks-where-possible

```suggestion
  if (!operationImports.length) {
```

packages/generator/src/operations/file.ts
```diff
@@ -17,9 +17,7 @@ import { operationFunction } from './operation';
 export function operationsSourceFile(
   service: VdmServiceMetadata
 ): SourceFileStructure {
-  const functions = service.functionImports;
-  const actions = service.actionImports || [];
-  const operations = [...functions, ...actions];
+  const operations = service.operationImports;
```
marikaner, line 20: [req] I think this is not necessary anymore: `operationImportDeclarations` should take just one argument and we can easily reach the operations through `service.operationImports`.

packages/generator/src/sdk-metadata/__snapshots__/generation-and-usage.spec.ts.snap
```diff
@@ -1,23 +1,23 @@
...
+"import { dummyActionImport, dummyParamType } from './generated/undefined/operation-imports';
 
 const parameter: dummyParamType = { dummyParam1: 'dummyParam1' };
     
 const resultPromise = dummyActionImport(parameter).execute({ destinationName: 'myDestinationName' });"
 `;
 
 exports[`generation-and-usage create api specific usage for function import 1`] = `
-"import { dummyFunc, dummyParamType } from './generated/undefined/function-imports';
+"import { dummyFunc, dummyParamType } from './generated/undefined/operation-imports';
```
marikaner, line 12: [req] This really should be `operations` not `operation-imports`.

marikaner: [q] Why is there `undefined` in this path. I can see that it was there before, but it is kind of a red flag.

marikaner (changes requested):
Mostly minor things but I have 2 main requests:
1. I would really appreciate it, if we could rename every occurrence of "operationImport" (or "operation-import") to "operation".
2. I think there are a few places where we could merge even more.

Issue #3846, opened 2023-05-06
Failed to get CSRF token after upgrade to v3


thulium3:
**Describe the bug**
After upgrading my generated OData v2 client from v2 to v3, my create is not working any longer. I always get error 403, failed to get CSRF token.

**To Reproduce**
I execute following code:
`const { retrieveJwt } = require("@sap-cloud-sdk/connectivity");
const { ssfRequest } = require('./client/ssf-request');
const { postingRequestSetApi } = ssfRequest();
app.use(express.json({ limit: '50mb' }));

// create Ticket in SSF
app.post("/Posting", async function (req, res, next) {
  const posting = postingRequestSetApi.entityBuilder().fromJson(req.body);
  const myTicket = await postingRequestSetApi.requestBuilder()
    .create(posting)
    .execute({ destinationName: 'xxx', jwt: retrieveJwt(req) })
    .catch(err => {
      console.log('Error:', err.message);
      console.log('Cause:', err.cause?.message);
      console.log('Root cause:', err.rootCause?.message);
      next(err);
    });`
before upgrading it worked, now always getting the error.

**Expected behavior**
create still works.

**Screenshots**

**Used Versions:**

- node version via `node -v` v18.13.0
- npm version via `npm -v` v8.19.3
- SAP Cloud SDK version you used as dependency v3.1.1

**Code Examples**
`const { retrieveJwt } = require("@sap-cloud-sdk/connectivity");
const { ssfRequest } = require('./client/ssf-request');
const { postingRequestSetApi } = ssfRequest();
app.use(express.json({ limit: '50mb' }));

// create Ticket in SSF
app.post("/Posting", async function (req, res, next) {
  const posting = postingRequestSetApi.entityBuilder().fromJson(req.body);
  const myTicket = await postingRequestSetApi.requestBuilder()
    .create(posting)
    .execute({ destinationName: 'xxx', jwt: retrieveJwt(req) })
    .catch(err => {
      console.log('Error:', err.message);
      console.log('Cause:', err.cause?.message);
      console.log('Root cause:', err.rootCause?.message);
      next(err);
    });`

**Log file**

**Impact / Priority**

Cannot update my app.

Affected development phase: Production

Impact: Blocked

tomfrenken:
Hi @thulium3,

without the error logs it's hard to tell what the root cause of your issue is, however after a brief investigation the `default` maximum length of header values seems to be 8KB, how exactly do you send the large attachment?
[...]

PR #3879, opened 2023-05-17
fix: csrf token fetching

tomfrenken:
With this PR, we stop sending the request's data when requesting a CSRF token, which caused issues for users who were sending large requests (see [issue](https://github.com/SAP/cloud-sdk-js/issues/3846))

### DIFF
--- a/.changeset/tasty-ties-give.md
+++ b/.changeset/tasty-ties-give.md
@@ -2,4 +2,4 @@
 '@sap-cloud-sdk/http-client': minor
 ---
 
-[Fixed Issue] CSRF token fetching now supports large requests.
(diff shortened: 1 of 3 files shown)


.changeset/tasty-ties-give.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/http-client': minor
+---
+
+[Fixed Issue] CSRF token fetching now supports large requests.
```
marikaner, line 5: [req] As usual, this should be in the imperative.

marikaner (approved):
LGTM, just please try to improve the changeset based on our style guide.

Issue #2712, opened 2022-07-27
Unable to compile OData v2 client of service with inheritance


lennart-m:
**Describe the bug**
Hi, I am trying to generate a client for a service that uses inheritance (has entities with a `BaseType`).
The client generates typescript files but they won't compile as the inheriting entities don't have key properties.

**To Reproduce**
Use this MWE.

`service-specs/service-mapping.json`:  
```json
{
  "test": {
    "directoryName": "test",
    "servicePath": "/test",
    "npmPackageName": "test"
  }
}
```

`service-specs/test.edmx`:  
```xml
<?xml version='1.0' encoding='UTF-8'?>
<edmx:Edmx 
  xmlns:edmx="http://schemas.microsoft.com/ado/2007/06/edmx" Version="1.0">
  <edmx:DataServices m:DataServiceVersion="1.0" 
    xmlns:m="http://schemas.microsoft.com/ado/2007/08/dataservices/metadata">
    <Schema Namespace="ns" 
      xmlns="http://schemas.microsoft.com/ado/2008/09/edm">
      <EntityType Name="MyBaseType">
        <Key>
          <PropertyRef Name="id"/>
        </Key>
        <Property Name="id" Type="Edm.String" Nullable="false"/>
      </EntityType>
      <EntityType Name="MyType" BaseType="ns.MyBaseType">
        <Property Name="name" Type="Edm.String" Nullable="true"/>
      </EntityType>
      <EntityContainer Name="DefaultContainer" m:IsDefaultEntityContainer="true">
        <EntitySet Name="MyType" EntityType="ns.MyType"/>
      </EntityContainer>
    </Schema>
  </edmx:DataServices>
</edmx:Edmx>
```

- Entity `MyType` inherits from `MyBaseType`
- `MyBaseType` has the key
- `MyType` only adds one non-key property

`package.json`:  
```json
{
  "name": "test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "generate-client": "generate-odata-client -i service-specs -o generated --forceOverwrite --clearOutputDir"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@sap-cloud-sdk/connectivity": "^2.6.0",
    "@sap-cloud-sdk/odata-common": "^2.6.0",
    "@sap-cloud-sdk/odata-v2": "^2.6.0",
    "@sap-cloud-sdk/util": "^2.6.0"
  },
[...]

PR #3899, opened 2023-05-30
fix(generator): don't generate `delete()`, `update()`, and `getByKeys()` methods in RequestBuilder class when the entity has no keys

netux:
For entities without a key, the overload would fail because the first `delete()` method has no parameters.
I have added a example test to showcase this.

Also, added the missing `Entity` import.

Relates to SAP/cloud-sdk-js#2712

**Additional notes:** I tried running `yarn checks` and `yarn lint:fix` as the contribution guide says, but both failed with scripts in the e2e-tests package for some reason. So I just ran the scripts local to the generator package.


netux:
> How would you delete something without a key? Do you have an example of an existing service with an entity that can be deleted and does not have a key?

I'm pretty sure it isn't standard. I found the case of [a service in the SuccessFactors LMS API](https://help.sap.com/docs/SAP_SUCCESSFACTORS_LEARNING/5aab9bef78fc4c4fa199c1f7aa142720/bebb64ac3e6641f9b5f4ebf91b23ca79.html#entity-types) that exposes an entity with fields that act like keys, but aren't marked as such.
Granted, that service has other issues, like exposing an entity more than once, which also confuse this tool. So we could tick that as an issue with the service, not the generator.

> I am wondering whether technically removing the key from the function in those cases is the right approach or whether we should rather not provide a delete request builder for those cases at all?

I simply chose the least destructive approach since I'm not too knowledgeable on the intricacies of OData.
I did think it'd also make sense to not provide a delete method at all, but I also thought of these non-standard services that could, for example, treat an entity like a singleton that can either have only one instance or not have an instance at all (or just be implemented wrong).

marikaner:
Hey @netux,

I have just discussed this issue with a colleague and we decided the following: The best approach would be to allow setting your own keys somehow. This would require an API discussion before implementation, however we envisioned something in the direction of this: Instead of calling `.delete('keyValue1', 'keyValue2')` there could be the option to define other keys `.delete(keys('fakeKey1', 'fakeKey2'), 'fakeKeyValue1', 'fakeKeyValue2')`.
This is not straight forward and currently does not seem super high-priority. Therefore it could take a bit of time until this will be implemented.

However, I assume you created this pull request because you tried to use the generator and it fails? So you need an immediate fix. With no keys and no workaround to create keys we could also just remove the `delete` function for entities without keys. The same actually applies to `getByKey` and `update`.

Would you like to update this PR with these changes?

netux:
> I have just discussed this issue with a colleague and we decided the following: The best approach would be to allow setting your own keys somehow. This would require an API discussion before implementation, however we envisioned something in the direction of this: Instead of calling `.delete('keyValue1', 'keyValue2')` there could be the option to define other keys `.delete(keys('fakeKey1', 'fakeKey2'), 'fakeKeyValue1', 'fakeKeyValue2')`. This is not straight forward and currently does not seem super high-priority. Therefore it could take a bit of time until this will be implemented.

Sure! I would maybe suggest using objects instead? `.delete({ fakeKey1: 'fakeKeyValue1', fakeKey2: 'fakeKeyValue2' })`
Or why did you choose to have a `keys()` function?

> However, I assume you created this pull request because you tried to use the generator and it fails? So you need an immediate fix. With no keys and no workaround to create keys we could also just remove the `delete` function for entities without keys. The same actually applies to `getByKey` and `update`.
> 
> Would you like to update this PR with these changes?

For my solution, I already had to monkey patch the service edmx files to solve other issues. So I could also patch some `<Key>`s in 😅.
Regardless, I updated the generator to exclude the methods as described.

marikaner:
I was so free to update the code.

marikaner:
I shortened the tests a bit and removed the `Entity` import. Please let me know if you think this is wrong.

netux:
> I shortened the tests a bit and removed the Entity import. Please let me know if you think this is wrong.

The `Entity` import was required because it library could generate a `delete(entity: Entity)` method. I guess it's safe to remove this then?

https://github.com/netux/SAP-cloud-sdk-js/blob/9aac8fa802479739a20dde1e3eb32a4cbc119c6d/packages/generator/src/generator-without-ts-morph/request-builder/class.ts#L217-L222

mr-flannery (approved):
pair-reviewed with @marikaner

PR #3925, opened 2023-06-12
chore: Remove xsuaa dependency when caching service tokens

marikaner:
Related to SAP/cloud-sdk-backlog#865.

This is the second of three PRs to remove the SDK's dependency on an XSUAA service binding.
Instead of relying on the XSUAA service to be there, the URL from the credentials of the service that the token is retrieved or is used for caching.

### DIFF
--- a/packages/connectivity/src/scp-cf/token-accessor.spec.ts
+++ b/packages/connectivity/src/scp-cf/token-accessor.spec.ts
@@ -22,6 +22,7 @@ import {
 } from '../../../../test-resources/test/test-util/xsuaa-service-mocks';
 import { clientCredentialsTokenCache } from './client-credentials-token-cache';
 import { jwtBearerToken, serviceToken } from './token-accessor';
+import { ClientCredentialsResponse } from './xsuaa-service-types';
 
 describe('token accessor', () => {
   describe('serviceToken()', () => {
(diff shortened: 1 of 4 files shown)


packages/connectivity/src/scp-cf/token-accessor.spec.ts
```diff
@@ -283,16 +284,47 @@ describe('token accessor', () => {
...
       process.env.VCAP_SERVICES = JSON.stringify({
         destination: [destinationBindingClientSecretMock]
       });
+      const token = signedJwt({ dummy: 'content' });
 
-      await expect(
-        serviceToken('destination')
-      ).rejects.toThrowErrorMatchingInlineSnapshot(
-        '"Could not find binding to service \'xsuaa\', that includes credentials."'
+      clientCredentialsTokenCache.clear();
```
deekshas8, line 293: [info] the cache is cleared in `afterEach()`, we can skip the call here

deekshas8 (approved):
Minor comment. LGTM

packages/connectivity/src/scp-cf/token-accessor.spec.ts
marikaner, line 293:
```suggestion
```

PR #3943, opened 2023-06-19
chore: Refactor token exchange

marikaner:
Refactor token exchange and do not exchange tokens that are issued by XSUAA.


deekshas8 (approved):
LGTM.

PR #3964, opened 2023-06-27
chore: Improve error handling when registering destinations

marikaner:
Improve error handling and don't create dummy tenant


packages/connectivity/src/scp-cf/destination/destination-from-registration.spec.ts
```diff
@@ -261,10 +262,36 @@ describe('register-destination without xsuaa binding', () => {
...
+    );
+  });
+
+  it('registers destination with a dummy id, if no JWT is given', async () => {
+    const logger = createLogger('register-destination');
+    jest.spyOn(logger, 'debug');
+
+    await registerDestination(testDestination);
+    expect(
+      await getDestination({
```
deekshas8, line 288: [nitpick] Checking the registerDestinationCache.destination.retrieveDestination.. with the said dummy id instead of calling getDestination here be simpler and more explicit what is being tested.

deekshas8 (approved):
LGTM.

PR #3968, opened 2023-06-28
chore: add `x-sap-security-session: create` for SAMLAssertion

YuanZhuang107:
Closes SAP/cloud-sdk-backlog#1112.

### DIFF
--- a/packages/connectivity/src/scp-cf/authorization-header.ts
+++ b/packages/connectivity/src/scp-cf/authorization-header.ts
@@ -150,7 +150,12 @@ function getProxyRelatedAuthHeaders(
 
 async function getAuthenticationRelatedHeaders(
   destination: Destination
-): Promise<AuthenticationHeaderCloud | AuthenticationHeaderOnPrem | AuthenticationHeaderSAMLAssertion | undefined> {
+): Promise<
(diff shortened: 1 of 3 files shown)


packages/connectivity/src/scp-cf/authorization-header.ts
```diff
@@ -133,6 +133,9 @@ interface AuthenticationHeaders {
   'Proxy-Authorization'?: string;
   'SAP-Connectivity-Authentication'?: string;
 }
+interface AuthenticationHeaderSAMLAssertion {
```
marikaner, line 136: [pp] Add a `\n` here ;) 
```suggestion

interface AuthenticationHeaderSAMLAssertion {
```

packages/connectivity/src/scp-cf/authorization-header.ts
```diff
@@ -133,6 +133,9 @@ interface AuthenticationHeaders {
...
+  'x-sap-security-session': string;
```
marikaner, line 137: [nth] This could be more explicit:
```suggestion
  'x-sap-security-session': 'create';
```

marikaner (approved):
LGTM

PR #4011, opened 2023-07-21
chore: read prettier config once

marikaner:
I noticed, that the prettier config file is read once per schema file. This is not necessary as it is read once before and can be reused.

### DIFF
--- a/packages/openapi-generator/src/generator.spec.ts
+++ b/packages/openapi-generator/src/generator.spec.ts
@@ -2,10 +2,11 @@ import { resolve } from 'path';
 import { existsSync, promises } from 'fs';
 import mock from 'mock-fs';
 import { readJSON } from '@sap-cloud-sdk/util';
-import prettier from 'prettier';
 import { getInputFilePaths } from '@sap-cloud-sdk/generator-common/internal';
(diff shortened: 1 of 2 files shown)


packages/openapi-generator/src/generator.spec.ts
```diff
@@ -2,10 +2,11 @@ import { resolve } from 'path';
 import { existsSync, promises } from 'fs';
 import mock from 'mock-fs';
 import { readJSON } from '@sap-cloud-sdk/util';
```
marikaner, line 4:
```suggestion
import { readJSON } from '@sap-cloud-sdk/util';
import prettier from 'prettier';
```

packages/openapi-generator/src/generator.spec.ts
```diff
@@ -6,6 +6,8 @@ import prettier from 'prettier';
 import { getInputFilePaths } from '@sap-cloud-sdk/generator-common/internal';
 import { emptyDocument } from '../test/test-util';
 import { generate } from './generator';
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const prettier = require('prettier');
```
marikaner, line 10:
```suggestion
```

PR #4020, opened 2023-07-25
fix: serialize Edm.Decimal as number within safe range

YuanZhuang107:
fix deserialization for decimal values

Closes SAP/cloud-sdk-backlog#1087.


packages/odata-common/src/de-serializers/default-de-serializers.ts
```diff
@@ -108,7 +108,17 @@ export const defaultDeSerializersRaw: DefaultDeSerializers = {
   /* DeSerializers with v2 and v4 specific URI serializer defaults. */
   'Edm.Decimal': {
     deserialize: deserializeToBigNumber,
-    serialize: serializeFromBigNumber
+    serialize: value => {
+      const primitiveNumber =
+        value instanceof BigNumber ? value.toNumber() : value;
```
marikaner, line 113: [req] We had issues with `instanceof` in the past when you have multiple "instances" of the class `BigNumber` in your `node_modules`. Therefore I would just turn it around.
```suggestion
      const primitiveNumber =
        typeof value === 'number' ? value : value.toNumber();
```

marikaner (approved):
LGTM

.changeset/wise-dodos-pump.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-cloud-sdk/odata-common': patch
+'@sap-cloud-sdk/odata-v4': patch
+---
+
+[Fixed Issue] Serialize Edm.Decimal as number within safe range.
```
marikaner, line 6:
```suggestion
[Fixed Issue] Serialize properties of type "Edm.Decimal" to numbers when they are within the safe range and strings when they are not. 
```

PR #4027, opened 2023-07-28
feat: Always add names to destinations from service bindings

marikaner:
The name of the service binding is known an can be added automatically, if it was forgotten in the transformation function. This is needed for caching.
The name should never differ from the automatic assignment, however, if the transformation function adds a name, it will not be overwritten.

Closes https://github.com/SAP/cloud-sdk-backlog/issues/1124


tomfrenken (approved):
LGTM

Issue #4033, opened 2023-08-01
Error: Failed to build headers.


tonymathew17sap:
**Describe the bug**
Getting the error "Error: Failed to build headers." when connecting with destination of authentication type "OAuth2SAMLBearerAssertion". I'm using open-api client 'execute()' method with 'addCustomHeaders()' method to add headers

**To Reproduce**

      const response = await OpenApi.Client('id')
        .addCustomHeaders({
          ['custom-header']: `dummy-value`,
        })
        .execute({
          destinationName: 'test',
        });

This is giving me the error "Error: Failed to build headers."

**Used Versions:**

- node version via `node -v` v18.16.0
- npm version via `npm -v` 9.7.1
- SAP Cloud SDK version you used as dependency:     "@sap-cloud-sdk/openapi": "^3.4.0",

**Impact / Priority**

Affected development phase: Production

Impact: Blocked

tomfrenken:
Hey @tonymathew17sap,

For destinations of type `OAuth2SAMLBearerAssertion` you need to provide a user jwt, otherwise, building the headers will fail. You may want to check out [our documentation](https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destinations#authentication-and-json-web-token-jwt-retrieval) regarding the different auth types and their associated auth flows.

In your case, you need to provide the jwt in the execute method, here is an example:
```
  const response = await OpenApi.Client('id')
    .addCustomHeaders({
      ['custom-header']: `dummy-value`,
    })
    .execute({
      destinationName: 'test',
      jwt: 'yourJWT'
    });
```

tonymathew17sap:
@tomfrenken after passing JWT like you said, getting the following error:
<img width="1491" alt="Screenshot 2023-08-02 at 8 10 49 PM" src="https://github.com/SAP/cloud-sdk-js/assets/134997532/946f2a11-7e45-4837-8435-2e2a3d573283">

Do you have any idea why I'm getting this error?

tonymathew17sap:
@tomfrenken yes the JWT is coming from approuter

tonymathew17sap:
@tomfrenken as suggested in the fix above, I gave the iss property directly:
  `const response = await OpenApi.Client('id')
    .execute({
      destinationName: 'test',
      jwt: 'yourJWT',
     iss:'<host>/oauth/token'
    });`
    
    With this change, I'm getting 401 error:
    
[...]

PR #4050, opened 2023-08-11
chore: improve error handling

tomfrenken:
This PR improves how we log a mismatch between a user's jwt and any given service binding, this should especially help issues like https://github.com/SAP/cloud-sdk-js/issues/4033 in finding the cause of their error.

### DIFF
--- a/packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
+++ b/packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
@@ -16,7 +16,7 @@ const logger = createLogger({
  * Throws an error if there are no bindings of the given type.
  * @param service - The service type.
  * @param token - Either an encoded or decoded JWT.
- * @returns The credentials for a match, otherwise `null`.
+ * @returns The credentials for a match, otherwise `undefined`.
  */
 export function getServiceCredentials<
   ServiceCredentialsT extends ServiceCredentials
@@ -43,13 +43,12 @@ export function getServiceCredentials<
     if (credentials) {
       return credentials;
     }
-
     logger.debug(
-      'Could not match binding service instance using JWT client id or audience.'
+      "Could not match the service instance to the JWT's client id or audience."
     );
-  } else {
-    logger.debug(`No JWT given to select binding to service '${service}.`);
+    return undefined;
   }
+  logger.debug(`No JWT given to select binding to service '${service}'.`);
 
   return getCredentialsWithoutJwt(service, credentialsList);
 }
--- a/packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
+++ b/packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
@@ -17,8 +17,10 @@ export function getXsuaaServiceCredentials(
     'xsuaa',
     token
   );
-  if (!credentials) {
-    throw new Error('Could not find binding to the XSUAA service.');
+  if (!credentials && token) {
+    throw new Error('Could not find a XSUAA service binding matching the token.');
+  } else if (!credentials) {
+    throw new Error('Could not find a XSUAA service binding.');
   }
   return credentials;
 }


packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
```diff
@@ -16,7 +16,7 @@ const logger = createLogger({
  * Throws an error if there are no bindings of the given type.
```
marikaner, line 16: [req] Can you also adjust the rest of the comment - this is still wrong.

packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
```diff
@@ -43,13 +43,12 @@ export function getServiceCredentials<
     if (credentials) {
       return credentials;
     }
-
     logger.debug(
-      'Could not match binding service instance using JWT client id or audience.'
+      "Could not match the service instance to the JWT's client id or audience."
```
marikaner, line 47: [pp] "the" implies that there is only one and we know which one how about
```suggestion
      'Could not match any of the bound service instances to the given JWT's client id nor audience.'
```

packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
```diff
@@ -43,13 +43,12 @@ export function getServiceCredentials<
...
     );
-  } else {
-    logger.debug(`No JWT given to select binding to service '${service}.`);
+    return undefined;
```
marikaner, line 49: [req] Instead of this you could write this code a bit more readable:

```ts

if (!credentials) {
    // log debug message
}

return credentials
```

packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
```diff
@@ -94,7 +93,8 @@ function getCredentialsWithJwt<ServiceCredentialsT extends ServiceCredentials>(
     c => matchesClientId(c, token) || matchesAudience(c, token)
   );
 
-  logResult(service, eligibleCredentials, true);
+  // eslint-disable-next-line no-unused-expressions
+  eligibleCredentials[0] ? logResult(service, eligibleCredentials, true) : logger.warn(`Found no service binding for service '${service}' matching either the token's client id or audience.`);
```
marikaner, line 97: [req|pp] Nooooooo! 😱 🤪

Please just add this to `logResult()` instead 😊

packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
```diff
@@ -17,8 +17,10 @@ export function getXsuaaServiceCredentials(
     'xsuaa',
     token
   );
-  if (!credentials) {
-    throw new Error('Could not find binding to the XSUAA service.');
+  if (!credentials && token) {
+    throw new Error('Could not find a XSUAA service binding matching the token.');
+  } else if (!credentials) {
+    throw new Error('Could not find a XSUAA service binding.');
```
marikaner, line 23: [pp] This checks the same condition (`!credentials`) twice, I think this could be easier to read with the following:

```suggestion
  if (!credentials) {
    throw new Error(token ? 'Could not find a XSUAA service binding matching the token.' : 'Could not find a XSUAA service binding.');
```

marikaner (changes requested):
Some codestyle and readability issues, but other than that looks good.

packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
tomfrenken: This was intentional, the logResult() method is used for both `getCredentialsWithoutJwt` and `getCredentialsWithJwt`, this specific logic only applies to `...WithJwt`.

tomfrenken: Now that I look at it again, we already pass a "usedToken" boolean, I'm an idiot 🤦

packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
```diff
@@ -12,11 +12,11 @@ const logger = createLogger({
 /**
  * @internal
  * Retrieves a service binding of the given type and tries to match it based on the JWT, if given.
- * If no match was found or no JWT is given, the first binding is returned.
- * Throws an error if there are no bindings of the given type.
+ * If no match between a given JWT and a service binding is found, undefined is returned.
```
marikaner, line 15:
```suggestion
 * Returns `undefined` if no match is found.
```

packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
```diff
@@ -12,11 +12,11 @@ const logger = createLogger({
...
+ * If no JWT is given, the first binding or undefined is returned, depending on if a binding exists.
```
marikaner, line 16:
```suggestion
 * If no JWT is given, returns the first binding of the given service type, if available.
```

packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
```diff
@@ -40,16 +40,9 @@ export function getServiceCredentials<
...
-    if (credentials) {
-      return credentials;
-    }
-
-    logger.debug(
-      'Could not match binding service instance using JWT client id or audience.'
-    );
-  } else {
-    logger.debug(`No JWT given to select binding to service '${service}.`);
+    return credentials ?? undefined;
```
marikaner, line 43: Is this ever `null`? If so, why? If not, why do we need the `??` check? (Did I suggest this in the review? ;))

packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
```diff
@@ -110,7 +103,10 @@ function logResult<ServiceCredentialsT extends ServiceCredentials>(
   credentials: ServiceCredentialsT[],
   usedToken: boolean
 ) {
-  if (credentials.length === 1) {
+  if (!credentials && usedToken){
```
marikaner, line 106: [pp] Not a fan of the order of the conditions. I would propose the following:

```ts
if (credentials?.length === 1) {}
else if (credentials?.length > 1) {}
else if (usedToken) {}
```

packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
```diff
@@ -18,7 +18,7 @@ export function getXsuaaServiceCredentials(
     token
   );
   if (!credentials) {
-    throw new Error('Could not find binding to the XSUAA service.');
+    throw new Error(token ? 'Could not find a XSUAA service binding matching the token.' : 'Could not find a XSUAA service binding.');
```
marikaner, line 21: [req] Either use "an" or remove the "a". I am in favor of removing.
```suggestion
    throw new Error(token ? 'Could not find XSUAA service binding matching the token.' : 'Could not find XSUAA service binding.');
```

marikaner (approved):
LGTM just one grammatical issue, that is [req] everything else is just suggestions.

packages/connectivity/src/scp-cf/environment-accessor/service-credentials.ts
tomfrenken: Good catch, I assumed one of the 2 functions which can be called could return null, but it could be undefined at most, I guess I can remove it (no you didn't suggest it 😂)

PR #4057, opened 2023-08-15
fix: Support ClientCertificateAuthentication without password

marikaner:
Allow MTLS destinations to not have password. Closes https://github.com/SAP/cloud-sdk-backlog/issues/1125.

### DIFF
--- a/packages/connectivity/src/http-agent/http-agent.ts
+++ b/packages/connectivity/src/http-agent/http-agent.ts
@@ -123,18 +123,22 @@ function getTrustStoreOptions(
  */
 function getKeyStoreOptions(destination: Destination): Record<string, any> {
   if (
-    destination.keyStoreName &&
-    destination.keyStorePassword &&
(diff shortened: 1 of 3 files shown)


packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -123,18 +123,22 @@ function getTrustStoreOptions(
...
-    !mtlsIsEnabled(destination)
+    !mtlsIsEnabled(destination) &&
+    destination.keyStoreName
   ) {
     const certificate = selectCertificate(destination);
 
     logger.debug(`Certificate with name "${certificate.name}" selected.`);
 
+    if (!destination.keyStorePassword) {
+      logger.debug(
```
tomfrenken, line 138: From my limited understanding a passphrase provides additional security, therefore leaving it out makes it more vulnerable to attacks, do you think debug is enough, or would a warning be more appropriate? 
(I don't really have an opinion here)

tomfrenken (approved):
LGTM, just one comment.

PR #4065, opened 2023-08-18
fix: Fix caching for destinations from vcap

deekshas8:
Closes SAP/cloud-sdk-backlog#1131.

Fix caching implementation.


packages/connectivity/src/scp-cf/destination/destination-from-vcap.spec.ts
```diff
@@ -158,6 +159,29 @@ describe('vcap-service-destination', () => {
     ).toBeDefined();
   });
 
+  it('returns undefined if cached destination jwt has expired', async () => {
```
marikaner, line 162:
```suggestion
  it('returns undefined if cached destination JWT has expired', async () => {
```

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -53,28 +53,31 @@ export async function getDestinationFromServiceBinding(
     : undefined;
 
   const retrievalOptions = { ...options, jwt: decodedJwt };
-  const destination =
-    (await retrieveDestinationFromCache(retrievalOptions)) ||
-    (await retrieveDestinationWithoutCache(retrievalOptions));
+  let destination = await retrieveDestinationFromCache(retrievalOptions);
```
marikaner, line 56: I wonder what I was thinking when writing this code. Seems super-weird.
~[req]~ I think that this is not enough though, if we have two requests with the same destination and the first is called with `useCache = true`, the second with `useCache = false`, and the token is not yet expired it will use the destination from the cache, which is not supposed to do, right?

I guess an additional test for this would be helpful. 

*EDIT*: I just saw that `retrieveDestinationCache()` checks for `useCache`. This is not obvious and I think it might be confusing at this point in the code (again: why did I build it this way? 🤪)
[nth] I would prefer to drag the `useCache` check up to here - maybe we just don't need the `retrieveDestinationWithoutCache()` function.

marikaner (approved):

PR #4095, opened 2023-08-30
fix(destinations): fixed selectionStrategy compare function

ThomasSwolfs:
Closes SAP/cloud-sdk-backlog#4090 .

Fixes the issue where selectionStrategy is not correctly compared when imported from different locations.


mr-flannery:
The `lint` step is currently failing. You should be able to fix it by running `yarn run lint:fix` from the repo root.

mr-flannery (approved):
Sorry that it took so long to merge this and thank you again for your contribution!

PR #4144, opened 2023-09-22
fix: Add support for `useDefineForClassFields: true`

deekshas8:
This PR adds support for `useDefineForClassFields: true` ,the default value when target in tsconfig is set to ES2022(recommended for node18). 
- Add ES2022 in compiler options for odata generator.
- Add `declare` keyword to field properties of v2/v4 entity to [avoid initialization to `undefined`](https://github.com/microsoft/TypeScript/issues/37640#issuecomment-604995672) 
- Move initialization of `_oDataVersion` to constructor
- Add tests for checking nonEnumerable properties
- Add tests to check requestConfig.payload does not contain properties that were not explicitly set 

Closes SAP/cloud-sdk-backlog#1136.


packages/generator/src/operations/request-builder-arguments.ts
```diff
@@ -12,7 +12,7 @@ export function getRequestBuilderArguments(
   const sharedParameters = [getTransformer(operation), 'params'];
   if (operation.isBound) {
     return [
-      'this._entityApi',
+      `this._entityApi as ${operation.entityClassName}Api<T>`,
```
marikaner, line 15: [q] Why is this needed? Of course without `as` it would be nicer, if possible.

scripts/check-public-api.ts
```diff
@@ -122,7 +122,6 @@ export async function checkApiOfPackage(pathToPackage: string): Promise<void> {
     const { pathToSource, pathCompiled } = paths(pathToPackage);
     mockFileSystem(pathToPackage);
     const opts = await getCompilerOptions(pathToPackage);
-    console.log(JSON.stringify(opts));
```
marikaner, line 125: good catch!

marikaner (commented):
Looks good, I left 1 minor comment only.
I would like to still try it manually.

Please note, that this is definitely not `chore` but `fix` and the changeset is missing. I think this should be `minor`. 😊

marikaner (commented):
@deekshas8 Do we have a test that checks what `requestConfig.payload` is when calling `new CreateRequestBuilder(...)`? Or something comparable? I'm pretty sure we have that, but if not it would be great to see that this fails with the old state and works with the new state somehow (even though it is difficult to switch to es2022 in the tests...).

packages/generator/src/operations/request-builder-arguments.ts
deekshas8: Without explicit typecast, I got an error:
 ```
Type 'DefaultDeSerializers' is not assignable to type 'T'.
          'DefaultDeSerializers' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'DeSerializers<any, any, any, any, any, any, any, any, any, any, any, any, any, any, any, any, any, any, any>'.
```

**Comment by deekshas8** [comment]
> @deekshas8 Do we have a test that checks what `requestConfig.payload` is when calling `new CreateRequestBuilder(...)`? 

We do kinda test it [here](https://github.com/SAP/cloud-sdk-js/blob/959255385b85400f013fbb4530ee501ae83361d5/packages/odata-v2/src/request-builder/create-request-builder.spec.ts#L22) . When I ran this test with ES2022 setup (I generated the `TestEntity` with ES2022), the test failed because the `postBody` no longer matched (had all other properties with `undefined` ). I will check if we can somehow add an explicit test for this.

File: packages/generator/src/operations/request-builder-arguments.ts (hunk 1 above)
Comment (marikaner) [reply packages/generator/src/operations/request-builder-arguments.ts:15]: I think this is an issue with how the `defaultDeSerialisers` are used here. Instead of passing them around as `deSerializers || defaultDeserializers` we could add this `deSerializers: T = defaultDeSerializers as T` to the parameters instead of `deSerializers?: T`. We still have an explicit type cast, but it is on a more defined property and thus more controllable (less error prone).

File: .changeset/perfect-toes-behave.md [hunk 3]
```diff
@@ -0,0 +1,13 @@
...
+'@sap-cloud-sdk/test-services-odata-v4': minor
+'@sap-cloud-sdk/test-services-e2e': minor
+'@sap-cloud-sdk/generator-common': minor
+'@sap-cloud-sdk/odata-common': minor
+'@sap-cloud-sdk/generator': minor
+'@sap-cloud-sdk/odata-v2': minor
+'@sap-cloud-sdk/odata-v4': minor
+---
+
+[Fixed Issue] Add support for `useDefineForClassFields` which defaults to `true` when target is `ES2022`.
```
marikaner, line 13: [nth] Consider describing the fixed issue, that the user had (that null values were sent for properties that were not set when writing entities) instead of the technical underlying issue.

packages/generator/src/operations/operation.ts
```diff
@@ -33,9 +33,11 @@ export function operationFunctionBase(
         }>`
       },
       {
-        name: isBound ? 'deSerializers?' : 'deSerializers',
+        name: 'deSerializers',
         type: isBound ? 'T' : 'DeSerializersT',
```
marikaner, line 37: [q] Is this line correct? `T` seems like a leftover.

marikaner (approved):
I left minor comments, although one of them could even be a bug. Otherwise, LGTM!

Issue #4152, opened 2023-09-25
Not able to fetch X-CSRF-Token from onPremise system


drvup:
**Describe the bug**
Dear SDK Team, 

I am not able to fetch X-CSRF-Tokens anymore within `@sap-cloud-sdk/http-client` in version `3.5.0`.
The "csrf-middleware" results every time in 500 http status error:
<img width="1327" alt="xcsrfTokenFailed" src="https://github.com/SAP/cloud-sdk-js/assets/9018930/a55dff8f-46da-4586-87f6-bebb7b2865ce">

If I do fire the same `HEAD` request, the SDK is generating to fetch the X-CSRF-Token within postman, I do get a 200 status code within an X-CSRF-TOKEN back from the onPremise system (see expected behavior below).

This seems to be an issue between the SDK and axios. But I am not able to debug the boilerplate efficiently to identify the reason for it. 
Can you please try to replicate fetching X-CSRF-Tokens from an onPremise system (via Cloud Connector) with your "csrf-middleware"?

**To Reproduce**

1. Create a VDM within Entities you can post something to 
2. Set up a Create or Update statement ( without `skipCsrfTokenFetching()` ;-) )
3. Execute the request and you'll receive a http 500 status error

**Expected behavior**
Instead of failing within http 500 status, the middleware should get the X-CSRF-TOKEN from the onPremise system, like it's done and feasible with postman (btw: via the same cloud-connector-ssh tunnel acting as proxy). I did copy exact every header-property, the baseurl and url to postman and did use the proxy via the SAP CC:
<img width="965" alt="xcsrfTokenPostman" src="https://github.com/SAP/cloud-sdk-js/assets/9018930/090b656f-7aa4-40ed-977f-95c1e397bffd">
Success!

**Used Versions:**
- node version via 18.15.0
- npm version: 9.5.0
- SAP Cloud SDK version : 3.5.0

**Impact / Priority**
Affected development phase: Release
Impact: Blocked
Timeline: UAT in 4 weeks

Kind Regards,
Cedric

marikaner:
Hey @drvup,
thank you for the thorough replication steps, I will try to reproduce this and let you know once I know more.

drvup:
Hey @marikaner ,
thanks for the support. I did roll back to `@sap-cloud-sdk/http-client` (and connectivity) `3.1.1` and there it's working, as soon as I delete the "data" property on the "requestConfig" within the `csrf-token-middleware.js`:
[...]

PR #4161, opened 2023-09-27
fix: fix csrf token retrieval

marikaner:
Remove the parameter encoder from the CSRF request. It seems to cause issues when making requests against OnPremise systems and it is not needed.

Relates to [Not able to fetch X-CSRF-Token from onPremise system](https://github.com/SAP/cloud-sdk-js/issues/4152#top)


packages/http-client/src/csrf-token-middleware.spec.ts
```diff
@@ -124,13 +124,11 @@ describe('CSRF middleware', () => {
...
+
     expect(spy).toHaveBeenNthCalledWith(
       1,
-      expect.objectContaining({
-        method: 'head',
-        headers: csrfFetchHeader,
-        url: 'some/path/',
-        data: {}
+      expect.not.objectContaining({
+        data: expect.anything()
```
marikaner, line 131:
```suggestion
        data: expect.anything(),
        params: expect.anything(),
        parameterEncoder: expect.anything()
```

tomfrenken (approved):
LGTM

marikaner:
This seems to work and was confirmed by an internal user, however, the request itself also fails, so this is not "done" yet.

PR #4177, opened 2023-10-02
Fix write cc 2

marikaner:

### DIFF
--- a/packages/http-client/src/http-client-types.ts
+++ b/packages/http-client/src/http-client-types.ts
@@ -4,6 +4,7 @@ import type {
   Middleware,
   MiddlewareOptions
 } from '@sap-cloud-sdk/resilience';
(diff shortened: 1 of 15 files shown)


packages/http-client/src/http-client-types.ts
```diff
@@ -43,25 +44,29 @@ export type HttpMiddleware = Middleware<
...
   httpAgent?: http.Agent;
   /**
-   * `httpsAgent` defines a custom agent to be used when performing https requests.
+   * Agent to be used when performing HTTPS requests.
    */
   httpsAgent?: http.Agent;
+  /**
+   * Proxy configuration, when going through a proxy like the SAP Cloud Connector.
+   */
+  proxy: BasicProxyConfiguration | false;
```
jjtang1985, line 69: This seems to be a breaking change.

Issue #4198, opened 2023-10-11
How to avoid warning: The provided JWT payload does not include a 'zid' property.


hosssa:
We are successfully performing an executeHttpRequest via BTP destination. 

```js
executeHttpRequest(
    { 
      destinationName: "NAME",
      useCache: true
    },
    requestConfig
)
```

However, we constantly get the following warning, which fills up our logs:

`(jwt): Warning JWT: The provided JWT payload does not include a 'zid' property.`

How can we avoid this warning? Is there a way to disable it?

marikaner:
Hey @hosssa, thanks for bringing that up. Does this occur when use disable the cache with `useCache: false`?
I believe this warning occurs because you are using the cache. By default the cache is isolated by user and tenant. As you don't pass a JWT, there is no tenant information, so that the destination cannot be safely cached with the given isolation. You are probably also seeing a second warning shortly after: `Could not build destination cache key. Isolation strategy 'tenant-user' is used, but tenant id or user id is undefined in JWT.`, right?

I might be wrong though, so let me know if you see the other warning as well. I will take a closer look at this tomorrow.

nocin:
Hi,

the messages also occurs with `useCache: false`. Also, there is no second warning after it. 
![grafik](https://github.com/SAP/cloud-sdk-js/assets/35531968/be2fe373-48fd-4020-8512-f98a4007656b)

I'm testing locally in BAS with a destination setup in a .env file and without multitenancy. 

BR, Nico

marikaner:
Hey @hosssa, I am working on improving the warning, it is quite misleading.

For your other request: The default isolation strategy is "tenant-user" if you pass a token that has a `user_id`. If you pass no token or a token without user ID, the strategy is `tenant`. If you don't have multitenancy, you basically have only one tenant - so it is a special case of isolation strategy "tenant".
If you don't pass a JWT, the SDK retrieves the subaccount ID and uses this as the tenant and assumes isolation strategy "tenant" => this is also why you don't see the second error message that I erroneously expected.
Did you have issues with caching or did I just confuse you with my previous message?

hosssa:
Hi @marikaner, 

Thanks for the details about the caching mechanism. I assume caching works fine, as the first request throws 2 'zid' warning messages but subsequent requests only one.

marikaner:
I opened a PR to remove the logs alltogether, I currently don't see a good reason to keep them. We will soon release a new version to include this, stay tuned ;)

PR #4208, opened 2023-10-18
chore: Refactor JWT access

marikaner:
Closes https://github.com/SAP/cloud-sdk-js/issues/4198.

This PR removes some overcomplicated JWT access and changes logs around that from warnings to debug messages.
The `zid` property is probably the most relevant to look out for and I couldn't find a place where we don't have an alternative to the `zid` property. This renders warnings irrelevant.


packages/connectivity/src/scp-cf/jwt.ts
```diff
@@ -15,6 +15,124 @@ const logger = createLogger({
...
+}
+
+/**
+ * Get the tenant name of a decoded JWT.
+ * @param jwtPayload - Token payload to read the tenant name from.
+ * @returns The tenant name, if available.
+ * @internal
+ */
+export function tenantName({ ext_attr }: JwtPayload): string {
+  return ext_attr?.zdn;
```
deekshas8, line 113: [q] I don't see any usage internally, do we need this?

packages/connectivity/src/scp-cf/jwt.ts
```diff
@@ -15,6 +15,124 @@ const logger = createLogger({
...
+  return [];
+}
+
+/**
+ * Extracts the custom attributes from the JWT.
+ * @param payload - Token payload to read the custom attributes from.
+ * @returns Custom attributes added by the XSUAA service to the issued JWT.
+ * @internal
+ */
+export function customAttributes(payload: JwtPayload): Record<string, any> {
```
tomfrenken, line 92: As discussed, perhaps move these into their own object so that a user can access all properties from one location with nice code completion in their IDE.
Other than that LGTM

packages/connectivity/src/scp-cf/jwt.ts
```diff
@@ -15,6 +15,124 @@ const logger = createLogger({
...
+ * @param jwtPayload - Token payload to read the user's scopes from.
+ * @returns The user's scopes if available.
+ * @internal
+ */
+export function userScopes(jwtPayload: JwtPayload): string[] {
+  const scopes = jwtPayload.scope;
+  if (Array.isArray(scopes)) {
+    return scopes.map(scope =>
+      scope.includes('.')
+        ? scope.substr(scope.indexOf('.') + 1, scope.length)
```
tomfrenken, line 79: Also if you find the time, make sure this is not just arbitrary 😅

packages/connectivity/src/scp-cf/jwt.ts
marikaner: This is a very good question. This was intended to be public API, but it is hasn't been since version 3 and no one complained. I discussed with Tom, that we will expose this functionality in a slightly different format publicly and remove all those unused functions from the internal scope.

packages/connectivity/src/scp-cf/jwt.spec.ts
```diff
@@ -252,37 +252,37 @@ describe('jwt', () => {
...
     });
 
-    it('returns a set of the entries of the "scope" claim iff the "aud" claim is not present and if the entry contains a dot (and then we again only take everything before the dot', () => {
+    it('returns audiences from scope, if no "aud" property exists', () => {
       const token = {
         scope: ['one', 'two.one', 'three.two.one']
       };
 
-      expect(audiences(token)).toEqual(new Set(['two', 'three']));
+      expect(audiences(token)).toEqual(['two', 'three']);
```
deekshas8, line 275: Not sure why 'one' is not included

deekshas8 (approved):
Left a question. Otherwise LGTM

packages/connectivity/src/scp-cf/jwt.spec.ts
marikaner: Because it only takes those that are in front of the "."

PR #4249, opened 2023-11-02
chore: support location ID in mail destinations

mr-flannery:
~Closes~ contributes to SAP/cloud-sdk-backlog#1126.

requires extending the e2e tests


packages/mail-client/src/socket-proxy.ts
```diff
@@ -18,8 +27,9 @@ export async function customAuthRequestHandler(jwt: string): Promise<Buffer> {
     // The actual value of the JWT in its encoded form
     ...jwtBytes,
     // Length of the Cloud Connector location ID (0 if no Cloud Connector location ID is used)
-    0x00
+    0x00,
```
marikaner, line 30: [q] Shouldn't the length also be adjusted?

mr-flannery: oh, duh, yeah, it probably should 🤦 thanks

packages/mail-client/src/socket-proxy.ts
```diff
@@ -5,11 +5,26 @@
...
   const jwtBytesLength = jwtBytes.length;
+
   const jwtLengthBuffer = Buffer.alloc(4);
   jwtLengthBuffer.writeUInt32BE(jwtBytesLength, 0);
+
+  const cloudConnectorLocationIdBytes = cloudConnectorLocationId
+    ? [...Buffer.from(cloudConnectorLocationId)]
+    : [];
+
+  const cloudConnectorLocationIdLengthBuffer = Buffer.alloc(1);
```
marikaner, line 22: [q] Is it intended that this is different than the `jwtLengthBuffer` - I would expect the same format is needed and that the functionality can be extracted into a separate function then. Something with a signature similar to this.

```ts
function createBuffer(): { buffer; length; }
```

marikaner (approved):
👍

PR #4295, opened 2023-11-21
fix: add relativeToRequestURI BatchSubRequestPathType

### LINKED ISSUE #4293: Support OData V4 Batch Request with "Resource path relative to the batch request URI."
**Is your feature request related to a problem? Please describe.**
When using OData V4 Batch Requests with a custom RAP service, we encounter an issue where the request is rejected with the error message "Die Data-Services-Anforderung wurde wegen einer deformierten Syntax nicht verstanden." (The data services request was not understood due to a malformed syntax.)
Upon troubleshooting, we discovered that this issue is caused by the resource path in the batch payload.

Both `POST /Entity` and `POST <absolute_Path/Entity>` are rejected, while `POST Entity` works but is not supported by the SDK. Related documention [Configure Subrequest Serialization](https://sap.github.io/cloud-sdk/docs/js/features/odata/v4-client#configure-subrequest-serialization).

The [OData specification](https://docs.oasis-open.org/odata/odata/v4.01/csprd02/part1-protocol/odata-v4.01-csprd02-part1-protocol.html#sec_MultipartBatchRequestBody) provides an example (Example 97) that demonstrates how the entity should be addressed.

**Describe the solution you'd like**
To address this issue, we suggest enhancing the `batch-request-serializer.ts` [getUrl method](https://github.com/SAP/cloud-sdk-js/blob/66b8bef46a64eefa6dfbddf1e44fbcc1996aa681/packages/odata-common/src/request-builder/batch/batch-request-serializer.ts#L93) to support relative paths without leading slashes.

**Describe alternatives you've considered**
An alternative would be to manually build the Batch request body and use the `executeHttpRequest`.

**Impact / Priority**

Affected development phase: Development

Impact: Impaired

Timeline: N/A

**Additional context**

cschuerings:
Closes #4293 

- [X] * Tests created/adjusted for your changes.
- [X] * PR title adheres to [conventional commit guidelines](https://www.conventionalcommits.org).
- [X] * Created a changeset `yarn changeset`
- [ ] * If applicable:
- [ ]   * Documented public API (TypeDoc).
- [X]   * Checked that `yarn run doc` still works.

### DIFF
--- a/.changeset/wet-chairs-enjoy.md
+++ b/.changeset/wet-chairs-enjoy.md
@@ -2,4 +2,4 @@
 '@sap-cloud-sdk/odata-common': patch
 ---
 
-add relativeToRequestURI BatchSubRequestPathType
+[Fixed Issue] Add support for subrequest serialization where a batch request may only contain an entity with no path.
--- a/.changeset/wet-chairs-enjoy.md
+++ b/.changeset/wet-chairs-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/odata-common': patch
+---
+
+add relativeToRequestURI BatchSubRequestPathType
--- a/packages/odata-common/src/request-builder/batch/__snapshots__/batch-request-serializer.spec.ts.snap
+++ b/packages/odata-common/src/request-builder/batch/__snapshots__/batch-request-serializer.spec.ts.snap
@@ -207,6 +207,17 @@ Accept: application/json
 "
 `;
 
+exports[`batch request serializer serializeRequest serializes getAll request with entity relative to request URI 1`] = `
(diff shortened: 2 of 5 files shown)


tomfrenken:
Hey @cschuerings,

already thanks for your nice contribution, it can take a day or two for the review, as we are a little short on people right now.

tomfrenken (approved):
LGTM!
The code & tests look clean, after a small discussion with the team I only made a few adjustments regarding the naming of certain variables.
While our `noPath` option may not be the very best choice naming-wise, our preferred choice would have introduced breaking changes to the current set of options, therefore we think this might be the most straight forward solution for now.

PR #4371, opened 2023-12-21
chore: Fetch one destination instead of all

marikaner:
Closes SAP/cloud-sdk-backlog#1158.

Fetch (and cache, if needed) only one destination instead of all available destinations. With this we can currently make just one request for the metadata, instead of 2.

### DIFF
--- a/.changeset/funny-weeks-warn.md
+++ b/.changeset/funny-weeks-warn.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Compatibility Note] The SAP Cloud SDK used to get the metadata for a destination (without potentially required authentication flows), through two requests to the destination service `/subaccountDestinations` and `/instanceDestinations`. While this approach has a some advantages when caching, it can cause severe performance issues when not caching. From now on, only one destination is retrieved per requested destination through `/destinations`.
--- a/.changeset/funny-weeks-warn.md
+++ b/.changeset/funny-weeks-warn.md
@@ -2,4 +2,5 @@
 '@sap-cloud-sdk/connectivity': minor
 ---
 
-[Compatibility Note] The SAP Cloud SDK used to get the metadata for a destination (without potentially required authentication flows), through two requests to the destination service `/subaccountDestinations` and `/instanceDestinations`. While this approach has a some advantages when caching, it can cause severe performance issues when not caching. From now on, only one destination is retrieved per requested destination through `/destinations`.
+[Compatibility Note] The SAP Cloud SDK used to get all subaccount and instance destinations, that are available with your JWT (without potentially required token retrieval), through two requests to the destination service (`/subaccountDestinations` and `/instanceDestinations`). While this approach can have advantages when caching, it can cause severe performance issues without caching. Therefore, from now on, only one destination is retrieved per requested destination through `/destinations`.
+You can no longer rely on the SDK to automatically cache all destinations on the first request. If needed, you can call `getAllDestinationsFromDestinationService()` with cache enabled instead.
--- a/.changeset/gorgeous-dogs-beam.md
+++ b/.changeset/gorgeous-dogs-beam.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Improvement] Retrieve only one destination per requested destination instead of all subaccount and instance destinations.
--- a/.changeset/gorgeous-dogs-beam.md
+++ b/.changeset/gorgeous-dogs-beam.md
@@ -2,4 +2,4 @@
 '@sap-cloud-sdk/connectivity': minor
 ---
 
-[Improvement] Retrieve only one destination per requested destination instead of all subaccount and instance destinations.
+[Improvement] Retrieve only one destination per requested destination instead of all subaccount and instance destinations. (See compatibility notes.)
--- a/packages/connectivity/src/scp-cf/destination/destination-service.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-service.ts
@@ -23,7 +23,10 @@ import {
   parseCertificate,
   parseDestination
 } from './destination';
-import { DestinationFetchOptions } from './destination-accessor-types';
+import {
+  DestinationFetchOptions,
+  DestinationsByType
+} from './destination-accessor-types';
 import { destinationServiceCache } from './destination-service-cache';
 import {
   Destination,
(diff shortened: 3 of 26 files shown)


packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -240,30 +235,43 @@ function getTenantFromTokens(token: AuthAndExchangeTokens | string): string {
   return tenant;
 }
 
-async function fetchDestinationByTokens(
+/**
+ * Fetches a specific destination by name from the given URI, including authorization tokens.
+ * For destinations with authenticationType OAuth2SAMLBearerAssertion, this call will trigger the OAuth2SAMLBearerFlow against the target destination.
+ * In this pass the access token as string.
```
mr-flannery, line 241: image

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -150,26 +124,47 @@ export interface AuthAndExchangeTokens {
...
+      return {
+        instance: [],
+        subaccount: []
+      };
+    }
+    throw new ErrorWithCause(
+      `Failed to fetch destination.${errorMessageFromResponse(err)}`,
+      err
+    );
+  }
```
mr-flannery, line 167: Is my understanding correct that you've changed the implementation to "standalone" instead of calling `fetchDestinationByTokens`? The difference that stands out to me is `?$skipTokenRetrieval=true`. Can you provide some context what the idea here is?

Also I don't understand what is meant by "fetching destination _by token_". Could you explain this to me please?

mr-flannery (commented):
Ok some after some confusion I think I understood what's going on and why your change addresses the issue:
- destination-accessor calls getDestinationFromDestinationService
- after like three levels of indirection, `getSubscriberDestinationService` and `getProviderDestinationService` get called
- these now call `fetchDestination` instead of `fetchEverythingAndTheirGrandma`

Is my understanding correct?

I haven't looked at the tests yet.

.changeset/gorgeous-dogs-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Improvement] Retrieve only one destination per requested destination instead of all subaccount and instance destinations.
```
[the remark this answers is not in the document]
marikaner: It doesn't matter and I am not sure if you can put multiple messages into one ;) (you probably can, but I didn't bother trying it)

test-resources/test/test-util/destination-service-mocks.ts
```diff
@@ -58,20 +59,168 @@ export function mockSubaccountDestinationsCall(
...
 
-export function mockSingleDestinationCall(
-  nockRef: nockFunction,
-  response: any,
-  responseCode: number,
-  destName: string,
-  headers: Record<string, any>,
-  options?: { uri?: string; badheaders?: string[] }
+type MockedDestinationConfiguration =
+  | (DestinationJson & {
```
mr-flannery, line 63: this leading `|` tripped me. I double checked it in the typescript playground, which doesn't complain about it, so I guess it's fine. wasn't aware this syntax was possible.

test-resources/test/test-util/destination-service-mocks.ts
```diff
@@ -58,20 +59,168 @@ export function mockSubaccountDestinationsCall(
...
+   * Mock a destination call that includes the auth flow. If set to `true`, the metadata configuration will be reused.
+   * If a configuration is given, it overwrites the default.
+   */
+  mockAuthCall?:
+    | boolean
+    | {
+        response?: any;
+        responseCode?: number;
+        headers?: Record<string, any>;
+      };
```
mr-flannery, line 92: After looking at this and the implmentation below in more detail, I'd offer a different suggestion.

Firstly, as discussed, I wouldn't use the name "metadata" because it's not part of the Destination service's model. 
Secondly, I would actually consider simplifying what the function is supposed to do and put more of the complexity of the caller. Hear me out:

The production code has two functions, which are currently called `fetchDestination` and `fetchDestinationBytoken`. The mock function attempts to do both, depending on what parameters are required. This already could be considered as a mental mismatch. If I _know_ that there are two calls potentially being made in the backend, then I think it's fair two call `mockFindDestinationCall` two times in the test. This has the added benefit of communicating cleary to whoever reads the test afterwards, that there are two calls being made. In scenarios where I don't care specifically assert things on the individual nock scopes, this is a detail that could be missed more easily when the function mocks two calls.

In addition, you could simplify the implementation, because you don't need to check whether different http params are set for the mockAuthCall, you can just use the ones that are being passed via the existing parameters.

This would mean that instead of having two params `mockMetadataCall` and `mockAuthCall`, you'd only have one, which could be called `skipTokenRetrieval`. It could still default to true, and if it's set, you throw away the `authTokens` from the response and add the param to the uri. And if it's not set, you basically just mock the calls as defined by the params.

One more thing I find worth thinking about is aligning this API with the API of the production code. As said above, there are two separate functions in the production code, while there's only one mock function (even if you would refactor it the way I suggested). I think there's value in having an analogous API in both branches of the code. If I understood how the productive code works, I don't have to mentally map it to the test API. It's just the same. This would imply, though, that you either unify the production code, or split the test code into two mocking functions (at least in terms of its public API).

Finally, I'd align the names with the production code. I.e. either rename the prod code to `findDestination`, or rename the test code to `mockFetchDestinationCall`.

🖐️ 🎤

mr-flannery: The only thing I feel strongly about is the `metadata` naming thing. If you want to keep the function and API as is, I honestly don't care that much. But then I'd at least call the params `mockCallWithTokenRetrieval` and `mockCallWithoutTokenRetrieval` or something among those lines.

test-resources/test/test-util/destination-service-mocks.ts
marikaner: This is done automatically by the linter. Initially I also stumbled over that, but when the types are longer, it is easier to grasp this way.

.changeset/funny-weeks-warn.md
```diff
@@ -2,4 +2,5 @@
 '@sap-cloud-sdk/connectivity': minor
 ---
 
-[Compatibility Note] The SAP Cloud SDK used to get the metadata for a destination (without potentially required authentication flows), through two requests to the destination service `/subaccountDestinations` and `/instanceDestinations`. While this approach has a some advantages when caching, it can cause severe performance issues when not caching. From now on, only one destination is retrieved per requested destination through `/destinations`.
+[Compatibility Note] The SAP Cloud SDK used to get all subaccount and instance destinations, that are available with your JWT (without potentially required token retrieval), through two requests to the destination service (`/subaccountDestinations` and `/instanceDestinations`). While this approach can have advantages when caching, it can cause severe performance issues without caching. Therefore, from now on, only one destination is retrieved per requested destination through `/destinations`.
+You can no longer rely on the SDK to automatically cache all destinations on the first request. If needed, you can call `getAllDestinationsFromDestinationService()` with cache enabled instead.
```
mr-flannery, line 6: 👍

mr-flannery (approved):
lak mach

PR #4465, opened 2024-02-05
chore: Remove manual mocks from tests

marikaner:
We still had a left-over from ancient times, where we mocked fs manually. Using `mock-fs` instead.


packages/test-util/src/test-destination-provider.ts
```diff
@@ -244,7 +244,7 @@ function readCredentials(filePath: string): CredentialsFile {
 function readJson(filePath: string): { [key: string]: any } {
   let content;
   try {
-    content = readFileSync(filePath, 'utf8');
+    content = readFileSync(filePath, 'utf-8');
```
tomfrenken, line 247: Might as well change it to async 👀

packages/test-util/src/test-destination-provider.ts
```diff
@@ -167,7 +167,7 @@ function readCredentialsOrFail(
 function mergeSystemsAndCredentials(systems, credentials) {
   return systems.map(system => ({
     ...system,
-    ...credentials.find(cred => cred.alias === system.alias)
+    ...credentials.find(({ alias }) => alias === system.alias)
```
tomfrenken, line 170: Didn't know you can destructure from a spread object like this, that's cool!

tomfrenken (approved):
LGTM, one minor suggestions though.

packages/test-util/src/test-destination-provider.ts
marikaner: No, because this would change the public API. ¯\_(ツ)_/¯

tomfrenken: Indeed 😬 - I just saw `test` in the file path and my brain jumped to private API.

PR #4473, opened 2024-02-06
feat: Support PEM certificates for ClientCertificateAuthentication

marikaner:
Currently I am using the whole file for both the key and certificate. That way we don’t have to parse the content of the certificate and it seems that the https agent can parse it on its own.

### DIFF
--- a/packages/connectivity/src/http-agent/http-agent.ts
+++ b/packages/connectivity/src/http-agent/http-agent.ts
@@ -114,24 +114,29 @@ function getTrustStoreOptions(destination: HttpDestination): {
 
 /**
  * @internal
- * The http agents (proxy and default) use node tls for the certificate handling. This method creates the options with the pfx and passphrase.
+ * The http agent uses node tls for the certificate handling. This method creates the options with the pfx and passphrase or key, cert and passphrase, depending on the format of the certificate.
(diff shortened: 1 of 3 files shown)


packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -114,24 +114,29 @@ function getTrustStoreOptions(destination: HttpDestination): {
...
-function getKeyStoreOptions(destination: Destination): {
-  pfx?: Buffer;
-  passphrase?: string;
-} {
+function getKeyStoreOptions(destination: Destination):
+  | {
+      pfx?: Buffer;
+      passphrase?: string;
+    }
+  | {
```
tomfrenken, line 127: Currently, every single parameter is optional, if one of the two options is chosen, isn't every single parameter required? Or can you have e.g. a pfx cert without passphrase or cert & key but no passphrase?

tomfrenken: Might even create a type for these two objects to make it a little more concise in the function API.

tomfrenken (commented):
LGTM, just a minor suggestion

tomfrenken (approved):

PR #4481, opened 2024-02-08
fix: fix default encoding in openapi

marikaner:
Currently the OpenAPI requestbuilder erroneously does not encode query parameters by default. 
This PR fixes that.

Closes SAP/cloud-sdk-backlog#1182.

### DIFF
--- a/packages/http-client/src/http-client.ts
+++ b/packages/http-client/src/http-client.ts
@@ -148,7 +148,7 @@ export function buildHttpRequestConfigWithOrigin(
  * @returns The parameters as they are without encoding.
  * @internal
  */
-export const odataTypedClientParameterEncoder: ParameterEncoder = (
+export const oDataTypedClientParameterEncoder: ParameterEncoder = (
   params: Record<string, any>
 ) => params;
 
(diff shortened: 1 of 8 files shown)


packages/http-client/src/http-client.ts
```diff
@@ -142,12 +143,12 @@ export function buildHttpRequestConfigWithOrigin(
 }
 
 /**
- * This method does nothing and is only there to indicate that the call was made by an OData or OpenApi client and encoding is already done on filter and key parameters.
+ * This method does nothing and is only there to indicate that the call was made by an typed OData client and encoding already happened on in the client.
```
deekshas8, line 146:
```suggestion
 * This method does nothing and is only there to indicate that the call was made by a typed OData client and encoding already happened in the client.
```

packages/http-client/src/http-client.ts
```diff
@@ -142,12 +143,12 @@ export function buildHttpRequestConfigWithOrigin(
...
  * @param params - Parameters which are returned.
  * @returns The parameters as they are without encoding.
  * @internal
  */
-export const encodeTypedClientRequest: ParameterEncoder = (
+export const odataTypedClientParameterEncoder: ParameterEncoder = (
```
deekshas8, line 151:
```suggestion
export const oDataTypedClientParameterEncoder: ParameterEncoder = (
```

I noticed in other places in the code, we write oData (e.g. commonODataUri). I would keep the casing consistent. I also noticed this in few other places in the PR

deekshas8 (commented):
Left some minor comments.

deekshas8 (approved):

Issue #4488, opened 2024-02-12
OpenAPI generator/client - custom request config has inaccurate type


sven-petersen:
**Describe the bug**

The `addCustomRequestConfiguration` method on a request builder accepts an object in the form of `Record<string, string>`. 
The params from this object **seem to be passed** (please correct if thats wrong) to axios which accepts more than just strings. Also there is no value help available which indicates which properties exist.

```typescript
const abortController = new AbortController();
await openApiClient.myApiGroup.doSomething({})
  .addCustomRequestConfiguration({
    timeout: 123, // Error: Type 'number' is not assignable to type 'string'
    signal: abortController.signal // Error: Type 'AbortSignal' is not assignable to type 'string'
  })
  .execute({ url: "foo.com " });
```

**Expected behavior**

When looking at https://sap.github.io/cloud-sdk/docs/js/features/openapi/execute-request#setting-a-custom-request-configuration I would expect the type to be something like the following:

```typescript
type CustomRequestConfig = Omit<
  AxiosRequestConfig<unknown>,
  "method" | "url" | "baseURL" | "data" | "headers" | "params"
>;
```
Also it would be great to expose the type from the SDK in order to easily use it for typing variables:

```typescript
const reusedConfig: CustomRequestConfig = { timeout: 2000 };
await openApiClient.myApiGroup.doSomething({})
  .addCustomRequestConfiguration(reusedConfig)
  .execute({ url: "foo.com " });
```

**Screenshots**

**Used Versions:**

- node version 20.11.0
- npm version: 10.2.4
- SAP Cloud SDK version you used as dependency

**Impact / Priority**

It is difficult to use the described APIs as it is required to explicitly add wrong type castings which is highly undesirable. 

Affected development phase: Development

deekshas8:
Hi @sven-petersen ,

I have created a PR to fix the type in `addCustomRequestConfiguration` to accept more than just string values.

Regarding the value help for existing properties, thank you for the suggestion. I will need to look into it a bit more.

marikaner:
Hey @sven-petersen this was fixed in version 13.12.1.

PR #4493, opened 2024-02-13
fix: Fix custom request config type

deekshas8:
Relates to https://github.com/SAP/cloud-sdk-js/issues/4488

The `addCustomRequestConfiguration` method on a request builder accepts an object in the form of Record<string, string>. However axios accepts other types than string as well.


marikaner (approved):
LGTM

PR #4508, opened 2024-02-19
fix: Fetch destination array by type in `fetchDestinationWithoutTokenRetrieval` 

deekshas8:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/1190

Fix check based on which subaccount destination array is returned. This leads to the warning log appearing if there is only one destination at instance level and not at sub-account level.

### DIFF
--- a/.changeset/seven-shoes-do.md
+++ b/.changeset/seven-shoes-do.md
@@ -2,4 +2,4 @@
 "@sap-cloud-sdk/connectivity": patch
 ---
 
-[Fixed Issue] Fix `fetchDestinationWithoutTokenRetrieval` to correctly return instance and subaccount destinations array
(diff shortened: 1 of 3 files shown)


.changeset/seven-shoes-do.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-cloud-sdk/connectivity": patch
+---
+
+[Fixed Issue] Fix `fetchDestinationWithoutTokenRetrieval` to correctly return instance and subaccount destinations array
```
marikaner, line 5:
```suggestion
[Fixed Issue] Fix `fetchDestinationWithoutTokenRetrieval` to correctly return instance and subaccount destinations.
```

marikaner (approved):
LGTM

PR #4541, opened 2024-03-05
chore: Remove `toAuthorizationHeader` and `assoc` function

marikaner:
I noticed this unnecessary function while investigating a user issue and I found it harder to read than necessary.

### DIFF
--- a/packages/util/src/object.ts
+++ b/packages/util/src/object.ts
@@ -88,7 +88,7 @@ export const exclude = <T extends Record<string, unknown>>(
  * @param value - Value to be added.
  * @param obj - Object the key value pair is added to.
  * @returns The object with the key value pair added.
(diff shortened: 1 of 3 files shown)


packages/util/src/object.ts
```diff
@@ -88,6 +88,7 @@ export const exclude = <T extends Record<string, unknown>>(
  * @param value - Value to be added.
  * @param obj - Object the key value pair is added to.
  * @returns The object with the key value pair added.
+ * @deprecated  Since v3.13.0. Will not be replaced.
```
marikaner, line 91:
```suggestion
 * @deprecated  Since v3.12.1. Will not be replaced.
```

tomfrenken (approved):
LGTM

PR #4564, opened 2024-03-14
feat: retrieve service bindings by tag to support user-provided services

mr-flannery:
Closes SAP/cloud-sdk-backlog#1169


.changeset/curly-buses-push.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-cloud-sdk/connectivity": minor
+---
+
+Service bindings will now be matched by tag if no match is found by label. This should support user-provided services if no other instances of the same type are bound.
```
marikaner, line 5: [req] This should be imperative.

packages/connectivity/src/scp-cf/environment-accessor/service-bindings.ts
```diff
@@ -8,23 +8,43 @@ const logger = createLogger({
...
 export function getServiceBindings(service: string): Service[] {
-  return xsenv.filterServices({ label: service });
+  const servicesByLabel = xsenv.filterServices({ label: service });
+
+  if (servicesByLabel.length) {
+    return servicesByLabel;
+  }
+
+  logger.debug(
+    'Found no service bindings for "%s" by label. Trying to match by tag instead.',
```
marikaner, line 25: [pp]
```suggestion
    'Could not find service bindings for "%s" by label. Trying to match by tag instead.',
```

marikaner (changes requested):
I am not sure how this is used, but I am wondering why we are looking at tags and not the name. Tags are not very unique aren't they?

marikaner (commented):
Also, don't forget to add this to the docs once you're done.

mr-flannery:
> I am not sure how this is used, but I am wondering why we are looking at tags and not the name. Tags are not very unique aren't they?

There seem to be use cases with CAP where matching user-provided services is required. I don't think this is necessarily the best contract, but the Java SDK has already implemented it, and it's non-breaking, so Junjie and I agreed to do it like this. Also I've never seen anyone use tags for anything. While not empirically useful, my concern that we accidentally introduce weird behavior is low, since per the old contract of the function, why would anyone pass a string that has not matched anything previously and now matches a tag.

mr-flannery:
> Also, don't forget to add this to the docs once you're done

somewhere here probably: Also, don't forget to add this to the docs once you're done.

PR #4565, opened 2024-03-14
chore: try to fix build

marikaner:
I probably broke the PR checks. This should fix it.


marikaner:
I am now in fact adding it to the other PR

PR #4571, opened 2024-03-15
chore: (really) fix canary build

marikaner:
I found that we were using the events api (`github.event.inputs`), which always gives you a string and I replaced it with the context api (`intputs`) that preserves the type.
This does not explain why it was previously necessary to compare to a boolean on strings, but I guess it has something to do with how it is compared and that `github.event.inputs.skip-checks` was falsy.

Also, I remoed the skip-check and always skip checks when running the canary build

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Show input from dispatch event
         run: echo "Canary release flag ${{ inputs.canary-release }}"
   tests:
-    if: inputs.canary-release == 'false'
+    if: inputs.canary-release == true
     runs-on: ubuntu-latest
     timeout-minutes: 15
     strategy:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,8 @@ on:
         description: 'Release canary version (skips tests and checks)'
         type: boolean
         required: false
-        default: false
+        # This looks like a boolean, but it's actually a string
+        default: 'false'
 
 jobs:
   log:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -12,19 +12,20 @@ on:
...
-        description: 'Canary release after build'
-        type: boolean
-        required: false
-        default: false
-      skip-checks:
-        description: 'Skip tests and checks'
+        description: 'Release canary version (skips tests and checks)'
         type: boolean
         required: false
+        # This looks like a boolean, but it's actually a string
```
mr-flannery, line 18: Isn't this comment now redundant/misleading when you've also switched to the inputs api and do actually check on booleans now?

.github/workflows/build.yml
```diff
@@ -12,19 +12,20 @@ on:
...
 
 jobs:
+  log:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Show input from dispatch event
+        run: echo "Canary release flag ${{ inputs.canary-release }}"
   tests:
-    if: github.event.inputs.skip-checks == false
+    if: inputs.canary-release == false
```
mr-flannery, line 28: Why are we only running the tests when not releasing a canary version?

.github/workflows/build.yml
```diff
@@ -49,7 +50,7 @@ jobs:
       - run: yarn test:build-packages
       - run: yarn test:type
   checks:
-    if: github.event.inputs.skip-checks == false
+    if: inputs.canary-release == false
```
mr-flannery, line 53: same question

.github/workflows/build.yml
```diff
@@ -78,7 +79,7 @@ jobs:
       - run: yarn check:license
         name: License Check
   e2e-tests:
-    if: github.event.inputs.skip-checks == false
+    if: inputs.canary-release == false
```
mr-flannery, line 82: same question

.github/workflows/build.yml
mr-flannery, line 18:
```suggestion
```

deekshas8 (approved):

PR #4630, opened 2024-04-10
fix: Openapi-generator array schema with nested array

### LINKED ISSUE #4618: OpenAPI generator - array type for property in schema not correctly generated
**Describe the bug**

There is a small error in the generated types affecting "nested array objects". E.g. when generating the type for an object of this shape:

```ts
{
  content: [{ name: "foo" }]
}
```

**To Reproduce**

1. Use this OpenAPI schema and save it to `minimal-example-api.yaml`:

```yaml
openapi: "3.0.3"
info:
  description: minimal example
  version: "0.0.1"
  title: "sample api"
paths:
  /my-path:
    get:
      summary: do something crazy
      operationId: myOperation
      responses:
        "200":
          description: "yay it worked"
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/MySchema"
components:
  schemas:
    MySchema:
      type: object
      required:
        - content
      properties:
        content:
          type: array
          items:
            type: object
            required:
              - name
            properties:
              name:
                type: string
```

2. Generate the OpenAPI client with

```bash
npm init -y && npm i @sap-cloud-sdk/openapi-generator@3.13.0
npx openapi-generator --input minimal-example-api.yaml --outputDir gen --overwrite
```

3. check the generated schema for `#/components/schemas/MySchema` which looks like this:

```ts
// ...
export type MySchema = {
  content: {
    name: string;
  } & Record<string, any>[];
} & Record<string, any>;
```

5. The type for `.content` is incorrect as `[]`/array only applies to `Record<string, any>` but should actually wrap the whole object type for the `content` property aka `{ name: string } & Record<string, any>`. So the full type should look like this:

```ts
// ...
export type MySchema = {
  // note the surrounding "( ... )" and that the [] is after the closing ")"
  content:  ( {
    name: string;
  } & Record<string, any> ) [];
} & Record<string, any>;

```

**Expected behavior**

The `content` should be properly typed as an array.

**Used Versions:**

- node version 20.11.1
- npm version 10.2.4
- SAP Cloud SDK version 3.13.0

**Impact / Priority**

The generated types are incorrect making it necessary to use `as any` or `as unknown` type casts defeating the purpose of this feature.

Affected development phase: e.g. Development

Impact: Impaired

shibeshduw:
Fixes bug in OpenAPI generator where array type for property in schema is incorrectly generated, specifically addressing nested array objects.

Closes #4618


.changeset/heavy-terms-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-cloud-sdk/openapi-generator": minor
+---
+
+Fix nested array type in generated OpenAPI schemas such that array wraps the whole object.
```
deekshas8, line 5:
```suggestion
[Fixed Issue] Fix nested array type in generated OpenAPI schemas such that array wraps the whole object.
```

packages/openapi-generator/src/file-serializer/schema.ts
```diff
@@ -29,7 +29,7 @@ export function serializeSchema(schema: OpenApiSchema): string {
   }
   if (isArraySchema(schema)) {
     const type = serializeSchema(schema.items);
-    return schema.uniqueItems ? `Set<${type}>` : `${type}[]`;
+    return schema.uniqueItems ? `Set<${type}>` : `(${type})[]`;
```
deekshas8, line 32: [req] I would prefer that we differentiate between a simple array type (e.g. string[]) or array of objects and based on that add the `(..)`

deekshas8: And also add tests that cover both cases

packages/openapi-generator/src/file-serializer/schema.spec.ts
```diff
@@ -108,6 +108,15 @@ describe('serializeSchema for object schemas', () => {
 });
 
 describe('serializeSchema for array schemas', () => {
+  it('serializes array schema', () => {
```
deekshas8, line 111: [req] Can you also add a test here for array of object type

packages/openapi-generator/src/file-serializer/schema.ts
deekshas8: If the difference is only in `items.properties`, then I think we can add the condition in the array check itself, insetad of another if

packages/openapi-generator/src/file-serializer/schema.spec.ts
shibeshduw: It is there just below this one.

packages/openapi-generator/src/file-serializer/schema.ts
shibeshduw: Yes I thought of doing this yesterday but i wasn't sure what the convention was. I have updated it as requested.

packages/openapi-generator/src/file-serializer/schema.spec.ts
```diff
@@ -108,13 +108,35 @@ describe('serializeSchema for object schemas', () => {
...
+        items: { type: 'string' }
+      })
+    ).toEqual('string[]');
+  });
+
   it('serializes array schema with nested array', () => {
     expect(
       serializeSchema({
         uniqueItems: false,
-        items: { items: { type: 'string' } }
```
deekshas8, line 115: [req] Please don't remove this test as it's still a valid case

packages/openapi-generator/src/file-serializer/schema.ts
```diff
@@ -27,10 +27,15 @@ export function serializeSchema(schema: OpenApiSchema): string {
   if (isReferenceObject(schema)) {
     return schema.schemaName;
   }
+
   if (isArraySchema(schema)) {
     const type = serializeSchema(schema.items);
+    if ('properties' in schema.items) {
```
deekshas8, line 33: [req] Can be made shorter as
```
return schema.uniqueItems ? `Set<${type}>` : ('properties' in schema.items ? `(${type})[]` : `${type}[]`)
```

packages/openapi-generator/src/file-serializer/schema.spec.ts
deekshas8: Also, I meant to add a test with an example along the lines of what the user has, i.e., an object with array items:

```
 serializeSchema({
        properties: [
          {
            name: 'objectArrayProp',
            required: true,
            schema: {
              uniqueItems: false,
              items: {
                properties: [
                  {
                    name: 'name',
                    required: true,
                    schema: { type: 'string' },
                    schemaProperties: {}
                  }
                ],
                additionalProperties: { type: 'any' }
              }
             },
            schemaProperties: {}
          }
        ],
        additionalProperties: { type: 'any' }
      })
    )
```
This test can go in the object schemas section

shibeshduw: I see. I didn't think we needed to match the exact case as the user ticket since we already have a test case for serializing object schema with properties. I have added it as requested.

deekshas8 (approved):
LGTM

Issue #4572, opened 2024-03-15
[Due: Mar] openapi-generator fails to generate client


lovis-ff:
**Describe the bug**
I want to create a open-api client using an open api schema from SAP Business Accelerator Hub, specifically `https://api.sap.com/api/OP_API_SALES_QUOTATION_SRV_0001/overview`. I am getting `ErrorWithCause: Could not write file "sales-quotation-header-api.ts".`. 
Digging a b it into the code I find that there is a syntax error in the created files
```
SyntaxError: ',' expected. (55:25)
  53 |   '$expand'?: Set<'*' | 'to_Item' | 'to_Partner' | 'to_PrecedingProcFlowDoc' | 'to_PricingElement' | 'to_RelatedObject' | 'to_SubsequentProcFlowDoc' | 'to_Text'>}) => new OpenApiRequestBuilder<APISALESQUOTATIONSRVASalesQuotationType>(
  54 |     'get',
> 55 |     '/A_SalesQuotation('{salesQuotation}')',
     |                         ^
  56 |     {
  57 |           pathParameters: { salesQuotation },
```
where it seems as if template literals are not used correctly, it should probably be
```
> 55 |     `/A_SalesQuotation('${salesQuotation}')`,
```

The generation does not work for `.json` or `.yaml`. The generation of the odata client via `.edmx` file works but is not sufficient for the use case.

**To Reproduce**

1. Download open api json spec from`https://api.sap.com/api/OP_API_SALES_QUOTATION_SRV_0001/overview`
2. Execute `npx openapi-generator --input <input-directory> --outputDir <output-directory> --clearOutputDir --skipValidation`. Note, I do have to add `--skipValidation` because the generation would otherwise not work because of some naming issues of the endpoints.
3. See error

**Expected behavior**
Successful generation of the open-api client.

**Used Versions:**

- node version: v18.16.0
- npm version: done with pnpm version `7.33.5`
- SAP Cloud SDK version you used as dependency: `3.12.1`

**Code Examples**
If applicable, add code snippets as examples to help explain your problem. Please remove sensitive information.

**Log file**
Original error
```
[2024-03-15T10:23:25.021Z] ERROR    (cli): ErrorWithCause: Generation of services failed.
[...]

PR #4634, opened 2024-04-11
fix: Fix quote issue in path pattern

deekshas8:
Closes [SAP/cloud-sdk-backlog#4572](https://github.com/SAP/cloud-sdk-js/issues/4572)

Fixes error in generating open api client. Api having path format `/path('{param}')/{id}` result in syntax error when client is generated. Generation fails with: `ErrorWithCause: Could not write file...`

### DIFF
--- a/packages/openapi-generator/src/file-serializer/operation.ts
+++ b/packages/openapi-generator/src/file-serializer/operation.ts
@@ -15,7 +15,7 @@ import { serializeSchema } from './schema';
 export function serializeOperation(operation: OpenApiOperation): string {
   const requestBuilderParams = [
     `'${operation.method}'`,
-    `'${operation.pathPattern}'`
+    `"${operation.pathPattern}"`
   ];
 
   const bodyAndQueryParams = serializeParamsForRequestBuilder(operation);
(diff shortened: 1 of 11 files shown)


packages/openapi-generator/src/file-serializer/operation.ts
```diff
@@ -15,7 +15,7 @@ import { serializeSchema } from './schema';
 export function serializeOperation(operation: OpenApiOperation): string {
   const requestBuilderParams = [
     `'${operation.method}'`,
-    `'${operation.pathPattern}'`
+    `"${operation.pathPattern}"`
```
mr-flannery, line 18: Let me check if I understood this correctly:
- there exist APIs that expect params to be passed like this: `example.com/api/entities('5')/something`
- because we used single quotes for the string, the generated code looked like this `'test('{id}')'`, which is syntactically incorrect?

deekshas8: Yes, correct

mr-flannery: The ideas some API providers have... 😒

deekshas8:
> lgtm, as far as I can tell. Have you manually run this on the API the author of the issue found the problem on?

Yes, it generates successfully now.

PR #2303, opened 2022-03-28
Support custom request config for method in OData and OpenAPI

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#71.

Remove `method` from `defaultDisallowedKeys` in `http-request` to support custom http request method. 
Overwrite http `method` in `odata-common` and `openapi` if it is defined in custom http configuration.

The usage should look like

```typescript
const { myEntityApi } = myEntityService();
myEntityApi
  .requestBuilder()
  .getAll()
  .addCustomRequestConfiguration({ method: 'YOUR-METHOD' });
```

By default, http method is set to `get` by `getAll()`. Now, it will be overwritten by `YOUR-METHOD`.

TODO:
- [x] Update change log

PR #4647, opened 2024-04-18
fix: Expose type for custom request parameter

shibeshduw:
Introduces and exposes a type for parameter accepted by `addCustomRequestConfiguration` with the format below as per the [SDK documentation](https://sap.github.io/cloud-sdk/docs/js/features/openapi/execute-request#setting-a-custom-request-configuration):

```
type CustomRequestConfig = Omit<
  HttpRequestConfig,
  "url" | "baseURL" | "data" | "headers" | "params"
>;
```

Using HttpRequestConfig instead of axiosRequestConfig suggested by the user ensures that we are not tightly coupled with Axios.

Closes SAP/cloud-sdk-backlog#1187

### DIFF
--- a/packages/openapi/src/openapi-request-builder.spec.ts
+++ b/packages/openapi/src/openapi-request-builder.spec.ts
@@ -323,10 +323,10 @@ describe('openapi-request-builder', () => {
     it('should overwrite default request config with filtered custom request config', async () => {
       const requestBuilder = new OpenApiRequestBuilder('get', '/test');
       requestBuilder.addCustomRequestConfiguration({
-        timeout: 100
+        method: 'merge'
       });
       const requestConfig = await requestBuilder['requestConfig']();
-      expect(requestConfig['timeout']).toBe(100);
+      expect(requestConfig['method']).toBe('merge');
     });
     it('test the exposed type from SDK', async () => {
       const reusedConfig: CustomRequestConfig = { timeout: 2000 };
--- a/packages/openapi/src/openapi-request-builder.spec.ts
+++ b/packages/openapi/src/openapi-request-builder.spec.ts
@@ -18,6 +18,7 @@ import {
   mockFetchDestinationCalls
 } from '../../../test-resources/test/test-util';
 import { OpenApiRequestBuilder } from './openapi-request-builder';
+import { CustomRequestConfig } from './types';
 
 const destination: HttpDestination = {
   url: 'http://example.com'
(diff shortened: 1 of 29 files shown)


shibeshduw:
Also found an inconsistency in our SDK [documentation](https://sap.github.io/cloud-sdk/docs/js/features/openapi/execute-request#setting-a-custom-request-configuration) in the topic `Setting a Custom Request Configuration`, it is mentioned that the SDK doesn't allow `method` option to be overwritten, when in fact it is allowed (this is also present in a [test case](https://github.com/SAP/cloud-sdk-js/blob/aa4e96ba1dbd4b78446e10a8c010fd5cba744ae1/packages/openapi/src/openapi-request-builder.spec.ts#L321)). @jjtang1985

tomfrenken:
Looking at the git history, support for the method option was specifically added [here](https://github.com/SAP/cloud-sdk-js/pull/2303) therefore it would also make sense for you to open a docs PR.

packages/openapi/src/types.ts
```diff
@@ -47,3 +48,12 @@ export type UnPromisify<T> = T extends Promise<infer U> ? U : T;
  * @internal
  */
 export type ConstructorType<T> = new (...args: any[]) => T;
+
+/**
+ * The type for parameter in Custom Request Configuration.
+ * @internal
+ */
+export type CustomRequestConfig = Omit<
```
tomfrenken, line 56: Looks like the correct type, however, youseem to have not added this type to the odata-request-builder, I would assume you will also have to use it there

packages/openapi/src/openapi-request-builder.spec.ts
```diff
@@ -327,5 +328,13 @@ describe('openapi-request-builder', () => {
       const requestConfig = await requestBuilder['requestConfig']();
       expect(requestConfig['method']).toBe('merge');
     });
+    it('test the exposed type from SDK', async () => {
```
tomfrenken, line 331: What exactly is this supposed to test?
If you were trying to test that you can not use the omitted options in the narrowed type of `CustomRequestConfig`, I do not think you can properly test this, as the omitted options are still usable, due to `HttpRequstConfigBase`'s type of `key: string]: any;`

tomfrenken (changes requested):

packages/openapi/src/openapi-request-builder.spec.ts
shibeshduw: Yeah I understand that the omitted options are still usable; the test was not intended for validation, but rather just to demonstrate the usability of the exposed type as users can now define custom configs using `CustomRequestConfig` type and pass them into `addCustomRequestConfiguration` method. Before this, only string values could be defined in the parameters for `addCustomRequestConfiguration`. I thought it would be worth adding this.

packages/openapi/src/types.ts
shibeshduw: I think the issue is only regarding openapi-generator and not odata generator. Unless I didn't get the full scope of this? I have imported it to `openapi-request-builder` and used in in this class.

shibeshduw: Okay I see what you mean now after going through the full scope. I will relocate this type in http-client-types and add it in `odata-request-builder` as well.

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -24,14 +24,15 @@ import {
   OriginOptions,
   HttpMiddleware
 } from '@sap-cloud-sdk/http-client/internal';
+import { CustomRequestConfig } from '@sap-cloud-sdk/http-client/src/http-client-types';
```
marikaner, line 27: [req] This should be imported from the toplevel (and if not possible, also exported).
Otherwise this will fail for users, because they don't have a `src` directory.

```suggestion
import { CustomRequestConfig } from '@sap-cloud-sdk/http-client';
```

tomfrenken: For it to be imported from the top level, you will also need to include the type in the index.ts 👀

packages/openapi/src/openapi-request-builder.spec.ts
tomfrenken: I see, for that I believe proper documentation should be sufficient, we can remove the test. 👍

tomfrenken (approved):
LGTM

PR #4663, opened 2024-05-02
fix: Odata generation doesn't fail if basePath is not set

shibeshduw:
If basePath cannot be determined, instead of odata generation failing altogether, basePath is set to `/`, and a warning message is logged. The previous behavior of setting `--skipValidation` will now be the default behavior if basePath is not set. 

Closes SAP/cloud-sdk-backlog#1105


.changeset/shy-days-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-cloud-sdk/generator": minor
+---
+
+[Improvement] If basePath cannot be determined, instead of generation failing altogether, basePath is set to `/`, and a warning message is logged.
```
marikaner, line 5:
```suggestion
[Improvement] If `basePath` cannot be determined, it is set to `/` and generation no longer fails.
```

PR #4664, opened 2024-05-02
fix: Filter for operations which do not have OperationImport 

shibeshduw:
Due to a wrong implementation of filtering operations without OperationImport, unnecessary warnings were being logged during odata generation. Instead of checking all `EdmxOperation[]`, only those with `isBound` flag were being checked. This is now rectified and aforementioned warnings are no longer generated.

Closes SAP/cloud-sdk-backlog#1199

### DIFF
--- a/.changeset/light-insects-bow.md
+++ b/.changeset/light-insects-bow.md
@@ -1,0 +1,5 @@
+---
+"@sap-cloud-sdk/generator": minor
+---
+
+[Fix] Filter for operations without OperationImport. Unnecessay warning logs are no longer generated.
--- a/packages/generator/src/edmx-to-vdm/v4/operation.ts
+++ b/packages/generator/src/edmx-to-vdm/v4/operation.ts
@@ -124,13 +124,18 @@ export function filterAndTransformOperations(
     operation => (operation.IsBound.toLowerCase() === 'true') === isBound
   );
 
-  const [withoutOperation, withOperation] = splitMissingOperation(
+  const [withoutOperation] = splitMissingOperation(
+    operationImports,
+    operations
+  );
+
+  const [, boundWithOperation] = splitMissingOperation(
     operationImports,
     filteredByBoundOperations
   );
 
   const [validOperations, operationsWithoutRequiredParameters] =
-    splitMissingParameter(withOperation);
+    splitMissingParameter(boundWithOperation);
 
   if (operationsWithoutRequiredParameters.length) {
     logger.warn(


.changeset/light-insects-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-cloud-sdk/generator": minor
+---
+
+[Fix] Filter for operations without OperationImport. Unnecessay warning logs are no longer generated.
```
marikaner, line 5: [req] It is the other way around, I think. The operation imports don't have an operation. Please note, that changelogs should be written in imperative and implementation details are not needed.

```suggestion
[Fix] Fix incorrect warning logs for operation imports that do have a matching operation in the service specification.
```

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -124,13 +124,18 @@ export function filterAndTransformOperations(
     operation => (operation.IsBound.toLowerCase() === 'true') === isBound
   );
 
-  const [withoutOperation, withOperation] = splitMissingOperation(
+  const [withoutOperation] = splitMissingOperation(
+    operationImports,
+    operations
+  );
+
+  const [, boundWithOperation] = splitMissingOperation(
```
marikaner, line 132: [req] I think it would be better to filter only the operations, that have an operation import. Otherwise you will get two logs for the same thing and this might be unnecessary.
[nth] You could consider switching the logic of `splitMissingOperation()` to return the operations with operation first, then line 132 would look slightly cleaner (and line 127 would be different with my previous comment anyways).

marikaner (changes requested):
You could also consider to extract the validation of the bound operation parameters as well as the validation of import and operation into separate functions for better readability.
There are multiple ways you could do this, let's talk if you need inspriration ;)

marikaner:
Oh and please add tests for this.

packages/generator/src/edmx-to-vdm/v4/operation.ts
shibeshduw: I did switch the logic of `splitMissingOperation()` as you said.

> I think it would be better to filter only the operations, that have an operation import.

Do you mean adding something like this to filter out operation imports that have an operation separately?
```
const withoutOperation = operationImports.filter(
    operation => !findOperationByImportName(operations, operation.operationName)
);
```

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -124,13 +124,31 @@ export function filterAndTransformOperations(
     operation => (operation.IsBound.toLowerCase() === 'true') === isBound
```
marikaner, line 124: [req] This name is confusing. It sounds like those are bound operations, but it could be either bound or unbound. How about `filteredOperations` or `relevantOperations`?

BTW, last time we talked about this we assumed this was only about bound operations, therefore we might have incorrect assumptions in the following.

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -32,11 +32,11 @@ const extractResponse = (response: string) => `${response}.value`;
 function splitMissingOperation(
```
marikaner, line 32: [req] This name is confusing, it sounds like this is about one operation that is split. Suggestion: `splitOperationsByMissingImports()` or similar.

marikaner (commented):
I realized that last time we discussed this, we misunderstood the `filteredByBoundOperations` variable. That is why I left you 2 minor comments on how to potentially rename things to make it easier to understand.

Also, I now don't see anymore why this is failing, therefore I would suggest that you write a test on the old state (maybe including renaming ;) ) that fails first to locate the problem. We will need a test anyways.

packages/generator/src/edmx-to-vdm/v4/operation.ts
shibeshduw: Okay this makes sense. It also makes sense for `splitOperationsByMissingImports` to run before this because we are only using `relevantOperations` to filter out all bound operations without a parameter and all operations without an operationImport are not required for this. I believe it would make sense for `relevantOperations` to only filter those operations with an operationImport. So something like:

```javascript
const [withoutOperation, withOperation] = splitOperationsByMissingImports(
  operationImports,
  operations
);

const relevantOperations = withOperation.filter(
  operation => (operation.IsBound) === isBound
);
```

shibeshduw: Okay there was a test case that checks for operations referenced by operation import. The probem was that the messageContext in the `createLogger` method was wrong. It should be 'operation' instead of 'action-import'. Making this change results in the test failing on the old state. 

The way to fix it is to make sure that we are only using `relevantOperations` to filter out all bound operations without a parameter. Operations without an operationImport are not required for this so it should not be related with `splitOperationsByMissingImports`. I believe it would make sense for `relevantOperations` to only filter those operations with an operationImport. I have implemented it this way.

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -120,17 +120,31 @@ export function filterAndTransformOperations(
...
 
-  const [withoutOperation, withOperation] = splitMissingOperation(
-    operationImports,
-    filteredByBoundOperations
+  const relevantOperations = withOperations.filter(
+    operation => operation.IsBound === isBound
   );
 
+  const validBoundOperations =
+    validateBoundOperationParameters(relevantOperations);
```
marikaner, line 133: [req] The function is called `validateBoundOperationParameters()`. However it is also called for unbound operations. It works because the underlying `split...()` function ignores unbound operations, but I believe it should not even get to this and we should filter it out here already. That would be way more obvious. (see comment below)

```suggestion
  if (isBound) {
    const validBoundOperations = validateBoundOperationParameters(relevantOperations);
  }
```

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -120,17 +120,31 @@ export function filterAndTransformOperations(
...
+
+  validateOperationImports(
+    withoutOperations,
+    withoutOperations[0]?.operationType
+  );
```
marikaner, line 138: [nth] If you move this up this whole section gets a little easier to implement. Also check the comments above and below for details.

```
  validateOperationImports(
    withoutOperations,
    withoutOperations[0]?.operationType
  );

  if (isBound) {
    return validateBoundOperationParameters(relevantOperations);
  }

  return relevantOperations;
```

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -120,17 +120,31 @@ export function filterAndTransformOperations(
...
+
+  return validBoundOperations;
```
marikaner, line 140: [req] The naming is very confusing here, because in the case of `isBound === false` those are not bound operations that we return. (see comment above)

```suggestion
  return validBoundOperations;
```

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -120,17 +120,31 @@ export function filterAndTransformOperations(
...
+}
+
+function validateBoundOperationParameters(
+  relevantOperations: EdmxJoinedOperation[]
+): EdmxJoinedOperation[] {
   const [validOperations, operationsWithoutRequiredParameters] =
-    splitMissingParameter(withOperation);
+    splitMissingParameter(relevantOperations);
```
marikaner, line 147: nitpick 1: The function is called `splitMissingParameter()` which sounds like a parameter is split. I know you did not introduce it, but it would be great if we could rename it so something speaking, like `splitBoundOperationsByMissingParameters()` or maybe you can think of something shorter.
nitpick 2: The `splitMissingParameter()` function basically ignores unbound operations. I think it should not be called for unbound operations, then the implementation would get simpler.

marikaner (changes requested):
mostly looks good to me, but I left a few comments mostly around naming because I think it is confusing an will make it hard to find issues (like it already did for us previously).

packages/generator/src/edmx-to-vdm/v4/operation.ts
```diff
@@ -75,7 +75,7 @@ function buildBoundOperation(
   };
 }
 
-function splitMissingParameter(
+function splitBoundOperationsByMissingParameters(
```
marikaner, line 78: [req] As we are not looking at unbound operations anymore, the `if (!curr.IsBound)` condition can be removed.

marikaner (approved):
Looks great. I only pointed out one minor issue, as discussed.

PR #4687, opened 2024-05-16
feat: Add support noImplicitOverride config option

shibeshduw:
Adds support for `"noImplicitOverride": true` in `tsconfig.json`. Generated clients will now comply with this config option. 

Closes SAP/cloud-sdk-backlog#1200.


packages/generator/src/entity/class.ts
```diff
@@ -78,6 +79,7 @@ function defaultBasePath(
     kind: StructureKind.Property,
     name: prependPrefix('defaultBasePath'),
     isStatic: true,
+    hasOverrideKeyword: true,
```
tomfrenken, line 82: Seems to work well for entityName & basePath, what is the reasoning for not adding it to keys as well? Is the assumption that they are never overwritten?

shibeshduw: I did not add it to keys because it is not defined in the base class `Entity`, but rather in a `Constructable` generic interface which uses a subclass of `EntityBase`. So it cannot be overwritten by principle.

tomfrenken (approved):
LGTM

PR #4708, opened 2024-05-24
feat: Parse OpenApi spec with in header parameters

shibeshduw:
With this change, a client user can pass all the "in header" parameters like resource groups, directly without using the custom header "workaround". OpenApi clients are generated with `headerParameters` according to the spec provided.

```Javascript
const responseData: ConfigurationList = await ConfigurationApi.configurationQuery({'AI-Resource-Group': 'default'})
    .execute(destination);
```

Closes [SAP/gen-ai-hub-sdk-js#26](https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/26)


packages/openapi-generator/src/parser/operation.ts
```diff
@@ -30,9 +30,10 @@ export function parseOperation(
     refs
   );
 
-  const [pathParams, queryParams] = partition(
+  const [pathParams, headerParams, queryParams] = partitionThree(
     relevantParameters,
-    parameter => parameter.in === 'path'
+    parameter => parameter.in === 'path',
+    parameter => parameter.in === 'header'
   );
```
marikaner, line 37: [pp] I am not a fan of this implementation. I think I created the old one, but I think this would be easier to read:

```suggestion
const pathParams = relevantParameters.filter(parameter => parameter.in === 'path');
const headerParams = relevantParameters.filter(parameter => parameter.in === 'header');
const queryParams = relevantParameters.filter(parameter => parameter.in === 'query');
```

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -160,10 +160,11 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
   private getHeaders(): OriginOptions {
-    if (Object.keys(this.customHeaders).length) {
-      return { custom: this.customHeaders, requestConfig: {} };
-    }
-    return { requestConfig: {} };
+    // if (Object.keys(this.customHeaders).length) {
+    //   return { custom: this.customHeaders, requestConfig: {} };
+    // }
+    // return { requestConfig: {} };
+    return { requestConfig: this.parameters?.headerParameters || {} };
```
marikaner, line 167: [req]
1. Please remove the old comment
2. I think this is incorrect because you want to send along custom headers if there are any. Please make sure this works for both cases.

```suggestion
const options = { requestConfig: this.parameters?.headerParameters || {} };
if (Object.keys(this.customHeaders).length) {
  return { custom: this.customHeaders, ...options };
}
return options;
```

packages/util/src/array.ts
```diff
@@ -137,6 +137,35 @@ export function partition<T>(
   );
 }
 
+/**
```
marikaner, line 140: [pp] With my other comments, this should not be needed.

marikaner (changes requested):

.changeset/happy-mayflies-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-cloud-sdk/openapi-generator": minor
+---
+
+OpenApi clients are generated with `in: header` parameters defined in the spec.
```
marikaner, line 5: [req] Use imperative: aka. Generate OpenAPI clients... (also OpenAPI not OpenApi)

packages/openapi-generator/src/file-serializer/api-file.spec.ts
```diff
@@ -122,14 +133,16 @@ describe('apiFile', () => {
          * Create a request builder for execution of get requests to the 'test/{id}' endpoint.
          * @param id - Path parameter.
          * @param queryParameters - Object containing the following keys: queryParam.
+         * @param headerParameters - Object containing the following keys: headerParam.
          * @returns The request builder, use the \`execute()\` method to trigger the request.
          */
-        getFn: (id: string, queryParameters: {'queryParam': QueryParameterType}) => new OpenApiRequestBuilder<string>(
+        getFn: (id: string, headerParameters?: {'headerParam'?: string}, queryParameters: {'queryParam': QueryParameterType}) => new OpenApiRequestBuilder<string>(
```
marikaner, line 139: [req] This seems problematic to me. The optional header parameters cannot come before the required query paramters. This is not allowed. I wonder if this is how our API is intended? I would also assume headers come last, however, it is possible that headers will be required and query parameters are optional. I somehow thought this was all in one object.

packages/openapi-generator/src/file-serializer/operation.ts
```diff
@@ -63,14 +70,14 @@ function serializeRequestBodyParamForSignature(
   }
 }
 
-function serializeQueryParamsForSignature(
-  operation: OpenApiOperation
+function serializeOtherParamsForSignature(
```
marikaner, line 73: [req] Why "other"?

```suggestion
function serializeParamsForSignature(
```

marikaner (changes requested):
Thanks for adjusting, looks very good. One thing I noticed though is that we can break the API if query parameters are mandatory and header parameters are optional. I would like to also see this change reflected in the test openAPI spec that we have.

packages/openapi-generator/src/file-serializer/api-file.spec.ts
shibeshduw: Should be good now. Also added a case where the optional modifier will be added to queryParameters object if there are no headerParameters defined as this was not covered.

.changeset/happy-mayflies-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sap-cloud-sdk/openapi-generator": minor
+---
+
+Generate OpenAPI clients with `in: header` parameters defined in the spec.
```
deekshas8, line 5: [req] Add a message type here.

packages/openapi-generator/src/file-serializer/operation.spec.ts
```diff
@@ -48,14 +57,67 @@ describe('serializeOperation', () => {
...
        */
-      getFn: (id: string, subId: string, queryParameters?: {'limit'?: number}) => new OpenApiRequestBuilder<string>(
+      getFn: (id: string, subId: string, queryParameters: {'limit'?: number}, headerParameters?: {'resource-group'?: string}) => new OpenApiRequestBuilder<string>(
         'get',
         "test/{id}/{subId}",
         {
               pathParameters: { id, subId },
-              queryParameters
+              queryParameters,
+              headerParameters
```
deekshas8, line 69: [req] the `headerParameters` need to be part of `OpenApiRequestParameters`

deekshas8:
Please also add e2e tests for the generation. You can find the test spec [here](https://github.com/SAP/cloud-sdk-js/blob/main/test-resources/openapi-service-specs/test-service.json) and tests for it [here](https://github.com/SAP/cloud-sdk-js/blob/main/test-packages/test-services-openapi/generate-openapi-services.ts)

packages/openapi-generator/src/file-serializer/operation.ts
```diff
@@ -63,23 +72,32 @@ function serializeRequestBodyParamForSignature(
...
           `'${param.name}'${param.required ? '' : '?'}: ${serializeSchema(
             param.schema
           )}`
       )
-      .join(',\n');
+      .join(', ');
+
+    const allOptional = parameters.every(param => !param.required);
+    const optionalModifier =
+      (paramType === 'headerParameters' && allOptional) ||
```
deekshas8, line 93: [req] I would rewrite it like:

```
const optionalModifier =
      allOptional && (
        (paramType === 'headerParameters') ||
        (paramType === 'queryParameters' && isHeaderParamsAllOptional)
      ) ? '?' : '';
```

packages/openapi/src/openapi-request-builder.spec.ts
```diff
@@ -206,33 +234,6 @@ describe('openapi-request-builder', () => {
     expect(response.data).toBe('iss token used on the way');
   });
 
-  it('addCustomHeaders', async () => {
```
deekshas8, line 209: [q] Isn't this still a valid test case?

deekshas8:
Can you maybe also check if there needs to be any update to the docs portal?

packages/openapi/src/openapi-request-builder.spec.ts
shibeshduw: Would the `addCustomHeaders` function still be required after this change? All mandatory headers are set by default as usual and users can add additional headers as normal parameters now.

deekshas8: I believe it is.
1. Internally, we still maintain normal request headers (api spec) and custom headers separately.
2. With these changes, the user can still only add request headers in the function that are part of the spec. I don't think the types provide extensibility. Any headers that are still not in the spec, need to be added via `addCustomHeaders` function.

@marikaner

marikaner (approved):
I made a few minor improvements, but if the tests pass this looks great :)

packages/openapi/src/openapi-request-builder.spec.ts
marikaner: yes it is 😊

PR #4714, opened 2024-05-27
chore: add noImplicitOverride keyword

marikaner:
Add the `noImplicitOverride` keyword. The transpilation errors we have seen before were a false alarm, no other properties were missed. However:

- pro: When this property is there, we cannot forget to add the override keyword to properties that are overriding. Will happen rarely to never, though.
- con: If users are changing something in the generated code to also override something, they are now forced to add the override keyword. We don't know how many people do this, but it is most likely an edge case.

Not sure it is worth adding this then.


deekshas8:
Closing this PR, due to reasons stated in the description. The client will compile if the user themselves decide to use this setting.

PR #4756, opened 2024-06-17
feat: Add API for service binding to destination transformation

KavithaSiva:
Closes : https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/40

Adding a new convenience function `transformServiceBindingToDestination` to easily build `OAuth2ClientCredentials` destinations for a set of service bindings.

The PR changes the visibility of `getServiceBindings`,`getServiceBinding`,`resolveServiceBinding`,`getServiceBindingByInstanceName` to public instead of internal functions.


packages/connectivity/src/scp-cf/environment-accessor/service-bindings.ts
```diff
@@ -9,19 +9,19 @@ const logger = createLogger({
 
 /**
  * Services getter for a given service.
+ * Internally uses, xsenv library to read configurations from VCAP_SERVICES.
```
KavithaSiva, line 12: [q] I chose to make all functions in this package public as I think they all could be used and don't see any behaviour change being introduced here until `xsenv` chnages something. WDY think?

Right now we only need `getServiceBinding` in the AI SDK.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -21,10 +27,48 @@ export const serviceToDestinationTransformers: Record<
...
+ * Throws an error if the provided service binding is not supported.
+ * @param serviceBinding - The service binding to transform.
+ * @param options - Options used during fetching destinations;includes a JWT payload {@link JwtPayload} and options to influence caching behavior see {@link CachingOptions}.
+ * @returns A promise returning the transformed OAuth2ClientCredentials destination on success.
+ */
+export async function transformServiceBindingToDestination(
+  serviceBinding: Service,
+  options?: ServiceBindingTransformOptions
+): Promise<Destination> {
+  if (serviceToDestinationTransformers[serviceBinding.label]) {
```
KavithaSiva, line 55: [q] For now I thought it is okay to return a Basic Auth destination only for 's4-hana-cloud' as it is documented well.

Should I throw an unsupported error for 's4-hana-cloud' instead?
WDY prefer?

.changeset/beige-bananas-divide.md
```diff
@@ -0,0 +1,15 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Add convenience function `transformServiceBindingToDestination` to create OAuth2ClientCredentials destination from service bindings.
```
marikaner, line 5: [req] According to our styleguide functions should have parantheses in documentation aka. `transformServiceBindingToDestination()`

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -21,10 +27,48 @@ export const serviceToDestinationTransformers: Record<
...
+ * - s4-hana-cloud (falls back to basic authentication for this service type)
+ * - saas-registry
+ * - workflow
+ * - service-manager
+ * - xsuaa
+ * - aicore.
+ *
+ * Throws an error if the provided service binding is not supported.
+ * @param serviceBinding - The service binding to transform.
+ * @param options - Options used during fetching destinations;includes a JWT payload {@link JwtPayload} and options to influence caching behavior see {@link CachingOptions}.
```
marikaner, line 48: [pp] I think we can be less verbose here. The type `ServiceBindingTransformOptions` is linked automatically and it includes the subtypes.
```suggestion
 * @param options - Options used for fetching the destination.
```

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -21,10 +27,48 @@ export const serviceToDestinationTransformers: Record<
   workflow: workflowBindingToDestination,
   'service-manager': serviceManagerBindingToDestination,
   xsuaa: xsuaaToDestination,
-  aicore: aicoreToDestiation
+  aicore: aicoreToDestination
 };
 
-async function aicoreToDestiation(
+/**
+ * Convenience function to create an OAuth2ClientCredentials destination from the provided service binding.
```
marikaner, line 34: [q] If one of them is basic authentication I wouldn't mention `OAuth2ClientCredentials` at all here.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
marikaner: [q] Why do we even need a warning here? The binding wants basic, so it is basic. To me, this is expected behavior. Therefore I would suggest to remove the warning and also adjust the description as pointed out above.

packages/connectivity/src/scp-cf/environment-accessor/service-bindings.ts
marikaner, line 12:
```suggestion
 * Internally uses xsenv library to read configurations from VCAP_SERVICES.
```

(Same below)

marikaner: I wouldn't expose anything that we don't need. We used to do that a lot and block ourselves with this. Even though it is unlikely that things will change, it also unnecessarily clutters the API, making it bigger and less accessible.

packages/connectivity/src/scp-cf/environment-accessor/service-bindings.ts
```diff
@@ -64,8 +63,11 @@ export function resolveServiceBinding(service: string | Service): Service {
 }
 
 /**
- * Filters services based on service instance name. Throws an error if no or multiple services exist.
- * @internal
+ * Filters services based on service instance name.
+ *
+ * Throws an error if no or multiple services exist.
+ * @param serviceInstanceName - A string represeting the service instance name.
```
marikaner, line 69:
```suggestion
 * @param serviceInstanceName - A string representing the service instance name.
```

marikaner (changes requested):
Generally LGTM, but I would prefer to just expose what we need and nothing more.

packages/connectivity/src/scp-cf/environment-accessor/service-bindings.ts
```diff
@@ -19,9 +19,9 @@ export function getServiceBindings(service: string): Service[] {
 
 /**
  * Returns the first found instance for the given service type.
+ * Internally uses, xsenv library to read configurations from VCAP_SERVICES.
```
marikaner, line 22:
```suggestion
 * Internally uses xsenv library to read configurations from VCAP_SERVICES.
```

marikaner (approved):

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
KavithaSiva: Added the auth flows in the description to make it more explicit.

PR #4774, opened 2024-06-21
chore: Improve client credentials cache and tenant retrieval

marikaner:
The client credentials cache now stores credentials per tenant id and client id instead of tenant aware authentication url and client id.
JWT retrieval checks for `zid` and if not available for `app_tid`.
Tenant retrieval for the provider checks the identity and destination service bindings before giving up.

Relates to SAP/cloud-sdk-backlog#1217 and SAP/cloud-sdk-backlog#1219.


packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -33,20 +36,27 @@ const ClientCredentialsTokenCache = (
 
 /** *
  * @internal
- * @param url - URL from where the token is fetched
+ * @param tenantId - URL from where the token is fetched
```
deekshas8, line 39: [req] the description needs an update

packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
```diff
@@ -66,11 +66,11 @@ function getJwtForCaching(options: RegisterDestinationOptions | undefined) {
   if (!jwt?.zid) {
     if (options?.jwt) {
       logger.error(
-        'Could neither determine tenant from JWT nor service binding to XSUAA, although a JWT was passed. Destination will be registered without tenant information.'
+        'Could neither determine tenant from JWT nor XSUAA, identity or destination service binding, although a JWT was passed. Destination is registered without tenant information.'
```
deekshas8: [pp] 
```suggestion
        'Could not determine tenant from the JWT, XSUAA, identity, or destination service binding, although a JWT was passed. Destination is registered without tenant information.'
```

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
marikaner, line 39:
```suggestion
 * @param tenantId - The ID of the tenant to cache the token for.
```
```suggestion
 * @param tenantId - URL from where the token is fetched
```

packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
marikaner: I adjusted it, but a bit differently.

deekshas8 (approved):

PR #4781, opened 2024-06-24
chore: do not retrieve another service token for the destination service

marikaner:
We used to retrieve service tokens and then retrieve them again using the already retrieved service token. I could not find any technical explanation why we are doing this, so I removed it for performance and better readability.


packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -256,13 +256,17 @@ export class DestinationFromServiceRetriever {
...
-          ? this.providerServiceToken.decoded
-          : this.subscriberToken?.serviceJwt?.decoded
-    });
-    return { authHeaderJwt: clientGrant, exchangeTenant };
+    const authHeaderJwt =
+      origin === 'provider'
+        ? this.providerServiceToken.encoded
+        : this.subscriberToken?.serviceJwt?.encoded;
+
+    if (!authHeaderJwt) {
```
deekshas8, line 264: [q] Just wondering if this check here is needed again?

deekshas8 (approved):
LGTM. Just one question.

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
marikaner: I guess not. 

provider case: token is there anyways, so **no**
subscriber case: technically token could not be there, but if it is not there, the origin cannot be "subscriber" either, so I guess **no**

marikaner: But actually, after I checked, this is a type problem and I prefer to solve it like this instead of using `!`.

PR #4799, opened 2024-06-28
feat: improve support for IAS tokens

marikaner:
This includes generalized handling of tokens so that XSUAA, IAS and other custom tokens.

I fixed some places to make sure the difference between subdomain (aka. tenant, aka. tenant name) and tenant ID (aka. zone ID, aka. app_tid) is hopefully more obvious and in some cases more correct.
Especially the subdomain needs to be handled differently between XSUAA and IAS.

Closes SAP/cloud-sdk-backlog#1219.
Closes SAP/cloud-sdk-backlog#1217.


packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -241,12 +240,10 @@ export class DestinationFromServiceRetriever {
...
     }
-    const subdomainSubscriber = getIssuerSubdomain(
-      this.subscriberToken?.userJwt?.decoded
-    );
-    const subdomainProvider = getIssuerSubdomain(
-      this.providerServiceToken?.decoded
-    );
+    const subdomainSubscriber =
+      getSubdomain(this.subscriberToken?.userJwt?.decoded) ||
+      getSubdomain(this.subscriberToken?.serviceJwt?.decoded);
```
marikaner, line 245:
```suggestion
      getSubdomain(this.subscriberToken?.serviceJwt?.decoded) ||
      getSubdomain(this.subscriberToken?.userJwt?.decoded);
```

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -241,12 +240,10 @@ export class DestinationFromServiceRetriever {
...
     }
-    const subdomainSubscriber = getIssuerSubdomain(
-      this.subscriberToken?.userJwt?.decoded
-    );
-    const subdomainProvider = getIssuerSubdomain(
-      this.providerServiceToken?.decoded
-    );
+    const subdomainSubscriber =
+      getSubdomain(this.subscriberToken?.serviceJwt?.decoded) ||
+      getSubdomain(this.subscriberToken?.userJwt?.decoded);
```
marikaner, line 245: Previously we only looked at the subdomain from the user JWT. However, if the user passed a separate `iss` property this should be the actual subdomain and it should take precedence (right?).

If the `iss` property was passed, the service JWT was fetched with that subdomain and should have the same subdomain as it is an XSUAA token. If it was not possible to fetch the service token with this subdomain, I assume it wouldn't work in this case either.

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -299,14 +296,15 @@ Possible alternatives for such technical user authentication are BasicAuthentica
...
 
-    // If subscriber token does not include service JWT (aka. originally passed JWT was not an XSUAA JWT) enforce the JWKS properties are there - destination service would do that as well. https://help.sap.com/docs/CP_CONNECTIVITY/cca91383641e40ffbe03bdc78f00f681/d81e1683bd434823abf3ceefc4ff157f.html
-    if (!this.subscriberToken.serviceJwt) {
+    const isXsuaaUserJwt = isXsuaaToken(this.subscriberToken.userJwt.decoded);
+    // If subscriber user token was not issued by XSUAA enforce the JWKS properties are there - destination service would do that as well. https://help.sap.com/docs/CP_CONNECTIVITY/cca91383641e40ffbe03bdc78f00f681/d81e1683bd434823abf3ceefc4ff157f.html
+    if (!isXsuaaUserJwt) {
       DestinationFromServiceRetriever.checkDestinationForCustomJwt(destination);
     }
 
     // Case 1 Destination in provider and JWT issued for provider account, but no custom JWT given -> no extra x-user-token header needed
```
marikaner, line 305:
```suggestion
    // Case 1: subscriber account is the provider accound, user JWT is from XSUAA
    // x-user-token header not needed
```

marikaner:
> Please take my comments with a grain of salt, since the code is so complicated and I don't understand much if any of it. Comments are only based on the diff, not much else 😅

No worries! Thanks for the review! 😊

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -299,14 +296,16 @@ Possible alternatives for such technical user authentication are BasicAuthentica
...
-    if (!this.subscriberToken.serviceJwt) {
+    const isXsuaaUserJwt = isXsuaaToken(this.subscriberToken.userJwt.decoded);
+    // If subscriber user token was not issued by XSUAA enforce the JWKS properties are there - destination service would do that as well. https://help.sap.com/docs/CP_CONNECTIVITY/cca91383641e40ffbe03bdc78f00f681/d81e1683bd434823abf3ceefc4ff157f.html
+    if (!isXsuaaUserJwt) {
       DestinationFromServiceRetriever.checkDestinationForCustomJwt(destination);
     }
 
-    // Case 1 Destination in provider and JWT issued for provider account, but no custom JWT given -> no extra x-user-token header needed
+    // Case 1: subscriber account is the provider accound, user JWT is from XSUAA
+    // x-user-token header not needed
```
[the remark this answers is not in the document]
marikaner: Just FYI: I did not change the logic, only improved the comment here.
I would also assume we just send the header in all the cases, however according to this documentation: https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/exchanging-user-jwts-via-oauth2usertokenexchange-destinations

In this exact case it doesn't happen and we have to do the exchange on our own and send it as authentication header. I remember we tried this once in the E2E tests and passing it in the header in fact didn't work.

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
```diff
@@ -72,34 +74,40 @@ async function retrieveUserToken(
...
+    try {
+      return getJwtPair(
+        await serviceToken('destination', {
+          ...options,
+          jwt
+        })
+      );
+    } catch (err) {
+      logger.warn(
+        'Failed to fetch subscriber service token for destination. This is only relevant if you are using subscriber destinations.'
```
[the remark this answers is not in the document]
marikaner: > currently this discards the error, maybe print it to debug to give users some options to get to the root cause, in case they need to? At least that's what we typically do in Java..

I think it is a nice idea to add the error. Debug would hide it even more. I could just add it to the warning. What do you think?


> Also, I'm wondering what the reason for this change is, considering that this threw before. What changed?

This is not obvious from the code at first glance, but previously this request was only executed (and thereby potentially throwing) when there was an `iss` property set or the given token was an XSUAA token. Now it would be executed for all tokens, if given. 
There might be cases where I don't care about subscriber destinations, therefore I wouldn't need a subscriber token and wouldn't care about it failing.
I considered rethrowing the error in case the `iss` property was set or the token was an XSUAA token instead. Do you think that would be better?

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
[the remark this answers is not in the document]
MatKuhr: I don't see any restrictions in the documentation why sending the header should not work for this auth type specifically. In fact, the Java SDK always sends the header, including for this auth type. So at least I am not aware of any reason, why this shouldn't work (assuming the OAuth service trusts and accepts the given user JWT).

If we have some test, where an XSUAA based token is not accepted in this manner, I'd like to look at it and potentially report that to the destination service.

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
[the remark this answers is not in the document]
MatKuhr: I still don't understand this "subscriber" aspect, so I don't know what this call is supposed to do, and in which cases this call failing can have a reasonable recovery..

packages/connectivity/src/scp-cf/xsuaa-service.ts
```diff
@@ -7,30 +7,29 @@ import {
...
+ * @param jwt - User JWT or object containing the `iss` property.
  * @returns Client credentials token.
  */
 export async function getClientCredentialsToken(
   service: string | Service,
-  userJwt?: string | JwtPayload
+  jwt?: string | JwtPayload
 ): Promise<ClientCredentialsResponse> {
-  const jwt = userJwt ? decodeJwt(userJwt) : {};
+  const decodedJwt = jwt ? decodeJwt(jwt) : {};
```
tomfrenken, line 29: I think this line is redundant, as it is already done in both getSubdomain and tenantId

packages/connectivity/src/scp-cf/xsuaa-service.ts
```diff
@@ -75,10 +74,10 @@ export function getUserToken(
   service: Service,
   userJwt: string
 ): Promise<string> {
-  const jwt = decodeJwt(userJwt);
+  const decodedUserJwt = decodeJwt(userJwt);
```
tomfrenken, line 77: Presumably also redundant

packages/connectivity/src/scp-cf/destination/get-subscriber-token.spec.ts
```diff
@@ -70,7 +70,7 @@ describe('getSubscriberToken()', () => {
     expect(serviceTokenSpy).toHaveBeenCalledWith(
       'destination',
       expect.objectContaining({
-        jwt: { iss: onlyIssuerXsuaaUrl }
+        jwt: { ext_attr: { zdn: 'subscriber-only-iss' } }
```
marikaner, line 73: Previously we mimicked a token with an issuer here. I changed this to be reflected in the `zdn` property, because the issuer should only be checked in the case we have an XSUAA token. If it is an IAS token, the subdomain of the tenant cannot be parsed from the issuer. Therefore we now first look at the `zdn` property and only if this is not there look at the issuer, if this is an XSUAA token.

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
```diff
@@ -49,12 +51,12 @@ export function isSubscriberToken(token: any): token is SubscriberToken {
 export async function getSubscriberToken(
   options: DestinationOptions
 ): Promise<SubscriberToken> {
-  const isXsuaaJwt =
-    !!options.jwt && isXsuaaToken(decodeJwtComplete(options.jwt));
+  const isXsuaaJwt = !!options.jwt && isXsuaaToken(decodeJwt(options.jwt));
+  const userJwt = await retrieveUserToken(options, isXsuaaJwt);
```
tomfrenken, line 55: Why move this to a separate line but leave serviceJwt the same?
Either create separate lines for both or for none of them.

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
```diff
@@ -72,34 +74,40 @@ async function retrieveUserToken(
 
 async function retrieveServiceToken(
   options: DestinationOptions,
-  isXsuaaJwt: boolean
+  decodedUserJwt: JwtPayload | undefined
 ): Promise<JwtPair | undefined> {
-  const jwt = getJwtForServiceToken(options, isXsuaaJwt);
+  const jwt = getJwtForServiceToken(options?.iss, decodedUserJwt);
```
tomfrenken, line 79: Why have a nullish check here, if the type of options isn't `options: DestinationOptions | undefined`?

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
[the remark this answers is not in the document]
marikaner: I am really not sure what the problem was, but I remember, that something did not work, I just don't know what. I would be willing to try this out, but definitely on a separate PR and probably after my vacation, if priorities allow it.

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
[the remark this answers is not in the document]
marikaner: This function is called to fetch a token for the subscriber. As we currently ALWAYS retrieve a subscriber token, when a user JWT is available, it is possible that we call this, although it isn't needed, because users are actually in the provider case. This is a very obvious flaw, but I don't dare touching this at the moment.

packages/connectivity/src/scp-cf/xsuaa-service.ts
marikaner: True, but then it is done only once instead.

packages/connectivity/src/scp-cf/xsuaa-service.ts
marikaner: same

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
marikaner: Because the userJwt is used to retrieve the serviceJwt.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -13,7 +16,8 @@ export async function exchangeToken(
   options: DestinationOptions
 ): Promise<string> {
   const xsuaaService = getXsuaaService({
-    disableCache: !options.cacheVerificationKeys
+    disableCache: !options.cacheVerificationKeys,
+    credentials: getDestinationServiceCredentials()
```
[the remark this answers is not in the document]
marikaner: because we are retrieving this token only for the destination service, so this should be enough. Honestly, we need to clarify, whether there are cases where we would need to do this for connectivity. I am even thinking splitting this off this PR.

packages/connectivity/src/scp-cf/xsuaa-service.ts
tomfrenken: Isn't it done multiple times anyway?
`getSubdomain` and `tenantId` do only check if a jwt exists, not if it is already decoded, and will then execute `decodeJwt` regardless, or am I missing something?
(I didn't see how exactly the underlying `decode` function works)

packages/connectivity/src/scp-cf/destination/destination-service.ts
```diff
@@ -231,15 +230,16 @@ export async function fetchCertificate(
...
-function getTenantFromTokens(token: AuthAndExchangeTokens | string): string {
+function getTenantIdFromTokens(token: AuthAndExchangeTokens | string): string {
   let tenant: string | undefined;
   if (typeof token === 'string') {
-    tenant = getTenantIdWithFallback(token);
+    // TODO: subdomain !== tenantID, the fallback is probably incorrect and should be removed or replaced with subdomain explicitly (in middlewarecontext)
+    tenant = getTenantId(token);
   } else {
     tenant =
       token.exchangeTenant || // represents the tenant as string already see https://api.sap.com/api/SAP_CP_CF_Connectivity_Destination/resource
```
marikaner, line 240:
```suggestion
      // represents the tenant as string already see https://api.sap.com/api/SAP_CP_CF_Connectivity_Destination/resource
      token.exchangeTenant || 
```

packages/connectivity/src/scp-cf/xsuaa-service.ts
tomfrenken: Looking further into the decode function, it is indeed only called once :)

.changeset/empty-eggs-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Support IAS token without the need to pass `iss` in the destination fetch options.
```
marikaner, line 5:
```suggestion
[New Functionality] Support IAS tokens without the need to pass `iss` in the destination fetch options.
```

packages/connectivity/src/scp-cf/jwt.ts
```diff
@@ -31,17 +32,36 @@ export function userId({ user_id }: JwtPayload): string {
...
+  const decodedJwt = jwt ? decodeJwt(jwt) : {};
+  logger.debug(
+    `JWT zid is: ${decodedJwt.zid}, app_tid is: ${decodedJwt.app_tid}.`
+  );
+  return decodedJwt.zid || decodedJwt.app_tid || undefined;
+}
+
+/**
+ * @internal
+ * Retrieve the subdomain from the decoded XSUAA JWT. If the JWT is not in XSUAA format, returns undefined.
```
marikaner, line 52:
```suggestion
 * Retrieve the subdomain from the decoded XSUAA JWT. If the JWT is not in XSUAA format, returns `undefined`.
```

PR #4807, opened 2024-07-03
chore: expose getTenantId function

tomfrenken:
With this PR, we are renaming tenantId to getTenantId and are exposing it to users, so that they have additional convenience, especially when setting tenant headers, which is a requirement for multi-tenant applications running against on-premise transparent proxy destinations.

### DIFF
--- a/.changeset/quick-points-return.md
+++ b/.changeset/quick-points-return.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Added getTenantId function for convenience.
--- a/.changeset/quick-points-return.md
+++ b/.changeset/quick-points-return.md
@@ -2,4 +2,4 @@
 '@sap-cloud-sdk/connectivity': minor
 ---
 
-[New Functionality] Added getTenantId function for convenience.
+[New Functionality] Add `getTenantId()` function for convenience.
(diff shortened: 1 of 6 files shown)


packages/connectivity/src/scp-cf/jwt.ts
```diff
@@ -32,12 +32,13 @@ export function userId({ user_id }: JwtPayload): string {
 }
 
 /**
- * @internal
  * Get the tenant ID of a decoded JWT, based on its `zid` or if not available `app_tid` property.
  * @param jwtPayload - Token payload to read the tenant ID from.
+ * @param jwtPayload.zid - The zone ID, representing the tenant in the XSUAA token.
+ * @param jwtPayload.app_tid - The app tenant ID, representing the tenant in the IAS token.
```
marikaner, line 38: [q] Have you checked how the docs are rendered with this? I think this doesn't work and maybe it is not needed (because you are describing what is in the `JwtPayload` and it is pointed out in the docs above.

marikaner (changes requested):
looks good, but please add a changelog

packages/connectivity/src/scp-cf/jwt.ts
tomfrenken: Mostly added it because vale is crying otherwise, I guess I can also exclude this instead.

.changeset/quick-points-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Added getTenantId function for convenience.
```
marikaner, line 5: [req] As usual, please follow our style guide and changelog guidelines.

```suggestion
[New Functionality] Add `getTenantId()` function for convenience.
```

marikaner (approved):

PR #4832, opened 2024-07-15
chore: Adjust client generated comments to follow eslint rules

deekshas8:
Some of the generated comments in the clients dont follow jsdoc complete sentence rule. Adjusted those. Also removed prettier recommended config from the new config, as it seems unnecessary.


tomfrenken (approved):
LGTM, just need to regenerate the test services.

Also for future reference; the reason we can just remove the prettierRecommended plugin is, because the same rules are already in our existing set of rules.

PR #4884, opened 2024-08-02
fix: Make generation slightly more resilient

marikaner:
I accidentally found that `oneOf` does not get generated correctly in some cases and the order matters. I think what we are doing is not a mistake but now we allow slightly incorrect specs as well.


deekshas8 (approved):

deekshas8:
I approved it already, but if it's not a good idea, then please feel free to close it.

marikaner:
It was fixed by the service owners for the service we were looking at, so I am fine to close it.

PR #4899, opened 2024-08-08
fix: Fix OpenAPI client's generated typedoc

tomfrenken:
With this PR, we check if the existing type doc is properly formated, and if not, do it ourselves.

### DIFF
--- a/.changeset/clever-hairs-flow.md
+++ b/.changeset/clever-hairs-flow.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/openapi-generator': patch
+---
+
+[Fixed Issue] Properly format generated typedoc if existing documentation doesn't align with our ESLint config.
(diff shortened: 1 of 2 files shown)


.changeset/clever-hairs-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/openapi-generator': patch
+---
+
+[Fixed Issue] Properly format generated typedoc if existing documentation doesn't align with our ESLint config.
```
deekshas8, line 5: [pp] I would avoid `our`, maybe SAP Cloud SDK or the actual package name.

packages/openapi-generator/src/file-serializer/schema-file.ts
```diff
@@ -48,9 +48,13 @@ function getImports(
  * @internal
  */
 export function schemaDocumentation(schema: OpenApiPersistedSchema): string {
+  const schemaDescription = schema.description
```
deekshas8, line 51: [pp]

```
const schemaDescription = schema.description
  ? `${schema.description}${schema.description.endsWith('.') ? '' : '.'}`
  : `Representation of the '${schema.schemaName}' schema.`;
```

deekshas8 (approved):
The services need to be regenerated, but other than that LGTM, only minor comments.

PR #53, opened 2020-03-22
chore: add doc 1.18.1 and adjust script for index based structure

jjtang1985:
## Context

- created index file with all links to the docs
- moved 1.18.0 doc to the `docs/1.18.0` folder
- added 1.18.1 doc in the `docs/1.18.1` folder and reflect it in the index file
- adjusted the script to 
    - move new docs to `docs/x.y.z` folder instead of under the `docs`
    - add new link of doc to the index file

## Definition of Done

Please consider all items and remove only if not applicable.

- [ ] Tests created/adjusted for your changes.
- [ ] Release notes updated.
  * Provide sufficient context so that each entry can be understood on its own.
  * Be specific about names of functions, classes, modules, etc.
  * Describe when or where this is relevant
  * Use indicative and present tense. For example, write "Provide function `name` that does X in order to Y" over "Now X can be done by calling a new function".
- [x] PR title adheres to [conventional commit guidelines](https://www.conventionalcommits.org) (please note that only `fix:` and `feat:` will end up in the release notes)
- [ ] If applicable: Properly documented (JSDoc of public API)

## Labels
We use labels to indicate the status of PRs. 
Please select `please review`, `please merge` or `don't merge` for your PR.

PR #4909, opened 2024-08-12
feat: Support parsing top level media types

shibeshduw:
This is a follow up ticket to [#53](https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/53). 
This PR adds support for additional media types, namely `application/octet-stream`, `text/plain` and `*/*` so that the generated clients are appropriately typed and warning messages are not logged during the generation process.

Closes [AI/gen-ai-hub-sdk-js#73](https://github.tools.sap/AI/gen-ai-hub-sdk-js-backlog/issues/73).


shibeshduw:
While testing this, I found that right now, we don't support generation for responses with multiple content types which [is allowed](https://swagger.io/docs/specification/media-types) in the official spec (see: Multiple Media Types). If we were to support this, then we would need to modify the `parseTopLevelMediaType` to parse multiple content types which are allowed. Probably not in the scope but we can create a separate ticket for this if required.

.changeset/spotty-bulldogs-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/openapi-generator': minor
+---
+
+[Improvement] Generated OpenAPI clients support `text/plain`, `application/octet-stream` and wildcard `*/*` content type within response headers.
```
marikaner, line 5: Please use the imperative (you can find examples in the previous release notes, e.g.).

packages/openapi-generator/src/parser/type-mapping.ts
```diff
@@ -26,6 +26,7 @@ const typeMapping = {
   map: 'any',
   date: 'string',
   DateTime: 'string',
+  Blob: 'Blob',
```
marikaner, line 29: [q] This is never used, is it?

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -66,6 +66,12 @@ export function parseSchema(
     return parseXOfSchema(schema, refs, 'anyOf', options);
   }
 
+  if (schema.type === 'string' && schema.format === 'binary') {
+    return {
+      type: 'Blob'
+    };
+  }
+
```
marikaner, line 74: [pp] I think I would like to push this down into `getType()` (internal) function.

It currently looks like this:
```ts
export function getType(originalType: string | undefined): string {
  const type = originalType ? typeMapping[originalType] : 'any';
  if (!type) {
    logger.verbose(
      `Could not map type '${originalType}' to a native type. Using any.`
    );
    return 'any';
  }
  return type;
}
```

but it might also just take the schema:

```ts
export function getType(schema: Pick<OpenApiSchema, 'type', 'format'>): string {
  if (schema.type === 'string' && schema.format === 'binary') {
    return {
      type: 'Blob'
    };
  }
  const type = originalType ? typeMapping[originalType] : 'any';
  if (!type) {
    logger.verbose(
      `Could not map type '${originalType}' to a native type. Using any.`
    );
    return 'any';
  }
  return type;
}
```

marikaner (commented):
Please add a test case to the test api, to make sure passing a blob to axios works without additional tweaks.

packages/openapi-generator/src/parser/type-mapping.ts
shibeshduw: Yes, its not used after the below change.

packages/openapi-generator/src/parser/schema.ts
shibeshduw: I opted to use a union type of schema and string because the type-mapping tests we have would need to be changed otherwise. But the current implementation is a bit cluttered whereas if we just change the tests, it would be cleaner. Thoughts @marikaner?

marikaner: I'd always go for changing the tests rather than keeping cluttered code 😊 This shouldn't affect tooooo many tests, right?

shibeshduw: Just a little as it turns out. I kept it simple with the type of schema in `getType()`. The reason we need both `NonArraySchemaObject`  and `OpenApiSimpleSchema` is because the function is called in both the parser and file-serializer with different schema typing.

packages/openapi-generator/src/parser/type-mapping.ts
```diff
@@ -26,6 +28,7 @@ const typeMapping = {
...
```
marikaner, line 31: [req] So it should be removed ;) 
```suggestion
```

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -51,13 +54,13 @@ export function parseMediaType(
 ): OpenApiSchema | undefined {
   const allMediaTypes = getMediaTypes(bodyOrResponseObject);
   if (allMediaTypes.length) {
-    const jsonMediaType = parseApplicationJsonMediaType(
+    const parsedMediaType = parseTopLevelMediaType(
```
marikaner, line 57: [q] Why is it "top level"?

shibeshduw: Since we are adding more than just application/json now, I thought using the [standard term](https://www.rfc-editor.org/rfc/rfc2046.html#section-2) for it was more appropriate.

marikaner: Interesting! never heard of it.

packages/openapi-generator/src/parser/type-mapping.ts
marikaner: ah ok, good to know

marikaner (approved):
yeah, finally 😊

PR #4915, opened 2024-08-14
fix: fix token exchange

marikaner:
The token exchange was looking at the token's audiences before and basically did a token verification for XSUAA in the case of IAS tokens. This is incorrect, the token should be merely exchanged.

### DIFF
--- a/packages/connectivity/src/scp-cf/destination/destination-accessor.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-accessor.ts
@@ -129,8 +129,8 @@ export async function getAllDestinationsFromDestinationService(
   logger.debug(
     'Attempting to retrieve all destinations from destination service.'
   );
-  if (shouldExchangeToken(options)) {
-    options.jwt = await exchangeToken(options);
+  if (shouldExchangeToken(options) && options.jwt) {
+    options.jwt = await exchangeToken(options.jwt);
   }
 
   const token =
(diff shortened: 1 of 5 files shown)


packages/connectivity/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -129,8 +129,8 @@ export async function getAllDestinationsFromDestinationService(
   logger.debug(
     'Attempting to retrieve all destinations from destination service.'
   );
-  if (shouldExchangeToken(options)) {
-    options.jwt = await exchangeToken(options);
+  if (shouldExchangeToken(options) && options.jwt) {
```
tomfrenken, line 132: I assume you only added this so the IDE doesn't complain?
Because shouldExchangeToken only returns true if options.jwt is set.

tomfrenken (approved):
Just one comment, other than that LGTM

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
tomfrenken: Alternatively you could use `options.jwt!` but I do not really have a [pp] here.

marikaner: exactly. Using `!` is worse in my opinion.

PR #4916, opened 2024-08-14
chore: Change modifier to protected

deekshas8:
This makes certain fields/ methods protected so they can be accessed within the subclass in sap-ai-sdk.


marikaner (approved):
lgtm

PR #4922, opened 2024-08-15
chore: Set default string as CC Token cache key when tenantId cannot be determined

deekshas8:
When fetching serviceToken for a service, if the cache key cannot be determined from either jwt or bindings, then the default `provider-tenant` is used.


marikaner (approved):
lgtm

PR #4927, opened 2024-08-19
chore: Update TS version to ~5.5.4

deekshas8:
Updates the typescript version to ~5.5.4 in all packages. I reverted ^5.5.4 since TS apparently does not support semver.

The update to the generated client is excluded as that will be a breaking change (the current version in clients is ~4.5 and ~4.1.2 in odata and openapi respectively). This will be handled in a separate [ticket](https://github.com/SAP/cloud-sdk-backlog/issues/1232) in the future


tomfrenken (approved):
LGTM

tomfrenken (changes requested):
Since we apparently need to make some type changes, I'll block it for now

package.json
```diff
@@ -110,8 +110,8 @@
     "ts-jest": "^29.2.4",
     "ts-node": "^10.9.2",
     "turbo": "^2.0.14",
-    "typedoc": "^0.25.13",
-    "typescript": "~5.4.5"
+    "typedoc": "^0.26.0",
```
deekshas8, line 113: [info] typedoc 0.26 added support for TS 5.4.

build-packages/check-pr/tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
   "compilerOptions": {
+    "target": "es2021",
+    "module": "Node16",
```
deekshas8, line 4: [info] Added these options explicitly here to resolve the error in `validator.ts`:
 `Named capturing groups are only available when targeting 'ES2018' or later`

packages/odata-common/src/request-builder/batch/batch-response-deserializer.ts
```diff
@@ -68,21 +68,25 @@ export class BatchResponseDeserializer<DeSerializersT extends DeSerializers> {
         this.responseDataAccessor,
         this.deserializer
       ),
-      isSuccess: () => true,
-      isError: () => false,
-      isReadResponse: () => true,
-      isWriteResponses: () => false
+      isSuccess: (): this is
```
deekshas8, line 71: I assume TS 5.5.4 enforces stricter checks on predicates compared to 5.4. The error here was: 

```
Type '() => true' is not assignable to type '() => this is ReadResponse<DeSerializersT> | WriteResponses<DeSerializersT>'.
  Signature '(): true' must be a type predicate.ts(2322)
batch-response.ts(15, 3): The expected type comes from property 'isSuccess' which is declared here on type 'ReadResponse<DeSerializersT>'
```

tomfrenken (approved):
LGTM, I took a look at the changelog from typescript, and the types seem to get narrowed stronger with the latest version, this also explains all the deletions in the generated clients.

Issue #4928, opened 2024-08-19
OpenAPI generator generates invalid code in specific combinations of required/optional header and query parameters


sven-petersen:
**Describe the bug**

The OpenAPI generator generates invalid JavaScript/TypeScript code when having both, optional and required, header parameters as well as only optional query parameters. The resulting typescript error is: `A required parameter cannot follow an optional parameter.ts(1016)`

**To Reproduce**

Generate an OpenAPI client using the OpenAPI schema from below.
Check the generated code using an IDE or try to build it with tsc to see the error I mentioned above.

Interestingly correct code will be generated when the `OptionalHeaderParam` in the **header** is removed. In this case the `queryParameter` becomes a required function parameter. In the generate code this looks like followed:

```ts
// with the "OptionalHeaderParam"
sampelOperation: (
    queryParameters?: { optionalQueryParam?: number },
    headerParameters: {
      "required-header-param": string;
      "optional-header-param"?: string;
    },
  ) => { /* ... */ }
```
becomes 

```ts
// without the OptionalHeaderParam
sampelOperation: (
    queryParameters: { optionalQueryParam?: number },
    headerParameters: { "required-header-param": string },
  ) => { /* ... */ }
```

**Expected behavior**

Generate valid TS code.

**Used Versions:**

- node version `20.11.1`
- npm version `10.2.4` 
- SAP Cloud SDK version: `3.18.1`

**Code Examples**

```yaml
openapi: "3.0.3"
info:
  version: "0.1.1"
  title: "Minimal example"
security:
  - mtls: []
paths:
  /samplePath:
    get:
      operationId: sampelOperation
      parameters:
        - $ref: "#/components/parameters/RequiredHeaderParam"
        - $ref: "#/components/parameters/OptionalHeaderParam"
        - $ref: "#/components/parameters/OptionalQueryParam"
      responses:
        "200":
          description: "A response"
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/SampleResponseSchema"
components:
  parameters:
    RequiredHeaderParam:
      name: required-header-param
      in: header
      required: true
      schema:
        type: string
    OptionalHeaderParam:
      name: optional-header-param
      in: header
      required: false
      schema:
        type: string
    OptionalQueryParam:
      name: optionalQueryParam
      in: query
      required: false
      schema:
        type: integer
  schemas:
    SampleResponseSchema:
      type: object
      properties:
        content:
          type: string
```

will generate this code

```typescript
[...]

PR #4935, opened 2024-08-22
fix: Prevent query parameters from being optional when header parameters are required.

shibeshduw:
With this change, the file serializer in `openapi-generator` will always ensure that query parameters are not marked optional when header parameters are required by conditionally passing the correct optional flag to `serializeParamsForSignature`.

Closes SAP/cloud-sdk-js#4928.


deekshas8:
@shibeshduw I am not sure if I am misunderstanding, but I understood the user's issue as: incorrect code is generated when there are optional query and required header params, I thought this was something we covered in our [tests](https://github.com/SAP/cloud-sdk-js/blob/5847e0af7e04286e58a6794bc297dffe5895430d/packages/openapi-generator/src/file-serializer/operation.spec.ts#L195). 

Is this not the case?

shibeshduw:
Yes you're correct. But the manner in which we were [checking for optional modifier](https://github.com/SAP/cloud-sdk-js/blob/5847e0af7e04286e58a6794bc297dffe5895430d/packages/openapi-generator/src/file-serializer/operation.ts#L47) was wrong. I was suggesting that we make the order of parameters dynamic but after speaking with @marikaner , I go to know that this would be a breaking change. So now I have reverted the changes and added a minimal fix for the issue.

packages/openapi-generator/src/file-serializer/operation.ts
```diff
@@ -36,15 +36,24 @@ ${operation.operationId}: (${serializeOperationSignature(
...
-    'headerParameters'
+    'headerParameters',
+    allOptionalHeaders
   );
 
   const queryParams = serializeParamsForSignature(
     operation,
     'queryParameters',
-    headerParams?.includes('?')
+    allOptionalHeaders ? allOptionalQuery : false
```
marikaner, line 56:
```suggestion
    allOptionalHeaders && allOptionalQuery
```

marikaner (approved):
LGTM

marikaner:
Please still correct the title 😊

PR #4937, opened 2024-08-23
chore: Improve check pr action

tomfrenken:
With this PR, the check-pr action fails if the change type ([Fixed Issue], ...) is written wrong, and logs proper types.

This will prevent our bump action from failing due to wrong change types when running `yarn version`.


build-packages/check-pr/validators.spec.ts
```diff
@@ -75,14 +78,38 @@ describe('check-pr', () => {
       expect(process.exitCode).toEqual(1);
     });
 
+    it('should fail if change type is false', async () => {
+      const fileContents = [
+        "'@sap-cloud-sdk/generator': major",
+        '[Fix] Something is fixed.'
+      ];
+      await validateChangesets('chore!', '', true, fileContents);
```
deekshas8, line 86: I mean technically a chore is not supposed to be used for a `Fixed Issue` ;)

deekshas8 (approved):

PR #4941, opened 2024-08-26
fix: Fix reading correct tenant for destination caching

marikaner:
Currently, for IAS tokens we do not read the correct key for the tenant, when registering destinations.


packages/connectivity/src/scp-cf/destination/destination-from-registration.ts
```diff
@@ -58,7 +63,7 @@ export async function registerDestination(
 
 function getJwtForCaching(options: RegisterDestinationOptions | undefined) {
   const jwt = decodeOrMakeJwt(options?.jwt);
-  if (!jwt?.zid) {
+  if (!getTenantId(jwt?.zid)) {
```
marikaner, line 66:
```suggestion
  if (!getTenantId(jwt)) {
```

tomfrenken (approved):
LGTM

PR #4945, opened 2024-08-27
chore: Adjust type of CustomRequestConfig

deekshas8:
Type is constructed using Pick instead of Omit


packages/http-client/src/http-client-types.ts
```diff
@@ -200,10 +200,16 @@ export interface HttpRequestOptions {
 /**
  * The type for parameter in Custom Request Configuration.
  */
-export type CustomRequestConfig = Omit<
+export type CustomRequestConfig = Pick<
   HttpRequestConfig,
-  'url' | 'baseURL' | 'data' | 'headers' | 'params'
->;
+  | 'middleware'
```
deekshas8, line 205: @marikaner I excluded `method` here because it is not an optional property which would then potentially be a breaking change. Also in req builders, users can set `method: 'merge'` which would not be possible if we pick'' it.

PR #4953, opened 2024-08-30
chore: Adjust openapi generator for azure spec

deekshas8:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/1238

- Schema with type object have 3 possible values in Azure spec: (Whether all are correct is another point). The below are now handled in the generation
  - The normal 
     ```
     type: object
          properties:
            error:
              $ref: '#/components/schemas/error'
    ```
  - multiple at same level (This is normalized as only allOf)
    ```
    error:
          type: object
          allOf:
            - $ref: '#/components/schemas/errorBase'
          properties:
            ...
    ```
  - allOf Inheritance (type: Object is ignored and is treated only as allOf)
      ```
      contentFilterDetectedResult:
            type: object
            allOf:
              - $ref: '#/components/schemas/contentFilterResultBase'
              - properties:
                  detected:
                    ....
           required:
              - detected
      ```
- In 3rd example, if `detected` is required, it is **not yet** considered in generation process.


packages/openapi-generator/src/file-serializer/schema.ts
```diff
@@ -32,13 +32,18 @@ export function serializeSchema(schema: OpenApiSchema): string {
     const type = serializeSchema(schema.items);
     return schema.uniqueItems
       ? `Set<${type}>`
-      : 'properties' in schema.items
+      : 'properties' in schema.items || 'allOf' in schema.items
```
marikaner, line 35: [req] This only solves it for `allOf`, but we should make sure this is also fixed for `oneOf` and `anyOf`.

packages/openapi-generator/src/file-serializer/schema.ts
```diff
@@ -32,13 +32,18 @@ export function serializeSchema(schema: OpenApiSchema): string {
...
   }
 
   if (isObjectSchema(schema)) {
-    return serializeObjectSchema(schema);
+    const types: string[] = [];
+    if (isAllOfSchema(schema)) {
+      types.push(schema.allOf.map(type => serializeSchema(type)).join(' & '));
+    }
+    types.push(serializeObjectSchema(schema));
+    return types.join(' & ');
```
marikaner, line 46: This is duplicate code, as we do this for the normal case, where we only have `allOf` as well. Also, we need to consider `oneOf` and `anyOf`.
I think it would be simpler to do this at the time of parsing instead of serializing (see my comment in the parser).

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -47,6 +47,14 @@ export function parseSchema(
...
     'additionalProperties' in schema
   ) {
+    if (isAllOfSchema(schema)) {
+      return {
+        ...(schema.allOf?.length &&
+          parseXOfSchema(schema, refs, 'allOf', options)),
+        ...((schema.properties || 'additionalProperties' in schema) &&
+          parseObjectSchema(schema, refs, options))
+      };
+    }
```
marikaner, line 57: [req] I still believe combining `allOf` and `type: object` is just incorrect, therefore if both are set and we know a few cases where services do that, I would allow that and treat it as `allOf` only. Parsing other properties and handling them a certain way is just guess work and as far as I know we haven't seen examples of that before, did we?

Therefore my suggestion is to just move this down below `anyOf` so objects are prioritized lower.

marikaner: [pp] I think it could be helpful to have a warning when generating this (incorrect) type. What do you think?

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -221,8 +229,11 @@ function parseXOfSchema(
   xOf: 'oneOf' | 'allOf' | 'anyOf',
   options: ParserOptions
 ): any {
+  const required = schema.required;
   return {
-    [xOf]: (schema[xOf] || []).map(entry => parseSchema(entry, refs, options))
+    [xOf]: (schema[xOf] || []).map(entry =>
+      parseSchema({ ...entry, required }, refs, options)
```
marikaner, line 235: [req] In this case the schema would overwrite anything that is required in the actual entry. We should at least merge this, but I think we shouldn't do this at all.

marikaner (changes requested):

packages/openapi-generator/src/parser/schema.ts
deekshas8: Removed this for now. Required will be considered in a followup

packages/openapi-generator/src/parser/schema.ts
deekshas8: Added info log if allOf and properties are at the same level. It is restructured into the "inheritance" type as discussed

packages/openapi-generator/src/options/options.ts
```diff
@@ -37,5 +42,11 @@ export const cliOptions = {
       'When enabled, all generated files follow the ECMAScript module syntax.',
     type: 'boolean',
     default: false
+  },
+  schemaPrefix: {
+    describe:
+      'When enabled, all generated files follow the ECMAScript module syntax.',
```
marikaner, line 48: [req] This is not the right description.

packages/openapi-generator/src/options/options.ts
```diff
@@ -37,5 +42,11 @@ export const cliOptions = {
...
+    type: 'string',
+    default: undefined
```
marikaner, line 50: [q] Would it be easier to handle if it was `''`? Then you wouldn't need any special handling when using the prefix.

packages/openapi-generator/src/parser/refs.ts
```diff
@@ -72,7 +72,7 @@ export class OpenApiDocumentRefs {
       (mapping, originalName, i) => ({
         ...mapping,
         [`#/components/schemas/${originalName}`]: {
-          schemaName: schemaNames[i],
+          schemaName: pascalCase(options.schemaPrefix ?? '') + schemaNames[i],
```
marikaner, line 75: [pp] I think if we give the user the flexibility to assign a prefix, we should leave it as given.

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -59,13 +51,28 @@ export function parseSchema(
...
+      delete schema.properties;
+    }
     return parseXOfSchema(schema, refs, 'allOf', options);
   }
 
   if (schema.anyOf?.length) {
     return parseXOfSchema(schema, refs, 'anyOf', options);
   }
 
+  if (
```
marikaner, line 68: [nth] Please consider adding a comment, so that other developers are aware that the order matters here

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -59,13 +51,28 @@ export function parseSchema(
   }
 
   if (schema.allOf?.length) {
+    if (schema.properties) {
```
marikaner, line 54: [req] I think this needs to be handled for anyOf and oneOf as well.

marikaner, line 54: IDEA: I am not sure what else could lie next to `allOf`. Wouldn't it be nice to "normalize" the schema through something along the lines of this:
```ts
const { allOf, ...objectSchema } = schema;
return parseXOfSchema({allOf: [...allOf, objectSchema] }, ref, 'allOf', options);
```

(I think you mentioned `required` could also be next to allOf, but then just add that as well.)

marikaner (changes requested):
I would prefer if we could split PRs like this one by concern: allof + properties and schema naming.

I have some more questions:
1. I assume the schema names including the prefix go through the unique name generator? Maybe a test would be nice?
2. I think the name prefix is not a hidden property. I think that is ok, but then please make sure to document it in the portal as well.
3. Consider not only supporting `properties` for the all of schemas, but also `additionalProperties`.

packages/openapi-generator/src/parser/schema.ts
deekshas8: @marika In case where `allOf` and `properties` are at the same level (and have required), this actually works quite well.
For the other case (which is the case in the spec):

```
allOf:
  - ref
  - type: object //there can be multiple of such objects i imagine
     properties:
       - ...
required:
 -
```

Here I would need to add `required` to(probably) every object item. I had handled this in `parseXOfSchema` before as:

```
return {
    [xOf]: (schema[xOf] || []).map(entry =>
      parseSchema({ ...entry, required }, refs, options)  };
```

Wouldn't this mean the same thing?

packages/openapi-generator/src/parser/schema.ts
deekshas8: I merge this now. Have also added a test for it.

packages/openapi-generator/src/parser/schema.ts
deekshas8: Sorry I had somehow missed this again. I moved the logic to the `parseXOfSchema` function now.

deekshas8: Updated it now (incl. merging the required properties from different levels). Please have a look

packages/openapi-generator/src/generator.ts
```diff
@@ -264,7 +264,9 @@ async function generateService(
   const parsedOpenApiDocument = await parseOpenApiDocument(
     openApiDocument,
     serviceOptions,
-    { strictNaming: !options.skipValidation }
+    {
+      strictNaming: !options.skipValidation
+    }
```
marikaner, line 269:
```suggestion
    { strictNaming: !options.skipValidation }
```

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -221,8 +223,34 @@ function parseXOfSchema(
...
+    normalizedSchema = { [xOf]: [...(schema[xOf] || []), objectSchema] };
+  }
   return {
-    [xOf]: (schema[xOf] || []).map(entry => parseSchema(entry, refs, options))
+    [xOf]: (normalizedSchema[xOf] || []).map(entry =>
+      parseSchema(
+        {
+          ...entry,
+          required: [
+            ...('required' in entry && entry.required ? entry.required : []),
```
marikaner, line 246: [q] Why is this necessary? I would have expected that this is enough:
```suggestion
            ...(entry.required || []),
```

test-resources/openapi-service-specs/swagger-yaml-service.yml
```diff
@@ -54,3 +54,23 @@ definitions:
         type: integer
         description: An integer property
         required: false
+  TestEntity_2:
```
marikaner, line 57: [nth] Consider giving those better names (what is the test case those entities are used for?)

marikaner (changes requested):

packages/openapi-generator/src/parser/schema.ts
deekshas8: required does not exist on type 'ReferenceObject'. But I guess I can also rewrite it as :

`...(entry['required'] || []),`

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -221,11 +223,46 @@ function parseXOfSchema(
...
 }
 
+function normalizeSchema(
+  schema: OpenAPIV3.NonArraySchemaObject,
+  xOf: 'oneOf' | 'allOf' | 'anyOf'
+): OpenAPIV3.NonArraySchemaObject {
+  if (
+    schema.properties ||
+    (schema.additionalProperties &&
+      typeof schema.additionalProperties === 'object')
```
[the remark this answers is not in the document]
marikaner: if the answer to the question above is yes, you could consider extracting a function to check whether the given schema is an object schema and reusing it for the object check above as well.

marikaner (approved):
lgtm, I just left one question

packages/openapi-generator/src/parser/schema.ts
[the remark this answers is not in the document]
deekshas8: So, if the object has properties and `schema.additionalProperties = true` then it will add it. If its:

```
type: object
allOf:
  - ...
additionalProperties: true
```

then I think not. I probably can just check for `schema.additionalProperties` instead of ```
schema.additionalProperties && typeof schema.additionalProperties === 'object'
```

and that should already fix it. Honestly I am forgetting now why I checked for object type. Let me try.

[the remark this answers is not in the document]
deekshas8: Tested this after removing `&& typeof schema.additionalProperties === 'object'` check. And it works.

PR #4993, opened 2024-09-12
chore: Move release relevant actions to GH actions

marikaner:
There is quite some logic going on when merging the changesets, so I didn't want to repeat it. Also bumping and tagging the version is the same action, so moved all three of those into actions, so that we can reuse them in our other project.

I had to refactor a few places to not duplicate the code and I changes smaller things along the way (removing `unixEOL` and executing file write and read options async instead of sync).

I left a few TODOs for potential follow ups (does not have to happen immediately, but I would like to attempt it on the friday ;) ). I would like to separate those from this already big PR.


.github/workflows/bump.yml
```diff
@@ -1,32 +1,54 @@
 name: bump
 
 on:
+  push:
+    branches:
+      - merge-changelogs
```
marikaner, line 6: This needs to be removed before merging!

.github/workflows/bump.yml
```diff
@@ -1,32 +1,54 @@
...
   bump:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           token: ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}
-          ref: 'main'
-      - uses: actions/setup-node@v3
+          # ref: 'main'
```
marikaner, line 20: This should be removed before merging!

deekshas8 (commented):
Overall, it looks like a huge improvement; I just have a few minor comments.

build-packages/merge-changelogs/index.ts
```diff
@@ -13,7 +16,7 @@ export const validMessageTypes = [
 interface Change {
   packageNames: string[];
   commit: string;
-  type: typeof validMessageTypes[number];
+  type: (typeof validMessageTypes)[number];
```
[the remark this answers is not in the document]
marikaner: I have no idea. But I am already working on a follow up where this won't matter anymore. I think this is not meant to be used manually.

build-packages/changesets-fixed-version-bump/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "changesets-fixed-version-bump",
+  "version": "1.0.0",
+  "description": "Bumps the version of the root package.json and packages according to changesets. Commits and pushes changes and creates a tag.",
```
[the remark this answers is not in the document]
marikaner: yes, not anymore! will fix

scripts/update-documentation-md.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+  await transformFile(resolve('DOCUMENTATION.md'), documentation =>
+    documentation
+      .split(unixEOL)
+      .map(line =>
+        line.startsWith('## Version:')
+          ? `## Version: ${process.env.NEXT_PACKAGE_VERSION}`
+          : line
+      )
+      .join(unixEOL)
```
tomfrenken, line 18: `get-changelog.ts` you decided to replace `unixEOL` with `\n` I suppose you can do the same here, since the action environment always is unix-like?

scripts/update-documentation-md.ts
```diff
@@ -0,0 +1,25 @@
...
+import { exit } from 'node:process';
+
+async function updateDocumentationMd() {
+  if (!process.env.NEXT_PACKAGE_VERSION) {
+    throw new Error('NEXT_PACKAGE_VERSION is not set!');
+  }
+  await transformFile(resolve('DOCUMENTATION.md'), documentation =>
+    documentation
+      .split(unixEOL)
+      .map(line =>
```
tomfrenken, line 13: Can't you use a simple .replace() instead of .split.map.join?

build-packages/merge-changelogs/index.ts
```diff
@@ -164,47 +142,40 @@ function mergeMessages(parsedMessages: Change[]): Change[] {
...
-      const packageJson = await readFile(resolve(workspace, 'package.json'), {
-        encoding: 'utf8'
-      });
-      return [!JSON.parse(packageJson).private, workspace] as const;
-    })
-  );
-  const pathsToPublicLogs = workspacesWithVisibility
-    .filter(([isPublic]) => isPublic)
-    .map(([, workspace]) => resolve(workspace, 'CHANGELOG.md'));
+  const { packages } = await getPackages(process.cwd());
```
tomfrenken, line 162: Nice replacement, out of curiosity, what made you look for an alternative?
/ What should trigger my search for replacements like this one in the future 🤔

scripts/before-bump.ts
```diff
@@ -1,68 +0,0 @@
...
-    /---\n((?<package>.+): (?<level>[a-z]+)\n)+---\n(?<summary>[\w\W]*)/;
-
-  // Explanation: https://regex101.com/r/1JpnOg/1
-  const parseSummary = /(\[(?<type>[\w ]+)\])? ?(?<summary>[\w\W]+)/;
-
-  for (const fileName of changesets) {
-    const content = await readFile(fileName, { encoding: 'utf-8' });
-    const match = content.match(parseCs);
-    if (!match) {
-      console.error(
```
tomfrenken, line 44: I did not find any replacement for this logic in your code.
I take it you deliberately removed this validation step?

scripts/write-changelog.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (!process.env.CHANGELOG) {
+    throw new Error('CHANGELOG environment variable not set.');
+  }
+  if (!process.env.VERSION) {
+    throw new Error('VERSION environment variable not set.');
+  }
+  const unifiedChangelog = await readFile('CHANGELOG.md', { encoding: 'utf8' });
+  await writeFile(
+    'CHANGELOG.md',
+    unifiedChangelog.split('\n').slice(0, 30).join('\n') +
```
tomfrenken, line 14: Can you explain this a bit?
This looks like the first (arbitrarily? 30 lines) are taken, put in front of the new changelog & version, and then the remaining lines are attached at the end??

build-packages/merge-changelogs/index.ts
```diff
@@ -25,26 +28,20 @@ interface ContentByVersion {
...
-  return changelog.split('\n## ').map(h2 => {
-    const [version, ...content] = h2.split('\n');
-    return {
-      version,
-      content: content.join('\n').trim()
-    };
-  });
+  return changelog
+    .split('\n## ')
+    .slice(1)
```
tomfrenken, line 37: Surprised this didn't cause issues before, I guess it returned an object looking like `{ version: '', content: '' }`

tomfrenken (approved):
LGTM, many nice refactors, before you merge though, make sure to read & answer my remaining comments.

scripts/before-bump.ts
deekshas8: The validate changeset message types was moved to the check PR action, but maybe validate structure (`parseCs`) should also be moved there?

scripts/update-documentation-md.ts
marikaner: Probably, I am not 100% what other implications this might have -> apparently there might be multiple lines that look like this. I don't dare touching it really, because there are no tests for it.

build-packages/merge-changelogs/index.ts
marikaner: I needed this information independently of the package manager (as we want to use this for the other SDK).

scripts/write-changelog.ts
marikaner: I can't really to be honest. I think this just takes out the comments that we have in the changelog file. Nobody is touching it manually anyways, so they are not helpful. I didn't touch this logic in this PR, only moved it to a different place.

PR #4996, opened 2024-09-13
chore: Add nullable option for schema properties and persisted Schema

deekshas8:
Closes SAP/cloud-sdk-backlog#1241
Add nullable option for schema properties and persisted Schema


packages/openapi-generator/src/parser/document.ts
```diff
@@ -46,7 +47,8 @@ export function parseSchemas(
       ...refs.getSchemaNaming(`#/components/schemas/${name}`),
       schema: parseSchema(schema, refs, options),
       description: refs.resolveObject(schema).description,
-      schemaProperties: parseSchemaProperties(schema)
+      schemaProperties: parseSchemaProperties(schema),
+      nullable: (!isReferenceObject(schema) && schema.nullable) ?? false
```
marikaner, line 51: [q] can a reference object not be nullable or is this a type issue again, because the reference type does not allow nullable? in that case I would prefer something like you did for the required properties: `'nullable' in schema && schema.nullable...`.

marikaner (approved):
looks good, but I left one question, because semantically it does not explain itself in my opinion.

packages/openapi-generator/src/parser/document.ts
deekshas8: yes, it's because refObj does not have this property. I rewrote it to:

`'nullable' in schema && schema.nullable ? true : false`

PR #5012, opened 2024-09-18
feat: initial implementation for discriminator

marikaner:
This provides an initial implementation for the discriminating properties of OpenAPI specs: https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

According to the documentation this can only happen on `oneOf` and `anyOf` schemas and there can either be an explicit mapping or it is implicitly derived from the referenced schema names.

Of course, the Azure OpenAPI spec does not use `oneOf` but `type: object`. 

This PR got all the above covered. There is however, one more thing:

The Azure OpenAPI spec has a circular reference like:
```yml
- SchemaA:
  discriminator:
    propertyName: discriminatorProp
    mapping:
      value1: => SchemaB
      value2: => SchemaC
- SchemaB:
  allOf:
    - SchemaA // this references back to the original type
    - ...
```

This would result in something like this: 
```ts
type SchemaA = { discriminatorProp: 'value1'} & SchemaB;
type SchemaB = SchemaA & ... // <-- ERROR
```

Currently, the solution that I have in mind is to clean this up before even really parsing the schemas by removing such references.

### DIFF
--- a/packages/openapi-generator/src/parser/schema.spec.ts
+++ b/packages/openapi-generator/src/parser/schema.spec.ts
@@ -4,122 +4,128 @@ import { createTestRefs, emptyObjectSchema } from '../../test/test-util';
 import { OpenApiObjectSchema } from '../openapi-types';
 import { parseSchema, parseSchemaProperties } from './schema';
 
-describe('parseSchema', () => {
-  const defaultOptions = { strictNaming: true, schemaPrefix: '' };
+describe('schema parser', () => {
+  describe('parseSchema()', () => {
+    const defaultOptions = { strictNaming: true, schemaPrefix: '' };
 
-  it('parses reference schema', async () => {
-    const schema = { $ref: '#/components/schemas/test' };
-    expect(
-      parseSchema(
-        schema,
-        await createTestRefs({ schemas: { test: { type: 'string' } } }),
-        defaultOptions
-      )
-    ).toEqual({
-      ...schema,
-      schemaName: 'Test',
-      fileName: 'test'
+    it('parses reference schema', async () => {
+      const schema = { $ref: '#/components/schemas/test' };
+      expect(
+        parseSchema(
+          schema,
+          await createTestRefs({ schemas: { test: { type: 'string' } } }),
+          defaultOptions
+        )
+      ).toEqual({
+        ...schema,
+        schemaName: 'Test',
+        fileName: 'test'
+      });
     });
-  });
 
-  it('parses simple schema', async () => {
-    const schema: OpenAPIV3.SchemaObject = { type: 'string' };
-    expect(parseSchema(schema, await createTestRefs(), defaultOptions)).toEqual(
-      schema
-    );
-  });
+    it('parses simple schema', async () => {
+      const schema: OpenAPIV3.SchemaObject = { type: 'string' };
+      expect(
+        parseSchema(schema, await createTestRefs(), defaultOptions)
+      ).toEqual(schema);
+    });
 
-  it('parses simple schema with description', async () => {
-    const schema: OpenAPIV3.SchemaObject = {
-      type: 'object',
-      properties: { prop1: { description: 'My Description', type: 'string' } }
-    };
-    expect(
-      (
-        parseSchema(
-          schema,
-          await createTestRefs(),
-          defaultOptions
-        ) as OpenApiObjectSchema
-      ).properties[0].description
-    ).toBe('My Description');
-  });
+    it('parses simple schema with description', async () => {
+      const schema: OpenAPIV3.SchemaObject = {
+        type: 'object',
+        properties: { prop1: { description: 'My Description', type: 'string' } }
+      };
+      expect(
+        (
+          parseSchema(
+            schema,
+            await createTestRefs(),
+            defaultOptions
+          ) as OpenApiObjectSchema
+        ).properties[0].description
+      ).toBe('My Description');
+    });
 
-  it('parses simple schema with nullable property', async () => {
-    const schema: OpenAPIV3.SchemaObject = {
-      type: 'object',
-      properties: {
-        prop1: { description: 'My Description', type: 'string', nullable: true }
-      }
-    };
-    expect(
-      (
-        parseSchema(
-          schema,
-          await createTestRefs(),
-          defaultOptions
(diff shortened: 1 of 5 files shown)


packages/openapi-generator/src/parser/document.spec.ts
```diff
@@ -103,6 +103,93 @@ describe('parseOpenApiDocument', () => {
     ]);
   });
 
+  it.only('parses discriminator schema with circular references', async () => {
```
deekshas8, line 106:
```suggestion
  it('parses discriminator schema with circular references', async () => {
```

packages/openapi-generator/src/parser/schema.spec.ts
```diff
@@ -518,18 +524,141 @@ describe('parseSchema', () => {
...
+            oneOf: [
+              { $ref: '#/components/schemas/SchemaA' },
+              { $ref: '#/components/schemas/SchemaB' }
+            ],
+            discriminator
+          },
+          await createTestRefs({
+            schemas: {
+              SchemaA: { properties: { c: { type: 'string' } } },
+              SchemaB: { properties: { c: { type: 'integer' } } }
```
MatKuhr, line 561: I might not fully understand this test, but I think it should be:

```suggestion
              SchemaA: { properties: { discriminatingProperty: { type: 'string' }, c: { type: 'string' } } },
              SchemaB: { properties: { discriminatingProperty: { type: 'string' }, c: { type: 'integer' } } }
```

The discriminator has to be a property of all schemas.

> It is important that all the models mentioned below anyOf or oneOf contain the property that the discriminator specifies.

Don't ask me why this redundancy is necessary, IDK 😄 
I guess it would also be possible for it to have different types, e.g. a string in `SchemaA` and a number in `SchemaB`.

Unfortunately, after checking again, the Azure spec seems to simply use this discriminator incorrectly. If you look at it in the swagger editor you can see that it is simply ignored and the type of the `messages` array is incorrect 😬

MatKuhr: Actually, they fixed this in the [preview 2024-08-01](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/AzureOpenAI/inference/preview/2024-08-01-preview/inference.json)

marikaner: Oh damn, good catch. I read this multiple times yesterday, but still ignored it ;) 
Unfortunately that makes things quite a bit more complicated, but let's see.

marikaner: ok, it doesn't make it more complicated. types should still work as expected.

jjtang1985:
5 cases with `discriminator:` detected from the new [spec](https://github.com/SAP/ai-sdk-js/blob/generate-azure-openai-types/packages/foundation-models/src/azure-openai/spec/inference.yaml), where 2 of them are not relevant (`onYourData...`)

test-packages/test-services-openapi/test-service/schema/discriminator-object-entity-workaround-child-only-type.d.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright (c) 2024 SAP SE or an SAP affiliate company. All rights reserved.
+ *
+ * This is a generated file powered by the SAP Cloud SDK for JavaScript.
+ */
+import type { DiscriminatorObjectEntityWorkaroundParent } from './discriminator-object-entity-workaround-parent';
+/**
+ * Representation of the 'DiscriminatorObjectEntityWorkaroundChildOnlyType' schema.
+ */
+export type DiscriminatorObjectEntityWorkaroundChildOnlyType =
+  DiscriminatorObjectEntityWorkaroundParent;
```
deekshas8, line 11: [q] This is still confusing to me, didn't we discuss that it should not reference the parent, but rather have properties of the parent?

deekshas8 (commented):
Overall LGTM. I still have one question, maybe we can discuss offline.

.changeset/gorgeous-toys-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/openapi-generator': minor
+---
+
+[New Feature] Add support for discriminator properties in"one of" and "all of" schemas. Also add support for incorrect usage of the discriminator property with schemas of type "object".
```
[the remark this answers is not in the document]
marikaner: We don't have a matching rule for this, but the rule for code font does not apply here I think (https://github.com/SAP/cloud-sdk/blob/main/STYLEGUIDE.md#use-code-font).

I will to rephrase it to make it apply better ;) 
```suggestion
[New Feature] Add support for discriminator properties in schemas that include `oneOf` and `anyOf`. Also add support for incorrect usage of the discriminator property with schemas of type `object`.
```

.changeset/gorgeous-toys-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/openapi-generator': minor
+---
+
+[New Feature] Add support for discriminator properties in schemas that include `oneOf` and `anyOf`. Also add support for incorrect usage of the discriminator property with schemas of type `object`.
```
marikaner, line 5:
```suggestion
[New Functionality] Add support for discriminator properties in schemas that include `oneOf` and `anyOf`. Also add support for incorrect usage of the discriminator property with schemas of type `object`.
```

packages/openapi-generator/src/parser/document.ts
```diff
@@ -24,16 +37,67 @@ export async function parseOpenApiDocument(
...
+    schemas
   };
 }
 
+type OpenApiPersistedSchemaWithDiscriminator = OpenApiPersistedSchema & {
+  schema: OpenApiOneOfSchema &
+    Required<Pick<OpenApiOneOfSchema, 'discriminator'>>;
+};
+
+async function sanitizeDiscriminatedSchemas(
```
marikaner, line 57:
```suggestion
// Some specs include incorrect discriminator definitions based on the schema type `object`, that circularly reference their parent type.
async function sanitizeDiscriminatedSchemas(
```

deekshas8 (approved):
LGTM.

PR #5021, opened 2024-09-24
feat: add types to type imports

marikaner:
Adds a new rule to the eslint config that requires type imports to be imported with `type`. This helps with avoiding failures in the cirucular deps check and separate TS and JS imports .


tomfrenken (approved):
LGTM

Issue #4969, opened 2024-09-04
Openapi generator: Could not parse 'application/json;charset=utf-8'


whydrae:
**Describe the bug**
When we generate a client using openapi-generator, we get the following warning: `Could not parse 'application/json;charset=utf-8', because it is not supported. Generation will continue with 'any'. This might lead to errors at runtime.`

It results in type `any` in the generated client.

**To Reproduce**

1. Go to https://api.sap.com/api/ExecutionsAPI/overview and download JSON or YAML api specification.
2. Execute `openapi-generator -i ExecutionsAPI.json -o generated`
3. Find `listExecutions` function and observe the following:

```ts
  listExecutions: (queryParameters?: { tag?: string[] }) =>
    new OpenApiRequestBuilder<any>('get', '/executions', {
      queryParameters,
    }),
```

**Expected behavior**
The function above shall be generated as follows (`any` -> `Execution[]`):
```ts
  listExecutions: (queryParameters?: { tag?: string[] }) =>
    new OpenApiRequestBuilder<Execution[]>('get', '/executions', {
      queryParameters,
    }),
```

**Used Versions:**

- node version via `node -v`: v18.20.4
- npm version via `npm -v`: 10.7.0
- SAP Cloud SDK version you used as dependency: `"@sap-cloud-sdk/openapi-generator": "^3.20.0"`

**Impact / Priority**

Low priority.

PR #5022, opened 2024-09-24
fix: Ignore charset parameter in media-types

shibeshduw:
Closes SAP/cloud-sdk-js#4969.

As we currently do not support handling charset, the changes in this PR simply ignores the provided charset in the spec and only considers the base media-type. So `application/json;charset=utf-8` would be read as `application/json` and underlying schema is parsed accordingly. Currently it is up to the user to provide any `Accept` and `Content-Type` headers as required by the individual API.

### DIFF
--- a/.changeset/plenty-vans-mix.md
+++ b/.changeset/plenty-vans-mix.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/openapi-generator': minor
+---
+
+[Fixed Issue] Ignore charset parameter in media-types given in OpenApi specification.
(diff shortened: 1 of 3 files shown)


.changeset/plenty-vans-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/openapi-generator': minor
+---
+
+[Fixed Issue] Ignore charset parameter in media-types given in OpenApi specification.
```
marikaner, line 5:
```suggestion
[Fixed Issue] Ignore charset parameter in media types given in OpenAPI specification.
```

marikaner (approved):

PR #5028, opened 2024-09-25
chore: Remove deprecated eslint rules in favor of stylistic rules

marikaner:
Some rules are deprecated (e.g. https://typescript-eslint.io/rules/member-delimiter-style/) and are not replaced by ESLint Stylistic. Those rules are removed in later versions of typescript eslint, so I removed them before the dependabot PR (there are other issues as well).


packages/eslint-config/package.json
```diff
@@ -25,16 +25,20 @@
...
     "eslint-plugin-unused-imports": "^4.1.4",
-    "typescript-eslint": "^7.18.0",
-    "@eslint/js": "^9.11.1"
+    "typescript-eslint": "^7.18.0"
   },
   "peerDependencies": {
     "eslint": "^8.0.0"
+  },
+  "devDependencies": {
+    "@stylistic/eslint-plugin-migrate": "^2.8.0"
```
deekshas8, line 42: [q] Is this needed? Briefly looking [here](https://eslint.style/guide/migration#eslint-migrate-plugin) it says we need to add comments at the top of the config file to automatically migrate rules.

test-resources/test/test-util/xsuaa-service-mocks.ts
```diff
@@ -87,7 +87,7 @@ export function mockRefreshTokenGrantCall(
 
 function xsuaaRequestHeaders(additionalHeaders: Record<string, string> = {}) {
   return {
-    Accept: 'application/json',
+    accept: 'application/json',
```
deekshas8, line 90: Scope creep ;)

packages/eslint-config/package.json
marikaner: no not needed anymore, also just realized this and removed it 😊

deekshas8 (approved):
LGTM. Left one question.

test-resources/test/test-util/xsuaa-service-mocks.ts
marikaner: yes 🫣

PR #5034, opened 2024-09-26
chore: Use correct eslint deps

marikaner:
We currently use the eslint import plugin incorrectly without the correct resolver and typescript eslint plugin. According to the documentation this is required: https://github.com/import-js/eslint-plugin-import/tree/main?tab=readme-ov-file#typescript

When adjusting the config many of the rules did report errors. I fixed them all except some that would otherwise fail the circular dependency check (@ZhongpinWang please take a look at this, maybe you remember some hidden issues that were not tested, but I hope we have everything covered by tests).
I also removed the "no-relative-parent-imports" rule because this actually breaks to many things and it has likely never really worked. If we would like to keep that rule we would need to make major changes to the code.

I found this because I thought there was an incorrect report for one specific dependency in the import plugin: https://github.com/import-js/eslint-plugin-import/issues/3069


packages/eslint-config/flat-config.js
```diff
@@ -155,8 +155,12 @@ const flatConfig = [
       ],
       'import/no-self-import': 'error',
       'import/no-cycle': 'error',
-      'import/no-useless-path-segments': 'error',
-      'import/no-relative-parent-imports': 'error',
+      'import/no-useless-path-segments': [
+        'error',
+        {
+          noUselessIndex: true
```
tomfrenken, line 161: I'm surprised this isn't this isn't the default

packages/odata-v4/package.json
```diff
@@ -47,7 +47,7 @@
   },
   "devDependencies": {
     "@sap-cloud-sdk/test-services-odata-v4": "^3.21.0",
-    "nock": "^14.0.0-beta.6",
+    "nock": "^14.0.0-beta.14",
```
tomfrenken, line 50: I assume this change is not directly related to the PR?

PR #5045, opened 2024-09-30
chore: add import/order rule groups

marikaner:
The rule currently uses the default, that ignores how types are treated. Now that we force types to be imported separately I think it is nicer to have this rule as well.

### DIFF
--- a/packages/eslint-config/flat-config.js
+++ b/packages/eslint-config/flat-config.js
@@ -162,7 +162,21 @@ const flatConfig = [
         }
       ],
       'import/export': 'error',
-      'import/order': 'error',
+      'import/order': [
+        'error',
+        {
+          groups: [
+            'builtin',
+            'external',
+            'internal',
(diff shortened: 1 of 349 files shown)


packages/eslint-config/flat-config.js
```diff
@@ -162,7 +162,21 @@ const flatConfig = [
         }
       ],
       'import/export': 'error',
-      'import/order': 'error',
+      'import/order': [
```
tomfrenken, line 165: Having a consistent order makes sense to me, I do not really have an opinion of which order is preferrable, out of curiosity, did you come up with the list yourself or is it some standard you follow here?

tomfrenken (approved):
LGTM

packages/eslint-config/flat-config.js
marikaner: this is the standard documented in the rule itself.

PR #5046, opened 2024-09-30
chore: switch workflows to use node 20 and 22 by default

marikaner:
Node 22 is the current node version, so all non-matrix workflows should run on this version by default.
Github actions seems to only allow `node16` and `node20` according to issues and [docs](https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#runsusing-for-javascript-actions) that I found. I didn't try 22 though.

The tests are now running on all three versions.


.github/workflows/build.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [18]
+        node-version: [20]
```
marikaner, line 28:
```suggestion
        node-version: [18, 20]
```

.github/workflows/blackduck.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - run: git fetch --depth=1
       - uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 22
           cache: 'yarn'
       - run: yarn install --frozen-lockfile
```
ZhongpinWang, line 19: `--ignore-engines` here as well?

marikaner, line 19:
```suggestion
      - run: yarn install --frozen-lockfile --ignore-engines
```

marikaner: yes! good catch

ZhongpinWang (approved):

PR #5079, opened 2024-10-09
fix: add workaround for “greeting never received”

### LINKED ISSUE #5015: sendMail trows Error: Greeting never received for Proxy Type OnPremise
**Describe the bug**

With the commit [add @sap-cloud-sdk/mail-client to send mails](https://github.com/gregorwolf/mtxs-bookshop/commit/677ea1bce8da28e594c13e7fb56cf60d9b4b0b7c) I've added the [Send e-mails](https://sap.github.io/cloud-sdk/docs/js/features/mail-client) functionality to my CAP Multitenant demo app [mtxs-bookshop](https://github.com/gregorwolf/mtxs-bookshop). With the REST Client Test Script [test/use-deployed.http](https://github.com/gregorwolf/mtxs-bookshop/blob/main/test/use-deployed.http) I can trigger the sendMail function in the deployed tenant. My SMTP test server is an instance of https://inbucket.org/ running in docker. When I configure a destination with Proxy Type "Internet" in the Subscriber Subaccount all works fine. But as soon as I switch to a destination with Proxy Type "OnPremise" I get the following error message:

```
OUT [odata|db0f6be2-a8bb-4576-9a67-c7cba89b99b9] - POST /odata/v4/catalog/sendmail
OUT [cat-service|db0f6be2-a8bb-4576-9a67-c7cba89b99b9] - Error: Greeting never received
OUT at SMTPConnection._formatError (/home/vcap/deps/0/node_modules/nodemailer/lib/smtp-connection/index.js:807:19)
OUT at SMTPConnection._onError (/home/vcap/deps/0/node_modules/nodemailer/lib/smtp-connection/index.js:793:20)
OUT at Timeout.<anonymous> (/home/vcap/deps/0/node_modules/nodemailer/lib/smtp-connection/index.js:727:22)
OUT at listOnTimeout (node:internal/timers:573:17)
OUT at process.processTimers (node:internal/timers:514:7) {
OUT code: 'ETIMEDOUT',
OUT command: 'CONN'
OUT }
ERR [cds|db0f6be2-a8bb-4576-9a67-c7cba89b99b9] - Error: Greeting never received
ERR at SMTPConnection._formatError (/home/vcap/deps/0/node_modules/nodemailer/lib/smtp-connection/index.js:807:19)
ERR at SMTPConnection._onError (/home/vcap/deps/0/node_modules/nodemailer/lib/smtp-connection/index.js:793:20)
ERR at Timeout.<anonymous> (/home/vcap/deps/0/node_modules/nodemailer/lib/smtp-connection/index.js:727:22)
ERR at listOnTimeout (node:internal/timers:573:17)
ERR at process.processTimers (node:internal/timers:514:7) {
ERR code: 'ETIMEDOUT',
ERR command: 'CONN',
ERR numericSeverity: 4,
ERR id: '1121184',
ERR level: 'ERROR',
ERR timestamp: 1726745835877
ERR }
```

A trace in SAP Could Connector is attached here: [scc_core.trc.txt](https://github.com/user-attachments/files/17058990/scc_core.trc.txt)

By searching for [Greeting never received](https://github.com/SAP/cloud-sdk-js/issues?q=%22Greeting+never+received%22) I found #3636 which was closed. But I think this might be a regression.

**To Reproduce**

1. Deploy my sample project [mtxs-bookshop](https://github.com/gregorwolf/mtxs-bookshop)
2. Maintain the credentials needed for the REST Client script [test/use-deployed.http](https://github.com/gregorwolf/mtxs-bookshop/blob/main/test/use-deployed.http) in the test/.env file
3. Authenticate by running [use-deployed.http#L3](https://github.com/gregorwolf/mtxs-bookshop/blob/main/test/use-deployed.http#L3)
4. Test with the request [use-deployed.http#L83](https://github.com/gregorwolf/mtxs-bookshop/blob/main/test/use-deployed.http#L83)

**Expected behavior**

E-Mails can be sent via SAP Cloud Connector using On Premise SMTP Server.

**Screenshots**

In the Cloud Connector I can see that a request is received to the virtual host of the SMTP Server:

Virtual Systems Usage

**Used Versions:**

- node version 20.12.2
- npm version 10.8.2
- SAP Cloud SDK version @sap-cloud-sdk/mail-client 3.21.0

**Code Examples**

See above.

**Log file**

See above.

**Impact / Priority**

Affected development phase: Production

Impact: Impaired

marikaner:
There seems to be a bug in the nodemailer library when using a socks proxy, because the library tries to connect to the socket, while the socket is already connected. This leads to nodemailer never receiving a greeting, which is clear, because the greeting was emitted before nodemailer comes into play. We have tried different approaches to work around that and ended up with saving the greeting and resending it until it was heard.
Preferably this would be fixed in the library directly, but as long as there is no action on the [issue](https://github.com/nodemailer/nodemailer/issues/1684), this hopefully is a robust enough fix.

Closes SAP/cloud-sdk-backlog#1246.
Closes #5015

The proposed changeset is at `patch` level right now, could also be `minor` because this changes some internal behavior. Let me know what you think.

### DIFF
--- a/packages/mail-client/src/mail-client.spec.ts
+++ b/packages/mail-client/src/mail-client.spec.ts
@@ -241,18 +241,15 @@ describe('mail client', () => {
     };
 
     it('should create transport/socket, send mails and close the transport/socket', async () => {
-      const { mockSocket, createConnectionSpy } = mockSocketConnection();
+      const { connection, createConnectionSpy } = mockSocketConnection();
       const spyCreateTransport = jest
         .spyOn(nodemailer, 'createTransport')
         .mockReturnValue(mockTransport as any);
-      const spySendMail = jest
-        .spyOn(mockTransport, 'sendMail')
-        .mockImplementation(() => {
-          mockSocket.socket.on('data', () => {});
-        });
+      const spySendMail = jest.spyOn(mockTransport, 'sendMail');
+
       const spyCloseTransport = jest.spyOn(mockTransport, 'close');
-      const spyEndSocket = jest.spyOn(mockSocket.socket, 'end');
-      const spyDestroySocket = jest.spyOn(mockSocket.socket, 'destroy');
+      const spyEndSocket = jest.spyOn(connection.socket, 'end');
+      const spyDestroySocket = jest.spyOn(connection.socket, 'destroy');
 
       await expect(
         sendMail(destination, mailOptions, { sdkOptions: { parallel: false } })
@@ -267,25 +264,22 @@ describe('mail client', () => {
     });
 
     it('should resend greeting', async () => {
-      const { mockSocket } = mockSocketConnection();
+      const { connection } = mockSocketConnection();
       jest
         .spyOn(nodemailer, 'createTransport')
         .mockReturnValue(mockTransport as any);
-      jest.spyOn(mockTransport, 'sendMail').mockImplementation(() => {
-        mockSocket.socket.on('data', () => {});
-      });
 
       const req = sendMail(destination, mailOptions, {
         sdkOptions: { parallel: false }
       });
 
       // The socket emits data for the first time before nodemailer listens to it.
-      // We re-emit the data a second time once there is a listener for it.
-      // In this test we listen for the data event to check that it in fact emits a second time.
+      // We re-emit the data until a listener listened for it.
+      // In this test we listen for the data event to check that we in fact re-emit the message.
       const emitsTwice = new Promise(resolve => {
         let dataEmitCount = 0;
         const collectedData: string[] = [];
-        mockSocket.socket.on('data', data => {
+        connection.socket.on('data', data => {
           dataEmitCount++;
           collectedData.push(data.toString());
           if (dataEmitCount === 2) {
@@ -300,6 +294,22 @@ describe('mail client', () => {
       ]);
       await expect(req).resolves.not.toThrow();
     });
+
+    it('should fail if nodemailer never listens to greeting', async () => {
+      mockSocketConnection();
+
+      jest
+        .spyOn(nodemailer, 'createTransport')
+        .mockReturnValue(mockTransport as any);
+
+      const req = sendMail(destination, mailOptions, {
+        sdkOptions: { parallel: false }
+      });
+      // jest.useFakeTimers();
+      // jest.advanceTimersByTime(1000);
+
+      await expect(req).rejects.toThrow();
+    }, 10000);
 
     it('should throw if greeting (really) was not received', async () => {
       mockSocketConnection(true);
@@ -376,7 +386,7 @@ function mockSocketConnection(fail = false) {
     destroy = jest.fn();
   }
 
-  const mockSocket = {
+  const connection = {
     socket: new MockSocket()
   };
   const createConnectionSpy = jest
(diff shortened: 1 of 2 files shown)


packages/mail-client/src/mail-client.spec.ts
```diff
@@ -258,16 +239,69 @@ describe('mail client', () => {
...
+
+      const req = sendMail(destination, mailOptions, {
+        sdkOptions: { parallel: false }
+      });
+
+      const emitsTwice = new Promise<void>(resolve => {
+        let dataEmitCount = 0;
+        mockSocket.socket.on('data', data => {
+          dataEmitCount += data.toString().includes('220') ? 1 : 0;
+          if (dataEmitCount === 2) {
```
ZhongpinWang, line 286: [q] Can you explain a bit why the on data event is emitted twice and write some comments here?

ZhongpinWang:
SInce we didn't really add functionalities here, I vote for a patch version release.

packages/mail-client/src/mail-client.spec.ts
```diff
@@ -327,3 +361,28 @@ function isValidSocksProxy(proxy) {
...
+
+function mockSocketConnection(fail = false) {
+  class MockSocket extends EventEmitter {
+    end = jest.fn();
+    destroy = jest.fn();
+  }
+
+  const mockSocket = {
+    socket: new MockSocket()
+  };
```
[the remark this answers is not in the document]
marikaner: I left it as it was before, but I guess `mockSocket` is not the correct name it is actually `mockConnection`. This corresponds to what `SocksClient.createConnection()` returns.

Does that clarify this?

packages/mail-client/src/mail-client.ts
```diff
@@ -140,20 +138,59 @@ async function createSocket(
...
+      socket.emit('data', greeting);
+      clearInterval(intervalId);
+      clearTimeout(timeoutId);
+    }
+  }, 1000);
+
+  const timeoutId = setTimeout(() => {
+    clearInterval(intervalId);
+    // default greeting timeout from nodemailer
+  }, 30 * 1000);
```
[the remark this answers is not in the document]
marikaner: great idea, much easier to read!

packages/mail-client/src/mail-client.spec.ts
marikaner: I did it in the test now, let me know if this is what you meant.

packages/mail-client/src/mail-client.spec.ts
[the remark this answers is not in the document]
ZhongpinWang: `mockConnection` makes total sense! It is a bit confusing with `mockSocket`. Leave to you to decide if you wanna rename it.

packages/mail-client/src/mail-client.spec.ts
ZhongpinWang: Exactly what I meant! Thanks for adding this.

ZhongpinWang (approved):

packages/mail-client/src/mail-client.ts
```diff
@@ -140,20 +139,57 @@ async function createSocket(
...
+  const greeting = await retrieveGreeting(socket);
+
+  retry(
+    () => {
+      // resend the greeting message until a listener is attached
+      // note: this is dangerous because there could be another listener that is not the mailer
+      if (socket.listenerCount('data')) {
+        socket.emit('data', greeting);
+        return;
+      }
```
ZhongpinWang, line 182: Wait, I am not sure but don't you need to throw some error here if no listener found? Otherwise will retry actually work?

ZhongpinWang (changes requested):
Wanna clarify retry first.

packages/mail-client/src/mail-client.ts
ZhongpinWang: I think since the return value of `socket.emit()` also tells you if there is a listener or not, can't we write

```ts
if (!socket.emit('data', greeting)) {
  throw new Error('Failed to re-emit event. No data listener found.');
}
```

ZhongpinWang (approved):
LGTM, let's fix this!

PR #5137, opened 2024-10-30
feat: Allow NoAuthentication for OnPremise MAIL destination

### LINKED ISSUE #5136: @sap-cloud-sdk/mail-client: Allow OnPremise MAIL destinations without basic authentication so that no user and password is required
**Is your feature request related to a problem? Please describe.**

OnPremise mail servers might not need any authentication. But with the check in [mail-client/src/mail-client.ts#L82](https://github.com/SAP/cloud-sdk-js/blob/main/packages/mail-client/src/mail-client.ts#L82) it is currently mandatory to provice a user and password by setting the MAIL destination to the Authentication type BasicAuthentication and maintaining User and Password.

**Describe the solution you'd like**

At least for destination of Proxy Type OnPremise the @sap-cloud-sdk/mail-client should allow destinations where Authentication is set to NoAuthentication.

**Describe alternatives you've considered**

The configuration of the internal mailserver can't be changed quickly.

**Impact / Priority**

Affected development phase: Production

Impact: Impaired

**Additional context**

Same customer as in #5015 is impacted.

gregorwolf:
This should fix #5136

### DIFF
--- a/packages/mail-client/src/mail-client.ts
+++ b/packages/mail-client/src/mail-client.ts
@@ -76,6 +76,12 @@ function getCredentials(
   destination: Destination,
   originalProperties: Record<string, any>
 ): { username: string; password: string } {
+  if (
+    destination.proxyType === 'OnPremise' &&
+    destination.authentication === 'NoAuthentication'
+  ) {
+    return { username: '', password: '' };
+  }
   const username = originalProperties['mail.user'];
   const password = originalProperties['mail.password'];
 
--- a/.changeset/angry-ways-kneel.md
+++ b/.changeset/angry-ways-kneel.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/mail-client': minor
+---
+
+Allow NoAuthentication for OnPremise MAIL destination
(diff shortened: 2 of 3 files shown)


packages/mail-client/src/mail-client.ts
```diff
@@ -76,6 +76,12 @@ function getCredentials(
   destination: Destination,
   originalProperties: Record<string, any>
 ): { username: string; password: string } {
+  if (
+    destination.proxyType === 'OnPremise' &&
+    destination.authentication === 'NoAuthentication'
+  ) {
+    return { username: '', password: '' };
```
ZhongpinWang, line 83: Not sure if this is the best way to set empty username and password. According to https://nodemailer.com/smtp/#authentication, if authentication not needed, maybe we should leave `auth` out in `createTransport`.

packages/mail-client/src/mail-client.ts
```diff
@@ -76,6 +76,12 @@ function getCredentials(
   destination: Destination,
   originalProperties: Record<string, any>
 ): { username: string; password: string } {
+  if (
+    destination.proxyType === 'OnPremise' &&
```
ZhongpinWang, line 80: Here, since internet smtp server can also have no username/password, I would leave this out.

PR #5165, opened 2024-11-11
chore: Improve check

deekshas8:
Improves getMessageOrStack function in logger local format


tomfrenken:
How about some utility function like:

```ts
function getErrorMessage(info: any): string {
  // Check if it's an error with a stack trace
  const hasStackTrace = info.stack && info.level === 'error';
  
  if (hasStackTrace && isString(info.stack)) {
    return info.stack;
  }
  
  if (isString(info.message)) {
    return info.message;
  }
  
  return '';
}
```

tomfrenken (approved):
LGTM

PR #5170, opened 2024-11-12
chore: Update ws

deekshas8:
Should address https://github.com/SAP/cloud-sdk-js/security/dependabot/68

The pm2 lib depends on pm2/agent ~2.0.0 (transitive dep) which in yarn lock resolved to 2.0.3. This version had the vulnerable ws@7.5.9. A newer version of pm2/agent was available with the fixed ws version but yarn does not automatically update transitive dependencies.

### DIFF
--- a/.github/actions/changesets-fixed-version-bump/index.js
+++ b/.github/actions/changesets-fixed-version-bump/index.js
@@ -18376,1289 +18376,6 @@ module.exports = (...arguments_) => {
 
 /***/ }),
 
-/***/ 184:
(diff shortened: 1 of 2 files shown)


tomfrenken:
Screenshot 2024-11-12 at 10 09 05
"minor changes"

tomfrenken (approved):
LGTM

tomfrenken:
I think some other dependencies were also bumped looking at the changes, but since our pipeline is still green it should be fine.

PR #5177, opened 2024-11-14
fix: Add nodenext option

deekshas8:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/1256

As explained in the ticket.

### DIFF
--- a/.changeset/chilled-colts-heal.md
+++ b/.changeset/chilled-colts-heal.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/generator-common': minor
+---
+
+[Fixed Issue] Fix ESM client code compilation by allowing the `module` option to be set to `nodenext` in a custom tsconfig.json file.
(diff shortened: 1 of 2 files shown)


.changeset/chilled-colts-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/generator-common': minor
+---
+
+[Fixed Issue] Fix ESM client code compilation by allowing the `module` option to be set to `nodenext` in a custom tsconfig.json file.
```
ZhongpinWang, line 5:
```suggestion
[Fixed Issue] Fix ESM client code compilation by allowing the `module` option to be set to `nodenext` in a custom `tsconfig.json` file.
```

ZhongpinWang (commented):
[q] Just curious, what about other module kinds like `es2022`? https://github.com/microsoft/TypeScript/blob/c043944b1434daa67b66573c66850d51c0a16944/src/server/protocol.ts#L3245

deekshas8:
> [q] Just curious, what about other module kinds like es2022?

`moduleResolution` nodenext requires that module is also nodenext. 
We can add it as an option, but it doesn't fix this specific issue.

PR #5179, opened 2024-11-14
fix: Sequential mail sending aborts after one mail is rejected

deekshas8:
Added error handling so other mails should be sent, even if one of them fails

### DIFF
--- a/packages/mail-client/src/mail-client.ts
+++ b/packages/mail-client/src/mail-client.ts
@@ -194,19 +194,27 @@ async function sendMailInSequential<T extends MailConfig>(
   mailConfigs: T[]
 ): Promise<MailResponse[]> {
   const response: MailResponse[] = [];
-  for (const mailConfigIndex in mailConfigs) {
-    logger.debug(
-      `Sending email ${mailConfigIndex + 1}/${mailConfigs.length}...`
-    );
-    response[mailConfigIndex] = await transport.sendMail({
-      ...mailConfigsFromDestination,
-      ...mailConfigs[mailConfigIndex]
-    });
-    logger.debug(
-      `...email ${mailConfigIndex + 1}/${mailConfigs.length} for subject "${
-        mailConfigs[mailConfigIndex].subject
-      }" was sent successfully.`
-    );
+  for (const [mailConfigIndex, mailConfig] of mailConfigs.entries()) {
+    try {
+      logger.debug(
+        `Sending email ${mailConfigIndex + 1}/${mailConfigs.length}...`
+      );
+      response[mailConfigIndex] = await transport.sendMail({
+        ...mailConfigsFromDestination,
+        ...mailConfig
+      });
+      logger.debug(
+        `...email ${mailConfigIndex + 1}/${mailConfigs.length} with subject "${mailConfig.subject}" was sent successfully.`
+      );
+    } catch (error) {
+      logger.error(
+        `Failed to send email ${mailConfigIndex + 1}/${mailConfigs.length} with subject "${mailConfig.subject}". Error: ${error.message}`
+      );
+      response[mailConfigIndex] = {
+        response: error.response,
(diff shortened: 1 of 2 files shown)


packages/mail-client/src/mail-client.ts
```diff
@@ -217,27 +225,36 @@ async function sendMailInParallel<T extends MailConfig>(
...
+        });
+        logger.debug(
+          `...email ${mailConfigIndex + 1}/${mailConfigs.length} with subject "${
+            mailConfig.subject
+          }" was sent successfully.`
+        );
+        return response;
+      } catch (error) {
+        logger.error(
+          `Failed to send email ${mailConfigIndex + 1}/${mailConfigs.length} for subject "${
```
deekshas8, line 245:
```suggestion
          `Failed to send email ${mailConfigIndex + 1}/${mailConfigs.length} with subject "${
```

.changeset/quick-chefs-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/mail-client': patch
+---
+
+[Fixed Issue] Fix email sending functionality to ensure that emails are sent to all valid addresses. Previously, if an email failed, all subsequent emails were not sent.
```
jjtang1985, line 5: I guess this is indeed a fix instead of a breaking change.

packages/mail-client/src/mail-client.ts
```diff
@@ -217,27 +225,36 @@ async function sendMailInParallel<T extends MailConfig>(
...
-    responses.forEach((_, responseIndex) =>
-      logger.debug(
-        `...email ${responseIndex + 1}/${mailConfigs.length} for subject "${
-          mailConfigs[responseIndex].subject
-        }" was sent successfully`
-      )
-    );
-    return responses;
-  });
+  return Promise.all(promises);
```
ZhongpinWang, line 257: [q] The reject error we could get here is always only the first error right?

packages/mail-client/src/mail-client.ts
```diff
@@ -194,19 +194,27 @@ async function sendMailInSequential<T extends MailConfig>(
...
-    response[mailConfigIndex] = await transport.sendMail({
-      ...mailConfigsFromDestination,
-      ...mailConfigs[mailConfigIndex]
-    });
-    logger.debug(
-      `...email ${mailConfigIndex + 1}/${mailConfigs.length} for subject "${
-        mailConfigs[mailConfigIndex].subject
-      }" was sent successfully.`
-    );
+  for (const [mailConfigIndex, mailConfig] of mailConfigs.entries()) {
```
ZhongpinWang, line 197: [pp] Maybe use map also here to keep them consistent?

tomfrenken (approved):
All in all LGTM, please add a test though, to cover the case where one of the mails fails and the others still are being send.

ZhongpinWang (approved):
Left two comments, but overall LGTM

packages/mail-client/src/mail-client.ts
deekshas8: No, this function will not reject with only the first error (or any unhandled error) because each individual error is handled in the catch and returns a MailResponse type
```
{
   response: error.response,
   rejected: error.rejected
}
```

packages/mail-client/src/mail-client.ts
deekshas8: But it will no longer be sequential then

KavithaSiva: I think, it will still be sequential even if you use a `map` as you don't have an async keyword used in this function block.

packages/mail-client/src/mail-client.ts
deekshas8: Actually I should write a test for parallel case as well.

PR #5195, opened 2024-11-20
feat: Add default transform function for service bindings

tomfrenken:
With this PR, we add a default transform function our users can use to build client credentials-based destinations.

Closes SAP/cloud-sdk-backlog#1258.

### DIFF
--- a/.changeset/neat-parents-search.md
+++ b/.changeset/neat-parents-search.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Add transform function for OAuth2ClientCredentials destinations.
--- a/.changeset/neat-parents-search.md
+++ b/.changeset/neat-parents-search.md
@@ -2,4 +2,4 @@
 '@sap-cloud-sdk/connectivity': minor
 ---
 
-[New Functionality] Add transform function for OAuth2ClientCredentials destinations.
+[New Functionality] Add transform function to create OAuth2ClientCredentials destinations from service bindings.
--- a/packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
+++ b/packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
@@ -58,19 +58,19 @@ export async function transformServiceBindingToDestination(
 
 /**
  * Transforms a service binding to a destination of type OAuth2ClientCredentials.
- * @param serviceBinding - The service binding to transform.
- * @param options - Options used for fetching the destination.
+ * @param service - The service binding to transform.
+ * @param options - Options used to transform the service binding.
  * @returns A promise returning the transformed destination on success.
  */
 export async function transformServiceBindingToClientCredentialsDestination(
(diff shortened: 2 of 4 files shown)


packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -56,6 +56,24 @@ export async function transformServiceBindingToDestination(
...
 }
 
+/**
+ * Transforms a service binding to a destination of type OAuth2ClientCredentials.
+ * @param serviceBinding - The service binding to transform.
+ * @param options - Options used for fetching the destination.
+ * @returns A promise returning the transformed destination on success.
+ */
+export async function transformServiceBindingToClientCredentialsDestination(
+  serviceBinding: Service,
```
deekshas8, line 66:
```suggestion
  service: Service,
```

[pp] To keep it consistent with other functions

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -56,6 +56,24 @@ export async function transformServiceBindingToDestination(
...
+  options?: ServiceBindingTransformOptions & { url: string }
```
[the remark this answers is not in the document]
tomfrenken: No that's a good point, I'll make it optional.
I'll also open a docs PR explaining proper use and assumptions regarding credential structure.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -56,6 +56,24 @@ export async function transformServiceBindingToDestination(
   );
 }
 
+/**
+ * Transforms a service binding to a destination of type OAuth2ClientCredentials.
```
deekshas8, line 60: [pp] Can we also add some hint on when to use this over  `transformServiceBindingToDestination`?

I also find this description from the other function helpful:

```
Convenience function to create a destination from the provided service binding.
 * If a JWT is provided as part of options, the tenant in the JWT is used for client credentials grant, else the provider tenant is used, wherever applicable.
```

.changeset/neat-parents-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Add transform function for OAuth2ClientCredentials destinations.
```
deekshas8, line 5: [q] Maybe a bit more context here? Like this is for custom service binding types that will be transformed to an OAuth2ClientCredentials destination.

deekshas8 (approved):
Some minor suggestions. Apart from that LGTM

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
tomfrenken: Yeah that's a good idea, I wasn't sure if putting it in our docs is enough or if I should put it here as well.
I'll add it!

PR #5203, opened 2024-11-24
chore: Refactor GH action for Public Api Check

shibeshduw:
Small change in the build package for public api check: we now use a regex path pattern for `ignored_path_patterns` instead of a list of paths that was previously used.

Pre-requisite to close: https://github.com/SAP/ai-sdk-js/pull/254


build-packages/check-public-api/index.ts
```diff
@@ -25,10 +25,6 @@ import { getPackages } from '@manypkg/get-packages';
 
 const { readFile, lstat, readdir } = promises;
 
-const localConfigPath = join(
-  process.cwd(),
-  'build-packages/check-public-api/local-config.json'
-);
```
shibeshduw, line 31: This is unused and leftover from a previous PR. It was missed in the review.

ZhongpinWang (approved):

PR #5237, opened 2024-12-05
chore: Stop further publish of mail-client

deekshas8:
Adding private :true to stop releasing mail-client versions


tomfrenken (approved):
LGTM

PR #5246, opened 2024-12-06
chore: Release major `next` 

deekshas8:
The release with tag next is version [3.24.1-20241205152930.0](https://www.npmjs.com/package/@sap-cloud-sdk/http-client/v/3.24.1-20241205152930.0) which is a bit misleading. 

https://www.npmjs.com/package/@sap-cloud-sdk/http-client?activeTab=versions

This PR adjusts build workflow so we release a version 4.0.0-20241207 with the tag next. This PR should be reverted once v4 is released. In case we forget, there is a check in the step so we dont mistakenly release v5 canary.

~~- [ ] If my change will be merged into the `main` branch (for v4), I've updated (V4-Upgrade-Guide.md)[./V4-Upgrade-Guide.md] in case my change has any implications for users updating to SDK v4~~

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -179,7 +179,10 @@ jobs:
         run: |
           date=`date +%Y%m%d%H%M%S`
           rm -f .changeset/*.md
-          cp canary-release-changeset.md .changeset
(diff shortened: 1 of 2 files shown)


tomfrenken (approved):
LGTM

PR #5247, opened 2024-12-06
feat: Support `basePath` in options-per-service for openapi generator - Option 2

KavithaSiva:
Closes SAP/cloud-sdk-backlog#1259.

This approach enables customising the `basePath` both before and after client generation.

It changes the generated client code to also add an additional `basePath` and pass it on to the `OpenApiRequestBuilder`.

```TS
export const CollectionsApi = {
  _defaultBasePath: '/base/path/to/service',
  
  createCollection: (body: CollectionRequest) =>
    new OpenApiRequestBuilder<any>(
      'post',
      '/collections',
      {
        body
      },
      // Forced to use the fully qualified name, as `this._default_basePath` results in Object is possibly 'undefined'.
      CollectionsApi._defaultBasePath
    )
};
```    

A new `setBasePath()` method is also introduced in the `OpenApiRequestBuilder`.
Usage from a user would look like this:

```TS
await CollectionsApi.CreateCollection({
  title,
  embeddingConfig: {},
}).setBasePath(`/some/base/path`).execute(destination);

```


packages/openapi-generator/src/file-serializer/operation.spec.ts
```diff
@@ -102,23 +104,24 @@ describe('serializeOperation', () => {
...
-       * Create a request builder for execution of delete requests to the 'test/{id}' endpoint.
+       * Create a request builder for execution of delete requests to the '/test/{id}' endpoint.
        * @param id - Path parameter.
        * @param headerParameters - Object containing the following keys: resource-group.
        * @returns The request builder, use the \`execute()\` method to trigger the request.
        */
       deleteFn: (id: string, headerParameters?: {'resource-group'?: string}) => new OpenApiRequestBuilder<Record<string, any>>(
         'delete',
-        "test/{id}",
+        "/test/{id}",
```
KavithaSiva, line 119: Drive by change, as all path parameters [should begin](https://swagger.io/specification/#paths-objec) with a `/`  according to Swagger v3.

packages/util/src/remove-slashes.ts
```diff
@@ -11,12 +11,18 @@ export function removeSlashes(path: string): string {
...
 
 /**
  * @internal
  */
 export function removeLeadingSlashes(path: string): string {
-  return path.startsWith('/') ? path.slice(1) : path;
+  while (path.startsWith('/')) {
+    path = path.startsWith('/') ? path.slice(1) : path;
+  }
+  return path;
```
KavithaSiva, line 27: Drive by change, to also account for multiple leading or trailing slashes.

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -39,11 +40,13 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
    * @param pathPattern - Path for the request containing path parameter references as in the OpenAPI specification.
    * @param parameters - Query parameters and or body to pass to the request.
+   * @param basePath - The path to be prefixed to the pathPattern.
    */
   constructor(
     public method: Method,
     private pathPattern: string,
-    private parameters?: OpenApiRequestParameters
+    private parameters?: OpenApiRequestParameters,
+    private basePath?: string | undefined
```
deekshas8, line 49:
```suggestion
    private basePath?: string
```

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -138,6 +141,18 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
 
+  /**
+   * Set the basePath that gets prefixed to the pathPattern before a request.
+   * @param basePath - Base path to be set.
+   * @returns The request builder itself, to facilitate method chaining.
+   */
+  setBasePath(basePath: string): this {
+    const sanitizedBasePath = basePath ? removeSlashes(basePath) : '';
+    // Add a leading slash if basePath is not empty, as all Path Items start with a slash according to Swagger 3
+    this.basePath = sanitizedBasePath ? '/' + sanitizedBasePath : undefined;
```
deekshas8, line 152: [req] Not needed, also the slash convention is for the spec, not our code.

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -175,11 +190,14 @@ export class OpenApiRequestBuilder<ResponseT = any> {
     // Get the innermost curly bracket pairs with non-empty and legal content as placeholders.
     const placeholders: string[] = this.pathPattern.match(/{[^/?#{}]+}/g) || [];
 
-    return placeholders.reduce((path: string, placeholder: string) => {
-      const strippedPlaceholder = placeholder.slice(1, -1);
-      const parameterValue = pathParameters[strippedPlaceholder];
-      return path.replace(placeholder, encodeURIComponent(parameterValue));
-    }, this.pathPattern);
+    return (
+      (this.basePath ?? '') +
```
deekshas8, line 194: [req] Remove slashes here in the base path.

deekshas8 (changes requested):

packages/util/src/remove-slashes.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+describe('removeSlashes', () => {
+  it('removes trailing slashes', () => {
+    expect(removeSlashes('/test/')).toBe('test');
+  });
+  it('removes leading slashes', () => {
+    expect(removeLeadingSlashes('/test')).toBe('test');
+  });
+  it('removes multiple leading slashes', () => {
+    expect(removeLeadingSlashes('///test/')).toBe('test/');
```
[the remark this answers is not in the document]
KavithaSiva: The spec would probably not have multiple leading or trailing slashes, but this is a utility function and the name is misleading if we are only removing one slash at the end or beginning.

packages/openapi-generator/src/file-serializer/api-file.spec.ts
```diff
@@ -124,6 +124,17 @@ describe('api-file', () => {
     expect(apiFile(multipleOperationApi, 'MyServiceName')).toMatchSnapshot();
   });
 
+  it('serializes api file with multiple operations and references and base path', () => {
+    expect(
+      apiFile(
+        multipleOperationApi,
+        'MyServiceName',
+        undefined,
+        '///base/path/to/service///'
```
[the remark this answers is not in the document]
KavithaSiva: Hmm, true, I didn't think about making the call fail. 

I was thinking it should work even if the user gives `basePath` as ` /sap/opu/odata/sap/` or ` sap/opu/odata/sap/` or `sap/opu/odata/sap`. WDY think? Or should we pick a format and ask them to always provide that?
How is it in OData?

packages/openapi-generator/src/file-serializer/api-file.ts
```diff
@@ -16,25 +21,33 @@ import type {
...
 export function apiFile(
   api: OpenApiApi,
   serviceName: string,
-  options?: CreateFileOptions
+  options?: CreateFileOptions,
+  basePath?: string
 ): string {
   const imports = serializeImports(getImports(api, options));
   const apiDoc = apiDocumentation(api, serviceName);
+  const sanitizedBasePath = basePath ? removeSlashes(basePath) : '';
```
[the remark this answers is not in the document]
KavithaSiva: Same as above, I was thinking it should work even if the user gives basePath as  `/sap/opu/odata/sap/` or  `sap/opu/odata/sap/` or `sap/opu/odata/sap`.  

I didn't check if we sanitise after building the path part in the OpenAPIRequestBuilder though, before merging it with the Destination URL.  I can't seem to easily find where this is done.

And should we skip doing anything if the basePath given is just `/` or should we fail?

packages/util/src/remove-slashes.spec.ts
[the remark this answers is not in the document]
KavithaSiva: Discussed in call, name change can be suggested in a separate PR.

packages/util/src/remove-slashes.ts
```diff
@@ -11,12 +11,18 @@ export function removeSlashes(path: string): string {
  * @internal
  */
 export function removeTrailingSlashes(path: string): string {
-  return path.endsWith('/') ? path.slice(0, -1) : path;
+  while (path.endsWith('/')) {
```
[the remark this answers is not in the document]
KavithaSiva: Discussed in call, would handle this function in a separate PR.

.github/actions/check-public-api/index.js
```diff
@@ -74871,13 +74871,19 @@ function removeSlashes(path) {
  * @internal
  */
 function removeTrailingSlashes(path) {
-    return path.endsWith('/') ? path.slice(0, -1) : path;
+    while (path.endsWith('/')) {
```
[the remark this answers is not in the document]
KavithaSiva: Done.

packages/util/src/remove-slashes.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+describe('removeSlashes', () => {
+  it('removes trailing slashes', () => {
+    expect(removeSlashes('/test/')).toBe('test');
+  });
+  it('removes leading slashes', () => {
+    expect(removeLeadingSlashes('/test')).toBe('test');
+  });
+  it('removes trailing slashes', () => {
+    expect(removeTrailingSlashes('/test/')).toBe('/test');
+  });
```
ZhongpinWang, line 16: Just want to comment on these two functions since you touched them. I also used them in my ai-sdk custom destination ticket. I find the naming of these two function however to be very misleading. The functions essentially only remove the **first** or **last** slash, instead of all slashes. The plural form is from my point of view wrong.

Furthermore, all these functions are marked with `@internal` but everything inside util package are exported in `index.ts` and published.

The implementation is 5 years old and `@internal` 3 years old though 😅 @deekshas8

packages/openapi-generator/src/file-serializer/api-file.ts
```diff
@@ -16,25 +21,29 @@ import type {
...
 export function apiFile(
   api: OpenApiApi,
   serviceName: string,
-  options?: CreateFileOptions
+  options?: CreateFileOptions,
+  basePath?: string
 ): string {
   const imports = serializeImports(getImports(api, options));
   const apiDoc = apiDocumentation(api, serviceName);
+  const defaultBasePath = basePath ? removeSlashes(basePath) : undefined;
```
ZhongpinWang, line 37: [req] Why are we removing slashes? Shouldn't path always start with '/'? If user provides a path ending with '/', it should be respected (may not be intentional but still a valid path). https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname

packages/openapi-generator/src/file-serializer/operation.spec.ts
ZhongpinWang: I agree. In general, path should contain at least one '/' at the beginning, even `http://example.com` with no trailing slash has path `/` instead of ''.

I suppose you adapted all occurrences.

packages/util/src/remove-slashes.spec.ts
ZhongpinWang: But if you change it now, it will break whoever used it. It is a huge breaking change unfortunately... Maybe something for v4.

deekshas8: You're right, in general util package functions are exported, so this was probably missed. 
I wasn't aware they are also exposed in `index.ts`, so renaming them would be a breaking change. Something for v4 then.

packages/openapi-generator/src/file-serializer/api-file.ts
deekshas8: [req] I feel we should not be sanitizing paths the user set in the generated code. 
We can remove slashes later when we are building the final url, just like we do in the OData request builder.

packages/openapi-generator/src/file-serializer/api-file.ts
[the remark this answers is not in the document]
deekshas8: Answered [here](https://github.com/SAP/cloud-sdk-js/pull/5247#discussion_r1882121119)

packages/util/src/remove-slashes.ts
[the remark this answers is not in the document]
deekshas8: Please see the comment [here](https://github.com/SAP/cloud-sdk-js/pull/5247#discussion_r1882110603). I thought this was an internal function, but it is also exported from index.ts, in short, its public. So rename can only happen in v4

.changeset/chilled-flowers-freeze.md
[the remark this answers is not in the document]
deekshas8: I think that's fine, both features are documented separately. They will be merged in final changelog anyay.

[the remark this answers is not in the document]
ZhongpinWang: Oh true, they are actually different features.

packages/openapi-generator/src/file-serializer/api-file.ts
KavithaSiva: It's just that we are sanitising it anyway later before execution so, why not make it evident on the generated code too?

ZhongpinWang: True, I also check OData before and it is not sanitized. I would not remove the slashes and the user should know what to provide.

KavithaSiva: Done.

packages/openapi-generator/src/file-serializer/operation.spec.ts
KavithaSiva: Yes, I adapted it everywhere in the this file's tests.

packages/util/src/remove-slashes.ts
KavithaSiva: We decided to do either a name change or implementation change for v4, as this is a breaking change.

packages/openapi/src/openapi-request-builder.spec.ts
```diff
@@ -76,6 +76,27 @@ describe('openapi-request-builder', () => {
     expect(response.data).toBe(dummyResponse);
   });
 
+  it('executeRaw executes a request without parameters and basePath explicitly set', async () => {
+    const requestBuilder = new OpenApiRequestBuilder(
+      'get',
+      '/test'
+    ).setBasePath('//base/path/to/service/');
```
ZhongpinWang, line 83: [q] Why '//'?

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -39,11 +40,13 @@ export class OpenApiRequestBuilder<ResponseT = any> {
    * @param method - HTTP method of the request to be built.
    * @param pathPattern - Path for the request containing path parameter references as in the OpenAPI specification.
    * @param parameters - Query parameters and or body to pass to the request.
+   * @param basePath - The path to be prefixed to the pathPattern.
```
ZhongpinWang, line 43:
```suggestion
   * @param basePath - The path to be prefixed.
```

ZhongpinWang: Similar to my previous comment but not that problematic this time. But still always point out what it is by saying the <name> [method | parameter | function | ...]

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -138,6 +141,16 @@ export class OpenApiRequestBuilder<ResponseT = any> {
     );
   }
 
+  /**
+   * Set the basePath that gets prefixed to the pathPattern before a request.
```
ZhongpinWang, line 145: [req] Similar. Don't use variable name directly. If people don't understand the context they will simply get lost.

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -175,11 +188,14 @@ export class OpenApiRequestBuilder<ResponseT = any> {
     // Get the innermost curly bracket pairs with non-empty and legal content as placeholders.
     const placeholders: string[] = this.pathPattern.match(/{[^/?#{}]+}/g) || [];
 
-    return placeholders.reduce((path: string, placeholder: string) => {
-      const strippedPlaceholder = placeholder.slice(1, -1);
-      const parameterValue = pathParameters[strippedPlaceholder];
-      return path.replace(placeholder, encodeURIComponent(parameterValue));
-    }, this.pathPattern);
+    return (
+      (this.basePath ? removeSlashes(this.basePath) : '') +
```
ZhongpinWang, line 192: [q] Again, why remove slashes, especially the prefix slash? And if basePath is only '/', it should be ignored.

So in short, consider the following cases (we only consider valid paths, if user provide trailing slash or no prefix slash, not our fault):

```txt
basePath: /
pathPattern: /api
=> /api, basePath ignored
```

```txt
basePath: /v1
pathPattern: /api
=> /v1/api, basePath simply prepended
```

ZhongpinWang (commented):
Already looks promising. I suggested some changes.

packages/openapi-generator/src/file-serializer/__snapshots__/api-file.spec.ts.snap
```diff
@@ -46,13 +49,16 @@ exports[`api-file creates an api file with documentation 1`] = `
...
  */
 export const TestApi = {
+  _defaultBasePath: undefined,
   /**
    * Create a request builder for execution of get requests to the 'test' endpoint.
    * @returns The request builder, use the \`execute()\` method to trigger the request.
    */
   getFn: () => new OpenApiRequestBuilder<any>(
     'get',
-    "test"
```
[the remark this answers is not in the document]
KavithaSiva: Oh okay, seems like I did miss a few, will fix.

packages/openapi-generator/src/file-serializer/api-file.spec.ts
```diff
@@ -124,6 +124,17 @@ describe('api-file', () => {
     expect(apiFile(multipleOperationApi, 'MyServiceName')).toMatchSnapshot();
   });
 
+  it('serializes api file with multiple operations and references and base path', () => {
+    expect(
+      apiFile(
+        multipleOperationApi,
+        'MyServiceName',
+        undefined,
+        '/base/path/to/service/'
```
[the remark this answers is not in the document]
KavithaSiva: Why? We discussed earlier [here](https://github.com/SAP/cloud-sdk-js/pull/5247#discussion_r1881635311) that if the user sets 

```
basePath: /sap/opu/odata/sap/  or 
basePath: sap/opu/odata/sap/   or 
basePath: sap/opu/odata/sap
pathPattern: /api
=> sap/opu/odata/sap/api, basePath prepended after cleanup.
```
the request should still work, and [decided](https://github.com/SAP/cloud-sdk-js/pull/5247#discussion_r1882121119) we wouldn't cleanup the input the `basePath` the user provides.
So, this is still a valid test, IMO.

packages/openapi/src/openapi-request-builder.ts
KavithaSiva: Again, I feel it is vague if I just say `Set the basePath that gets prefixed`, can I use {pathPattern} instead?

packages/openapi/src/openapi-request-builder.ts
KavithaSiva: Was added as per Deeksha's review [comment](https://github.com/SAP/cloud-sdk-js/pull/5247#discussion_r1880306891). 

We have to remove a suffix slash if it exists for sure as the `basePath` gets appended to the `pathPattern`.
We allow all these combinations now:
```
basePath: /sap/opu/odata/sap/  
basePath: sap/opu/odata/sap/ 
basePath: /sap/opu/odata/sap     
basePath: sap/opu/odata/sap 
pathPattern: /api
=> sap/opu/odata/sap/api, basePath prepended after cleanup.
```
And the calls only fail if the `basePath` contains an invalid path with multiple slashes.
Do you want only `/sap/opu/odata/sap/` and `/sap/opu/odata/sap` to work instead?

packages/openapi/src/openapi-request-builder.spec.ts
KavithaSiva: Was a leftover, cleaned this now.

packages/openapi/src/openapi-request-builder.ts
KavithaSiva: I also checked that for `executeHttpRequest(..)` that is used internally for executing OpenAPI requests, works if the `pathParameter` is supplied with our without the preceding slash.

```
executeHttpRequest(resolvedDestination,{method: 'get',url: 'sap/opu/odata/sap'}) => works
executeHttpRequest(resolvedDestination,{method: 'get',url: '/sap/opu/odata/sap'}) => works
executeHttpRequest(resolvedDestination,{method: 'get',url: 'sap/opu/odata/sap/'}) => fails for some reason
```

KavithaSiva: I went down the rabbit hole and interesting learning is that according to `WHATWG URL` Standard, which is a specification that defines how browsers and libraries like Axios resolve and parse URLs.

If:

```
baseURL: https://example.org/base/
relativePath: path/to/resource;
fullURL: https://example.org/base/path/to/resource
```
But if:

```
baseURL: https://example.org/base/
relativePath: /path/to/resource
fullURL: https://example.org/path/to/resource
```
So, I suggest we stick to still cleaning the prefix and suffix `/` in the `basePath`.

.changeset/chilled-flowers-freeze.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/openapi': minor
+---
+
+[New Functionality] Introduce `setBasePath()` method on the OpenAPI request builder which allows to set the base path dynamically. This gets prepended with the path parameter for every request to the client.
```
deekshas8, line 5:
```suggestion
[New Functionality] Introduce `setBasePath()` method on the OpenAPI request builder, allowing a custom base path URL to be set for a single request. This base path is prepended to the API path for that request.
```

packages/openapi/src/openapi-request-builder.spec.ts
```diff
@@ -76,6 +76,27 @@ describe('openapi-request-builder', () => {
     expect(response.data).toBe(dummyResponse);
   });
 
+  it('executeRaw executes a request without parameters and basePath explicitly set', async () => {
```
deekshas8, line 79: [req] Please follow the convention here, it should be a complete sentence beginning with `it`

packages/openapi/src/openapi-request-builder.ts
deekshas8: Maybe `to the API path parameter` or just the API path?

packages/openapi/src/openapi-request-builder.ts
deekshas8: [req] Also please mention (also in other places) that this is custom.

packages/util/src/remove-slashes.spec.ts
```diff
@@ -0,0 +1,17 @@
+import {
+  removeSlashes,
+  removeLeadingSlashes,
+  removeTrailingSlashes
+} from './remove-slashes';
+
+describe('removeSlashes', () => {
+  it('removes trailing slashes', () => {
```
deekshas8, line 8:
```suggestion
  it('removes both leading and trailing slashes', () => {
```

packages/util/src/remove-slashes.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+    expect(removeSlashes('/test/')).toBe('test');
+  });
+  it('removes leading slashes', () => {
```
deekshas8, line 11:
```suggestion
  it('removes leading slash', () => {
```

packages/util/src/remove-slashes.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+    expect(removeLeadingSlashes('/test')).toBe('test');
+  });
+  it('removes trailing slashes', () => {
```
deekshas8, line 14:
```suggestion
  it('removes trailing slash', () => {
```

test-packages/test-services-openapi/no-schema-service/default-api.d.ts
```diff
@@ -9,6 +9,7 @@ import { OpenApiRequestBuilder } from '@sap-cloud-sdk/openapi';
  * This API is part of the 'no-schema-service' service.
  */
 export declare const DefaultApi: {
+  _defaultBasePath: string;
```
deekshas8, line 12: [q] Why is it string here and undefined else where?

deekshas8 (changes requested):

.changeset/chilled-flowers-hide.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-cloud-sdk/openapi-generator': minor
+'@sap-cloud-sdk/openapi': minor
+'@sap-cloud-sdk/util': minor
+---
+
+[New Functionality] Support `basePath` option in the `options-per-service.json` file in the OpenAPI generator. It enables the base path to be prepended to the path parameter for every request of the generated client.
```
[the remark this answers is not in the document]
KavithaSiva: Adding API path parameter.

.changeset/chilled-flowers-freeze.md
KavithaSiva: Good catch!

test-packages/test-services-openapi/no-schema-service/default-api.d.ts
KavithaSiva: Good catch, I have no idea, why, let me check.

packages/openapi/src/openapi-request-builder.spec.ts
KavithaSiva: Seems like other tests in the file also have this problem, will fix.

packages/openapi/src/openapi-request-builder.ts
KavithaSiva: I like API path parameter, will use that instead.

packages/openapi/src/openapi-request-builder.ts
KavithaSiva: Used API path pattern instead of the variable name.

test-packages/test-services-openapi/no-schema-service/default-api.d.ts
KavithaSiva: I think this is expected, because in this example, we explicitly set `basePath` to a custom value.
So, the corresponding `.ts` file entry is `_defaultBasePath: '/base/path/to/service'`.

In all the other tests, the `basePath` value is `undefined` and hence the declaration files also have the same value.
Is this an issue?

deekshas8: Oops, I mistook this for a .ts file. Didn't realize its a declaration file. Then please ignore my comment,

packages/openapi-generator/src/file-serializer/__snapshots__/api-file.spec.ts.snap
```diff
@@ -72,33 +78,78 @@ import type { QueryParameterType, RefType, ResponseType } from './schema';
...
+  /**
+   * Create a request builder for execution of get requests to the '/test/{id}' endpoint.
+   * @param id - Path parameter.
+   * @param queryParameters - Object containing the following keys: queryParam.
+   * @param headerParameters - Object containing the following keys: headerParam.
+   * @returns The request builder, use the \`execute()\` method to trigger the request.
+   */
+  getFn: (id: string, queryParameters: {'queryParam': QueryParameterType}, headerParameters?: {'headerParam'?: string}) => new OpenApiRequestBuilder<string>(
+    'get',
+    "/test/{id}",
```
ZhongpinWang, line 133: Replying to my [other comment](https://github.com/SAP/cloud-sdk-js/pull/5247/files#r1884047673) here, why no trailing slash in base path. Because `/base/path/to/service/` + `/test/{id}` is wrong. Although the test will pass because we don't care what user provides as a base path, it can be a random string. But to make the test really "correct" I would leave the trailing slash out.

packages/openapi-generator/src/file-serializer/api-file.spec.ts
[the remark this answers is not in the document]
ZhongpinWang: Answered [here](https://github.com/SAP/cloud-sdk-js/pull/5247/files#r1888696768).

packages/openapi/src/openapi-request-builder.ts
ZhongpinWang: Alright, if it is the same behavior as the OData client, I am fine with it.

KavithaSiva: So if the user gives `_defaultBasePath` with `/` in the beginning it will be generated that way, but always before constructing the full url, we remove the `/` from the beginning and end of the base Path while appending it to the path pattern. Axios takes care of appending it properly, as mentioned [above](https://github.com/SAP/cloud-sdk-js/pull/5247#discussion_r1884134366).

packages/openapi/src/openapi-request-builder.spec.ts
```diff
@@ -76,7 +76,28 @@ describe('openapi-request-builder', () => {
...
+      'get',
+      '/test'
+    ).setBasePath('/base/path/to/service');
+    const response = await requestBuilder.executeRaw(destination);
+    expect(httpClient.executeHttpRequest).toHaveBeenCalledWith(
+      sanitizeDestination(destination),
+      {
+        method: 'get',
+        middleware: [],
+        url: 'base/path/to/service/test',
```
ZhongpinWang, line 90: @tomfrenken Fyi since it relates to some of your code changes like two years ago. Here without a leading slash is the same behaviour as in OData, where remove (leading/trailing) slashes functions were used at almost each and every step to ensure no slashes at all. But it feels kind inconsistent to me because without setting the base path, we use and probably also recommend adding a leading slash at the path.

ZhongpinWang (approved):
Overall LGTM. Thanks for this nice contribution.

Be sure to add this to v4.

PR #5308, opened 2024-12-20
fix: Tests in v4

KavithaSiva:
Fixing the `mock-fs` version to `5.3.0` as the latest version `5.4.1` causes some tests to fail.
(Ref: [Dependency updates that fail in v3](https://github.com/SAP/cloud-sdk-js/pull/5242))

- ~[ ] If my change will be merged into the `main` branch (for v4), I've updated (V4-Upgrade-Guide.md)[./V4-Upgrade-Guide.md] in case my change has any implications for users updating to SDK v4~


KavithaSiva:
After a discussion with @deekshas8, decided to close this and clean up the `main` branch by restoring it from a backup instead.

PR #5350, opened 2025-01-10
chore: Update mail client deprecation notes

jjtang1985:
cherry picked from v3: https://github.com/SAP/cloud-sdk-js/pull/5347

### DIFF
--- a/packages/mail-client/README.md
+++ b/packages/mail-client/README.md
@@ -6,6 +6,9 @@
 > [!caution]
 > This package has been deprecated and is no longer maintained.
 > Use at your own risk.
+>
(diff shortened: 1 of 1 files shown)


ZhongpinWang (approved):

deekshas8:
@jjtang1985 If I understood the discussion before, this package will be removed in v4 completely. Do we even need this note here?

jjtang1985:
> If I understood the discussion before, this package will be removed in v4 completely. Do we even need this note here?

Yes, @deekshas8.
Removing it completely was what we aligned in the meeting previously.
However, we need a final confirmation with the manager.
In order to be safe, I just updated it.

PR #5361, opened 2025-01-14
feat(v4)!: Respect precision for Odata V4 type `Edm.DateTimeOffset`

KavithaSiva:
Closes SAP/cloud-sdk-backlog#1261.


.changeset/funny-shrimps-visit.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-cloud-sdk/generator': minor
+'@sap-cloud-sdk/odata-v4': minor
+'@sap-cloud-sdk/odata-common': minor
```
KavithaSiva, line 4: [q] Should I write `major` here? What is the correct heading for the PR then? `feat` needs a changeset with `minor`.

packages/odata-v4/src/de-serializers/default-de-serializers.ts
```diff
@@ -14,6 +14,7 @@ import {
   serializeToTime,
   deserializeDateToMoment
 } from './converters';
+import type { Moment } from 'moment';
```
deekshas8, line 17: I think this is not the correct way to import moment.

test-packages/test-services-e2e/v4/test-service/TestEntity50PropApi.ts
```diff
@@ -603,7 +605,8 @@ export class TestEntity50PropApi<
         DECIMAL_PROPERTY_2: fieldBuilder.buildEdmTypeField(
           'DecimalProperty2',
           'Edm.Decimal',
-          true
+          true,
+          9
```
deekshas8, line 609: [q] What is the max allowed precision as per odata? Just want to make sure we're testing the correct values. If I remember cds allows only till 7

test-packages/test-services-odata-v4/test-service/TestEntityApi.ts
```diff
@@ -500,7 +504,8 @@ export class TestEntityApi<
         DURATION_PROPERTY: fieldBuilder.buildEdmTypeField(
           'DurationProperty',
           'Edm.Duration',
-          true
+          true,
+          0
```
deekshas8, line 508: [q] Is there an example where nothing is set? Couldn't find it

deekshas8 (changes requested):
I will probably have another look, but left some initial comments already,

packages/odata-v4/src/de-serializers/default-de-serializers.ts
KavithaSiva: I now removed the import as I am `moment` seems to be accessible without it in this file.

But, I don't understand how this works, and why we explicitly use `import { Moment } from 'moment'` in some places but use it directly in some.

test-packages/test-services-e2e/v4/test-service/TestEntity50PropApi.ts
KavithaSiva: From OData spec:
>For a decimal value: the maximum number of significant decimal digits of the property’s value; it MUST be
a positive integer.
>For a temporal value (datetime-with-timezone-offset, duration, or time-of-day): the number of decimal
places allowed in the seconds portion of the value; it MUST be a non-negative integer between zero and
twelve. 

Currently, `precision` values get generated for both temporal and decimal values.
But, I only use it for formatting `DateTimeOffset` fields.
I am okay with generating them for decimal fields as well, eventhough we don't use it, do you feel otherwise?

deekshas8: I think it would be nicer if we generate it for types where we actually support it. otherwise it might be confusing.

deekshas8: I would only update our test client code for cases where we actually support the behavior.

test-packages/test-services-odata-v4/test-service/TestEntityApi.ts
KavithaSiva: yes, for e.g, in the same file:
Screenshot 2025-01-27 at 11 35 50

packages/odata-v4/src/de-serializers/default-de-serializers.ts
deekshas8: I think it's likely due to moment being a global namespace here (which allows using it without import statement). But for consistency we should still import it.

test-packages/test-services-e2e/v4/test-service/TestEntity50PropApi.ts
KavithaSiva: It was updated automatically now during generation, as we always add `precision` if it's available in the EDMX file.
Let me check, how to restrict this during generation only for temporal fields.

packages/odata-v4/src/de-serializers/default-de-serializers.ts
deekshas8: If you're using moment only for its type then use:

import type moment else import moment

packages/odata-common/src/de-serializers/de-serializers.ts
```diff
@@ -152,9 +152,13 @@ export function createValueDeserializer<DeSerializersT extends DeSerializers>(
  */
 export function createValueSerializer<DeSerializersT extends DeSerializers>(
   deSerializers: DeSerializersT
-): <EdmT>(value: any, edmType: EdmT) => any {
-  return (value, edmType) => {
+): <EdmT>(value: any, edmType: EdmT, precision?: number) => any {
+  return (value, edmType, precision) => {
     const serialize = deSerializers[edmType as any]?.serialize;
+    // Check if the edmType is specifically Edm.DateTimeOffset as we can have precision, scale property defined
+    if (edmType === 'Edm.DateTimeOffset' && precision) {
```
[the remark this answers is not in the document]
KavithaSiva: >[req] If precision is 0 (which is a valid value) this will be falsy.
>Also, even if we just send precision without any checks, does it break anything? With JS it just ignores the extra params.
Addressed these comments removing the additional check and passing both parameters always.

>[q] Can we double check if it's only DateTimeOffset? or also valid for TimeStamp?
From the OData spec, this precision  is to be considered for datetime-with-timezone-offset, duration, or time-of-day as well, so we would have to modify the serialisers for `Edm.Duration` and 'Edm.TimeOfDay' too.
I will add a follow-up BLI item for this case as we would also have to check if CAP is facing compatability issues for `Edm.TimeOfDay` type too.

test-packages/test-services-e2e/v4/test-service/TestEntity50PropApi.ts
KavithaSiva: I have restricted to generate `precision` into  `buildEdmTypeField ` only if the EDM type is DateTimeOffset.

packages/odata-common/src/de-serializers/de-serializers.ts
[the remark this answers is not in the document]
KavithaSiva: Follow-up BLI: https://github.com/SAP/cloud-sdk-backlog/issues/1264

packages/odata-v4/src/de-serializers/converters.ts
```diff
@@ -42,8 +42,25 @@ export function deserializeDateTimeOffsetToMoment(
...
+export function serializeToDateTimeOffset(
+  value: moment.Moment,
+  precision?: number
+): string {
+  // For temporal values, OData specification restricts precision values to a non-negative integer between 0 and 12
+  if (precision !== undefined && !validatePrecision(precision)) {
+    throw new Error(
+      `Provided precision value: ${precision} is invalid. Precision must lie between 0 and 12`
+    );
+  }
```
KavithaSiva, line 54: @jjtang1985 We also decided to throw an error here (different from previous behaviour) as OData specification only permits precision be a non-negative integer between zero and twelve.

CAP seems to restrict the precision for Timestamp CDS(`Edm.DateTimeOffset` with Precision="7") type too, but we are are only enforcing what OData spec mentions.

Please let me know, if this is okay.

.changeset/funny-shrimps-visit.md
tomfrenken: Yes major is appropriate as this will be a breaking change and will be part of our next major release.

packages/odata-v4/src/de-serializers/default-de-serializers.ts
```diff
@@ -55,7 +56,8 @@ const defaultDeSerializersRaw: DefaultDeSerializers = {
   },
   'Edm.DateTimeOffset': {
     deserialize: deserializeDateTimeOffsetToMoment,
-    serialize: serializeToDateTimeOffset,
+    serialize: (value: moment.Moment, precision?: number) =>
```
tomfrenken, line 59: Interestingly enough, in the diff here in the UI it looks like it's moment.Moment, but you've already made the improvement of only importing { Moment } ...

packages/odata-common/src/selectable/field-builder.ts
```diff
@@ -142,7 +146,8 @@ export class FieldBuilder<
 
     return new ctor(fieldName, this.fieldOf, edmType, this.deSerializers, {
```
tomfrenken, line 147: Kind of bad naming here, but I guess it's always been there ...

tomfrenken (approved):
LGTM, but adjust the changeset.

PR #5429, opened 2025-01-30
 feat: Support resolve external option

tomfrenken:
Cherry-pick from #5360


packages/openapi-generator/src/options/options.ts
```diff
@@ -48,5 +52,12 @@ export const cliOptions = {
     type: 'string',
     default: '',
     hidden: true
+  },
+  resolveExternal: {
+    describe:
+      'By default, external $ref pointers will be resolved. If set to false, external $ref pointers will be ignored.',
+    type: 'boolean',
+    default: true,
+    hidden: true
```
deekshas8, line 61: Since it's hidden, I will not add a changeset

packages/openapi-generator/src/parser/document.spec.ts
```diff
@@ -54,7 +54,7 @@ describe('parseOpenApiDocument()', () => {
         packageName: '@sap/cloud-sdk-openapi-test-service',
         directoryName: 'test-service'
       },
-      { strictNaming: false, schemaPrefix: '' }
+      { strictNaming: false, schemaPrefix: '', resolveExternal: true }
```
deekshas8, line 57: Maybe as an improvement we can make as a follow-up, define defaultOptions for the function

```
defaultOptions = { strictName: true, schemaPrefix: '', resolveExternal: true }

export async function parseOpenApiDocument(
  fileContent: OpenAPIV3.Document,
  serviceOptions: ServiceOptions,
  options: ParserOptions = defaultOptions
): Promise<OpenApiDocument> {
```

packages/openapi-generator/src/parser/document.ts
```diff
@@ -35,7 +35,9 @@ export async function parseOpenApiDocument(
   options: ParserOptions
 ): Promise<OpenApiDocument> {
   const clonedContent = JSON.parse(JSON.stringify(fileContent));
-  const document = (await parseBound(clonedContent)) as OpenAPIV3.Document;
+  const document = (await SwaggerParser.parse(
```
deekshas8, line 38: [q] Why?

tomfrenken: parseBound was a workaround that is not necessary anymore

PR #5440, opened 2025-02-04
feat!: Enable destination cache by default

KavithaSiva:


.changeset/chatty-icons-accept.md
```diff
@@ -0,0 +1,9 @@
+---
```
KavithaSiva, line 1: I was not sure if I had to update the upgrade guide as the user doesn't have to take any steps, but just know that caching has been enabled by default.

We are updating documentation and also release notes. Please let me what you think.

deekshas8: I think we should document the new default behavior it as it is one of the breaking changes.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -158,17 +159,76 @@ function assertMockUsed(mock: nock.Scope, used: boolean) {
...
     mockServiceToken();
+    await destinationCache.clear();
   });
 
   afterEach(() => {
     nock.cleanAll();
     jest.clearAllMocks();
   });
 
+  it('should use cache by default', async () => {
```
deekshas8, line 174: [req] The destination-cache.spec.ts already checks all behavior regarding `useCache: true`. We should update those tests (not pass this property anymore) and test that they are still green. There are also tests for the other behavior where nothing should be cached. 

Please remove these tests here.

.changeset/chatty-icons-accept.md
KavithaSiva: Documented in upgrade guide.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -326,18 +330,20 @@ describe('JWT type and selection strategies', () => {
...
+    await destinationServiceCache.clear();
   });
 
   afterEach(() => {
     nock.cleanAll();
     jest.clearAllMocks();
   });
 
-  it('should fetch all subscriber destinations', async () => {
+  it('should fetch all subscriber destinations and cache destinations', async () => {
```
KavithaSiva, line 346: Added these `caching` tests here as `getAllDestinationsFromDestinationService()` doesn't allow passing a destination service url, and I did not find an easy way to test this function with the existing setup in `destination-service-cache.spec.ts`

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -121,6 +121,11 @@ function createDestinationFetchLogs(
 export async function getAllDestinationsFromDestinationService(
   options: AllDestinationOptions = {}
 ): Promise<DestinationWithoutToken[]> {
+  options = {
+    ...options,
+    // Enable caching by default
+    useCache: options.useCache === undefined ? true : options.useCache
```
deekshas8, line 127: options.useCache ?? true

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -343,7 +350,6 @@ describe('destination cache', () => {
       );
       const actual = await getDestination({
         destinationName: destName,
-        useCache: true,
```
deekshas8, line 346: [req] There is also one left in line 325

deekshas8: [req] Can you check the test `disables the cache by default`? This should not work anymore

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -210,11 +205,26 @@ describe('destination cache', () => {
...
+    it('does not use cache if caching is explicitly disabled', async () => {
+      mockDestinationsWithSameName();
+
+      await getDestination({
+        destinationName: 'SubscriberDest',
+        jwt: subscriberUserToken,
+        useCache: false
+      });
+
+      const c1 = await getSubscriberCache('tenant');
```
deekshas8, line 217: [req] This test is technically incorrect, because when you set a user jwt and no isolationStrategy, the `isolationStrategy` will be `tenant-user`. So even if something was cached, this line would still return undefined. Either explicitly set the strategy or call getCache methods with tenant-user

deekshas8: Also check if you really need this test, since there is another which tests for false behavior. Also please consider the ordering of tests and where you place them. Place similar tests together.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -326,18 +330,20 @@ describe('JWT type and selection strategies', () => {
 });
 
 describe('call getAllDestinations with and without subscriber token', () => {
-  beforeEach(() => {
+  beforeEach(async () => {
     mockServiceBindings();
     mockVerifyJwt();
     mockServiceToken();
+    await destinationCache.clear();
+    await destinationServiceCache.clear();
```
deekshas8, line 338: [q] Do you really need an await here?

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -347,19 +353,81 @@ describe('call getAllDestinations with and without subscriber token', () => {
...
+    expect(cacheSpy).not.toHaveBeenCalled();
+    expect(retrieveSpy).not.toHaveBeenCalled();
+
+    // Reset spies and mocks for second call
+    jest.clearAllMocks();
+    nock.cleanAll();
+    mockGetAllSubscriber();
+    mockGetAllProvider();
+
+    // Second call - should fetch again since cache was not used previously
```
deekshas8, line 422: The second setup is not really needed since we already estbalished: expect(cacheSpy).not.toHaveBeenCalled();
expect(retrieveSpy).not.toHaveBeenCalled();

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
KavithaSiva: The test was set up with the wrong values to cache, hence was working without failing. 
I have fixed this now and added tests for both caching by default and non-caching case.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
KavithaSiva: You are right, I didn't see that `destinationServiceCache.clear()` didn't return a Promise.

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
KavithaSiva: Removed the faulty test and added a test for disabling cache in the right group.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
KavithaSiva: Changed the test as per your [suggestion](https://github.com/SAP/cloud-sdk-js/pull/5440#discussion_r1957898898).

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
KavithaSiva: Done.

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -296,15 +287,21 @@ describe('destination cache', () => {
 
     const destName = destinationOne.name!;
 
-    it('disables the cache by default', async () => {
+    it('uses the cache by default', async () => {
```
deekshas8, line 290: [pp] I would place this test at the top.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -37,6 +38,8 @@ import {
   alwaysSubscriber,
   subscriberFirst
 } from './destination-selection-strategies';
+import { destinationServiceCache } from './destination-service-cache';
+import { destinationCache } from './destination-cache';
```
deekshas8, line 42: [req] Instead of importing this, you can change the fetchDestination function and set usecache false there. I believe in the original tests, that was the case we were testing under anyway.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -91,6 +94,8 @@ const parsedSubscriberDestination: DestinationWithoutToken = {
   url: 'http://subscriber.com'
 };
 
+const mockedDestinationUrlSubaccountDestinations =
+  'https://destination.example.com/destination-configuration/v1/subaccountDestinations';
```
deekshas8, line 98: [req] Can you import the `destinationServiceUri` from test util. I wasn't sure looking at it where the url came from

deekshas8 (commented):
Left some comments

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -296,15 +287,21 @@ describe('destination cache', () => {
...
       await destinationCache.cacheRetrievedDestination(
-        { user_id: 'user', zid: 'tenant' },
+        {
```
deekshas8, line 292: [q] Why not use the `providerUserToken` here?

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
```diff
@@ -409,6 +403,24 @@ describe('destination cache', () => {
         "Could not build destination cache key. Isolation strategy 'tenant-user' is used, but tenant id or user id is undefined in JWT."
       );
     }, 15000);
+
+    it('disables the cache if explicitly specified', async () => {
+      await destinationCache.cacheRetrievedDestination(
+        {
+          iat: 1739795395,
```
deekshas8, line 410: [q] Same question as above

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
KavithaSiva: Done.

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
KavithaSiva: Done.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
KavithaSiva: I imported `destinationServiceUri`, but the path `/destination-configuration/v1/subaccountDestinations` still has to be appended as I didn't find the path in constants.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -326,18 +335,20 @@ describe('JWT type and selection strategies', () => {
...
+    destinationServiceCache.clear();
   });
 
   afterEach(() => {
     nock.cleanAll();
     jest.clearAllMocks();
   });
 
-  it('should fetch all subscriber destinations', async () => {
+  it('should fetch all subscriber destinations and cache destinations', async () => {
```
deekshas8, line 351:
```suggestion
  it('should fetch and cache all subscriber destinations', async () => {
```

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -353,13 +364,36 @@ describe('call getAllDestinations with and without subscriber token', () => {
...
     );
+    expect(
+      destinationServiceCache.retrieveDestinationsFromCache(
+        mockedDestinationUrlSubaccountDestinations,
+        decodeJwt(subscriberUserToken)
+      )
+    ).toEqual([parsedSubscriberDestination]);
+  });
+
+  it('should fetch all subscriber destinations and not cache destinations', async () => {
```
deekshas8, line 381:
```suggestion
  it('should fetch all subscriber destinations without caching them', async () => {
```

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -91,6 +95,9 @@ const parsedSubscriberDestination: DestinationWithoutToken = {
   url: 'http://subscriber.com'
 };
 
+const mockedDestinationUrlSubaccountDestinations =
+  destinationServiceUri +
```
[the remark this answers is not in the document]
KavithaSiva: Done.

packages/connectivity/src/scp-cf/destination/destination-cache.spec.ts
KavithaSiva: Done.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
KavithaSiva: I have disabled caching for the first set of tests, still need it though for tests that cover fetching all destinations.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -326,18 +333,20 @@ describe('JWT type and selection strategies', () => {
 });
 
 describe('call getAllDestinations with and without subscriber token', () => {
-  beforeEach(() => {
+  beforeEach(async () => {
```
deekshas8, line 336: [q] Same here, await destinationCache.clear(); still needed?

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
```diff
@@ -121,6 +121,11 @@ function createDestinationFetchLogs(
 export async function getAllDestinationsFromDestinationService(
   options: AllDestinationOptions = {}
```
deekshas8, line 122:
```suggestion
  options: AllDestinationOptions = { useCache: true }
```

Tom gave a nice suggestion, this way you dont have to write:

```
options = {
    ...options,
    // Enable caching by default
    useCache: options.useCache ?? true
  };
```

deekshas8 (commented):
Sorry for one more round. After this we're ready to merge

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
```diff
@@ -158,7 +165,7 @@ function assertMockUsed(mock: nock.Scope, used: boolean) {
 }
 
 describe('JWT type and selection strategies', () => {
-  beforeEach(() => {
+  beforeEach(async () => {
```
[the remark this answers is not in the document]
KavithaSiva: Yeah, I just missed clearing it up, done now.

packages/connectivity/src/scp-cf/destination/destination-accessor-provider-subscriber-lookup.spec.ts
KavithaSiva: Yeah, I just missed clearing it up, done now.

KavithaSiva:
> Sorry for one more round. After this we're ready to merge

No issues.

packages/connectivity/src/scp-cf/destination/destination-accessor.ts
KavithaSiva: So, I had initially added this, but this default is only used, if the user doesn't explicitly provide any `options`.
So, code right now always adds `useCache` option regardless of if the user provided `options` or not.

PR #5488, opened 2025-02-14
chore: Automate v4 canary release

KavithaSiva:
Creating an workflow, that deploys v4 canary release with next tag nightly.

### DIFF
--- a/.github/workflows/automatic-v4-canary-release.yml
+++ b/.github/workflows/automatic-v4-canary-release.yml
@@ -1,0 +1,20 @@
+name: automatic-v4-canary-release
+
+on:
+  schedule:
+    - cron: '0 1 * * *'
+  workflow_dispatch:
+
+jobs:
+  trigger-v4-canary-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger v4 release from main branch
+        run: |
+          curl -L \
+           -X POST \
+           -H "Accept: application/vnd.github+json" \
+           -H "Authorization: Bearer ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}" \
+           -H "X-GitHub-Api-Version: 2022-11-28" \
+           https://api.github.com/repos/SAP/cloud-sdk-js/actions/workflows/build.yml/dispatches \
+           -d '{"ref":"main", "inputs": {"canary-release": "true"}}'


.github/workflows/automatic-v4-canary-release.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  trigger-v4-canary-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger v4 release from main branch
+        run: |
+          curl -L \
+           -X POST \
+           -H "Accept: application/vnd.github+json" \
+           -H "Authorization: Bearer ${{ secrets.GH_CLOUD_SDK_JS_ADMIN_WRITE_TOKEN }}" \
```
KavithaSiva, line 17: Not sure if this token has permission to trigger workflows, I don't have admin access in this repo, to manually check.

.github/workflows/automatic-v4-canary-release.yml
```diff
@@ -0,0 +1,20 @@
...
+           -H "X-GitHub-Api-Version: 2022-11-28" \
```
deekshas8, line 18: [q] What it this for?

.github/workflows/automatic-v4-canary-release.yml
```diff
@@ -0,0 +1,20 @@
...
+           https://api.github.com/repos/SAP/cloud-sdk-js/actions/workflows/build.yml/dispatches \
+           -d '{"ref":"main", "inputs": {"canary-release": "true"}}'
```
deekshas8, line 20: [req] We should not set canary-release to true. This would mean e2e-tests, checks etc are skipped. If it runs as scheduled, we check for a diff and then release. This behavior should be same

KavithaSiva: ah okay, so you mean to just run the `build.yml` from the `main` branch first and then try to do the canary release? 
But, how can we check for a `diff` then with the previous `next` ?

deekshas8: Maybe it makes sense to do it like we discussed before

deekshas8: > But, how can we check for a diff then with the previous next 
The worfklow already does that

KavithaSiva:
Closed in favour of https://github.com/SAP/cloud-sdk-js/pull/5501

PR #5501, opened 2025-02-17
chore: Automate v4 canary nightly release 

KavithaSiva:
Creating an workflow, that deploys v4 canary release with `next` tag nightly.
The workflow is essentially a copy paste of parts of the `build` worflow, with the `main` branch being checked out.


.github/workflows/next-build-nightly.yml
```diff
@@ -0,0 +1,155 @@
...
+          registry-url: 'https://registry.npmjs.org'
+      - run: yarn install --frozen-lockfile --ignore-engines
+      - name: Canary Release
+        run: |
+          date=`date +%Y%m%d%H%M%S`
+          rm -f .changeset/*.md
+          project_version=$(cat package.json | jq '.version' | tr -d '"')
+          echo The project version is: $project_version
+          [[ $project_version == 4* ]] && cp canary-release-changeset.md .changeset          
+          [[ $project_version == 3* ]] && cp major-release-changeset.md .changeset
```
deekshas8, line 149: [req] I dont see a major-release-changeset.md in this PR

KavithaSiva: I swapped the correct changeset files for the versions. But, I think in this workflow, we don't need the check for version 3 right?
Where do these changesets get published in `npm`? I wasn't easily able to find it now.

deekshas8: > I swapped the correct changeset files for the versions. 

I didn't understand this.

> I think in this workflow, we don't need the check for version 3 right
The project version will always be 3 here, as thats the version in package.json. Hence we use the major release changeset so a version 4 next is released. We probably dont need a check for version 4. But I would test this first, else leave it here.

.github/workflows/next-build-nightly.yml
```diff
@@ -0,0 +1,155 @@
...
+      - run: yarn test:self
+      - run: yarn test:build-packages
+      - run: yarn test:type
+      - if: failure()
+        name: Slack Notify
+        uses: rtCamp/action-slack-notify@v2.3.2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          SLACK_USERNAME: SDK Pipeline Bot
+          SLACK_TITLE: Build tests for v4 ${{ job.status }}
```
ZhongpinWang, line 38:
```suggestion
          SLACK_TITLE: Build tests for v4 ${{ job.status }} (${{ matrix.version }})
```
[req] missing ` (${{ matrix.version }})`?

.github/workflows/next-build-nightly.yml
```diff
@@ -0,0 +1,155 @@
...
+        name: API Doc Check
+      - uses: ./.github/actions/check-license
+        name: License Check
+      - if: failure()
+        name: Slack Notify
+        uses: rtCamp/action-slack-notify@v2.3.2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          SLACK_USERNAME: SDK Pipeline Bot
+          SLACK_TITLE: Build checks for v4 ${{ job.status }}
```
ZhongpinWang, line 83:
```suggestion
          SLACK_TITLE: Build checks for v4 ${{ job.status }} (${{ matrix.version }})
```

.github/workflows/next-build-nightly.yml
```diff
@@ -0,0 +1,155 @@
...
+          registry-url: 'https://registry.npmjs.org'
+      - run: yarn install --frozen-lockfile --ignore-engines
+      - name: Canary Release
+        run: |
+          date=`date +%Y%m%d%H%M%S`
+          rm -f .changeset/*.md
+          project_version=$(cat package.json | jq '.version' | tr -d '"')
+          echo The project version is: $project_version        
+          [[ $project_version == 4* ]] && cp major-release-changeset.md .changeset          
+          [[ $project_version == 3* ]] && cp canary-release-changeset.md .changeset
```
ZhongpinWang, line 149: [req] If this workflow is only triggered targeting on main / v4 branch, why are we checking the project version? changesets for v4 should be independent from v3 right?

ZhongpinWang (commented):
Some comments left for improvement.

One general question, I still don't fully get the point why we are not dispatching this workflow triggered by `pull_request` or `workflow_dispatch`. We can use `${{ github.event.pull_request.base.ref }}` to check if a PR is merged to `main` or `v3-main`.

deekshas8:
> I still don't fully get the point why we are not dispatching this workflow triggered by pull_request or workflow_dispatch.

We already have a workflow on the main branch that is triggered on pull_request and can be triggered manually. However, as I understood, scheduled workflows only run on the default branch(which is still v3-main). Therefore we had to duplicate the workflow here. This should therefore only be for releasing, as we dont want to trigger multiple workflows at once.

.github/workflows/next-build-nightly.yml
```diff
@@ -0,0 +1,155 @@
...
+          cache: 'yarn'
+          registry-url: 'https://registry.npmjs.org'
+      - run: yarn install --frozen-lockfile --ignore-engines
+      - name: Canary Release
+        run: |
+          date=`date +%Y%m%d%H%M%S`
+          rm -f .changeset/*.md
+          project_version=$(cat package.json | jq '.version' | tr -d '"')
+          echo The project version is: $project_version        
+          [[ $project_version == 4* ]] && cp major-release-changeset.md .changeset
```
deekshas8, line 148: [req] I released what you meant now, about swapping. This is wrong and please revert this to what it was in the original workflow.  The verison in package.json in main branch is 3.x.y. As such with your change, we'll release a next version with 3.x.y which is misleading and incorrect. That was the whole point of having this check

.github/workflows/next-build-nightly.yml
deekshas8: I created the original workflow on the main branch and there we need this check (1. if we create a manual canary, 2. for future after we do a v4 release) so that we don't mistakenly release wrong versions.
This was copied here to allow for scheduled runs. We ideally dont need to have this check: [[ $project_version == 4* ]] as currently we're still at version 3 (even on main) and after the release (version will be 4 then), we'll switch the default branch anyway and this workflow will be disabled.

.github/workflows/next-build-nightly.yml
KavithaSiva: Discussed in call, left it as is, as this job is a copy paste of the `build` workflow in main.

.github/workflows/next-build-nightly.yml
```diff
@@ -0,0 +1,157 @@
+name: next-build-nightly
+
+on:
+  schedule:
+    - cron: '0 2 * * *'
```
KavithaSiva, line 5: Core tests run at 3 am, so this should be good.
I don't see an option to use `next` release in smoke test, so I assume it runs only on latest canary release.

KavithaSiva:
> Some comments left for improvement.
> 
> One general question, I still don't fully get the point why we are not dispatching this workflow triggered by `pull_request` or `workflow_dispatch`. We can use `${{ github.event.pull_request.base.ref }}` to check if a PR is merged to `main` or `v3-main`.

As Deeksha pointed out, the idea was to trigger nightly canary releases for `v4`. As scheduled workflows only run on the default branch, we wanted a workflow that triggers a v4 release from the `v3-main` branch.

I initially thought about [triggering the action](https://github.com/SAP/cloud-sdk-js/pull/5488/files) for `main` branch directly from `v3-main`, but after discussions on PR settled for just copying the `build` workflow and running it by checking out `main` instead.

.github/workflows/next-build-nightly.yml
KavithaSiva: Did you mean `matrix.node-version` ? I have added that, other than that there is no other applicable `version` property here.

Issue #5131, opened 2024-10-29
Expired destinations through connectivity proxy remain in cache


carlonnheim:
**Describe the bug**
Destinations which use Authorization and Proxy-Authorization only consider the expiry time of the Authorization token when determining cache expiry.

**To Reproduce**
Use a destination which has for example `OAuth2JWTBearer` with a service going through the SAP Cloud Connector via the connectivity service. Configure it with `useCache = true`.

Make a request with one authenticated user. The framework obtains a Proxy-Authentication token to use with the connectivity service and the application token to use with the service being connected too. Both of these have an expiry time but only the application token is considered when determining the cache expiry. If the application token has a shorter expiry than the proxy token, the cached destination will start failing once the proxy token expires. Once the application token expires, the cache will be discarded and the next request will work again.

Make another request with a different user. The framework reuses the Proxy-Authentication token obtained during the first call, which can now have an arbitrary time-to-live left. A new application token for the other user is fetched. These are stored in the cache together and expires based on the expiry date of the application token. Once the proxy-auth expires, the destination fails until the application token eventually expires and the cache entry gets discarded.

**Expected behavior**
The cached destination must expire based on the shortest lived token (application or connectivity proxy). Alternatively, the proxy-auth needs to be refreshed on cached destinations when it expires.

**Screenshots**
We have worked around the issue with the following patch, which also further clarifies the issue. This includes excessive logging and is also not properly protecting for authorization tokens which are not jwt's (`decodeJwt` throws in that case), but I assume the proper solution to the problem might be different altogether anyway - this is just to clarify the issue.
```
diff --git a/node_modules/@sap-cloud-sdk/connectivity/dist/scp-cf/destination/destination-cache.js b/node_modules/@sap-cloud-sdk/connectivity/dist/scp-cf/destination/destination-cache.js
index 884f610..47972fc 100644
--- a/node_modules/@sap-cloud-sdk/connectivity/dist/scp-cf/destination/destination-cache.js
+++ b/node_modules/@sap-cloud-sdk/connectivity/dist/scp-cf/destination/destination-cache.js
@@ -105,9 +105,24 @@ async function cacheRetrievedDestination(token, destination, isolation, cache) {
     }
     const key = getDestinationCacheKey(token, destination.name, isolation);
     const expiresIn = (0, util_1.first)(destination.authTokens || [])?.expiresIn;
-    const expirationTime = expiresIn
+    let expirationTime = expiresIn
[...]

PR #5554, opened 2025-02-24
fix: check proxy and sap connectivity auth tokens for caching

ZhongpinWang:
Fix SAP/cloud-sdk-js#5131

Use the minimum of authentication token, `Proxy-Authorization` token, and `SAP-Connectivity-Authentication` token to determine the expiration time.

### DIFF
--- a/.changeset/thick-glasses-lick.md
+++ b/.changeset/thick-glasses-lick.md
@@ -2,4 +2,4 @@
 '@sap-cloud-sdk/connectivity': minor
 ---
 
-[Fix Issue] Check `Proxy-Authorization` and `SAP-Connectivity-Authentication` tokens to determine cache expiration time.
+[Fixed Issue] Check `Proxy-Authorization` and `SAP-Connectivity-Authentication` tokens to determine cache expiration time.
--- a/.changeset/thick-glasses-lick.md
+++ b/.changeset/thick-glasses-lick.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Fix Issue] Check `Proxy-Authorization` and `SAP-Connectivity-Authentication` tokens to determine cache expiration time.
--- a/packages/connectivity/src/scp-cf/destination/destination-cache.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination-cache.ts
@@ -1,5 +1,5 @@
 import { createLogger, first } from '@sap-cloud-sdk/util';
-import { getTenantId, userId } from '../jwt';
+import { decodeJwt, getTenantId, userId } from '../jwt';
 import { AsyncCache } from '../async-cache';
 import type { JwtPayload } from '../jsonwebtoken-type';
 import type { AsyncCacheInterface } from '../async-cache';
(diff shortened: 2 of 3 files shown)


.changeset/thick-glasses-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Fix Issue] Check `Proxy-Authorization` and `SAP-Connectivity-Authentication` tokens to determine cache expiration time.
```
tomfrenken, line 5:
```suggestion
[Fixed Issue] Check `Proxy-Authorization` and `SAP-Connectivity-Authentication` tokens to determine cache expiration time.
```

packages/connectivity/src/scp-cf/destination/destination-cache.ts
```diff
@@ -217,11 +217,40 @@ async function cacheRetrievedDestination<T extends DestinationCacheInterface>(
...
-  const key = getDestinationCacheKey(token, destination.name, isolation);
-  const expiresIn = first(destination.authTokens || [])?.expiresIn;
-  const expirationTime = expiresIn
-    ? Date.now() + parseInt(expiresIn) * 1000
+  const authToken = first(destination.authTokens || []);
+  const proxyAuthorizationToken =
+    destination.proxyConfiguration?.headers?.['Proxy-Authorization']?.match(
+      /^Bearer (.+)/
+    )?.[1];
+  const sapConnectivityAuthenticationToken =
```
deekshas8, line 225: [req/q] I'm not sure if we should cache based on `SAP-Connectivity-Authentication`? This token is basically the `subscriberToken?.userJwt.encoded`, which is passed by the user (would fail the verify jwt step if its expired). We have no control on how this is fetched and refreshed by the user. So basing our caching on this seems wrong.

ZhongpinWang: > [req/q] I'm not sure if we should cache based on `SAP-Connectivity-Authentication`. This token is basically the `subscriberToken?.userJwt.encoded`, which is passed by the user (would fail the verify jwt step if its expired). We have no control on how this is fetched and refreshed by the user. So basing our caching on this seems wrong.

I understand your concern. I don't have strong opinion on this. Only from a pure technical point of view, users are allowed to set destination cache with `setDestinationCache()` method. And both headers are defined as part of the `Destination` / `ProxyConfigurationHeaders` interface. Technically, it means we also can't be sure, in the end, how user sets `SAP-Connectivity-Authentication` token. Therefore, I would say we should consider both headers when calculating the expiration time. 

If exp of `SAP-Connectivity-Authentication` token is shorter than the `authTokens` for destination service and got expired, the `Destination` object itself is anyway invalid. 

> [pp] I would have preferred the approach to refresh only the proxy token if its expired. Its a CC grant token, whereas for fetching the whole destination again we make two additional calls to the destination service and the destination service itself might do additional exchange calls to the target system.
We're also mixing concerns between destinationCache and clientCredentialTokenCache.

Thanks for this cool suggestion, and ideally yes, we should try renew the proxy auth token. And I tried out. 

We can add a check when `retrieveDestinationFromCache`. But since we store only `expires` time in a `CacheEntry`, which is calculated based on the first `authTokens` `expiresIn`, it lacks expiration time for proxy authentication (and, if applicable, sap connectivity authentication). We will have to modify `CacheEntry` interface to store this additional information. But I feel `CacheEntry` is designed rather in a generic way, and adding information regarding proxy is not appropriate. 

So by cache design, we anyway need to determine a expiration time, which should invalidate the cache if its entry cannot be used anymore. Judging from a cached `Destination` entry, we have no clue, if the proxy authentication token has expired or not, since it only provides us a time interval and we don't know the start time when retrieving the destination.

Based on the above consideration, I would accept, at this moment, the fact that we may make few additional calls to Destination service. Unless in general, beyond the caching concept, we always renew the proxy auth token if it got expired.

deekshas8: Can you try what happens when we move the below line before we return the destination? I think this will automatically refresh the proxy auth and the other token in case they expired.

`const withProxySetting = await da.addProxyConfiguration(destination);`

ZhongpinWang:
Replaced by #5563

PR #5563, opened 2025-02-25
fix: `Proxy-Authorization` token expired issue for cached destination

ZhongpinWang:
We do not cache destination with `proxyConfiguration` any more. The proxy configuration will be added every time after a destination is fetched from or not from cache. 

Client credential token is cached anyway already so no performance issue when keep calling `serviceToken()` to set proxy authorization token.

Fix https://github.com/SAP/cloud-sdk-js/issues/5131

### DIFF
--- a/packages/connectivity/src/scp-cf/connectivity-service.ts
+++ b/packages/connectivity/src/scp-cf/connectivity-service.ts
@@ -1,7 +1,6 @@
 import { createLogger, ErrorWithCause } from '@sap-cloud-sdk/util';
 import { getServiceBindings } from './environment-accessor';
 import { serviceToken } from './token-accessor';
-import { decodeJwt } from './jwt';
 import type { JwtPayload } from './jsonwebtoken-type';
 import type { Protocol } from './protocol';
 import type {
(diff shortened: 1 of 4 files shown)


packages/connectivity/src/scp-cf/connectivity-service.ts
```diff
@@ -38,7 +39,28 @@ export async function addProxyConfigurationOnPrem(
...
 
-  const proxyConfiguration: ProxyConfiguration = {
+  // Get the proxy authorization token if it exists
+  const proxyAuthorizationToken =
+    destination.proxyConfiguration?.headers?.['Proxy-Authorization']?.match(
+      /^Bearer (.+)/
+    )?.[1];
+
+  let proxyConfiguration: ProxyConfiguration;
+  if (proxyAuthorizationToken) {
```
deekshas8, line 49: [q] Sorry I don't get why you need to check this here. Isn't the token validity already being checked in serviceToken function? The cache will returned undefined if the access_token is expired and trigger a clientCredentialToken call. So the refresh will happen if needed.

ZhongpinWang: Make totally sense. Thanks for pointing this out.

deekshas8 (approved):
LGTM 👍🏽

PR #5580, opened 2025-02-27
fix: `Proxy-Authorization` token expired issue for cached destination

ZhongpinWang:
Cherry pick #5563.

### DIFF
--- a/.changeset/salty-teams-follow.md
+++ b/.changeset/salty-teams-follow.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/salty-teams-follow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': patch
+---
+
+[Fixed Issue] Add `proxyConfiguration` on the fly to avoid expired proxy authorization token in cached destination if it lives shorter than the token for destination service.
```
deekshas8, line 5: not sure if we need a changeset for this in v4

ZhongpinWang: The thing is that if no changeset, our check-pr workflow will fail..

PR #5581, opened 2025-02-27
chore!: Remove `@sap-cloud-sdk/mail-client` package

ZhongpinWang:
Remove `@sap-cloud-sdk/mail-client` package as it is no longer supported in v4.

Closes SAP/cloud-sdk-backlog#230.

Doc: SAP/cloud-sdk#2029.


build-packages/check-pr/validators.ts
```diff
@@ -127,7 +127,7 @@ export function validateChangesets(
   }
 
   const changeTypes = fileContents.flatMap(content => {
-    const matches = content.match(/\[([^\]]+)\]/g);
+    const matches = content.match(/\[([^\]]+)\](?!\([^\]]*\))/g);
```
ZhongpinWang, line 130: [For reviewer] The `check-pr` workflow is failing because it is matching `text` from `[text](http://example.com)` as preamble.. I fixed it here but since it is not merged, the workflow still fails..

test-packages/e2e-tests/test/mail/mail-server.js
ZhongpinWang, line 1: [For reviewer] I removed the e2e tests for mail-client. But since in the pipeline, we use the previously built e2e-tests package. The changes does not affect the current workflow. So it is also failing.

tomfrenken (approved):
LGTM

PR #5585, opened 2025-03-03
fix: Json stringify Axios error response data

ZhongpinWang:
This issue was found when working on SAP/ai-sdk-js-backlog#235.

If `error.response.data` is an object, it leads to the following error stack, which is not helpful at all

```txt
...
Caused by:
HTTP Response: Request failed with status code 400 - [object Object]
```

Relates to SAP/ai-sdk-js-backlog#235.

### DIFF
--- a/packages/util/src/error-with-cause.spec.ts
+++ b/packages/util/src/error-with-cause.spec.ts
@@ -1,4 +1,5 @@
 import { ErrorWithCause } from './error-with-cause';
+import type { AxiosError } from 'axios';
 
 describe('ErrorWithCause', () => {
   it('creates an error', () => {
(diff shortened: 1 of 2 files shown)


tomfrenken:
I remember some fields in an axios error are frequently to breaking the JSON.stringify() method, for now this is fine, but we should consider omitting them from the stringified object, if we see this as an issue regularly.

tomfrenken (approved):
LGTM, but let's remember that we might need to adjust this further in the future, if stringification frequently fails.

ZhongpinWang:
> I remember some fields in an axios error are frequently breaking the JSON.stringify() method, for now this is fine, but we should consider omitting them from the stringified object, if we see this as an issue regularly.

Since I try catched stringify, in the worst case scenario, we get [object Object]. Also I saw that some fields in `error.response` has circular issue, thus, make `response` field even invisible when stringify just `error` object.

PR #5608, opened 2025-03-05
fix: Downgrade `@stylistic/eslint-plugin` to `v3.1.0`

ZhongpinWang:
`@stylistic/eslint-plugin` v4 (not Cloud SDK v4 😄 ) supports ESM only. [Release Note v4.0.0](https://github.com/eslint-stylistic/eslint-stylistic/releases/tag/v4.0.0) However, we are still on CJS. 

Also, starting from v4, they do not support legacy config but only flat config. But we currently support both config files. So if user uses legacy config and we upgrade to `@stylistic/eslint-plugin` v4, it will break.

Further discussion required for long-term support, but as for now, I will downgrade to v3.

---
I will create a PR for SDK v3 as it also breaks their..

### DIFF
--- a/.changeset/tame-spoons-kiss.md
+++ b/.changeset/tame-spoons-kiss.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/eslint-config': patch
+---
+
+[Fixed Issue] Downgrade `@stylistic/eslint-plugin` to v3 as v4 is EMS-only.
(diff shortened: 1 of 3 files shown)


deekshas8 (approved):

deekshas8:
Worth considering if we should we pin it somehow or ignore it from further dependabot updates?

ZhongpinWang:
> Worth considering if we should we pin it somehow or ignore it from further dependabot updates?

Thanks for mentioning this. I will add it to the dependabot ignore list.

tomfrenken:
If we ignore it entirely, do we not stop getting v3 patch updates?

tomfrenken:
(Assuming they continue pushing some security fixes for all the people that can't switch to ESM yet)

ZhongpinWang:
> If we ignore it entirely, do we not stop getting v3 patch updates?

We can ignore just the major version in dependabot.

Issue #5631, opened 2025-03-13
onPremise Connectivity via destination not working - results in getaddrinfo ENOTFOUND <<address-from-destination>>


drvup:
### Describe the Bug

Since version 3.26.2 of SAP Cloud SDK, we do face issues within the connectivity to onPremise systems maintained within destination.
The first request for a period is running / passing through with 200. All other requests to the endpoint do result in an error getaddrinfo ENOTFOUND <<address-from-destination>>

We initially raised an OSS incident as we thought it's about the SAP Cloud Connector. But, we were wondering why only applications on DEV-Subaccounts were effected. 
We identified `renovate` did his job and refreshed the dependencies to the latest minor version.

I did roll back to 3.26.1, and it's back working again. Since 3.26.2 the error is raised.

### Steps to Reproduce

Call a URL of a destination maintained within a onPremise system behind a SAP Cloud Connector (basic auth)

### Expected Behavior

All requests are running successfully and do pass through the destination.
I guess, the proxy to the connectivity service is lost after the first request by the SAP Cloud SDK.

### Used Versions

- Node version via `node -v`: 20.18.1
- NPM version via `npm -v: 10.8.2
- SAP Cloud SDK version: effected starting at 3.26.2

### Affected Development Phase

Development

### Impact

Impaired

ZhongpinWang:
Hi @drvup ,

Thanks a lot for reporting this. This might relates to #5563. Could you please help me understand the following points?

1. What do you mean by "first request **for a period**"? Does it mean that after a while, even the first request in the sequence won't go through?
2. Is the `useCache` option enabled when retrieving a destination.

drvup:
Hi @ZhongpinWang ,

yes, I guess it's related to #5563 :/ 

1. I can observe: If I did not use the application some hours (there's no jobscheduler active on `dev` tenants) and I do send a request to the app, the first request is passed through and works like expected. All other requests to the same endpoint fail with the shared message.
2. Yes, we're using the `useCache` option. So, this could be in co-relation to the scenario. 
3. Let me do that ... 

Best regards,
Cedric

ZhongpinWang:
@drvup I was able to reproduce the issue. The `proxyConfiguration` is indeed missing when retrieving the destination again. Working on it.

ZhongpinWang:
Hi @drvup ,

[...]

PR #5635, opened 2025-03-14
fix: Remove destination cache when getting destinations from service binding

ZhongpinWang:
Closes SAP/ai-sdk-js-backlog#256.

Fix user issue #5631.

In #5563, we cache only partial destination without `proxyConfiguration` to allow it being added on the fly after cache retrieval. This works well when calling `getDestinationFromDestinationService()` directly. 

However, since `searchServiceBindingForDestination()` and `getDestinationFromDestinationService()` are sharing one destination cache, and vcap search comes first, it intercepts the cache retrieval and causing `proxyConfiguration` being not added in `getDestinationFromDestinationService()`. 

My first thought was to separate the caches for both functions. But it is not possible because we export `setDestinationCache()` from `destination-cache.ts` to allow user using their own cache object. And we cannot clone such object. 

Therefore, I checked again and found out that, actually, destination cache is not needed for retrieval from service binding, because no request is sent to destination service. And the only request that potentially needs to be sent is inside the `serviceToken()` function to get a token from XSUAA service. And caching this token has already been covered with `clientCredentialsTokenCache`. 

Therefore, I don't see any performance downside when removing destination cache for getting destination from service binding. And this should solve the user issue.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1677,12 +1677,12 @@
   dependencies:
     "@sinonjs/commons" "^3.0.0"
 
-"@stylistic/eslint-plugin@^4.1.0":
-  version "4.1.0"
-  resolved "https://registry.npmjs.org/@stylistic/eslint-plugin/-/eslint-plugin-4.1.0.tgz#8882a6fe307cc94022e4495720373e8a5652d60c"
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -1677,12 +1677,12 @@
   dependencies:
     "@sinonjs/commons" "^3.0.0"
 
-"@stylistic/eslint-plugin@^4.1.0":
```
ZhongpinWang, line 1680: Oh.. I guess this is due to cherry pick.. Never yarn install again.

tomfrenken (approved):
LGTM

PR #5879, opened 2025-06-23
Remove XSUAA binding dependency from destination retrieval flow

KavithaSiva:


packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -4,12 +4,16 @@ import type { DestinationOptions } from './destination';
...
- * @param options - Configuration for how to retrieve destinations from the destination service.
- * @returns Exchanged token.
+ * Make a token exchange from IAS token to XSUAA token using credentials from destination service.
+ * @param jwt - The IAS JWT to exchange.
+ * @param service - The service used for the token exchange.
+ * @returns Exchanged XSUAA token.
  */
-export async function exchangeToken(jwt: string): Promise<string> {
-  return jwtBearerToken(jwt, 'xsuaa');
+export async function exchangeToken(
```
tomfrenken: What's the point of this change?
It looks like we exclusively use this function with the destination service, now it has a parameter more and each call needs to add 'destination'?

KavithaSiva: Right now we only use this during the destination fetch flow, so the credentials in the destination service are used for the token exchange.
If in the future for some other XSUAA based service if we need to do the exchange, then we could use the respective service credentials for the exchange. 
I did not want the exchange to always rely on the destination service, but rather do it based on the service involved.

tomfrenken: Given that we didn't change this since the inception of the SDK, I think it's fair to say we don't need to over-engineer this for now and keep it as is.
If we realize there is actually any need for this, we can still make it generic down the line, right now it's unnecessary complexity IMO.

KavithaSiva: Switched to using destination service by default.

packages/connectivity/src/scp-cf/identity-service.spec.ts
```diff
@@ -37,3 +38,41 @@ describe('shouldExchangeToken', () => {
     ).toBe(false);
   });
 });
+
+describe('exchangeToken', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should call jwtBearerToken with destination service', async () => {
```
tomfrenken: Not sure if these tests add much, since the wrapping function is essentially just a wrapper that sets "destination" as default, I leave it to you if you want to leave it in.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -4,12 +4,12 @@ import type { DestinationOptions } from './destination';
...
 /**
  * @internal
- * Make a token exchange from IAS token to XSUAA token.
- * @param options - Configuration for how to retrieve destinations from the destination service.
- * @returns Exchanged token.
+ * Make a token exchange from IAS token to XSUAA token using credentials from destination service.
+ * @param jwt - The IAS JWT to exchange.
+ * @returns Exchanged XSUAA token.
  */
 export async function exchangeToken(jwt: string): Promise<string> {
```
tomfrenken: [pp] I would consider renaming the function to `exchangeTokenToXsuaaToken` or something similar, as it narrows the function a little more.

tomfrenken (approved):
LGTM, some notes left, but I leave it up to you since it's only [pp]

packages/connectivity/src/scp-cf/identity-service.spec.ts
KavithaSiva: I will leave them in.

packages/connectivity/src/scp-cf/identity-service.spec.ts
```diff
@@ -37,3 +41,43 @@ describe('shouldExchangeToken', () => {
     ).toBe(false);
   });
 });
+
+describe('exchangeToken', () => {
```
ZhongpinWang:
```suggestion
describe('exchangeTokenToXsuaaToken', () => {
```

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -4,12 +4,12 @@ import type { DestinationOptions } from './destination';
...
- * Make a token exchange from IAS token to XSUAA token.
- * @param options - Configuration for how to retrieve destinations from the destination service.
- * @returns Exchanged token.
+ * Make a token exchange from IAS token to XSUAA token using credentials from destination service.
+ * @param jwt - The IAS JWT to exchange.
+ * @returns Exchanged XSUAA token.
  */
-export async function exchangeToken(jwt: string): Promise<string> {
-  return jwtBearerToken(jwt, 'xsuaa');
+export async function exchangeTokenToXsuaaToken(jwt: string): Promise<string> {
```
ZhongpinWang: [req] This function is located in `scp-cf` folder and was meant to be more generic to exchange IAS token to the bound XSUAA service token. Since now its purpose changed to only retrieve token for destination service, I would suggest to either keep the original one or move this to subfolder `scp-cf/destination`.

.changeset/early-rocks-dip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Compatibility Note] The incoming XSUAA JWT is not verified anymore while retrieving destinations using `getDestinationFromDestinationService`.
```
ZhongpinWang:
```suggestion
[Compatibility Note] The incoming XSUAA JWT is not verified anymore while retrieving destinations using `getDestinationFromDestinationService()` function.
```

.changeset/early-rocks-work.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Improvement] Remove dependency on XSUAA service binding while retrieving destinations using `getDestinationFromDestinationService` and `getAllDestinationsFromDestinationService`.
```
ZhongpinWang:
```suggestion
[Improvement] Remove dependency on XSUAA service binding while retrieving destinations using `getDestinationFromDestinationService()` and `getAllDestinationsFromDestinationService()` functions.
```

.changeset/early-rocks-dip.md
ZhongpinWang: [req] Style guide:  https://github.com/SAP/cloud-sdk/blob/main/STYLEGUIDE.md#include-a-descriptive-noun-before-or-after-code

ZhongpinWang (changes requested):
Mostly LGTM, some minors.

ZhongpinWang:
Before merging this, it would be nice if we create an E2E test that fails on the current release and passes on this PR / canary.

packages/connectivity/src/scp-cf/identity-service.ts
KavithaSiva: We discussed using `jwtBearerToken` directly instead of the wrapper function.

packages/connectivity/src/scp-cf/identity-service.spec.ts
KavithaSiva: No longer relevant.

packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
```diff
@@ -44,19 +17,15 @@ export function clearXsuaaServices(): void {
...
-export function getXsuaaService(options?: {
+export function getXsuaaInstanceFromServiceCredentials(options: {
   disableCache?: boolean;
-  jwt?: JwtPayload | string;
-  credentials?: ServiceCredentials;
+  credentials: ServiceCredentials;
 }): any {
-  const credentials =
-    options?.credentials || getXsuaaServiceCredentials(options?.jwt);
+  const { credentials } = options;
```
ZhongpinWang: [pp] Can we make each property directly a parameter instead of having a mandatory `options` object?

packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
```diff
@@ -44,19 +17,15 @@ export function clearXsuaaServices(): void {
...
- * @param options - Options on how to configure the XSUAA service.
  * @param options.disableCache - Value to enable or disable JWKS cache in xssec library. Defaults to false.
- * @param options.jwt - Either a JWT payload or an encoded JWT. Will be ignored if `credentials` are provided. If not provided, the first XSUAA service binding is used.
- * @param options.credentials - Xsuaa service credentials. If not provided, the credentials are fetched based on the JWT
- * @returns An instance of the xsuaa service that the application is bound to.
+ * @param options.credentials - Xsuaa service credentials. Required to create the xssec XSUAA instance.
+ * @returns An instance of {@code @sap/xssec/XsuaaService} for the provided credentials.
  */
-export function getXsuaaService(options?: {
+export function getXsuaaInstanceFromServiceCredentials(options: {
```
ZhongpinWang:
```suggestion
export function getXsuaaServiceFromServiceCredentials(options: {
```
[pp] Let's keep naming consistent.

.changeset/early-rocks-dip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Compatibility Note] The incoming XSUAA JWT is not verified anymore while retrieving destinations using `getDestinationFromDestinationService()` function.
```
ZhongpinWang:
```suggestion
[Compatibility Note] The incoming XSUAA JWT is not verified anymore with the bound XSUAA service while retrieving destinations using `getDestinationFromDestinationService()` function.
```

packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
KavithaSiva: I feel having XSUAA service in the name is misleading, as it makes as think we are dealing with the service, when in fact we are dealing with the UAA credentials in re-use services.

As xssec itself calls it XSUAA Service as well, adding instance to the name hints at creating an object of it instead.
The function is internal, so it should be okay to change it, WDY think?

packages/http-client/src/http-client.spec.ts
```diff
@@ -94,6 +95,21 @@ describe('generic http client', () => {
     name: 'FORWARD-TOKEN-NOAUTH'
   };
 
+  function responseWithPublicKey(key: string = publicKey) {
```
KavithaSiva, line 98: For reviewer, this was originally imported from the verify test file, I just moved the function in here as I deleted the verify test file.

packages/connectivity/src/scp-cf/jwt/verify.ts
```diff
@@ -1,43 +1,11 @@
...
-  }).catch(e => {
-    throw new ErrorWithCause('Failed to verify JWT.', e);
-  });
-
-  return token.payload;
-}
-
 /**
  * Options to control certain aspects of JWT verification behavior.
+ * @deprecated This property is no longer used as the incoming JWT is not verified during destination retrieval.
```
deekshas8: [req] Generally a good practice to mention `@deprecated since vX.Y.Z`. Also I would say `The property has no effect as the incoming JWT is no longer verified during destination retrieval.`

deekshas8: The interface deprecation should not say property

packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
```diff
@@ -73,6 +39,8 @@ export function getXsuaaService(options?: {
   const cacheKey = `${credentials.clientid}:${disableCache}`;
 
   if (!xsuaaServices[cacheKey]) {
+    // XsuaaService is a class that extends Service from @sap/xssec and is a representation of XSUAA credentials
```
deekshas8:
```suggestion
    // XsuaaService is a representation of XSUAA credentials extracted from a reuse service, e.g., destination
```

[q] Do we care what it extends? Its clear its from xssec

packages/connectivity/src/scp-cf/environment-accessor/xsuaa.ts
```diff
@@ -73,6 +39,8 @@ export function getXsuaaService(options?: {
...
+    // extracted from a reuse service like for e.g destination service
```
deekshas8:
```suggestion
```

packages/connectivity/src/scp-cf/destination/get-subscriber-token.ts
```diff
@@ -46,23 +46,13 @@ export function isSubscriberToken(token: any): token is SubscriberToken {
 export async function getSubscriberToken(
   options: DestinationOptions
 ): Promise<SubscriberToken> {
-  const isXsuaaJwt = !!options.jwt && isXsuaaToken(decodeJwt(options.jwt));
-  const userJwt = await retrieveUserToken(options, isXsuaaJwt);
-  const serviceJwt = await retrieveServiceToken(options, userJwt?.decoded);
+  const incomingUserJwtPair = options.jwt ? getJwtPair(options.jwt) : undefined;
```
deekshas8: [req] Please keep it as userJwt. I think the naming was clear as it is.

deekshas8 (changes requested):

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -90,10 +90,10 @@ export class DestinationFromServiceRetriever {
     options: DestinationFetchOptions
   ): Promise<Destination | null> {
     // TODO: This is currently always skipped for tokens issued by XSUAA
-    // in the XSUAA case no exchange takes place, but instead the JWT is verified
-    // in the future we should just let it verify here, but skip it later (get-subscriber-token)
+    // in the XSUAA case no exchange takes place
     if (shouldExchangeToken(options) && options.jwt) {
-      options.jwt = await exchangeToken(options.jwt);
+      // Exchange the IAS token to a XSUAA token using the destination service credentials
+      options.jwt = await jwtBearerToken(options.jwt, 'destination');
```
ZhongpinWang, line 96: We should keep in mind that 

Currently, `jwtBearerToken` means get access token **from the service's XSUAA** with service client credentials (of course issued by that service's XSUAA) and user jwt. 

[pp] Do you have an idea where to put this reminder? I feel we will forget again.

KavithaSiva: JSDoc of both `jwtBearerToken` and `getUserToken` have been modified, hopefully, this acts as a reminder.

PR #5945, opened 2025-07-24
chore: Skip `as` keyword and aliases in barrel files when checking public APIs

ZhongpinWang:
Having aliases defined by `as` keyword in barrel files was causing issue (found in SAP/ai-sdk-js/pull/880).

So previously, `export { foo as bar } from './somefile.js';` causes issue `fooasbar is exported but marked as internal`.

Now we use regex to match and replace `as` keyword and aliases. These should not be checked.

### DIFF
--- a/.github/actions/check-license/index.js
+++ b/.github/actions/check-license/index.js
@@ -6259,6 +6259,413 @@ module.exports = bind.call(call, $hasOwn);
 
 /***/ }),
 
+/***/ 4181:
+/***/ ((module) => {
+
+"use strict";
+
+
+var gitHosts = module.exports = {
+  github: {
+    // First two are insecure and generally shouldn't be used any more, but
+    // they are still supported.
+    'protocols': [ 'git', 'http', 'git+ssh', 'git+https', 'ssh', 'https' ],
+    'domain': 'github.com',
+    'treepath': 'tree',
+    'filetemplate': 'https://{auth@}raw.githubusercontent.com/{user}/{project}/{committish}/{path}',
+    'bugstemplate': 'https://{domain}/{user}/{project}/issues',
+    'gittemplate': 'git://{auth@}{domain}/{user}/{project}.git{#committish}',
+    'tarballtemplate': 'https://codeload.{domain}/{user}/{project}/tar.gz/{committish}'
+  },
+  bitbucket: {
+    'protocols': [ 'git+ssh', 'git+https', 'ssh', 'https' ],
+    'domain': 'bitbucket.org',
+    'treepath': 'src',
+    'tarballtemplate': 'https://{domain}/{user}/{project}/get/{committish}.tar.gz'
+  },
+  gitlab: {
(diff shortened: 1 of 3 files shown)


build-packages/check-public-api/check-public-api.spec.ts
```diff
@@ -235,9 +235,11 @@ export declare type MyType = {value:string}
 
 const dummyIndexFile = `export { o1 } from './bla';
     
-    export { o2, o3 } from './bla';
-    export { o4, 
-    
+    export { o2,  o3 as o3$Some_thing   } from './bla';  // :-) This is a comment
+    export { o4     as  o4Something, 
+    /****
+      * * This is a block comment **??? *
```
deekshas8, line 241: [q] Why would a barrel file have multi line comments?

build-packages/check-public-api/index.ts
```diff
@@ -268,13 +268,17 @@ export function parseExportedObjectsInFile(
...
- * /\/\*[\s\S]*?\*\/|\/\/.*|[\s]+/g - Matches single-line comments or multi-line comments or whitespaces.
+ * Aliases defined with 'as' keyword are removed.
  * @param fileContent - Content of the index file to be parsed.
  * @param regex - Regular expression used for matching exports.
  * @returns List of objects exported by the given index file.
  */
 export function parseBarrelFile(fileContent: string, regex: RegExp): string[] {
-  const normalized = fileContent.replace(/\/\*[\s\S]*?\*\/|\/\/.*|[\s]+/g, '');
+  // Remove block comments, single-line comments, 'as' keyword and aliases, and whitespace characters
+  const normalized = fileContent.replace(
```
deekshas8, line 278: IMO we should avoid `as` in `index.ts` and if it exists for whatever reason, it should not just be ignored. We still need to make sure if its not marked internal, it is exported and checked correctly.

build-packages/check-public-api/check-public-api.spec.ts
ZhongpinWang: The regex check was already implemented. I am not sure in which case we might have those comments, but well I added it for testing.

build-packages/check-public-api/index.ts
ZhongpinWang: IMO avoid or not is rather a style thing. Not handling it correctly resulting into something like `fooasbar` is a bug in the script. The original type/object is still checked against `@internal`, we just don't care about the alias after `as` since they do not exist anywhere else in the code.

ZhongpinWang: So for example, `export { foo as bar } from './somefile.js';`, we still check if `foo` is marked or not, but not `bar` because it's just a name.

deekshas8: Ah okay, I think the comment confused me. I thought if a line with as is encountered, we ignore it like comments line.

deekshas8 (approved):

PR #6007, opened 2025-08-28
fix: Avoid caching jwt if it needs to be forwarded

ZhongpinWang:
Closes SAP/cloud-sdk-backlog#1278.

To be forwarded jwt is no longer part of the destination cache to avoid provider account destination being cached with only provider tenant id as the cache key, and to be forwarded jwt can change next time.

Refactored a lot to make the flow more clear.


packages/connectivity/src/scp-cf/destination/forward-auth-token.ts
```diff
@@ -42,21 +42,26 @@ function validateToken(token: string | undefined): token is string {
...
       logger.debug(
         "Option 'forwardAuthToken' enabled on destination. Using the given token for the destination."
       );
-      destination.authTokens = buildDestinationAuthToken(token);
+      authTokens = buildDestinationAuthToken(token);
     }
+    return {
+      ...destination,
+      authTokens
+    };
```
marikaner: [pp] I always try to avoid using `let` and I think here it is easily avoidable.

```suggestion
if (validateToken(token)) {
logger.debug(
"Option 'forwardAuthToken' enabled on destination. Using the given token for the destination."
);
      return {
      ...destination,
      authTokens: buildDestinationAuthToken(token);
      }
}
```

packages/connectivity/src/scp-cf/destination/destination-from-env.ts
marikaner: [pp] For consistency with the new code maybe consider:
```suggestion
if (isHttpDestination(destination) &&
['internet', 'private-link'].includes(proxyStrategy(destination))) {
  destination = addProxyConfigurationInternet(destination);
}
return destination;
```

This also applies to the other places where we have the same code.

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -405,6 +398,71 @@ Possible alternatives for such technical user authentication are BasicAuthentica
...
+      authentication === 'OAuth2JWTBearer' ||
+      authentication === 'SAMLAssertion' ||
+      (authentication === 'OAuth2SAMLBearerAssertion' &&
+        !this.usesSystemUser(destination))
+    ) {
+      // VERY BAD...
+      // If origin is provider, next time subscriber jwt might change.
+      // -> It might be an invalid user jwt next time, and SDK won't throw as destination cached already.
+      // -> SDK will use auth token retrieved with the previous subscriber jwt for a different subscriber next time.
+      destination = await this.fetchDestinationWithUserExchangeFlows(
```
marikaner:
```suggestion
      return this.fetchDestinationWithUserExchangeFlows(
```

marikaner: (you can then also remove the `else`s)

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -405,6 +398,71 @@ Possible alternatives for such technical user authentication are BasicAuthentica
...
+      this.usesSystemUser(destination)
+    ) {
+      // SOMETIMES BAD!
+      // If origin is provider
+      // -> Auth token can be cached in destination cache as subscriber jwt is not used.
+      // -> UNLESS for `OAuth2ClientCredentials`, if `x-tenant` is used with `tokenServiceURLType` set to `Common` (see `getExchangeTenant()`),
+      //   then the subdomain of the tenant is sent to destination service and jwt will be exchanged based on the templated token service url.
+      // If origin is subscriber
+      // -> Auth token can be cached in destination cache as destination is tenant-isolated.
+      destination = await this.fetchDestinationWithNonUserExchangeFlows(
```
marikaner:
```suggestion
      return this.fetchDestinationWithNonUserExchangeFlows(
```

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -405,6 +398,71 @@ Possible alternatives for such technical user authentication are BasicAuthentica
...
+        destination,
+        origin
+      );
+    } else if (authentication === 'OAuth2RefreshToken') {
+      // OK!
+      // If origin is provider, provider jwt + refresh token is used.
+      // -> Auth token can be cached in destination cache as subscriber is not used.
+      // If origin is subscriber, subscriber jwt + refresh token is used.
+      // -> Auth token can be cached in destination cache as destination is tenant-isolated.
+      destination = await this.fetchDestinationWithRefreshTokenFlow(
```
marikaner:
```suggestion
      return this.fetchDestinationWithRefreshTokenFlow(
```

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
```diff
@@ -405,6 +398,71 @@ Possible alternatives for such technical user authentication are BasicAuthentica
...
+
+    if (
+      authentication === 'OAuth2UserTokenExchange' ||
+      authentication === 'OAuth2JWTBearer' ||
+      authentication === 'SAMLAssertion' ||
+      (authentication === 'OAuth2SAMLBearerAssertion' &&
+        !this.usesSystemUser(destination))
+    ) {
+      // VERY BAD...
+      // If origin is provider, next time subscriber jwt might change.
```
marikaner: [q] I don't understand this sentence.

packages/connectivity/src/scp-cf/destination/forward-auth-token.spec.ts
```diff
@@ -1,20 +1,20 @@
...
 
 describe('forward auth token', () => {
   it('sets an auth token, when `forwardAuthToken` is set', () => {
     const jwt = signedJwt({});
-    const destination = setForwardedAuthTokenIfNeeded(
-      { forwardAuthToken: true },
-      jwt
-    );
+    let destination: Destination = { forwardAuthToken: true };
+    destination = addForwardedAuthTokenIfNeeded(destination, jwt);
```
marikaner, line 10:
```suggestion
    const destination = addForwardedAuthTokenIfNeeded({ forwardAuthToken: true }, jwt);
```

packages/connectivity/src/scp-cf/destination/forward-auth-token.spec.ts
```diff
@@ -1,20 +1,20 @@
...
     expect(destination.authTokens?.[0]).toMatchObject({ value: jwt });
   });
 
   it('does not set an auth token, when `forwardAuthToken` is not set', () => {
     const jwt = signedJwt({});
-    const destination = setForwardedAuthTokenIfNeeded({}, jwt);
+    let destination: Destination = {};
+    destination = addForwardedAuthTokenIfNeeded(destination, jwt);
```
marikaner, line 17:
```suggestion
    const destination = addForwardedAuthTokenIfNeeded({}, jwt);
```

marikaner (changes requested):
I don't fully understand the comments, maybe let's talk about them.

packages/connectivity/src/scp-cf/destination/forward-auth-token.ts
ZhongpinWang: Very true 👍 I changed it.

packages/connectivity/src/scp-cf/destination/destination-from-service.ts
ZhongpinWang: I think I meant

If destination is fetched from the provider subaccount (origin = 'provider'), then we use provider tenant id to build the cache key.

But these auth flows need specific user jwt, which comes from certain subscriber tenant, and the auth token will be cached together with the destination using provider tenant id + user id only (no subscriber tenant id). This is a problem for multi-tenant app as next get destination may be for a user from a different subscriber.

packages/connectivity/src/scp-cf/destination/forward-auth-token.spec.ts
ZhongpinWang: This part I wanted to make the code style similar to our code logic, have a let destination first, then add things on top of it. Feels easier for me to understand the purpose of the code.

packages/connectivity/src/scp-cf/destination/forward-auth-token.spec.ts
ZhongpinWang: Same as above, I made a tradeoff here to help better understand the logic that we add things on top of an existing destination.

.changeset/mean-flies-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': patch
+---
+
+[Fixed Issue] Avoid caching jwt if `forwardAuthToken` is set to `true`.
```
marikaner:
```suggestion
[Fixed Issue] Avoid caching JWT if `forwardAuthToken` is set to `true`.
```

marikaner (approved):

PR #6061, opened 2025-09-17
feat: esm support on Odata

BrigittaK307:


.github/actions/check-public-api/index.js
```diff
@@ -71791,6 +71791,17 @@ function packageJsonBase(options) {
             url: ''
         }
     };
+    if (options.moduleType === 'esm') {
```
deekshas8, line 71800: [q] Why do we need this?

test-packages/test-services-odata-v2/generate-odata-services.ts
```diff
@@ -1,19 +1,19 @@
...
   transpile: true,
   useSwagger: false,
   readme: false,
   clearOutputDir: false,
   packageJson: false,
-  generateSdkMetadata: false,
   skipValidation: true,
-  verbose: false
+  verbose: false,
+  generateESM: true
```
deekshas8:
```suggestion
  generateESM: false
```

Don't set it true here as the generate tests will fail. I believe the default is false anyway, so you can also leave it out.

packages/generator/src/generator-without-ts-morph/request-builder/imports.ts
```diff
@@ -1,34 +1,74 @@
...
-    ...entityKeyImportDeclaration(entity.keys)
+    {
+      kind: StructureKind.ImportDeclaration,
+      moduleSpecifier: options?.generateESM
+        ? `./${entity.className}.js`
+        : `./${entity.className}`,
+      namedImports: [entity.className]
+    },
+    {
+      kind: StructureKind.ImportDeclaration,
```
deekshas8: [q] Why did we change this? I now see an additional import statement `import { RequestBuilder } from '@sap-cloud-sdk/odata-v4'`

packages/generator/src/service/package-json.spec.ts
```diff
@@ -35,7 +35,8 @@ describe('package-json', () => {
       npmPackageName: `my-${oDataVersion}-package`,
       description: `my ${oDataVersion} package description`,
       oDataVersion,
-      sdkVersion: '1.2.3'
+      sdkVersion: '1.2.3',
+      moduleType: 'commonjs'
```
deekshas8, line 39: Why do we change this? The default moduleType is generally commonjs

packages/generator/src/service/package-json.ts
```diff
@@ -1,16 +1,31 @@
 import { unixEOL } from '@sap-cloud-sdk/util';
 import { packageJsonBase } from '@sap-cloud-sdk/generator-common/internal';
 import type { ODataVersion } from '@sap-cloud-sdk/util';
-import type { PackageJsonOptions as PackageJsonOptionsBase } from '@sap-cloud-sdk/generator-common/internal';
 
 /**
  * @internal
  */
-export interface PackageJsonOptions extends PackageJsonOptionsBase {
+export interface PackageJsonOptions {
```
deekshas8: generator-common package is shared between odata and openapi generators. Have we check if this affects openapi generator?

packages/generator-common/src/file-writer/package-json.ts
```diff
@@ -39,4 +43,17 @@ export function packageJsonBase(
...
     }
   };
+
+  if (options.moduleType === 'esm') {
+    basePackageJson.type = 'module';
+    basePackageJson.main = './index.mjs';
+    basePackageJson.exports = {
+      '.': {
+        import: './index.mjs',
+        require: './index.js'
```
deekshas8: Hybrid export is not needed. Lets first just keep it simple.

packages/generator/src/options/options.ts
```diff
@@ -17,6 +21,10 @@ export interface GeneratorOptions extends CommonGeneratorOptions {
    * Number of node processes used for transpilation of JavaScript files.
    */
   transpilationProcesses?: number;
+  /**
+   * Whether to generate ECMAScript modules instead of CommonJS modules.
+   */
+  generateESM?: boolean;
```
deekshas8: This should be moved to the common generator options since its duplicated in both generators

packages/generator/src/options/options.ts
```diff
@@ -44,5 +52,27 @@ export const cliOptions = {
...
 } as const satisfies Options<GeneratorOptions>;
+
+/**
+ * Build a tsconfig.json file as string.
+ * If the given options include a tsConfig setting, this config is read and returned.
+ * @param options - Options passed to the generator.
+ * @returns The serialized tsconfig.json contents.
+ * @internal
+ */
+export async function tsconfigJson({
```
deekshas8: Also duplicated in both generators, should be moved to generator-common package

packages/generator/src/generator.ts
```diff
@@ -466,12 +530,15 @@ function projectOptions(): ProjectOptions {
     },
     compilerOptions: {
       target: ScriptTarget.ES2021,
-      module: ModuleKind.CommonJS,
+      module: moduleType === 'esm' ? ModuleKind.ESNext : ModuleKind.CommonJS,
```
deekshas8: Should be nodenext: https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-compiling-and-running-the-outputs-in-nodejs

packages/generator/src/generator.ts
```diff
@@ -181,10 +180,13 @@ export async function generateProject(
     emptyDirSync(options.outputDir.toString());
   }
 
-  const project = new Project(projectOptions());
+  const tsConfig = await tsconfigJson(options);
```
deekshas8: I didn't get why this was moved here?

packages/generator/src/generator.ts
```diff
@@ -310,37 +314,54 @@ export async function generateSourcesForService(
...
-          description: packageDescription(service.speakingModuleName),
-          oDataVersion: service.oDataVersion
-        }),
-        createFileOptions
-      )
-    );
+    await generatePackageJson(serviceDirPath, service, options);
   }
 
-  if (options.transpile || options.tsconfig) {
```
deekshas8, line 332: transpile options still needs to be checked for generating tsconfig.

packages/generator/src/generator.ts
```diff
@@ -310,37 +314,54 @@ export async function generateSourcesForService(
...
+    };
+    await transpileDirectory(serviceDirPath, transpileOptions);
+  }
+
+  if (options.readme) {
+    await generateReadme(serviceDirPath, service, options);
   }
+}
+
+async function generateMandatorySources(
```
deekshas8: I see that we're refactoring the code, but it has become harder to review what changes are related to supporting esm, vs what is refactored code. I would suggest to always create separate PRs for this

packages/generator/src/generator.ts
```diff
@@ -310,37 +314,54 @@ export async function generateSourcesForService(
...
+  }
+
+  if (tsConfig) {
+    await createFile(
+      serviceDirPath,
+      'tsconfig.json',
+      tsConfig,
+      createFileOptions
     );
+    const transpileOptions = {
```
deekshas8: Revert this. transpile is not supposed to be called here. Its called after all source files are generated. See line 97 in this file

packages/generator/src/generator.ts
```diff
@@ -310,37 +314,54 @@ export async function generateSourcesForService(
...
+      compilerOptions: await readCompilerOptions(serviceDirPath),
+      createFileOptions
+    };
+    await transpileDirectory(serviceDirPath, transpileOptions);
+  }
+
+  if (options.readme) {
+    await generateReadme(serviceDirPath, service, options);
```
deekshas8: Please keep the original order in which the files were generated.

packages/generator/src/imports.ts
```diff
@@ -184,21 +187,52 @@ export function mergeImportDeclarations(
...
+      : `./${prop.jsType}`,
     namedImports: [prop.jsType]
   };
 }
+
+/**
+ * @internal
+ */
+export function getImportsWithESM(
+  schemaName: string,
```
deekshas8: I didn't get the point of this function?

packages/generator/src/complex-type/imports.spec.ts
```diff
@@ -94,4 +95,49 @@ describe('complex type imports', () => {
...
+        commonjsOptions
+      );
+
+      expect(
+        actual.find(imp => imp.moduleSpecifier === './ComplexDesert')
+          ?.moduleSpecifier
+      ).toBe('./ComplexDesert');
+    });
+
+    it('importDeclarations with ESM', () => {
```
deekshas8: Great that you added tests for this. I would restructure them to test for default (which should be commonjs) or esm.

deekshas8 (changes requested):

packages/generator/src/generator-without-ts-morph/request-builder/imports.spec.ts
```diff
@@ -25,4 +25,28 @@ describe('imports', () => {
...
+          'GetAllRequestBuilder',
+          'GetByKeyRequestBuilder',
+          'RequestBuilder',
+          'Time',
+          'UpdateRequestBuilder'
+        ]
+      },
+      {
+        names: ['Breakfast'],
+        moduleIdentifier: './Breakfast.js'
```
deekshas8: [pp] If we only care about whether esm extension was generated, then I will only test for moduleIdentifier, instead of the whole object

.changeset/honest-candles-kick.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-cloud-sdk/generator': minor
+'@sap-cloud-sdk/generator-common': minor
+---
+
+[New Functionality] Introduce option `generateESM` in generator package to generate ESM compatible code.
```
deekshas8:
```suggestion
[New Functionality] Introduce option `generateESM` in OData generator to generate ESM compatible code.
```

packages/generator-common/src/file-writer/package-json.ts
```diff
@@ -39,4 +43,17 @@ export function packageJsonBase(
       url: ''
     }
   };
+
+  if (options.moduleType === 'esm') {
+    basePackageJson.type = 'module';
+    basePackageJson.main = './index.mjs';
+    basePackageJson.exports = {
```
deekshas8: As discussed, please remove the `.exports` property.

packages/generator-common/src/options.ts
```diff
@@ -6,6 +6,8 @@ import {
   resolveRequiredPath
 } from './options-parser';
 import type { ServiceType } from './options-parser';
+import type { ParsedOptions } from './options-parser';
```
deekshas8:
```suggestion
```

There is already a type import from `options-parser`.

packages/generator-common/src/options.ts
```diff
@@ -6,6 +6,8 @@ import {
   resolveRequiredPath
 } from './options-parser';
 import type { ServiceType } from './options-parser';
```
deekshas8, line 8:
```suggestion
import type { ServiceType, ParsedOptions } from './options-parser';
```

packages/generator-common/src/options.ts
```diff
@@ -6,6 +6,8 @@ import {
...
+import type { ParsedOptions } from './options-parser';
+import { formatTsConfig, readCustomTsConfig } from './ts-config';
```
deekshas8:
```suggestion
```

remove un-used imports from the file.

packages/openapi-generator/src/options/tsconfig-json.ts
```diff
@@ -1,21 +1,3 @@
...
- */
-export async function tsconfigJson({
-  transpile,
-  tsconfig: tsConfig
-}: ParsedGeneratorOptions): Promise<string | undefined> {
-  if (transpile || tsConfig) {
-    return tsConfig ? readCustomTsConfig(tsConfig) : formatTsConfig();
-  }
-}
+export { tsconfigJson };
```
deekshas8: Delete this file completely. It's not needed.

packages/openapi-generator/src/generator.ts
```diff
@@ -85,7 +86,10 @@ export async function generateWithParsedOptions(
   const inputFilePaths = options.input;
 
   const optionsPerService = await getOptionsPerService(inputFilePaths, options);
-  const tsConfig = await tsconfigJson(options);
+  const tsConfig = await tsconfigJson({
```
deekshas8: `const tsConfig = await tsconfigJson(options);` is correct, please revert

packages/openapi-generator/src/generator.ts
```diff
@@ -168,6 +171,10 @@ async function generateSources(
     await copyFiles(options.include, serviceDir, options.overwrite);
   }
 
+  const tsConfig = await tsconfigJson({
```
deekshas8: Revert this change. the parameter is passed to this function.

packages/openapi-generator/src/generator.ts
```diff
@@ -149,7 +153,6 @@ async function generateSources(
   serviceDir: string,
   openApiDocument: OpenApiDocument,
   inputFilePath: string,
-  tsConfig: string | undefined,
```
deekshas8, line 152: Why are we removing this parameter? Only the import needed to be changed in this file. Add the parameter back.

packages/openapi-generator/src/generator.ts
```diff
@@ -282,7 +289,6 @@ async function generateService(
     serviceDir,
     parsedOpenApiDocument,
     inputFilePath,
-    tsConfig,
```
deekshas8, line 285: Revert

packages/generator/src/imports.spec.ts
```diff
@@ -1,207 +0,0 @@
...
-      namedImports: ['1']
-    } as ImportDeclarationStructure;
-
-    const declaration2FromModule = {
-      kind: StructureKind.ImportDeclaration,
-      moduleSpecifier: './module',
-      namedImports: ['2']
-    } as ImportDeclarationStructure;
-
-    it('merges named imports', () => {
```
deekshas8, line 177: [q] Why did we delete this file?

packages/generator/src/generator.ts
```diff
@@ -330,16 +332,20 @@ export async function generateSourcesForService(
...
-      createFile(
-        serviceDirPath,
-        'tsconfig.json',
-        options.tsconfig
-          ? await readCustomTsConfig(options.tsconfig)
-          : formatTsConfig(),
-        createFileOptions
-      )
-    );
+    const tsConfig = await tsconfigJson({
```
deekshas8: the function `tsconfigJson` should take the full options obj . `await tsconfigJson(options)`

packages/generator/src/generator.ts
```diff
@@ -456,7 +462,9 @@ export async function generateSourcesForService(
   await Promise.all(filePromises);
 }
 
-function projectOptions(): ProjectOptions {
+function projectOptions(
+  moduleType: 'commonjs' | 'esm' = 'commonjs'
```
deekshas8: Maybe it will make sense to create a separate type and reuse it, instead of typing it like this everywhere

packages/openapi-generator/src/options/tsconfig-json.ts
deekshas8: Also delete the ts-config.spec.ts next to it, since you already moved all the tests to the `generator-common` package.

packages/generator-common/src/ts-config.ts
```diff
@@ -46,3 +45,20 @@ export async function readCustomTsConfig(configPath: string): Promise<string> {
...
+ * Build a tsconfig.json file as string.
+ * If transpile is true or tsconfig is provided, return the appropriate config.
+ * @param transpile - Whether to transpile.
+ * @param tsconfig - Path to custom tsconfig file.
+ * @returns The serialized tsconfig.json contents.
+ * @internal
+ */
+export async function tsconfigJson(
+  transpile: boolean,
+  tsconfig: string | undefined
```
deekshas8:
```suggestion
  tsconfig?: string
```

With this syntax, you won't have to pass undefined everytime as second param

deekshas8: Also set `transpile: boolean = false,` so you only need to pass it when its true.

packages/generator/src/complex-type/imports.spec.ts
```diff
@@ -94,4 +95,49 @@ describe('complex type imports', () => {
...
+  describe('ESM support', () => {
+    const commonjsOptions = {
+      generateESM: false
+    } as CreateFileOptions;
+
+    const esmOptions = {
+      generateESM: true
+    } as CreateFileOptions;
+
+    it('importDeclarations with CommonJS', () => {
```
deekshas8: [req] Please change this to read like a sentence. (I know some others in the file read incorrect)
Please include only 1 test: when the option is explicitly set to true.

packages/generator/src/complex-type/imports.spec.ts
```diff
@@ -94,4 +95,49 @@ describe('complex type imports', () => {
       }
     ]);
   });
+
+  describe('ESM support', () => {
```
deekshas8: Don't add the extra describe. There is already a test for the default behavior (`importDeclarations for a complex type that includes a complex type`). Move the esm related test to the original describe.

packages/generator/src/complex-type/imports.spec.ts
```diff
@@ -94,4 +95,49 @@ describe('complex type imports', () => {
...
+    const commonjsOptions = {
+      generateESM: false
```
deekshas8: You don't need this. that's the default anyway.

packages/generator/src/complex-type/imports.spec.ts
```diff
@@ -94,4 +95,49 @@ describe('complex type imports', () => {
...
+        esmOptions
+      );
+
+      expect(
+        actual.find(imp => imp.moduleSpecifier === './ComplexDesert.js')
+          ?.moduleSpecifier
+      ).toBe('./ComplexDesert.js');
+    });
+
+    it('maintains backward compatibility when options is undefined', () => {
```
deekshas8: [req] Its not backwards compatible, its the default behavior.

deekshas8: Remove this test

packages/generator/src/entity/imports.spec.ts
```diff
@@ -31,6 +32,85 @@ describe('imports', () => {
...
+    const commonjsOptions = {
+      generateESM: false
+    } as CreateFileOptions;
+
+    const esmOptions = {
+      generateESM: true
+    } as CreateFileOptions;
+
+    it('importDeclarations with CommonJS', () => {
+      const actual = entityImportDeclarations(
```
deekshas8: This is already tested in `otherEntityImports`. Remove commonjs related tests. Only add for ESM and in the original describe sections.

packages/generator/src/entity/imports.spec.ts
```diff
@@ -31,6 +32,85 @@ describe('imports', () => {
...
+        foodService,
+        esmOptions
+      );
+
+      expect(actual.map(imports => imports.moduleSpecifier)).toEqual([
+        './Brunch.js'
+      ]);
+    });
+
+    it('maintains backward compatibility when options is undefined', () => {
```
deekshas8: Same as above, its not backwads compatible but the default behavior. Remove this test

packages/generator/src/generator-without-ts-morph/request-builder/file.ts
```diff
@@ -4,16 +4,18 @@ import { requestBuilderClass } from './class';
...
  */
 export function requestBuilderSourceFile(
   entity: VdmEntity,
-  oDataVersion: ODataVersion
+  oDataVersion: ODataVersion,
+  options?: CreateFileOptions
 ): string {
   const imports = serializeImports(
-    requestBuilderImportDeclarations(entity, oDataVersion)
+    requestBuilderImportDeclarations(entity, oDataVersion, options?.generateESM)
```
deekshas8: The API is a bit inconsistent. Everywhere else we pass the full options object, but in some methods we pass generateESM?

packages/generator/src/generator-without-ts-morph/request-builder/imports.ts
```diff
@@ -54,20 +58,34 @@ function requestBuilderImports(entity: VdmEntity) {
   return imports;
 }
 
-function entityImportDeclaration(entity: VdmEntity): Import {
+/**
+ * @internal
+ */
+function entityImportDeclaration(
+  entity: VdmEntity,
+  generateESM?: boolean
```
deekshas8: here as well.

packages/generator/src/options/options.ts
```diff
@@ -44,5 +44,11 @@ export const cliOptions = {
     hidden: true,
     replacedBy: 'processesJsGeneration'
   },
+  generateESM: {
```
deekshas8: This option should be in the generator-common right? in getCommonCliOptions?

packages/generator/src/imports.spec.ts
deekshas8: ?

packages/generator-common/src/file-writer/package-json.ts
```diff
@@ -39,4 +43,11 @@ export function packageJsonBase(
       url: ''
     }
   };
+
+  if (options.moduleType === 'esm') {
+    basePackageJson.type = 'module';
+    basePackageJson.main = './index.mjs';
```
deekshas8: I believe this line is also not needed.

packages/generator/src/imports.spec.ts
BrigittaK307: I remembered that we do not these tests anymore, but I will revert it.

packages/generator-common/src/internal.ts
```diff
@@ -1,3 +1,11 @@
+/**
+ * @internal
+ */
+export type ModuleType = 'commonjs' | 'esm';
```
deekshas8: A barel file is only for exporting. You shouldn't define types here. Please move it to a more appropriate file.

packages/generator-common/src/internal.ts
```diff
@@ -1,3 +1,11 @@
...
+
+export { getCommonCliOptions } from './options';
```
deekshas8:
```suggestion
export * from './options';
```


The `internal` barel file exports everything. `index.ts` determines the public api and hence only exports public facing types, functions, etc.

packages/generator-common/src/ts-config.spec.ts
```diff
@@ -1,24 +1,18 @@
...
 describe('tsconfigJson', () => {
   afterEach(() => {
     mock.restore();
   });
 
   it('returns the default tsconfig if transpilation is enabled', async () => {
-    const tsConfig = await tsconfigJson({
-      transpile: true
-    } as ParsedGeneratorOptions);
+    const tsConfig = await tsconfigJson(true, undefined);
```
deekshas8:
```suggestion
    const tsConfig = await tsconfigJson(true);
```

Same in other places.

packages/generator-common/src/ts-config.spec.ts
```diff
@@ -1,24 +1,18 @@
...
     expect(JSON.parse(tsConfig!)).toEqual(defaultTsConfig);
   });
 
   it('returns undefined if transpilation is disabled', async () => {
-    const tsConfig = await tsconfigJson({
-      transpile: false
-    } as ParsedGeneratorOptions);
+    const tsConfig = await tsconfigJson(false, undefined);
```
deekshas8:
```suggestion
    const tsConfig = await tsconfigJson();
```

false and undefined are implied.

packages/generator/src/complex-type/imports.spec.ts
deekshas8: You don't need an extra describe here. Please move this to the existing one, and just add it as a test.

packages/generator/src/entity/imports.spec.ts
```diff
@@ -31,6 +32,40 @@ describe('imports', () => {
       }))
     ).toEqual(expectedOtherEntityImports);
   });
+
+  describe('ESM support', () => {
```
deekshas8: Again, please no extra describes.

packages/generator/src/entity/imports.spec.ts
```diff
@@ -31,6 +32,40 @@ describe('imports', () => {
...
+    const esmOptions = {
+      generateESM: true
+    } as CreateFileOptions;
+
+    it('importDeclarations when ESM flag is set to true', () => {
```
deekshas8:
```suggestion
    it('adds .js in import declarations when ESM flag is true', () => {
```

It should be a complete sentence. I see existing ones are incorrect as well, but it shouldn't mean we continue the wrong pattern.

deekshas8: Also fix in other places.

packages/generator/src/entity/imports.spec.ts
```diff
@@ -31,6 +32,34 @@ describe('imports', () => {
...
+
+    const breakfastApiImport = actual.find(
+      imp =>
+        Array.isArray(imp.namedImports) &&
+        imp.namedImports.includes('BreakfastApi')
+    );
+    expect(breakfastApiImport?.moduleSpecifier).toBe('./BreakfastApi.js');
+  });
+
+  it('add .js in other entity imports when ESM flag is true', () => {
```
deekshas8:
```suggestion
  it('adds .js in other entity imports when ESM flag is true', () => {
```

packages/generator/src/complex-type/imports.spec.ts
```diff
@@ -94,4 +95,16 @@ describe('complex type imports', () => {
       }
     ]);
   });
+
+  const esmOptions = {
```
deekshas8, line 99: Please declare this inside the `it`

deekshas8: Only common variables should be declared within `describe`

packages/generator/src/generator.ts
deekshas8:
```suggestion
          oDataVersion: service.oDataVersion,
          generateESM: options.generateESM
```

This needs to be passed here as well, else the package.json doesn't have type: module for esm

packages/generator-common/src/ts-config.ts
deekshas8: This is always `commonjs` which is a problem if you set generateESM to true. Because the generated ESM code gets compiled to commonjs. both `module` and `moduleResolution` should also be conditional.

deekshas8 (changes requested):
I found two issues during compilation, having to do with incorrect tsconfig and missing type in package.json.

Issue #6042, opened 2025-09-10
OData V4 deserialize error for decimal fields if we pass decimal value in "" as string


prashant4be:
https://github.com/SAP/cloud-sdk-js/blob/751dd139de0810838182f020e8594ae84012ab05/packages/odata-v4/src/de-serializers/default-de-serializers.ts#L84

OData V4 deserialize error for decimal fields if we pass decimal value in "", eg "10.6", same field values works fine with OData V2 service

TypeError: value.toNumber is not a function
    at serialize (default-de-serializers.js:41:79)
    at default-de-serializers.js:103:20
    at de-serializers.js:28:28
    at serializeField (entity-serializer.js:38:20)
    at entity-serializer.js:70:37
    at Array.reduce (<anonymous>)
    at serializeEntityNonCustomFields (entity-serializer.js:67:44)
    at serializeEntity (entity-serializer.js:29:16)
    at entity-serializer.js:44:51
    at Array.map (<anonymous>)

Why we don't have Number(value) instead of value.toNumber() ? in serialize method in default-de-serializers.js

davidkna-sap:
We have investigated this and only `BigNumber` is supported as an input by EDM.Decimal with OData v4.

If you cannot provide a `BigNumber` you can use a[ custom deserialiser/serialiser](https://sap.github.io/cloud-sdk/docs/js/features/odata/v4-client#customize-de-serialization) to accommodate different inputs and outputs.

PR #6142, opened 2025-10-28
fix: OData v4 default de-serializer number conversion

ZhongpinWang:
Closes SAP/cloud-sdk-js#6042.

Replace the mistakenly used `value.toNumber()` with `Number(value)` in OData V4 default de-serializers.

### DIFF
--- a/packages/connectivity/src/scp-cf/xsuaa-service.ts
+++ b/packages/connectivity/src/scp-cf/xsuaa-service.ts
@@ -18,7 +18,7 @@ interface XsuaaParameters {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


deekshas8:
> Replace the mistakenly used `value.toNumber()`

I don't think this is a mistake. The expected input type for `Edm.Decimal` in Odata v4 is BigNumber, which has a `.toNumber()` function. For OData V4 service, if a number is within safe range, it should be sent as a number. Only if its outside the limit, we serialize to string.

deekshas8:
Closing as resolution is proposed in the main ticket.

PR #6152, opened 2025-11-12
chore(ci): Switch to npm trusted publishing

davidkna-sap:
Closes SAP/cloud-sdk-backlog#1280.

* Updated workflows to have a single entry-point (`release-entry.yml`) in the manner of https://github.com/SAP/ai-sdk-js/pull/1272

>  Some GitHub Actions workflows use `workflow_call` to invoke other workflows that run npm publish, or use `workflow_dispatch` for manual publishing. When this happens, validation checks the calling workflow's name instead of the workflow that actually contains the publish command, which can cause configuration mismatches.

(via. [npm/documentation#1755](https://www.github.com/npm/documentation/issues/1755) and https://docs.npmjs.com/trusted-publishers#troubleshooting)

* For the non-private `package.json` files I added the `directory` for `package.json` `repository` option

### DIFF
--- a/.github/workflows/release-entry.yml
+++ b/.github/workflows/release-entry.yml
@@ -23,7 +23,7 @@ jobs:
     secrets: inherit
 
   delegate_to_canary_job:
-    if: ${{ github.event_name != 'release' }}
+    if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
     uses: ./.github/workflows/build.yml
     secrets: inherit
     with:


.github/workflows/release-entry.yml
```diff
@@ -0,0 +1,30 @@
...
+        required: false
+        default: false
+jobs:
+  delegate_to_release_job:
+    if: ${{ github.event_name == 'release' }}
+    uses: ./.github/workflows/release.yml
+    secrets: inherit
+
+  delegate_to_canary_job:
+    if: ${{ github.event_name != 'release' }}
```
deekshas8: [pp] I prefer positive checks over negations, so the condition reads as the actual intent.

davidkna-sap: I updated it, for the record this was supposed to be a makeshift `else` to highlight that it is a binary decision.

PR #6153, opened 2025-11-12
feat: Add jks to supported formats

emincihangeri:
Closes SAP/ai-sdk-js-backlog#358.


packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -132,6 +133,25 @@ function getKeyStoreOptions(destination: Destination):
...
+    if (getFormat(certificate) === 'jks') {
+      if (!destination.keyStorePassword) {
+        throw Error('Keystore password is required for JKS format');
+      }
+      const pemKeystore = jks.toPem(certBuffer, destination.keyStorePassword);
+      const aliases = Object.keys(pemKeystore);
+      if (aliases.length === 0) {
+        throw Error('No entries found in JKS keystore');
+      }
+      const alias = aliases[0]; // Use first alias
```
davidkna-sap: [q] How does the java sdk handle multiple keys?
[pp] If supporting multiple keys is not a requirement, I would also throw if `aliases.length > 1`.

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -132,6 +133,30 @@ function getKeyStoreOptions(destination: Destination):
 
     const certBuffer = Buffer.from(certificate.content, 'base64');
 
+    if (getFormat(certificate) === 'jks') {
```
davidkna-sap: [req] I had a look at the other code again and the format-value for jks-certs can also be `keystore`

```suggestion
    if (getFormat(certificate) === 'jks' || getFormat(certificate) === 'keystore') {
```

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -132,6 +133,33 @@ function getKeyStoreOptions(destination: Destination):
 
     const certBuffer = Buffer.from(certificate.content, 'base64');
 
+    if (
+      getFormat(certificate) === 'jks' ||
+      getFormat(certificate) === 'keystore'
+    ) {
+      if (!destination.keyStorePassword) {
+        throw Error('Keystore password is required for JKS format');
+      }
```
davidkna-sap: AFAICT this is optional and removing it might help with the most recent update by the user.

```suggestion
```

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -207,7 +235,7 @@ function mtlsIsEnabled(destination: Destination) {
 /*
  The node client supports only these store formats https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions.
  */
-const supportedCertificateFormats = ['p12', 'pfx', 'pem'];
+const supportedCertificateFormats = ['p12', 'pfx', 'pem', 'jks'];
```
davidkna-sap: Not completely sure, but I think this can be listed there too now:
```suggestion
const supportedCertificateFormats = ['p12', 'pfx', 'pem', 'jks', 'keystore'];
```

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -132,6 +133,28 @@ function getKeyStoreOptions(destination: Destination):
...
+        throw Error('No entries found in JKS keystore');
+      }
+      const alias = aliases[0];
+      const entry = pemKeystore[alias];
+      if (!entry.cert || !entry.key) {
+        throw Error('Invalid JKS entry: missing cert or key');
+      }
+      return {
+        cert: Buffer.from(entry.cert, 'utf8'),
+        key: Buffer.from(entry.key, 'utf8')
```
KavithaSiva, line 164: [q] Should we also a pass the keystorePassword as passphrase if available? I see that this is being done in the `pem` case.

packages/connectivity/src/http-agent/http-agent.ts
KavithaSiva: [req] Remove this line as it is confusing and no longer relevant.

KavithaSiva (changes requested):
I have left two comments to work on.

packages/connectivity/src/http-agent/http-agent.ts
davidkna-sap: The library does appear to decrypt keys (returns value includes "decrypted private key"), but this implementation does not handle the case where the keystore and certificate passwords are different yet (but this could be added as needed).

packages/connectivity/src/http-agent/http-agent.ts
KavithaSiva: I checked this in Java, all private keys are loaded into the key manager.
For now, as the original issue is solved with the current implementation, we can create a follow-up to handle multiple aliases.
IMO, all the aliases can be returned as Node.JS TLS https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options seems to support multiple cert chains.

PR #6156, opened 2025-11-14
fix: Handle arrays of enum strings correctly

daniel-kullmann:
Without this change an array of enums will result in e.g.:

export type Model1 = {
  values: 'A' | 'B' | 'C'[];
} & Record<string, any>;

which means the value can be either "A", "B", or "C"[], which is wrong.


daniel-kullmann:
I tested this with the following API json:

```json
{
  "openapi": "3.0.3",
  "info": {
    "version": "1.0.0",
    "title": "Api",
    "description": "API"
  },
  "components": {
    "schemas": {
      "Model1": {
        "type": "object",
        "required": ["values"],
        "properties": {
          "values": {
            "type": "array",
            "items": {
              "type": "string",
              "uniqueItems": true,
              "minItems": 0,
              "enum": [
                "A",
                "B",
                "C"
              ]
            }
          }
        }
      }
    }
  },
  "paths": {
    "/values": {
      "get": {
        "summary": "",
        "operationId": "getValues",
        "responses": {
          "200": {
            "description": "A list of all values",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/Model1"
                }
              }
            }
          }
        }
      }
    }
  }
}
```

daniel-kullmann:
Ah, `uniqueItems`should be one level higher, next to `type: "array"` 

```json
      "Model1": {
        "type": "object",
        "required": ["values"],
        "properties": {
          "values": {
            "type": "array",
            "uniqueItems": true,
            "items": {
              "type": "string",
              "minItems": 0,
              "enum": [
                "A",
                "B",
                "C"
              ]
            }
          }
        }
      }
```

packages/openapi-generator/src/file-serializer/schema.spec.ts
```diff
@@ -303,6 +303,18 @@ describe('serializeSchema for xOf schemas', () => {
...
+it('serializes array schemas for enum', () => {
+  expect(
+    serializeSchema({
+      type: 'array',
+      items: {
+        type: 'string',
+        enum: ["'One'", "'Two'"]
+      }
+    })
+  ).toEqual("('One' | 'Two')[]");
```
deekshas8, line 315: 👍🏽

PR #6164, opened 2025-11-18
chore(ci): Fix release-entry canary-release option handling

davidkna-sap:
Closes SAP/cloud-sdk-backlog#1280.

### DIFF
--- a/.github/workflows/release-entry.yml
+++ b/.github/workflows/release-entry.yml
@@ -30,5 +30,5 @@ jobs:
     uses: ./.github/workflows/build.yml
     secrets: inherit
     with:
-      canary-release-skip-checks: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.canary-release == 'Skip Tests and Checks (Force Release)' || github.event_name != 'schedule' }}
+      canary-release-skip-checks: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.canary-release == 'Skip Tests and Checks (Force Release)' }}
       is-workflow-call: true


.github/workflows/release-entry.yml
```diff
@@ -27,4 +27,4 @@ jobs:
     uses: ./.github/workflows/build.yml
     secrets: inherit
     with:
-      canary-release: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.canary-release || github.event_name != 'schedule' }}
+      canary-release: ${{ github.event_name == 'workflow_dispatch' && (github.event.inputs.canary-release == 'true') || github.event_name != 'schedule' }}
```
ZhongpinWang: Why wrapped with ()? And I feel `github.event_name != 'schedule'` should be negated because scheduled task is always releasing canary?

```suggestion
      canary-release: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.canary-release == 'true' || github.event_name == 'schedule' }}
```
Also since `canary-release` is defined as a boolean, is it necessary to compare it to a string `true` which is truthy?

davidkna-sap: I added the `()` to avoid any possible surprises in the evaluation order.
The naming is not great, but if `canary-release` is `true` it will skip an additional check to confirm if the previous canary release was up-to-date, so this should be `false` for `schedule`-events.

ZhongpinWang: Okay after checking it thoroughly I get what you mean. The naming `canary-release` is extremely confusing. Because true or false, we are releasing canary. The only difference is if we skip tests and pre-checkes and force the canary release. 

So if it is set to true, it means `build.yml` will skip everything tests and checks and force a canary release. Can you improve the namings? Make it clear that whenever using workflow dispatch, it is canary, and checking the box is just skipping things. Only release event do the real official releases.

ZhongpinWang: What about replacing the check box with a dropdown, title "Release Canary", list ['Without Tests and Checks (Force Release)', 'With Tests and Checks']

This at least indicate us when running the workflow dispatch from the UI, that this is about canary release only, even when the filename says `release-entry`. It was pretty clear before because the file was called `canary-release`.

Maybe you have better ideas.

deekshas8 (approved):
LGTM

PR #6172, opened 2025-11-24
fix: added generateESM flag to packageJson options

BrigittaK307:
Added generateESM flag to openApi generator in order to generate the ECMAScript/CommonJS style files correctly.

### DIFF
--- a/.changeset/cool-tigers-make.md
+++ b/.changeset/cool-tigers-make.md
@@ -2,4 +2,4 @@
 '@sap-cloud-sdk/openapi': minor
 ---
 
-[Fixed Issue] Added generateESM to packageJson options
+[Fixed Issue] Fix missing `type: module` in generated `package.json` when `--generateESM` option is enabled.
--- a/packages/openapi-generator/src/file-serializer/package-json.spec.ts
+++ b/packages/openapi-generator/src/file-serializer/package-json.spec.ts
@@ -33,4 +33,15 @@ describe('packageJson', () => {
     );
     expect(parsed.type).toBe('module');
   });
+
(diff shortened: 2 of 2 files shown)


.changeset/cool-tigers-make.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/openapi': minor
+---
+
+[Fixed Issue] Added generateESM to packageJson options
```
ZhongpinWang:
```suggestion
[Fixed Issue] Fix missing `type: module` in generated `package.json` when `--generateESM` option is enabled.
```

Describe user-facing issue instead of what the fix actually was since users will also read this and won't be able to understand from the details.

packages/openapi-generator/src/file-serializer/package-json.spec.ts
```diff
@@ -21,4 +21,16 @@ describe('packageJson', () => {
...
+    const parsed = JSON.parse(
+      packageJson({
+        npmPackageName: 'workflow-service',
+        description: 'description',
+        sdkVersion: '1.35.0',
+        generateESM: true
+      })
+    );
+    expect(parsed.type).toBe('module');
+  });
```
ZhongpinWang, line 35: [pp] One negative test can be helpful here to avoid always generating the type module.

```ts
  it('includes type module when generateESM is undefined', () => {
    const parsed = JSON.parse(
      packageJson({
        npmPackageName: 'workflow-service',
        description: 'description',
        sdkVersion: '1.35.0'
      })
    );
    expect(parsed.type).toBeUndefined();
  });
```

PR #6185, opened 2025-12-02
feat: IAS App-To-App Auth

davidkna-sap:
Handles SAP/ai-sdk-js-backlog#347.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "eslint": "^9.39.2",
     "execa": "^5.0.0",
     "glob": "^13.0.0",
-    "globals": "^17.0.0",
+    "globals": "^17.1.0",
     "jest": "^30.2.0",
     "jest-junit": "^16.0.0",
     "license-checker": "^25.0.1",
@@ -113,7 +113,7 @@
     "semver": "^7.7.2",
     "ts-jest": "^29.4.6",
     "ts-node": "^10.9.2",
-    "turbo": "^2.7.4",
+    "turbo": "^2.7.6",
     "typedoc": "^0.28.16",
     "typescript": "~5.9.3"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -4829,10 +4829,10 @@ globals@^14.0.0:
   resolved "https://registry.npmjs.org/globals/-/globals-14.0.0.tgz#898d7413c29babcf6bafe56fcadded858ada724e"
   integrity sha512-oahGvuMGQlPw/ivIYBjVSrWAfWLBeku5tpPE2fOPLi+WHffIWbuh2tCjhyQhTBPMf5E9jDEH4FOmTYgYwbKwtQ==
 
-globals@^17.0.0:
-  version "17.0.0"
-  resolved "https://registry.npmjs.org/globals/-/globals-17.0.0.tgz#a4196d9cfeb4d627ba165b4647b1f5853bf90a30"
-  integrity sha512-gv5BeD2EssA793rlFWVPMMCqefTlpusw6/2TbAVMy0FzcG8wKJn4O+NqJ4+XWmmwrayJgw5TzrmWjFgmz1XPqw==
+globals@^17.1.0:
+  version "17.1.0"
+  resolved "https://registry.npmjs.org/globals/-/globals-17.1.0.tgz#03448e667caa3914312a7937eef48c4a59dc6859"
+  integrity sha512-8HoIcWI5fCvG5NADj4bDav+er9B9JMj2vyL2pI8D0eismKyUvPLTSs+Ln3wqhwcp306i73iyVnEKx3F6T47TGw==
 
 globalthis@^1.0.4:
   version "1.0.4"
@@ -9087,47 +9087,47 @@ tunnel@^0.0.6:
   resolved "https://registry.npmjs.org/tunnel/-/tunnel-0.0.6.tgz#72f1314b34a5b192db012324df2cc587ca47f92c"
   integrity sha512-1h/Lnq9yajKY2PEbBadPXj3VxsDDu844OnaAo52UVmIzIvwwtBPIuNvkjuzBlTWpfJyUbG3ez0KSBibQkj4ojg==
 
-turbo-darwin-64@2.7.4:
-  version "2.7.4"
-  resolved "https://registry.npmjs.org/turbo-darwin-64/-/turbo-darwin-64-2.7.4.tgz#ea3f00f7f02474c253e59275d26078053c3eec99"
-  integrity sha512-xDR30ltfkSsRfGzABBckvl1nz1cZ3ssTujvdj+TPwOweeDRvZ0e06t5DS0rmRBvyKpgGs42K/EK6Mn2qLlFY9A==
+turbo-darwin-64@2.7.6:
+  version "2.7.6"
+  resolved "https://registry.npmjs.org/turbo-darwin-64/-/turbo-darwin-64-2.7.6.tgz#d344cc03893615bfa33613190aee4798795b79ef"
+  integrity sha512-bYu0qnWju2Ha3EbIkPCk1SMLT3sltKh1P/Jy5FER6BmH++H5z+T5MHh3W1Xoers9rk4N1VdKvog9FO1pxQyjhw==
 
-turbo-darwin-arm64@2.7.4:
-  version "2.7.4"
-  resolved "https://registry.npmjs.org/turbo-darwin-arm64/-/turbo-darwin-arm64-2.7.4.tgz#b35acee0f009648e1205efb4c37b6f97e38a78cb"
-  integrity sha512-P7sjqXtOL/+nYWPvcDGWhi8wf8M8mZHHB8XEzw2VX7VJrS8IGHyJHGD1AYfDvhAEcr7pnk3gGifz3/xyhI655w==
+turbo-darwin-arm64@2.7.6:
+  version "2.7.6"
+  resolved "https://registry.npmjs.org/turbo-darwin-arm64/-/turbo-darwin-arm64-2.7.6.tgz#36517372a51b789af9ff976d79779ba1e273b995"
+  integrity sha512-KCxTf3Y1hgNLYIWRLw8bwH8Zie9RyCGoxAlXYsCBI/YNqBSR+ZZK9KYzFxAqDaVaNvTwLFv3rJRGsXOFWg4+Uw==
 
-turbo-linux-64@2.7.4:
-  version "2.7.4"
-  resolved "https://registry.npmjs.org/turbo-linux-64/-/turbo-linux-64-2.7.4.tgz#e649cbd50b5c7e5fd0a303a2c6fb9686bd0e44d0"
-  integrity sha512-GofFOxRO/IhG8BcPyMSSB3Y2+oKQotsaYbHxL9yD6JPb20/o35eo+zUSyazOtilAwDHnak5dorAJFoFU8MIg2A==
+turbo-linux-64@2.7.6:
+  version "2.7.6"
+  resolved "https://registry.npmjs.org/turbo-linux-64/-/turbo-linux-64-2.7.6.tgz#8062f31689ffffe18f48969d8b72fed57baa34bd"
+  integrity sha512-vjoU8zIfNgvJR3cMitgw7inEoi6bmuVuFawDl5yKtxjAEhDktFdRBpGS3WojD4l3BklBbIK689ssXcGf21LxRA==
 
-turbo-linux-arm64@2.7.4:
-  version "2.7.4"
-  resolved "https://registry.npmjs.org/turbo-linux-arm64/-/turbo-linux-arm64-2.7.4.tgz#3a72cfb73c09cc417ff8cfb30103334d3597568e"
-  integrity sha512-+RQKgNjksVPxYAyAgmDV7w/1qj++qca+nSNTAOKGOfJiDtSvRKoci89oftJ6anGs00uamLKVEQ712TI/tfNAIw==
+turbo-linux-arm64@2.7.6:
+  version "2.7.6"
+  resolved "https://registry.npmjs.org/turbo-linux-arm64/-/turbo-linux-arm64-2.7.6.tgz#88224a60faed697e3c586f4ed140143c565a0311"
+  integrity sha512-TcMpBvTqZf+1DptrVYLbZls7WY1UVNDTGaf0bo7/GCgWYv5eZHCVo4Td7kCJeDU4glbXg67REX0md0S0V6ghMg==
 
-turbo-windows-64@2.7.4:
-  version "2.7.4"
-  resolved "https://registry.npmjs.org/turbo-windows-64/-/turbo-windows-64-2.7.4.tgz#da0957806407567501bbf48ee707bccff42e3af4"
-  integrity sha512-rfak1+g+ON3czs1mDYsCS4X74ZmK6gOgRQTXjDICtzvR4o61paqtgAYtNPofcVsMWeF4wvCajSeoAkkeAnQ1kg==
+turbo-windows-64@2.7.6:
+  version "2.7.6"
+  resolved "https://registry.npmjs.org/turbo-windows-64/-/turbo-windows-64-2.7.6.tgz#bea938ee7c0d2907342bdf518c5c3403da7c2a44"
+  integrity sha512-1/MhkYldiihjneY8QnnDMbAkHXn/udTWSVYS94EMlkE9AShozsLTTOT1gDOpX06EfEW5njP09suhMvxbvwuwpQ==
 
-turbo-windows-arm64@2.7.4:
-  version "2.7.4"
-  resolved "https://registry.npmjs.org/turbo-windows-arm64/-/turbo-windows-arm64-2.7.4.tgz#6bad15321282af636fb5a81eb22e17a777e882dc"
-  integrity sha512-1ZgBNjNRbDu/fPeqXuX9i26x3CJ/Y1gcwUpQ+Vp7kN9Un6RZ9kzs164f/knrjcu5E+szCRexVjRSJay1k5jApA==
+turbo-windows-arm64@2.7.6:
+  version "2.7.6"
+  resolved "https://registry.npmjs.org/turbo-windows-arm64/-/turbo-windows-arm64-2.7.6.tgz#8e1c878f614b840ac383808c0a5f45208d574c57"
+  integrity sha512-0wDVnUJLFAWm4ZzOQFDkbyyUqaszorTGf3Rdc22IRIyJTTLd6ajqdb+cWD89UZ1RKr953+PZR1gqgWQY4PDuhA==
 
-turbo@^2.7.4:
-  version "2.7.4"
-  resolved "https://registry.npmjs.org/turbo/-/turbo-2.7.4.tgz#9088187ff36bc212cea8cb68e2b48a0691a042c4"
-  integrity sha512-bkO4AddmDishzJB2ze7aYYPaejMoJVfS0XnaR6RCdXFOY8JGJfQE+l9fKiV7uDPa5Ut44gmOWJL3894CIMeH9g==
+turbo@^2.7.6:
+  version "2.7.6"
+  resolved "https://registry.npmjs.org/turbo/-/turbo-2.7.6.tgz#bbdb35ecb804c1bfbbe4077ef818ae45233d5eff"
+  integrity sha512-PO9AvJLEsNLO+EYhF4zB+v10hOjsJe5kJW+S6tTbRv+TW7gf1Qer4mfjP9h3/y9h8ZiPvOrenxnEgDtFgaM5zw==
   optionalDependencies:
-    turbo-darwin-64 "2.7.4"
-    turbo-darwin-arm64 "2.7.4"
-    turbo-linux-64 "2.7.4"
-    turbo-linux-arm64 "2.7.4"
-    turbo-windows-64 "2.7.4"
-    turbo-windows-arm64 "2.7.4"
+    turbo-darwin-64 "2.7.6"
+    turbo-darwin-arm64 "2.7.6"
+    turbo-linux-64 "2.7.6"
+    turbo-linux-arm64 "2.7.6"
+    turbo-windows-64 "2.7.6"
+    turbo-windows-arm64 "2.7.6"
 
 tv4@^1.3.0:
   version "1.3.0"


packages/connectivity/src/scp-cf/jwt/jwt.ts
```diff
@@ -27,12 +27,18 @@ function makeArray(val: string | string[] | undefined): string[] {
...
+ * For IAS tokens, this is `user_uuid`.
  * @param jwtPayload - Token payload to read the user ID from.
  * @returns The user ID, if available.
  */
-export function userId({ user_id }: JwtPayload): string {
-  logger.debug(`JWT user_id is: ${user_id}.`);
-  return user_id;
+export function userId(jwtPayload: JwtPayload): string {
+  // IAS tokens use user_uuid, XSUAA tokens use user_id
+  const id = jwtPayload.user_uuid || jwtPayload.user_id;
```
davidkna-sap: TODO `user_uuid` has not been included in the JWTs so far.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +45,134 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+    // Determine authentication method
+    if (certificate && key) {
+      // mTLS authentication
+      logger.debug('Using certificate-based authentication for IAS token.');
+      requestConfig.httpsAgent = new https.Agent({
+        cert: certificate,
+        key
+      });
+    } else if (clientsecret) {
+      logger.debug('Using client secret authentication for IAS token.');
```
davidkna-sap: Unable to test this so far, because even if multi-tenant is disabled (requirement) it does not let me create client-secret style keys.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -179,6 +182,37 @@ async function xfS4hanaCloudBindingToDestination(
...
+  }
+): Promise<Destination> {
+  const { access_token } = await getIasClientCredentialsToken(service, {
+    appName: options?.appName,
+    appTenantId: options?.appTenantId,
+    extraParams: options?.extraParams
+  });
+  const destination = buildClientCredentialsDestination(
+    access_token,
+    options?.targetUrl ?? service.credentials.url,
```
KavithaSiva: [q] `targetUrl` is  for IAS App2App, but taking `service.credentials.url` takes the IAS service credentials. 
Did you also think about supporting the IAS App2Service use case here? Do they have to supply the service's url as target url?

packages/connectivity/src/scp-cf/jwt/jwt.ts
davidkna-sap: Included in buisiness-user jwts.

packages/connectivity/src/scp-cf/identity-service.ts
davidkna-sap: tested successfully.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -81,8 +92,77 @@ export interface DestinationFromServiceBindingOptions {
...
+   * Additional parameters to be sent along with the token request.
+   */
+  extraParams?: Record<string, string>;
+}
+
+/**
+ * IAS options for technical user authentication.
+ */
+type IasOptionsTechnical = IasOptionsBase & {
+  actAs?: 'technical-user';
```
davidkna-sap: Instead of `actAs` (like `onBehalfOf` in java SDK) it might be possible to re-use `AuthenticationType` here.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -81,8 +95,97 @@ export interface DestinationFromServiceBindingOptions {
...
+  {
+    /**
+     * The client ID of the application resource.
+     */
+    clientId: string;
+    /**
+     * The tenant ID of the application resource (Optional).
+     */
+    tenantId?: string;
+  }
```
KavithaSiva: [q] What use case does this support, shouldn't we always have a name of IAS App2App scenario?

packages/connectivity/src/scp-cf/environment-accessor/environment-accessor-types.ts
```diff
@@ -19,6 +21,10 @@ export interface Service {
    * The service credentials.
    */
   credentials: ServiceCredentials;
+  /**
+   * IAS Resource to request for App-to-App communication.
+   */
+  iasResource?: IasResource | IasResource[];
```
KavithaSiva: [req] This a representation of a service, VCAP_SERVICES in BTP, and the `iasResource` property doesn't fit in this interface.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -81,8 +95,97 @@ export interface DestinationFromServiceBindingOptions {
...
+  resource?: IasResource | IasResource[];
+  /**
+   * The consumer (BTP) tenant ID of the application.
+   * May be required for multi-tenant communication.
+   */
+  appTid?: string;
+  /**
+   * Additional parameters to be sent along with the token request.
+   */
+  extraParams?: Record<string, string>;
```
KavithaSiva: [q] Why are all fields in this interface optional? Technically, if `targetUrl` is present isn't the `resource` also required?

KavithaSiva (commented):
Review in progress, added a few comments for now.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
davidkna-sap: I'm not sure exactly but this is a feature that the Java SDK ended up having to support and integrating it now ensures the API accounts for it. [These docs](https://github.wdf.sap.corp/pages/CPSecurity/sci-dev-guide/docs/Authentication/Exposing-APIs#consumer-consuming-apis) just talk about "certain cases".

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
KavithaSiva: Discussed this already, so consider it resolved.

davidkna-sap: I'm not sure if this is the best way to handle this would be for non app2app. Being able to supply this in general may be useful because in general there is not standard way to either discover both the IAS dependency name to use with the identity service or the identity service instance to use.

packages/connectivity/src/scp-cf/environment-accessor/environment-accessor-types.ts
davidkna-sap: Removed.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
davidkna-sap: - `targetUrl` has a fallback to `service.credentials.url`
- `resource` is only required for App2App
- `appTid` is an optional parameter
- `extraParams` should be optional

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -12,6 +45,157 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+    context: {
+      uri: fnArgument.serviceCredentials.url,
+      tenantId: fnArgument.serviceCredentials.tenantid
+    }
+  }).catch(err => {
+    throw new Error(
+      `Could not fetch IAS client credentials token for service of type ${resolvedService.label}: ${err.message}`
+    );
+  });
+  return token;
```
davidkna-sap, line 124: `xssec` also provides an `IdentityToken` class which could be returned here instead.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -12,6 +45,157 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+    access_token: response.access_token,
+    token_type: response.token_type,
+    expires_in: response.expires_in,
+    // IAS tokens don't have scope property
+    scope: '',
+    jti: decodedJwt.jti ?? '',
+    aud: decodedJwt.aud,
+    // Added if resource parameter was specified
+    ias_apis: decodedJwt?.ias_apis
+  };
```
davidkna-sap, line 241: Same as above.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
davidkna-sap: This file has some more comments/explanations on the different options: https://github.com/SAP/cloud-sdk-java/blob/main/cloudplatform/cloudplatform-connectivity/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/BtpServiceOptions.java

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -12,6 +45,157 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+  const identityService = getIdentityServiceInstanceFromCredentials(
+    arg.serviceCredentials
   );
+
+  const authenticationType =
+    arg.authenticationType || 'OAuth2ClientCredentials';
+
+  const tokenOptions: IdTokenFetchOptions = {
+    token_format: 'jwt'
+  };
```
davidkna-sap: In the e2e multi-tenant testing with user token exchange this was required because otherwise IAS returns an opaque token.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
KavithaSiva: For the App2App case, both `targetUrl` and `resource` are required. 
Even if they are optional here, we have to document this. Consider adding to JSDoc and to the documentation portal.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -81,8 +95,97 @@ export interface DestinationFromServiceBindingOptions {
    * Custom transformation function to control how a {@link Destination} is built from the given {@link Service}.
    */
   serviceBindingTransformFn?: ServiceBindingTransformFunction;
+  /**
+   * Options for IAS token retrieval.
+   */
+  iasOptions?: IasOptions;
```
KavithaSiva: [req] As this interface is used by `getDestination` and couple of other places but is only used for one specific functionality which is for transforming a service binding to a destination, I would recommend that you move it to `getDestinationFromServiceBinding` instead.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
KavithaSiva, line 85: [req] Pass `iasOptions` here alongside `useCache` and `jwt`.

packages/connectivity/src/scp-cf/jwt/jwt.ts
```diff
@@ -62,18 +68,39 @@ export function getTenantId(
...
+
 /**
  * Check if the given JWT is not an IAS token.
  * Currently, there are only two domains for IAS tokens:
- * `accounts.ondemand.com` and `accounts400.onemand.com`.
+ * `accounts.ondemand.com` and `accounts400.ondemand.com`.
  * @param decodedJwt - The decoded JWT to check.
  * @returns Whether the given JWT is not an IAS token.
  */
 function isNotIasToken(decodedJwt: JwtPayload): boolean {
```
KavithaSiva: [req] Remove this function and replace it's usages with `!isIasToken()` instead.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
KavithaSiva: [req] Can you rename this fields to `providerClientId` and `providerTenantId` instead? This is how they are named in Java.

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -34,15 +37,38 @@ const ClientCredentialsTokenCache = (
...
+  if (!resource) {
+    return undefined;
+  }
+  if ('name' in resource) {
+    return `name=${resource.name}`;
+  }
+
+  let normalized = `clientid=${resource.clientId}`;
+  if (resource.tenantId) {
+    normalized += `:tenantid=${resource.tenantId}`;
```
KavithaSiva: [req] Use `provider-clientId` and `provider-tenantId` instead as it can be a bit confusing as we already have `clientId` and `tenantId` in the cache key.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -1,5 +1,37 @@
...
+ */
+export interface IasClientCredentialsResponse extends ClientCredentialsResponse {
+  /**
+   * Audience claim from the JWT token.
+   */
+  aud?: string | string[];
+  /**
+   * IAS API resources. Present when resource parameter is specified in the token request.
+   */
+  ias_apis?: string[];
```
KavithaSiva: [q] Should this be an array? Given we only pass one resource name?

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -12,6 +44,157 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+}
+
+type IasParameters = {
+  serviceCredentials: ServiceCredentials;
+} & IasOptions;
+
+type FirstArg<T> = T extends (arg1: infer U, ...args: any[]) => any ? U : never;
+type UnwrapPromise<T> = T extends Promise<infer U> ? U : T;
+
+// xssec does not properly export these
```
KavithaSiva: [suggestion] Please also raise this request with the xssec team.

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -75,15 +92,40 @@ export async function serviceToken(
...
+ * For IAS services, you can pass IAS-specific options like `resource` and `appTid` in `options.iasOptions`.
+ *
  * Throws an error if there is no instance of the given service type.
  * @param jwt - The JWT of the user for whom the access token should be fetched.
  * @param service - The type of the service or an instance of {@link Service}.
+ * @param options - Optional options to modify token fetching behaviour.
+ * @param options.iasOptions - Options to change IAS token fetching (see {@link IasOptions}).
  * @returns A JWT bearer token.
  */
 export async function jwtBearerToken(
```
KavithaSiva, line 84: [req] The changes to this function are to support  using an IAS token to fetch destinations using the destination service, and I would keep this out of this PR and do it in the the dedicated PR instead.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -12,6 +44,157 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+  return urn;
+}
+
+/**
+ * Implementation of the IAS client credentials token retrieval using @sap/xssec.
+ * @param arg - The parameters for IAS token retrieval.
+ * @returns A promise resolving to the client credentials response.
+ * @internal
+ */
+async function getIasClientCredentialsTokenImpl(
```
KavithaSiva: [req] I think in this function, we should also consider the `subdomain` from the incoming jwt and add that to token fetching options, Let's clarify this with Matthias next week.
Let's also clarify if `grant_type` needs to be added.

KavithaSiva (changes requested):
Added few more requests to be worked on.

packages/connectivity/src/scp-cf/identity-service.ts
davidkna-sap: Even for single items it returns an array in the jwt, and when a `providerClientId` is given instead of name it's possible for it's supposed to return a token for everything the `providerClientId` applies to (possibly multiple depencies with different API permissions on the same provider).

packages/connectivity/src/scp-cf/jwt/jwt.ts
```diff
@@ -62,18 +68,39 @@ export function getTenantId(
...
+ * Currently, there are only two domains for IAS tokens:
+ * `accounts.ondemand.com` and `accounts400.ondemand.com`.
+ * @param decodedJwt - The decoded JWT to check.
+ * @returns Whether the given JWT is an IAS token.
+ * @internal
+ */
+export function isIasToken(decodedJwt: JwtPayload): boolean {
+  if (!decodedJwt.iss) {
+    return false;
+  }
```
davidkna-sap, line 84: This could potentially check for `ias_apis`, `ias_iss` or `scim_id` but I'm not sure if exchanged-from-IAS XSUAA tokens could have these properties.

packages/connectivity/src/scp-cf/token-accessor.ts
davidkna-sap: Should the changes to `serviceToken` also be reverted? It is responsible for (some of) the caching.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
davidkna-sap: Updated the documentation to highlight the requirement for App-to-App.

packages/connectivity/src/scp-cf/identity-service.ts
davidkna-sap: `xssec` already handles the `grant_type`, after talking with Matthias I have implemented extracting the subdomain from the jwt.

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
```diff
@@ -0,0 +1,82 @@
...
+    // Reason: We don't want to blindly trust the URL in the assertion
+    const credentialsUrl = tryParseUrl(credentials.url, 'Identity Service');
+    const credentialsSplit = credentialsUrl.hostname.split('.');
+    credentialsUrl.hostname = [subdomain, ...credentialsSplit.slice(1)].join(
+      '.'
+    );
+    let normalizedUrl = credentialsUrl.toString();
+    if (normalizedUrl.endsWith('/')) {
+      normalizedUrl = normalizedUrl.slice(0, -1);
+    }
```
davidkna-sap, line 60: Instead of this it might be better to use `IdentityService.getSafeUrlFromTokenIssuer(token, credentials.domains)` and not rewrite the URL.

packages/connectivity/src/scp-cf/destination/destination-accessor-types.ts
```diff
@@ -51,6 +51,7 @@ export interface DestinationAccessorOptions {
    * ATTENTION: The property is mandatory in the following cases:
    * - User-dependent authentication flow is used, e.g., `OAuth2UserTokenExchange`, `OAuth2JWTBearer`, `OAuth2SAMLBearerAssertion`, `SAMLAssertion` or `PrincipalPropagation`.
    * - Multi-tenant scenarios with destinations maintained in the subscriber account. This case is implied if the `selectionStrategy` is set to `alwaysSubscriber`.
+   * - IAS token exchange with `iasOptions.authenticationType` set to `'OAuth2JWTBearer'`. In this case, the JWT is automatically used as the assertion for IAS token exchange.
```
KavithaSiva: [q] Where is `iasOptions.authenticationType` set? It's not clear from the comment.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -83,6 +101,100 @@ export interface DestinationFromServiceBindingOptions {
...
+   */
+  targetUrl?: string;
+  /**
+   * The application resource(s) for which the token is requested.
+   * The token will only be usable to call the requested application(s).
+   * Either provide the app name (common case) or the provider client ID
+   * and tenant ID (optional).
+   *
+   * It is recommended to also provide the targetUrl parameter, otherwise
+   * the destination will point to the identity service URL from the service bindingm,
```
KavithaSiva:
```suggestion
   * the destination will point to the identity service URL from the service binding,
```

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -83,6 +101,100 @@ export interface DestinationFromServiceBindingOptions {
...
+/**
+ * Base options shared by all IAS authentication modes.
+ */
+interface IasOptionsBase {
+  /**
+   * The target URL of the destination that the IAS token is requested for.
+   * It is recommended to provide this for App-to-App communication (when resource parameter is used),
+   * otherwise the destination will point to the identity service URL from the service binding,
+   * instead of the actual target application. This function is not able to to infer
+   * the target application from the information available.
```
KavithaSiva:
```suggestion
   * instead of the actual target application. 
```
[pp] Sentence is verbose and can be omitted.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -83,6 +101,100 @@ export interface DestinationFromServiceBindingOptions {
   serviceBindingTransformFn?: ServiceBindingTransformFunction;
 }
 
+/**
+ * The application resource for which the token is requested for App-to-App communication.
+ * The token will only be usable to call the requested application.
+ * Either provide the app name (common case) or the provider client ID
+ * and tenant ID (optional).
+ */
+export type IasResource = Xor<
```
KavithaSiva: [req] Consider moving all IAS related types into a dedicated file.

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
KavithaSiva: I don't understand, how does the xssec function help here? Wouldn't it just help in validation? The actual URL manipulation still needs to be done right?

packages/connectivity/src/scp-cf/identity-service.ts
KavithaSiva: I checked the usages of this function, and right now we use `ClientCredentialsResponse` base interface during caching. So, we would also have to change the caching function.

We can consider migrating to token classes available from xssec, with caching changes in a follow-up.

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -21,12 +30,17 @@ import type { JwtPayload } from './jsonwebtoken-type';
...
+ * @param options.iasOptions - Options to change IAS token fetching (see {@link IasOptions}).
  * @returns Access token.
  */
 export async function serviceToken(
   service: string | Service,
   options?: CachingOptions & {
     jwt?: string | JwtPayload;
+    iasOptions?: Omit<
+      IasOptions,
+      'authenticationType' | 'assertion' | 'destinationUrl'
```
KavithaSiva:
```suggestion
      'authenticationType' | 'assertion' | 'targetUrl'
```

packages/connectivity/src/scp-cf/token-accessor.ts
KavithaSiva: Yes. I think we can also revert the changes to `serviceToken` function for now as it also is mostly used with functions calling a service using an access token. Other usages might require more checks.

The `transformServiceBindingToClientCredentialsDestination` seems to be an exception, but I think we can add IAS support to this function later.

.changeset/slow-cars-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Support IAS (App-to-App) authentication (experimental)
```
KavithaSiva: [req] Also add the functions that support this use-case: `transformServiceBindingToDestination` and `getDestinationFromServiceBinding`

packages/connectivity/src/scp-cf/destination/destination-from-vcap.spec.ts
```diff
@@ -267,6 +267,26 @@ describe('vcap-service-destination', () => {
 
     expect(destination?.authTokens?.[0]).toMatchObject({ value: jwt });
   });
+
```
KavithaSiva, line 271: [pp] Add tests that also show the IAS service binding (`destinationName`) and assertion with resource-name being transformed, as this would be one of the recommended App2App ways.

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
davidkna-sap: xssec uses the function to also determine the issuer for token verification, so in this case it could be enough verification to use the issuer url as is in this case. Additionally it has the advantage of catching cases of `accounts400` vs `accounts` and any potential custom domains for ias.

packages/connectivity/src/scp-cf/destination/destination-accessor-types.ts
davidkna-sap: I tried to clarify the comment, though arguably the IAS-case is covered by "User-dependent authentication flow".

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -71,6 +88,92 @@ export async function serviceToken(
...
+ * If a JWT is passed, the tenant of the JWT will be used when performing the grant.
+ * When no JWT is passed, the grant will be performed using the provider tenant.
+ *
+ * Throws an error if there is no instance of the given service type or if the request to the IAS service fails.
+ * @param service - The type of the service or an instance of {@link Service}.
+ * @param options - Options to influence caching behavior (see {@link CachingOptions}) and IAS-specific options (see {@link IasOptions}).
+ * @returns Access token.
+ * @experimental This function is experimental and may change in future versions.
+ */
+export async function serviceTokenIas(
```
KavithaSiva: [req] I looked at the code again, and the `serviceToken`(although the naming is misleading) is only used for fetching the token using client credentials grant and never the jwt bearer grant. 

Currently, you are changing the behaviour of this function, please remove this code for now as the `serviceToken` is only used during App2Service workflows.

Your current implementation only works if the service parameter is given as `identity` and doesn't work if the service is for e.g. `destination`, in this case, XSUAA tokens are always obtained.

We should also support fetching IAS tokens for the destination service or other services.

packages/connectivity/src/scp-cf/token-accessor.ts
davidkna-sap: I've moved the IAS-specific parts of `serviceToken` to `serviceTokenIas` for now and marked it as experimental to keep the client credentials cache for IAS around.

.changeset/slow-cars-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[New Functionality] Support IAS (App-to-App) authentication (experimental). Use `transformServiceBindingToDestination` and `getDestinationFromServiceBinding` to create a destination from an IAS service binding.
```
KavithaSiva:
```suggestion
[New Functionality] Support IAS (App-to-App) authentication (experimental). Use `transformServiceBindingToDestination()` function or `getDestinationFromServiceBinding()` function to create a destination targeting an IAS application.
```

packages/connectivity/src/scp-cf/destination/ias-types.ts
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Additional parameters to be sent along with the token request.
+   */
+  extraParams?: Record<string, string>;
+}
+
+/**
+ * IAS options for technical user authentication (client credentials).
+ */
+type IasOptionsTechnical = IasOptionsBase & {
```
KavithaSiva: [suggestion] In Java, we have the option to set an Onbehalf value to `TECHNICAL_USER_CURRENT_TENANT` or `TECHNICAL_USER_PROVIDER`.
To reflect these options here, what about introducing a boolean field that differentiates between the two?

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +64,159 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+  };
+
+  // Stringify resource
+  if (arg.resource) {
+    tokenOptions.resource = convertResourceToUrn(arg.resource);
+  }
+
+  if (arg.appTid) {
+    tokenOptions.app_tid = arg.appTid;
+  }
```
KavithaSiva: [req] I would use the suggested bool flag above in `IasOptionsTechnical` to determine if the provider tenant id(app_tid from the ias credentials) or the tenant id(from incoming jwt) needs to be supplied here.

davidkna-sap: Done, but it still needs unit-tests.

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -37,8 +37,9 @@ const ClientCredentialsTokenCache = (
 /** *
  * @internal
  * @param tenantId - The ID of the tenant to cache the token for.
- * @param clientId - ClientId to fetch the token
- * @returns the token
+ * @param clientId - ClientId to fetch the token.
```
marikaner, line 83:
```suggestion
 * @param clientId - Client ID to fetch the token.
```

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -37,8 +37,9 @@ const ClientCredentialsTokenCache = (
...
+ * @param resource - Optional resource parameter (for IAS app2app scenarios).
```
marikaner: [req] There is no resource parameter 😅

packages/connectivity/src/scp-cf/destination/destination-accessor-types.ts
```diff
@@ -51,6 +51,7 @@ export interface DestinationAccessorOptions {
    * ATTENTION: The property is mandatory in the following cases:
    * - User-dependent authentication flow is used, e.g., `OAuth2UserTokenExchange`, `OAuth2JWTBearer`, `OAuth2SAMLBearerAssertion`, `SAMLAssertion` or `PrincipalPropagation`.
    * - Multi-tenant scenarios with destinations maintained in the subscriber account. This case is implied if the `selectionStrategy` is set to `alwaysSubscriber`.
+   * - IAS business user authentication (OAuth2JWTBearer `authenticationType`). In this case, the JWT is used as the assertion for IAS token exchange. The `authenticationType` is set via the `iasOptions` parameter when used with {@link getDestinationFromServiceBinding}.
```
marikaner: [req] Use a descriptive noun when using inline code (see [styleguide](https://github.com/SAP/cloud-sdk/blob/main/STYLEGUIDE.md#include-a-descriptive-noun-before-or-after-code))
```suggestion
      * - IAS business user authentication (OAuth2JWTBearer). In this case, the JWT is used as the assertion for the IAS token exchange. The authentication type is set via the `iasOptions` parameter when used with {@link getDestinationFromServiceBinding}.
```

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -24,22 +25,33 @@ const logger = createLogger({
  * Throws an error if no services are bound at all, no service with the given name can be found, or the service type is not supported.
  * The last error can be circumvent by using the second parameter to provide a custom function that transforms a service binding to a destination.
  * @param options - Options to customize the behavior of this function.
+ * @param options.iasOptions - Options for IAS token retrieval in case of IAS authentication.
```
marikaner: I think it is a little weird to single out one specific property of the options. It should be described in the type itself. The documentation generation used to have issues with this style of referencing properties, not sure if this improved, though. (also see next comment)

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -24,22 +25,33 @@ const logger = createLogger({
...
  * @returns A destination.
  */
 export async function getDestinationFromServiceBinding(
   options: Pick<
     DestinationFetchOptions,
     'jwt' | 'iss' | 'useCache' | 'destinationName'
   > &
-    DestinationFromServiceBindingOptions
+    DestinationFromServiceBindingOptions & { iasOptions?: IasOptions }
```
marikaner, line 35: [q] Why do we need the extra `iasOptions` property here? It is already in the `DestinationFromServiceBindingOptions` type.

packages/connectivity/src/scp-cf/destination/ias-types.ts
```diff
@@ -0,0 +1,98 @@
...
+     */
+    name: string;
+  },
+  {
+    /**
+     * The client ID of the application resource.
+     */
+    providerClientId: string;
+    /**
+     * The tenant ID of the application resource (Optional).
```
marikaner:
```suggestion
     * The tenant ID of the application resource.
```

packages/connectivity/src/scp-cf/destination/ias-types.ts
```diff
@@ -0,0 +1,98 @@
+import type { Xor } from '@sap-cloud-sdk/util';
+import type { AuthenticationType } from './destination-service-types';
+
+/**
+ * The application resource for which the token is requested for App-to-App communication.
+ * The token will only be usable to call the requested application.
+ * Either provide the app name (common case) or the provider client ID
+ * and tenant ID (optional).
```
marikaner: [pp] The type already suggests, that this is optional. In my opinion it only makes sense to mention that, if the consequence of making it optional is also explained. Aka. something along the lines of "Omit the tenant ID to use the provider account". I wouldn't do that either unless it is something unusual.

```suggestion
 * and tenant ID.
```

packages/connectivity/src/scp-cf/destination/ias-types.ts
```diff
@@ -0,0 +1,98 @@
...
+   * It is recommended to provide this for App-to-App communication (when resource parameter is used),
+   * otherwise the destination will point to the identity service URL from the service binding.
+   * @default The (identity service) URL from the service binding.
+   */
+  targetUrl?: string;
+  /**
+   * The application resource(s) for which the token is requested.
+   * The token will only be usable to call the requested application(s).
+   * Either provide the app name (common case) or the provider client ID
+   * and tenant ID (optional).
```
marikaner:
```suggestion
   * and tenant ID.
```

packages/connectivity/src/scp-cf/destination/ias-types.ts
```diff
@@ -0,0 +1,98 @@
...
+   *
+   * It is recommended to also provide the targetUrl parameter, otherwise
```
marikaner:
```suggestion
   * It is recommended to also provide the `targetUrl` parameter, otherwise
```

packages/connectivity/src/scp-cf/destination/ias-types.ts
```diff
@@ -0,0 +1,98 @@
...
+   */
+  resource?: IasResource;
+  /**
+   * The consumer (BTP) tenant ID of the application.
+   * May be required for multi-tenant communication.
+   */
+  appTid?: string;
+  /**
+   * Specifies whether the token request is made in the context of the current tenant or the provider tenant.
+   * @default 'current-tenant'
```
marikaner: `@default` seems to be incorrect: https://typedoc.org/documents/Tags._defaultValue.html
```suggestion
   * @defaultValue 'current-tenant'
```

packages/connectivity/src/scp-cf/destination/ias-types.ts
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Additional parameters to be sent along with the token request.
+   */
+  extraParams?: Record<string, string>;
+}
+
+/**
+ * IAS options for technical user authentication (client credentials).
+ */
+type IasOptionsTechnical = IasOptionsBase & {
```
marikaner: For consistency I would prefer this to be `IasOptionsTechnicalUser`, because the other type is called  `IasOptionsBusinessUser`.

packages/connectivity/src/scp-cf/destination/ias-types.ts
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Authentication type. Use 'OAuth2ClientCredentials' for technical user (default).
```
marikaner: [nth] Use `@defaultValue`.

packages/connectivity/src/scp-cf/destination/ias-types.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * IAS options for business user authentication (JWT bearer).
+ */
+type IasOptionsBusinessUser = IasOptionsBase & {
+  /**
+   * Authentication type. Use 'OAuth2JWTBearer' for business user authentication.
+   */
+  authenticationType: Extract<AuthenticationType, 'OAuth2JWTBearer'>;
+  /**
+   * The JWT assertion string to use for business user authentication (required).
```
marikaner:
```suggestion
   * The JWT assertion string to use for business user authentication.
```

packages/connectivity/src/scp-cf/destination/ias-types.ts
```diff
@@ -0,0 +1,98 @@
...
+   */
+  assertion: string;
+};
+
+/**
+ * Options for IAS token retrieval with type-safe authenticationType/assertion relationship.
```
marikaner:
```suggestion
 * Options for IAS token retrieval with type-safe authentication type/assertion relationship.
```

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
```diff
@@ -0,0 +1,82 @@
+import { IdentityService, IdentityServiceToken } from '@sap/xssec';
+import { ErrorWithCause } from '@sap-cloud-sdk/util';
+import type { IdentityServiceCredentials } from './environment-accessor-types';
+
+const identityServices: Record<string, IdentityService> = {};
+
+/**
+ * @internal
+ * Clears the cache of Identity services.
```
marikaner:
```suggestion
 * Clears the cache of identity services.
```

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Should only be used for testing purposes.
```
marikaner: [req] Then it really should be located in the tests, not the productive code. I know this can be tricky, how about you export the `identityServices` internally instead?

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
```diff
@@ -0,0 +1,82 @@
...
+  try {
+    return new URL(url);
+  } catch (err) {
+    throw new ErrorWithCause(`Could not parse ${name} URL: ${url}`, err);
+  }
+}
+
+/**
+ * @internal
+ * @param credentials - Identity service credentials extracted from a service binding or re-use service. Required to create the xssec IdentityService instance.
```
marikaner:
```suggestion
 * @param credentials - Identity service credentials extracted from a service binding or re-use service. Required to create the xssec `IdentityService` instance.
```

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
```diff
@@ -0,0 +1,82 @@
...
+ * @param assertion - Optional JWT assertion to extract the issuer URL for bearer assertion flows.
+ * @param disableCache - Value to enable or disable JWKS cache in xssec library. Defaults to false.
```
marikaner:
```suggestion
 * @param disableCache - Value to enable or disable JWKS cache in the xssec library. Defaults to false.
```

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
```diff
@@ -0,0 +1,82 @@
...
+ * @returns An instance of {@code @sap/xssec/IdentityService} for the provided credentials.
```
marikaner: [req] `@code` does not seem to be supported: https://typedoc.org/documents/Tags.html

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
marikaner: @davidkna-sap Are you aware of this module? => `packages/connectivity/src/scp-cf/subdomain-replacer.ts`

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +65,169 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+    throw new ErrorWithCause(
+      message + (err.message ? `: ${err.message}` : '.'),
+      err
+    );
+  });
+  return token;
+}
+
+/**
+ * Converts an IasResource to the URN format expected by @sap/xssec.
```
marikaner:
```suggestion
 * Converts an IAS resource to the URN format expected by @sap/xssec.
```

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -21,12 +22,17 @@ import type { JwtPayload } from './jsonwebtoken-type';
  * Throws an error if there is no instance of the given service type or the XSUAA service, or if the request to the XSUAA service fails.
  * @param service - The type of the service or an instance of {@link Service}.
  * @param options - Options to influence caching behavior (see {@link CachingOptions}) and a JWT. By default, caching and usage of a circuit breaker are enabled.
+ * @param options.iasOptions - Options to change IAS token fetching (see {@link IasOptions}).
```
marikaner: I believe it is best to describe this in the `options` parameter description. Otherwise the documentation generation will probably look bad. (feel free to check though)

marikaner (changes requested):
LGTM API-wise, I left quite a lot of comments (many minor things) and I didn't look at the tests for now. If there is anything you would like to have reviewed specifically, let me know.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
davidkna-sap: `iasOptions` only is in `ServiceBindingTransformOptions`, `DestinationFromServiceBindingOptions` does not include `iasOptions`.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
davidkna-sap: Removed, that said the generated documentation seemed fine.

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
davidkna-sap: I migrated the code to use `packages/connectivity/src/scp-cf/subdomain-replacer.ts` with some IAS-specific changes.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
marikaner: Oops, then I saw that incorrectly 👓

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const logger = createLogger({
+  package: 'connectivity',
+  messageContext: 'ias'
+});
+
+/**
+ * @internal
+ * A cache for IdentityService instances.
+ * Should only be used for testing purposes.
```
marikaner: [pp] I believe this should not only be used for testing purposes ;) only direct access outside this module is intended for testing purposes. Maybe that's splitting hairs though.

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
marikaner: > xssec uses the function to also determine the issuer for token verification, so in this case it could be enough verification to use the issuer url as is in this case. Additionally it has the advantage of catching cases of `accounts400` vs `accounts` and any potential custom domains for ias.

Why don't we use it then? is there a downside to it?

packages/eslint-config/index.js
```diff
@@ -220,7 +220,16 @@ module.exports = {
...
       'error',
-      { definedTags: ['packageDocumentation', 'typeParam', 'experimental'] }
+      {
+        definedTags: [
+          'packageDocumentation',
+          'typeParam',
+          'experimental',
+          'defaultValue'
+        ],
+        // The other default-allowed tags are not supported by tsdoc
```
marikaner, line 230: 🥴

marikaner (commented):
I left one question. Good work!

packages/connectivity/src/scp-cf/environment-accessor/ias.ts
davidkna-sap: I didn't use it because for better consistency with the handling in XSUAA and the Java-SDK (where it seems to work well).

marikaner: Agreed, let's avoid scope creep. 👍

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -179,6 +182,31 @@ async function xfS4hanaCloudBindingToDestination(
   };
 }
 
+async function iasBindingToDestination(
+  service: Service,
+  options?: ServiceBindingTransformOptions
+): Promise<Destination> {
+  const { access_token } = await getIasClientCredentialsToken(service, {
```
KavithaSiva: [req] The fetched token is not cached, but the user can provide caching options, can you please also add caching for these tokens?
Use the `clientCredentialsTokenCache`, we use it for all other transformation functions.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +65,163 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+    let message = `Could not fetch IAS client credentials token for service "${serviceName}" of type ${resolvedService.label}`;
+
+    // Add contextual hints based on error status code (similar to Java SDK)
+    if (err.response?.status === 401) {
+      message +=
+        '. In case you are accessing a multi-tenant BTP service on behalf of a subscriber tenant, ensure that the service instance is declared as dependency to SaaS Provisioning Service or Subscription Manager (SMS) and subscribed for the current tenant';
+    }
+
+    throw new ErrorWithCause(
+      message + (err.message ? `: ${err.message}` : '.'),
```
marikaner, line 120:
```suggestion
      message + (err.message || '.'),
```

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +65,163 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+    tokenOptions.resource = convertResourceToUrn(arg.resource);
+  }
+
+  if (arg.appTid) {
+    tokenOptions.app_tid = arg.appTid;
+  }
+
+  // Add any extra parameters
+  if (arg.extraParams) {
+    Object.assign(tokenOptions, arg.extraParams);
```
marikaner: [q] Why `Object.assign` and not the spread operator? (It would need to be `let` instead of `const` then)

```suggestion
    tokenOptions = {...tokenOptions, ...(arg.extraParams || {})}; 
```

packages/connectivity/src/scp-cf/subdomain-replacer.ts
```diff
@@ -34,3 +51,31 @@ function isValidUrl(url: string): boolean {
...
+    throw new Error(`Base URL is not a valid URL: "${baseUrl}".`);
+  }
+
+  const url = new URL(baseUrl);
+  const hostParts = getHost(url).split('.');
+  url.hostname = [newSubdomain, ...hostParts.slice(1)].join('.');
+
+  let result = url.toString();
+  // Remove trailing slash for consistency
+  if (result.endsWith('/')) {
```
marikaner: I am quite sure that there is a util function for this (`removeTrailingSlash()`)

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -34,15 +37,38 @@ const ClientCredentialsTokenCache = (
   getCacheInstance: () => cache
 });
 
+/**
+ * Normalizes the IAS resource parameter to a consistent string format for cache key.
+ * @param resource - The resource parameter from iasOptions.
+ * @returns Normalized resource string or empty string if not provided.
```
marikaner, line 58: This is never an empty string only undefined.

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -56,7 +82,12 @@ export function getCacheKey(
     );
     return;
   }
-  return [tenantId, clientId].join(':');
+  const parts = [tenantId, clientId];
+  const resourceStr = normalizeResource(resource);
+  if (resourceStr) {
+    parts.push(resourceStr);
+  }
+  return parts.join(':');
```
marikaner: [pp] I find it a little odd, that the `normalizeResource()` function transforms the resource to a string already and potentially already include the separator(`":"`). If we were to change the separator, we would have to do this in two places. Anyways, it could be helpful to instead change `normalizeResource()`, so that it returns an array instead. Then you wouldn't have to check for undefined here. Only my opinion, though.

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -12,16 +13,18 @@ const ClientCredentialsTokenCache = (
...
+    clientId: string,
+    resource?: IasResource
   ): ClientCredentialsResponse | undefined =>
-    cache.get(getCacheKey(tenantId, clientId)),
+    cache.get(getCacheKey(tenantId, clientId, resource)),
 
   cacheToken: (
     tenantId: string | undefined,
     clientId: string,
+    resource: IasResource | undefined,
```
marikaner: With 4 parameters it might be time to think about a different API, e.g. passing an object, but deciding/doing that might be out of scope of this PR...

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -44,7 +45,8 @@ export async function serviceToken(
   if (opts.useCache) {
     const cachedToken = clientCredentialsTokenCache.getToken(
       tenantForCaching,
-      serviceCredentials.clientid
+      serviceCredentials.clientid,
+      undefined
```
marikaner: [req] This is unnecessary.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -179,6 +187,88 @@ async function xfS4hanaCloudBindingToDestination(
...
+  } as IasOptions & CachingOptions;
+
+  let accessToken: string | undefined;
+  let iasTenant: string | undefined;
+  let tenantCacheKey: string | undefined;
+
+  // Technical user client credentials grant preperation
+  if (iasOptions.authenticationType === 'OAuth2ClientCredentials') {
+    if (!iasOptions.appTid) {
+      const requestAs = iasOptions.requestAs ?? 'current-tenant';
```
marikaner: [pp] I noticed, that you often use `??` which is more exact and I guess most of the time correct. However, I believe `||` could cover cases that we don't forsee sometimes, e.g. if this was an empty string. I would suggest to only use `??` where it would be wrong to use `||`.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -179,6 +187,88 @@ async function xfS4hanaCloudBindingToDestination(
...
+      if (requestAs === 'provider-tenant') {
+        iasOptions.appTid = service.app_tid;
+      } else if (requestAs === 'current-tenant') {
+        iasOptions.appTid = options?.jwt?.app_tid;
+      }
```
marikaner: [pp] I think it is not necessary to assign the `requestAs` variable, which is subsequently unused.

```suggestion
      if (iasOptions.requestAs === 'provider-tenant') {
        iasOptions.appTid = service.app_tid;
      } else if (iasOptions.requestAs === 'current-tenant' || !iasOptions.requestAs) {
        iasOptions.appTid = options?.jwt?.app_tid;
      }
```

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -179,6 +187,88 @@ async function xfS4hanaCloudBindingToDestination(
...
+
+    if (iasOptions.useCache) {
+      iasTenant = parseUrlAndGetHost(service.credentials.url);
+      tenantCacheKey = new URLSearchParams({
+        iasTenant,
+        ...(iasOptions.appTid && { appTid: iasOptions.appTid })
+      }).toString();
+
+      const cached = clientCredentialsTokenCache.getToken(
+        tenantCacheKey,
```
marikaner: [req] This feels very hacky to me. Normally you should parse the tenant ID, now we are passing our own cache key to create a new key from. I would recommend to add another function to the caching to differentiate different types of tokens (if I am not mistaken those are XSUAA and IAS). Currently it seems to me that we are trying to generalize over two very different things.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -179,6 +187,88 @@ async function xfS4hanaCloudBindingToDestination(
...
+        iasOptions.resource
+      );
+      if (cached) {
+        accessToken = cached.access_token;
+      }
+    }
+  }
+
+  let response: IasClientCredentialsResponse | undefined;
+  if (!accessToken) {
```
marikaner: [req/pp] I find the flow of this function quite hard to read, because it mixes many concerns. I think it would help to extract a few functions and apply a more functional style to this.

To understand, that `accessToken` could only be defined if there was a cached token needs you to read the code above. Instead, I think it would be nicer to extract all code related to getting the token into one function that returns the token. That way you would never have to assign the token to a variable and could return it directly, when you find it in the cache.

Generally, I think it is better style if you try to avoid `let` as much as possible. If you have to use `let` too often, it is a code smell in my opinion.

packages/connectivity/src/scp-cf/identity-service.ts
davidkna-sap: This would lead to the error message being appended without a space to separate it from the preceding  message.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
davidkna-sap: I split off the code for determinig `app_tid` and building the destination from options and the token to ease readability.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
davidkna-sap: I migrated IAS token caching functions to IAS-specific `getTokenIas` / `cacheTokenIas` methods.

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -28,17 +29,59 @@ const ClientCredentialsTokenCache = (
         : undefined
     });
   },
+
+  getTokenIas: (
```
marikaner, line 33: [pp] I would call it `getIasToken()` and `cacheIasToken()`.

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -59,6 +102,60 @@ export function getCacheKey(
...
+  if (!clientId) {
+    logger.warn(
+      'Cannot create cache key for client credentials token cache. The given client ID is undefined.'
+    );
+    return;
+  }
+
+  const output = [iasInstance, appTid || '', clientId];
+  const normalizedResource = normalizeResource(resource);
+  output.push(...normalizedResource);
```
marikaner: [pp] shorter and easier to read, but possibly harder to debug. matter of taste 😊 
```suggestion
  const output = [iasInstance, appTid || '', clientId, ...normalizeResource(resource)];
```

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -59,6 +102,60 @@ export function getCacheKey(
...
+
+/** *
+ * @internal
+ * @param data.iasInstance - The IAS instance (tenant) hostname the token is fetched from.
+ * @param data.appTid - The BTP instance (tenant) id (App-Tid)
+ * @param data.clientId - ClientId to fetch the token.
+ * @param data.resource - The App-To-App resource the token is scoped for.
+ * @returns The cache key.
+ */
+export function getCacheKeyIas(
```
marikaner: [pp] 
```suggestion
export function getIasCacheKey(
```

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -178,6 +185,113 @@ async function xfS4hanaCloudBindingToDestination(
...
   };
 }
+/**
+ * Tries to resolve `app_tid` based on supplied IAS options.
+ * @param iasOptions - IAS technical user options.
+ * @param service - Service binding for identity service.
+ * @param options - Service bidning transform options.
+ * @returns The BTP app_tid based on `requestAs` configuration.
+ */
+function iasGetAppTid(
```
marikaner: [req] Please start the function name with the verb: `getIasAppTid()`. Reads more natural. The [style guide ](https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-verbs-in-function-names) doesn't explicitly state that it should be the first word, but at least all examples follow this.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -178,6 +185,113 @@ async function xfS4hanaCloudBindingToDestination(
...
+): string {
+  const { requestAs } = iasOptions;
+  if (requestAs === 'provider-tenant') {
+    return service.app_tid;
+  }
+  if (requestAs === 'current-tenant' || !requestAs) {
+    return options?.jwt?.app_tid;
+  }
+
+  requestAs satisfies never;
```
marikaner, line 191: [q] What is the purpose of this line?

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -178,6 +185,113 @@ async function xfS4hanaCloudBindingToDestination(
...
+
+/**
+ * Builds destination based on supplied IAS options.
+ * Uses `targetUrl` as the destination URL if supplied and adds `mtlsKeyPair` if available.
+ * @param accessToken - The JWT token to access the service.
+ * @param service - Service binding for identity service.
+ * @param options - Service bidning transform options.
+ * @returns A destination object.
+ */
+function iasBuildDestination(
```
marikaner: (see above)

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -178,6 +185,113 @@ async function xfS4hanaCloudBindingToDestination(
...
+  if (service.credentials.certificate && service.credentials.key) {
+    destination.mtlsKeyPair = {
+      cert: service.credentials.certificate,
+      key: service.credentials.key
+    };
+  }
+  return destination;
+}
+
+async function iasBindingToDestination(
```
marikaner: [req] Add a verb 😊 => style guide

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -178,6 +185,113 @@ async function xfS4hanaCloudBindingToDestination(
...
+        clientId: service.credentials.clientid,
+        resource: options?.iasOptions?.resource
+      });
+      if (cached) {
+        return iasBuildDestination(cached.access_token, service, options);
+      }
+    }
+  }
+
+  const response = await getIasClientCredentialsToken(service, {
```
marikaner: [q] Do I understand correctly that we fetch a client credentials token, no matter what authentication type is used, but only cache it for "OAuth2ClientCredentials"? If so maybe a comment or so would be nice. Could also just be me not understanding. 🫣

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -178,6 +185,113 @@ async function xfS4hanaCloudBindingToDestination(
...
+    jwt: options?.jwt,
+    ...(options?.iasOptions || {})
```
marikaner: [q] Could we use `iasOptions` here?
```suggestion
    iasOptions
```

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -178,6 +185,113 @@ async function xfS4hanaCloudBindingToDestination(
...
+    iasOptions.authenticationType === 'OAuth2ClientCredentials' &&
+    iasOptions.useCache &&
+    response
+  ) {
+    clientCredentialsTokenCache.cacheTokenIas(
+      {
+        iasInstance,
+        appTid: iasOptions.appTid,
+        clientId: service.credentials.clientid,
+        resource: options?.iasOptions?.resource
```
marikaner: [q] same here, why not?
```suggestion
        resource: iasOptions?.resource
```

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
marikaner: Also, I think this reads a lot better now 👍

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -1,5 +1,59 @@
...
+  /**
+   * Custom issuer claim from the JWT token.
+   */
+  custom_iss?: string;
+  /**
+   * Application tenant ID claim from the JWT token.
+   */
+  app_tid?: string;
+  /**
+   * IAS tokens don't have scope property.
```
marikaner, line 49: [q] Should it then be something like `never` or `undefined`?

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +69,174 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+
+  if ('name' in resource) {
+    return `urn:sap:identity:application:provider:name:${resource.name}`;
+  }
+
+  let urn = `urn:sap:identity:application:provider:clientid:${resource.providerClientId}`;
+  if (resource.providerTenantId) {
+    urn += `:apptid:${resource.providerTenantId}`;
+  }
+  return urn;
```
marikaner: When I see `let` I get the urge to remove it. Not sure this is better though. Just an idea. Feel free to keep your version.

```suggestion
  const resourceUrn = [resource.provderClientId, resource.providerTenantId].filter(val=>val).join(':apptid');
  return `urn:sap:identity:application:provider:clientid:${resourceUrn}`;
```

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +69,174 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+}
+
+/**
+ * Implementation of the IAS client credentials token retrieval using @sap/xssec.
+ * @param arg - The parameters for IAS token retrieval.
+ * @returns A promise resolving to the client credentials response.
+ * @internal
+ */
+async function getIasClientCredentialsTokenImpl(
```
marikaner: [req] Every function is an implementation of something. The name currently sounds like we get some implementation. I assume this is not what is meant. Can we just remove the `Impl`?

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +69,174 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+  }
+
+  if (arg.appTid) {
+    tokenOptions.app_tid = arg.appTid;
+  }
+
+  // Add any extra parameters
+  if (arg.extraParams) {
+    tokenOptions = { ...tokenOptions, ...arg.extraParams };
+  }
```
marikaner: Please extract this to a separate function. This will again improve readability and allows you to remove the `let` altogether.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +69,174 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+    if (arg.resource && tokenOptions.refresh_expiry === undefined) {
+      tokenOptions.refresh_expiry = 0;
+    }
+
+    // Extract appTid from assertion if not provided
+    const token = new IdentityServiceToken(arg.assertion);
+    if (!tokenOptions.app_tid) {
+      // Set to `null` if not prevent xssec from also trying to extract it internally
+      tokenOptions.app_tid = token.appTid ?? null;
+    }
```
marikaner, line 187: Please see my other comment about extracting a function here. I would try add this code to this function (you would have to repeat `if (arg.authenticationType === 'OAuth2JWTBearer')` but I think that's ok).

marikaner: Something like that maybe?
```ts
function buildToken(...) {
const tokenOptions: IdentityService.TokenFetchOptions &
    IdentityService.IdentityServiceTokenFetchOptions = {
    token_format: 'jwt',
    ...arg.resource && { resource: convertResourceToUrn(arg.resource) },
    ...arg.appTid && { app_tid: arg.appTid },
    ...(arg.extraParams || {})
  };

  if (arg.authenticationType === 'OAuth2JWTBearer') {
    // JWT bearer grant for business user propagation
    if (!arg.assertion) {
      throw new Error(
        'JWT assertion required for authenticationType: "OAuth2JWTBearer". Provide iasOptions.assertion.'
      );
    }

    // Disable refresh token for App-To-App JWT bearer token exchange (recommended for better performance)
    if (arg.resource && tokenOptions.refresh_expiry === undefined) {
      tokenOptions.refresh_expiry = 0;
    }

    // Extract appTid from assertion if not provided
    const token = new IdentityServiceToken(arg.assertion);
    if (!tokenOptions.app_tid) {
      // Set to `null` if not prevent xssec from also trying to extract it internally
      tokenOptions.app_tid = token.appTid || null;
    }

    // Workaround for IAS bug
    // JAVA SDK: https://github.com/SAP/cloud-sdk-java/blob/61903347b607a8397f7930709cd52526f05269b1/cloudplatform/connectivity-oauth/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/OAuth2Service.java#L225-L236
    // Issue: https://jira.tools.sap/browse/SECREQ-5220
    if (tokenOptions.app_tid) {
      tokenOptions.refresh_expiry = 0;
    }
  }
  return tokenOptions;
}
```

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
davidkna-sap: It should type-check that all possible values of `requestAs` are handled before reaching `else`.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
davidkna-sap: Yes only technical user tokens are cached, afaict they are not cached elsewhere either (and the client credentials token cache is not keyed on `user_id`/`user_uuid`.

packages/connectivity/src/scp-cf/identity-service.ts
davidkna-sap: I wanted to stay compatible with the existing `ClientCredentialsResponse`, but considered that something of a breaking change (also XSUAA will likely always have this).

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
marikaner: Ah, I see. Thanks!

packages/connectivity/src/scp-cf/identity-service.ts
marikaner: ok

packages/connectivity/src/scp-cf/identity-service.ts
davidkna-sap: I think the name is fine `getIasClientCredentialsToken` is the inner implementation of `getIasClientCredentialsToken` which iirc calls it with additional resilience.

packages/connectivity/src/scp-cf/identity-service.ts
davidkna-sap: I changed it with a slightly different approach.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -178,11 +185,118 @@ async function xfS4hanaCloudBindingToDestination(
...
+        iasInstance,
+        appTid: iasOptions.appTid,
+        clientId: service.credentials.clientid,
+        resource: iasOptions?.resource
+      },
+      response
+    );
+  }
+
+  return buildIasDestination(response.access_token, service, options);
```
KavithaSiva: [req] I checked the flow again here and when `iasOptions.authenticationType` is set to  `OAuth2JWTBearer`, the built IasDestination is still always one with `OAuth2ClientCredentials` which is incorrect as we do the access token retrieval using the jwt bearer token flow.

So, the authentication type `OAuth2JWTBearer` should be retained.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +69,177 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+/**
+ * Make a client credentials request against the IAS OAuth2 endpoint.
+ * Supports both certificate-based (mTLS) and client secret authentication.
+ * @param service - Service as it is defined in the environment variable.
+ * @param options - Options for token fetching, including authenticationType to specify authentication mode, optional resource parameter for app2app, appTid for multi-tenant scenarios, and extraParams for additional OAuth2 parameters.
+ * @returns Client credentials token response.
+ * @internal
+ * @experimental
+ */
+export async function getIasClientCredentialsToken(
```
KavithaSiva: [suggestion] I think it is better if we rename this function to `getIasAccessToken` instead as it performs both client credentials and jwt bearer grant flows.

Same for `getIasClientCredentialsTokenImpl` as well, `getIasAccessTokenImpl` is better.

davidkna-sap: I renamed the function.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
davidkna-sap: I've addressed this by allowing `authenticationType` overrides for `buildClientCredentials`, and forwarding it in `buildIasDestination`.

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -178,11 +168,130 @@ async function xfS4hanaCloudBindingToDestination(
...
+ * @param name - The name of the destination.
+ * @param authentication - The authentication type for the destination. Defaults to 'OAuth2ClientCredentials'.
+ * @returns A destination object.
+ */
+function buildDestination(
   token: string,
   url: string,
-  name
+  name: string,
+  authentication: AuthenticationType = 'OAuth2ClientCredentials'
```
KavithaSiva: [pp] Restricting the authentication types to what we build currently with `bearer` tokens.

```suggestion
  authentication: AuthenticationType = 'OAuth2ClientCredentials'
  authentication: Extract<AuthenticationType, 'OAuth2ClientCredentials' | 'OAuth2JWTBearer'> = 'OAuth2ClientCredentials'
```

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
```diff
@@ -178,11 +168,130 @@ async function xfS4hanaCloudBindingToDestination(
...
+  const iasInstance = parseUrlAndGetHost(service.credentials.url);
+
+  // Technical user client credentials grant preperation
+  if (iasOptions.authenticationType === 'OAuth2ClientCredentials') {
+    if (!iasOptions.appTid) {
+      iasOptions.appTid = getIasAppTid(iasOptions, service, options);
+    }
+
+    if (iasOptions.useCache) {
+      const cached = clientCredentialsTokenCache.getTokenIas({
```
KavithaSiva, line 244: [suggestion] I would recommend that we also cache the tokens if it's obtained via jwt bearer flow. 
But, this can be done in a follow-up as I see that we have not cached such tokens before in other token fetch flows.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -1,5 +1,59 @@
...
+import type { JwtPayload } from './jsonwebtoken-type';
+
+export { identityServicesCache } from './environment-accessor';
+
+/**
+ * @internal
+ * Represents the response to an IAS client credentials request.
+ * Extends the XSUAA response with IAS-specific fields.
+ */
+export interface IasClientCredentialsResponse extends ClientCredentialsResponse {
```
KavithaSiva: [suggestion] It's better to be renamed as `IasTokenResponse` as currently you use it to represent the output of both client credential and jwt bearer flows. Adjust the JSDoc too.

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -15,3 +69,177 @@ export function shouldExchangeToken(options: DestinationOptions): boolean {
...
+    fn: getIasTokenImpl,
+    fnArgument,
+    context: {
+      uri: fnArgument.serviceCredentials.url,
+      tenantId: fnArgument.serviceCredentials.app_tid
+    }
+  }).catch(err => {
+    const serviceName =
+      typeof service === 'string' ? service : service.name || 'unknown';
+    let message = `Could not fetch IAS client credentials token for service "${serviceName}" of type ${resolvedService.label}`;
```
KavithaSiva: [pp]
```suggestion
    let message = `Could not fetch IAS token for service "${serviceName}" of type ${resolvedService.label}`;
```

packages/connectivity/src/scp-cf/destination/service-binding-to-destination.ts
davidkna-sap: I would rather handle that in a follow-up, I've added it to the BLI.

packages/connectivity/src/scp-cf/identity-service.ts
davidkna-sap: I've renamed it, FWIW it still extends `ClientCredentialsResponse`, but XSUAA also extends `ClientCredentialsResponse` for `UserTokenResponse`.

PR #6233, opened 2026-01-05
chore: Update to eslint 9

marikaner:
While integrating dependabot PRs, I noticed, that we are weirdly using two versions of eslint (8 + 9). I don't see a reason why this was done and I assume, that this was by mistake. Let me know, if I missed something.

### DIFF
--- a/.changeset/funny-adults-fold.md
+++ b/.changeset/funny-adults-fold.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/eslint-config': patch
+---
+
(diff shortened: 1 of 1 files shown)


packages/eslint-config/package.json
```diff
@@ -42,6 +42,6 @@
     "typescript-eslint": "^8.51.0"
   },
   "peerDependencies": {
-    "eslint": "^8.0.0"
+    "eslint": ">=8.0.0"
```
KavithaSiva, line 45: [q] Can this still remain 8 or should we bump this to 9 ?

marikaner: If I didn't make a mistake this should still work with 8. Technically bumping to 9 would be a breaking change. 🤓

PR #6276, opened 2026-01-27
feat: Support multipart requests

marikaner:
This PR adds the capability to use formdata in the openapi requestbuilder.

### DIFF
--- a/packages/openapi-generator/src/parser/media-type.ts
+++ b/packages/openapi-generator/src/parser/media-type.ts
@@ -68,7 +68,7 @@ export function parseMediaType(
       logger.warn(
         `Could not parse '${allMediaTypes}', because it is not supported. Generation will continue with 'any'. This might lead to errors at runtime.`
       );
-      return { schema: { type: 'any' }, mediaType: 'application/json ' };
+      return { schema: { type: 'any' }, mediaType: 'application/json' };
     }
 
     // There is only one media type


packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -51,32 +55,36 @@ export function parseMediaType(
...
       options
     );
 
-    if (!parsedMediaType) {
+    if (!parsedSchema) {
       logger.warn(
         `Could not parse '${allMediaTypes}', because it is not supported. Generation will continue with 'any'. This might lead to errors at runtime.`
       );
-      return { type: 'any' };
+      return { schema: { type: 'any' }, mediaType: 'application/json ' };
```
davidkna-sap:
```suggestion
      return { schema: { type: 'any' }, mediaType: 'application/json' };
```

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -182,6 +183,22 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
+    if (
+      pickValueIgnoreCase(this.parameters?.headerParameters, 'content-type') ===
+      'multipart/form-data'
+    ) {
+      const formData = new FormData();
+      for (const key in body) {
+        // TODO: is it enough to just append the body[key] or does it need to be stringified?
+        formData.append(key, body[key]);
+      }
+      return formData;
```
davidkna-sap, line 197: The parquet API requires filenames (`formData.append(key, body[key], 'name.parquet');`). One way to avoid this issue would be to use `File` (which supports filenames).

We also need to consider using the `form-data` package which has wide support for streaming input data.

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -28,15 +29,18 @@ export function parseTopLevelMediaType(
...
   if (bodyOrResponseObject) {
-    const mediaType = getMediaTypeObject(
+    const mediaTypeObject = getMediaTypeObject(
       bodyOrResponseObject,
       allowedMediaTypes
     );
-    const schema = mediaType?.schema;
-    if (schema) {
-      return parseSchema(schema, refs, options);
+
```
davidkna-sap, line 38: IIRC the spec has some rules for inferring the media type, otherwise we should perhaps keep the media type optional (also to avoid breaking changes).

PR #6278, opened 2026-01-27
feat: Improve support for HTTP-client request cancellation

davidkna-sap:
Required for SAP/ai-sdk-js-backlog#458.

This PR exposes `signal` in the public API, ensures the `signal` is forwarded to the CSRF-token fetch and adds unit tests to ensure it works as expected.

- [x] PR title adheres to [conventional commit guidelines](https://www.conventionalcommits.org).
- [x] Created a changeset `yarn changeset`
- [x] If applicable:
  - [x] Documented public API (TypeDoc).
  - [x] Checked that `yarn run doc` still works.


packages/http-client/src/csrf-token-middleware.ts
```diff
@@ -170,14 +175,21 @@ async function makeCsrfRequests(
...
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const { data, params, parameterEncoder, ...requestConfigWithoutData } =
-    requestConfig;
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  const {
+    data,
+    params,
+    parameterEncoder,
+    signal,
+    ...requestConfigWithoutData
```
ZhongpinWang: [q/req] Should `signal` not be part of the `requestConfigWithoutData`?

packages/http-client/src/csrf-token-middleware.ts
```diff
@@ -137,6 +137,11 @@ async function makeCsrfRequest(
     });
     return findCsrfHeader(response.headers);
   } catch (error) {
+    // Avoid repeat CSRF attempts if request was cancelled
+    if (error.code === 'ERR_CANCELED') {
+      throw error;
+    }
```
ZhongpinWang: [q] I think previously, the `signal` property was not explicitly exposed in the type, but it was usable and was passed to axios. If someone before used signal, this would be a behaviour change for them.

PR #6291, opened 2026-01-30
feat: Add Middleware for HTTP-request compression

davidkna-sap:
Required for SAP/ai-sdk-js-backlog#470


packages/http-client/src/compress-request-middleware.ts
```diff
@@ -0,0 +1,227 @@
...
+  | 'brotli'
+  | 'deflate'
+  | 'zstd';
+
+/**
+ * Options for different request compressors to configure their behavior (e.g., compression level).
+ */
+export interface RequestCompressorOptions {
+  /** Options to control gzip compression. */
+  gzip: zlib.ZlibOptions;
```
marikaner, line 60: [req] I think this is a copy paste error and should be GzipOptions?

packages/http-client/src/compress-request-middleware.ts
```diff
@@ -0,0 +1,227 @@
...
+  /**
+   * Options for the chosen compression algorithm, e.g. to control the compression effort.
+   */
+  compressOptions?: RequestCompressorOptions[C];
+  /**
+   * Compression mode. Can be 'auto', 'passthrough' or a boolean.
+   * - In 'auto' mode, the payload is compressed based on the `autoCompressMinSize` threshold.
+   * - In 'passthrough' mode, it is assumed that the payload is already compressed.
+   * - If `true` is provided, the payload will always be compressed.
+   * - If `false` is provided, the payload will never be compressed.
```
marikaner: [pp] I find it a little odd, that this is a combination of boolean and strings. Consider "always" and "never" as values. That is more explicit. If that is common practice, it's fine.

marikaner (approved):
I like this😊

packages/http-client/src/compress-request-middleware.ts
davidkna-sap: Both `zlib` and `deflate` use a common `zlib.ZlibOptions` type.

packages/http-client/src/compress-request-middleware.ts
davidkna-sap: I've renamed it to `always` and `never`. In addition, I've also renamed `passthrough` to `header-only` WDYT?

packages/http-client/src/compress-request-middleware.ts
```diff
@@ -0,0 +1,231 @@
...
+  const mode = options?.mode ?? 'auto';
+  if (mode === 'header-only') {
+    return 'header-only';
+  }
+  if (mode === 'never') {
+    return false;
+  }
+  if (mode === 'always') {
+    return true;
+  }
```
ZhongpinWang: [pp] Minor preference. Since we are using 'always' and 'never', maybe also return here string literal instead of true false. Just a pp.

packages/http-client/src/compress-request-middleware.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  const minSize = options?.autoCompressMinSize ?? 1024;
+  let payloadSize: number;
+  try {
+    payloadSize = Buffer.byteLength(payload as any);
+  } catch (e: any) {
+    throw new ErrorWithCause(
+      "Could not determine payload size for 'auto' compression decision.",
+      e
+    );
```
ZhongpinWang: [pp/req] Should we throw or rather do a soft warning log and fallback to not compressing option? I find it a bit annoying if people use `auto` but the whole app crashes without fallback because of our implementation failed.
[q] When will `Buffer.byteLength()` throw?

davidkna-sap: It will mainly throw if the input is not `<string> | <Buffer> | <TypedArray> | <DataView> | <ArrayBuffer> | <SharedArrayBuffer> `

packages/http-client/src/compress-request-middleware.ts
```diff
@@ -0,0 +1,236 @@
...
+   * Please note that not all servers support all algorithms.
+   * @defaultValue 'gzip'
+   */
+  algorithm?: C;
+  /**
+   * Options for the chosen compression algorithm, e.g. to control the compression effort.
+   */
+  compressOptions?: RequestCompressorOptions[C];
+  /**
+   * Compression mode. Can be 'auto', 'header-only' or a boolean.
```
marikaner: outdated
```suggestion
   * Compression mode. 
```

packages/http-client/src/compress-request-middleware.ts
marikaner: I think it's good.

packages/http-client/src/compress-request-middleware.ts
```diff
@@ -0,0 +1,236 @@
...
+      requestConfig.headers,
+      'content-encoding'
+    );
+    const algorithmValue = getContentEncodingValue(algorithm);
+    const targetValue = currentValue
+      ? `${currentValue}, ${algorithmValue}`
+      : algorithmValue;
+    requestConfig.headers = mergeIgnoreCase(requestConfig.headers, {
+      'content-encoding': targetValue
+    });
```
marikaner: [pp] I would again extract this to a separate function (e.g. `addContentEncodingHeader()`). That way the function name would replace the comment. I also think that the name `currentValue` is not very speaking in the context of this function, but good enough in an extracted function, that deals with the headers only.

davidkna-sap: I tried to change this, but did not like how it ended up looking, so I opted to rename the variables here.

packages/http-client/src/compress-request-middleware.ts
ZhongpinWang: I guess other types does not make sense to be sent in the context of the RPT service?

But looking at the function name `checkIfNeedsCompression()` it is supposed to just check if needed and this is Cloud SDK which is supposed to support all request `data` types. I still feel bad if we throw something in a check function. Not a clean design I would say. Also we are technically allowing user sending for example `number` to remote if using `never`. Just want to make the responsibility of this function clear.

ZhongpinWang: I would fallback to `return false` in case we cannot compress.

davidkna-sap: @ZhongpinWang 

> I guess other types does not make sense to be sent in the context of the RPT service?

For the parquet-endpoint `FormData` will be used too, but parquet is a binary format which can have internal compression. Compression with `FormData` might be per-item anyway.

> I would fallback to `return false` in case we cannot compress.

The compress functions have the same `<string> | <Buffer> | <TypedArray> | <DataView> | <ArrayBuffer> | <SharedArrayBuffer>` types, so I moved the payload size check to also run in `always`-mode to catch incompatible types.

In think throwing for incompatible compression algorithms is fine.

packages/http-client/src/compress-request-middleware.ts
```diff
@@ -0,0 +1,242 @@
...
+      if (needsCompression === 'skip') {
+        return middlewareOptions.fn(requestConfig);
+      }
+
+      // If the payload already has a Content-Encoding header, we need to preserve it and append our encoding to it.
+      const currentContentEncoding = pickValueIgnoreCase(
+        requestConfig.headers,
+        'content-encoding'
+      );
+      const ourContentEncoding = getContentEncodingValue(algorithm);
```
marikaner: [pp] `our` is weird. How about:
```suggestion
      const algorithmContentEncoding = getContentEncodingValue(algorithm);
```

packages/http-client/src/compress-request-middleware.ts
marikaner: [pp] Sorry to reopen this can... I disagree. This only makes things more complicated... This function should return a boolean, but only do what its name suggests: check if the payload should be compressed. Later, when it is used, we can then do this:

```ts
if (needsCompression || options.mode === 'header-only') {
 // add header
}

if (needsCompression) {
  // compressPayload
}

// call middleware chain
```

marikaner: [pp] Also, for better readability and brevity I would like to suggest the following flow for this function:

```ts
if (mode === "auto") {
// do the compression decision
return decision; // true or false
}

return mode === "always";

```

packages/http-client/src/compress-request-middleware.ts
```diff
@@ -0,0 +1,242 @@
...
+  /**
+   * Options for the chosen compression algorithm, e.g. to control the compression effort.
+   */
+  compressOptions?: RequestCompressorOptions[C];
+  /**
+   * Compression mode.
+   * - In 'auto' mode, the payload is compressed based on the `autoCompressMinSize` threshold.
+   * - In 'header-only' mode, it is assumed that the payload is already compressed. The middleware will only set the appropriate Content-Encoding header without modifying the payload.
+   * - If 'always' is provided, the payload will always be compressed.
+   * - If 'never' is provided, the payload will never be compressed.
```
marikaner:
```suggestion
   * - 'auto' - The payload is compressed based on the `autoCompressMinSize` threshold.
   * - 'header-only' - It is assumed that the payload is already compressed. The middleware will only set the appropriate Content-Encoding header without modifying the payload.
   * - 'always' - The payload will always be compressed.
   * - never' - The payload will never be compressed.
```

PR #6301, opened 2026-02-04
feat: Support multipart requests

davidkna-sap:
Related SAP/ai-sdk-js-backlog#469.

Continuation of #6276

Documentation: https://github.com/SAP/cloud-sdk/pull/2374

### DIFF
--- a/packages/openapi/src/openapi-request-builder.ts
+++ b/packages/openapi/src/openapi-request-builder.ts
@@ -79,8 +79,8 @@ class FormDataBuilder {
 
       const encoded: string | Blob =
         value instanceof Blob
-          ? this.encodeBlob({ key, value, metadata })
-          : this.encodeString({ key, value, metadata, allowedTypes });
+          ? this.encodeBlob({ key, value }, metadata)
+          : this.encodeString({ key, value }, metadata, allowedTypes);
 
       formData.append(key, encoded);
     }
@@ -103,21 +103,22 @@ class FormDataBuilder {
 
   /**
    * Encode a Blob value for FormData with appropriate content type handling.
-   * @param params - The parameters object.
+   * @param params - The key and value to encode.
    * @param params.key - The field name.
    * @param params.value - The Blob value to encode.
-   * @param params.metadata - Encoding metadata for this field.
+   * @param metadata - Encoding metadata for this field.
    * @returns Blob - The encoded Blob value.
    */
-  private encodeBlob({
-    key,
-    value,
-    metadata
-  }: {
-    key: string;
-    value: Blob;
-    metadata: EncodingMetadata;
-  }): Blob {
+  private encodeBlob(
+    {
+      key,
+      value
+    }: {
+      key: string;
+      value: Blob;
+    },
+    metadata: EncodingMetadata
+  ): Blob {
     const {
       contentType: targetContentType,
       isImplicit: targetIsImplicit,
@@ -171,24 +172,24 @@ class FormDataBuilder {
 
   /**
    * Encode a string value for FormData with appropriate content type and charset handling.
-   * @param params - The parameters object.
+   * @param params - The key and value to encode.
    * @param params.key - The field name.
    * @param params.value - The value to encode.
-   * @param params.metadata - Encoding metadata for this field.
-   * @param params.allowedTypes - Set of allowed content types for this field.
+   * @param metadata - Encoding metadata for this field.
+   * @param allowedTypes - Set of allowed content types for this field.
    * @returns Blob or string - The encoded value.
    */
-  private encodeString({
-    key,
-    value,
-    metadata,
-    allowedTypes
-  }: {
-    key: string;
-    value: any;
-    metadata: EncodingMetadata;
-    allowedTypes: Set<string>;
-  }): Blob | string {
+  private encodeString(
+    {
+      key,
+      value
+    }: {
+      key: string;
+      value: any;
+    },
+    metadata: EncodingMetadata,
+    allowedTypes: Set<string>
+  ): string | Blob {
     // Handle string data
     // Only use JSON.stringify for application/json content type - otherwise may unduly escape e.g. stringified XML
     // To avoid stringifying pre-stringified JSON, users should use `Blob` or raw `FormData`


packages/openapi-generator/package.json
```diff
@@ -30,29 +30,32 @@
...
     "test": "yarn test:unit",
     "test:unit": "yarn node --experimental-vm-modules ../../node_modules/jest/bin/jest.js",
     "coverage": "jest --coverage",
     "lint": "eslint --ext .ts . && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -c",
     "lint:fix": "set TIMING=1 && eslint --ext .ts . --fix --quiet && prettier . --config ../../.prettierrc --ignore-path ../../.prettierignore -w --log-level error",
     "check:dependencies": "depcheck . --ignores='@sap-cloud-sdk/openapi'",
-    "readme": "ts-node ../../scripts/replace-common-readme.ts"
+    "readme": "ts-node ../../scripts/replace-common-readme.ts",
+    "prepack": "test -f dist/index.js || tsc -p tsconfig.for-prepack.json || true"
   },
```
davidkna-sap: TODO:

To be reverted, temporary change to allow using as a git dependency in AI SDK.

packages/openapi-generator/src/file-serializer/operation.ts
```diff
@@ -121,13 +121,33 @@ function serializeParamsForRequestBuilder(
   }
   if (operation.requestBody) {
     params.push('body');
+    if (
+      operation.requestBody.encoding &&
+      Object.keys(operation.requestBody.encoding).length > 0
```
marikaner:
```suggestion
      Object.keys(operation.requestBody.encoding).length
```

packages/openapi-generator/src/file-serializer/operation.ts
```diff
@@ -121,13 +121,33 @@ function serializeParamsForRequestBuilder(
...
+          `headerParameters: {${contentTypeStr}, ...headerParameters}`
+        );
+      } else {
+        params.push(`headerParameters: {${contentTypeStr}}`);
+      }
+    } else if (operation.headerParameters.length) {
+      params.push('headerParameters');
+    }
+  } else if (operation.headerParameters.length) {
+    params.push('headerParameters');
```
marikaner: [pp] Consider extracting this into a separate function `getHeaderParameters()` or so. That way you could reduce some of the complexity:
```ts
  function getHeaderParameters(operation) {
    if (operation.requestBody?.mediaType) {
      const contentTypeStr = `'content-type': '${operation.requestBody.mediaType}'`;
      return operation.headerParameters.length ? `headerParameters: {${contentTypeStr}, ...headerParameters}` : `headerParameters: {${contentTypeStr}}`);
    }
    if (operation.headerParameters.length) {
      return 'headerParameters';
    }
```

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,164 @@ const allowedMediaTypes = [
...
+    {
+      contentType: string;
+      isImplicit: boolean;
+      contentTypeParsed: ParsedMediaType[];
+    }
+  > = {};
+
+  if (!schema) {
+    const joined = { ...autoEncoding, ...explicitEncoding };
+    return Object.keys(joined).length > 0 ? joined : undefined;
```
marikaner:
```suggestion
    return Object.keys(joined).length ? joined : undefined;
```

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,164 @@ const allowedMediaTypes = [
...
+function parseEncoding(
+  mediaTypeObject: OpenAPIV3.MediaTypeObject | undefined,
+  refs: OpenApiDocumentRefs
+):
+  | Record<
+      string,
+      {
+        contentType: string;
+        isImplicit: boolean;
+        contentTypeParsed: ParsedMediaType[];
```
marikaner: [pp] Please rename this to `parsedContentTypes` the other way around it sounds like a boolean to me. Also it is a list, right?

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,164 @@ const allowedMediaTypes = [
...
+    string,
+    {
+      contentType: string;
+      isImplicit: boolean;
+      contentTypeParsed: ParsedMediaType[];
+    }
+  > = {};
+
+  if (!schema) {
+    const joined = { ...autoEncoding, ...explicitEncoding };
```
marikaner: At this point `autoEndoding` is always empty, right?

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,164 @@ const allowedMediaTypes = [
...
+              contentTypeParsed: [contentTypeParsed]
+            };
+          }
+        }
+      }
+    );
+  }
+
+  const combined = { ...autoEncoding, ...explicitEncoding };
+  return Object.keys(combined).length > 0 ? combined : undefined;
```
marikaner:
```suggestion
  return Object.keys(combined).length ? combined : undefined;
```

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,164 @@ const allowedMediaTypes = [
...
+        }
+
+        if (!propSchema || typeof propSchema !== 'object') {
+          return;
+        }
+
+        // Resolve $ref for property schema
+        const resolvedPropSchema = refs.resolveObject(propSchema);
+
+        const inferContentTypeFromSchema = (
```
marikaner: [q] Why not extract this on the toplevel?

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,164 @@ const allowedMediaTypes = [
...
+
+/**
+ * Parse encoding object from a media type, extracting contentType for each property.
+ * Also automatically infers content types for properties with binary format.
+ * @param mediaTypeObject - The media type object containing encoding and schema.
+ * @param refs - Object representing cross references throughout the document.
+ * @returns Encoding configuration mapping property names to contentType, or undefined.
+ * @internal
+ */
+function parseEncoding(
```
marikaner: Generally please note, that this is quite a long function. There is probably potential to shorten it, for quicker grasping of its contents (e.g. when you need to find a bug in the future).

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,181 @@ const allowedMediaTypes = [
...
+  return contentType
+    .split(',')
+    .map(ct => ct.trim())
+    .map(ct => {
+      try {
+        return parseContentType(ct);
+      } catch (error: any) {
+        throw new ErrorWithCause(
+          `Invalid content-type '${ct}' for property '${propName}' in OpenAPI specification. ` +
+            "Content types must follow the format 'type/subtype' (e.g., 'image/png', 'text/plain'). " +
```
marikaner: minor inconsistency: "content-type" and "content type". I guess in text it is "content type".

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,181 @@ const allowedMediaTypes = [
...
+    });
+}
+
+/**
+ * Infer content type based on OpenAPI schema type.
+ * @param s - The schema object to infer content type from.
+ * @returns The inferred content type, or undefined if cannot be determined.
+ */
+function inferContentTypeFromSchema(
+  s: OpenAPIV3.SchemaObject
```
marikaner, line 60: just a suggestion and matter of taste
```suggestion
  { type, items }: OpenAPIV3.SchemaObject
```

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,181 @@ const allowedMediaTypes = [
...
+ * @param refs - Object representing cross references throughout the document.
+ * @returns Encoding map with inferred content types added, or undefined if no encodings.
+ * @internal
+ */
+function buildInferredEncodings(
+  resolvedEncodings: string[],
+  resolvedSchema: OpenAPIV3.SchemaObject,
+  refs: OpenApiDocumentRefs
+): EncodingMap | undefined {
+  if (!('properties' in resolvedSchema) || !resolvedSchema.properties) {
```
marikaner: [q] wouldn't the second part suffice?
```suggestion
  if (!resolvedSchema.properties) {
```

or is this a type issue? then maybe you could use "as any", but i guess then again it's a matter of taste.

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,181 @@ const allowedMediaTypes = [
...
+      if (resolvedEncodings.includes(propName)) {
+        return;
+      }
+      if (!propSchema || typeof propSchema !== 'object') {
+        return;
+      }
+
+      // Resolve $ref for property schema
+      const resolvedPropSchema = refs.resolveObject(propSchema);
+      if (Object.prototype.hasOwnProperty.call(resolvedPropSchema, '$ref')) {
```
marikaner: [q] why is `hasOwnProperty` relevant here, wouldn't it be enough to check `'$ref' in resolvedPropSchema`? If it is relevant why not `resolvedPropSchema.hasOwnProperty('$ref')`?

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,181 @@ const allowedMediaTypes = [
...
+
+/**
+ * Add auto-inferred content types for schema properties.
+ * @param resolvedEncodings - Explicitly defined encodings that have already been resolved.
+ * @param resolvedSchema - The resolved schema object.
+ * @param refs - Object representing cross references throughout the document.
+ * @returns Encoding map with inferred content types added, or undefined if no encodings.
+ * @internal
+ */
+function buildInferredEncodings(
```
marikaner: [pp] Consider adding "properties" to the function name. I think it could help understand when this is used.

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,181 @@ const allowedMediaTypes = [
...
+              parsedContentTypes: parseContentTypes(
+                encodingObj.contentType!,
+                propName
+              )
+            }
+          ])
+      )
+    : {};
+
+  // Auto-infer content types based on schema types
```
marikaner: [q] is this comment correct?

packages/openapi-generator/src/parser/media-type.ts
marikaner: much better 😊

packages/openapi-generator/package.json
marikaner: don't forget this before merging 😊

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -22,13 +22,15 @@ const logger = createLogger('openapi-generator');
...
+ * @param mediaType - Optional media type context for proper type mapping (e.g., 'multipart/form-data').
  * @returns The parsed schema.
  * @internal
  */
 export function parseSchema(
   schema: OpenAPIV3.SchemaObject | OpenAPIV3.ReferenceObject | undefined,
   refs: OpenApiDocumentRefs,
-  options: ParserOptions
+  options: ParserOptions,
+  mediaType?: string
```
marikaner: I was hoping we could avoid dragging it through everywhere... Can you point me to, where this is ultimately used, i don't see it 👓

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -182,6 +190,135 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
+
+      // Content type is provided by the generator in _encoding based on the schema
+      const {
+        contentType: targetContentType,
+        isImplicit: targetIsImplicit,
+        parsedContentTypes
+      } = encoding[key];
+      // Use the first parsed content type (primary type)
+      const allowedTypes = new Set(
+        // TODO: compcase?
```
marikaner: TODO

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -182,6 +190,135 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
+        // unless the target content type is complex (multiple choices or wildcards)
+        // or the encoding is implicit (in which case we are more lenient as there was no specific request from the spec)
+        const valueContentTypeBase = value.type.split(';')[0].trim();
+        if (
+          // Don't warn about implicit encodings (less likely to be relevant)
+          !targetIsImplicit &&
+          // We do not handle more complex content types
+          !isFlexibleContentType &&
+          // Do the actual comparison
+          valueContentTypeBase.localeCompare(
```
marikaner: [q] why this and not just lowercase as above?

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -182,6 +190,135 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
+
+    // Handle multipart/form-data body unless the body is already a FormData instance
+    if (contentType === 'multipart/form-data' && !(body instanceof FormData)) {
+      return this.buildFormData(body);
+    }
+
+    return body;
+  }
+
+  private buildFormData(body: Record<string, any>): FormData {
```
marikaner: [q] Do you see potential to split some of the content into functions, e.g. around each of the `continue`s? I find this cumbersome to read. My attention span is too short for the whole function. 🤪

packages/openapi/src/openapi-request-builder.ts
davidkna-sap: Technically this is a bit more correct in case of unicode. This could use [`compcase`](https://github.com/SAP/cloud-sdk-js/pull/6301#discussion_r2793406652) instead to achieve the same thing, but that is not readily available in JS. Edit: will just replace everything with `toLowerCase`, this should be ASCII-only anyway.

packages/openapi-generator/src/parser/media-type.ts
davidkna-sap: Per the recent axios CVE this can be used to avoid prototype pollution, but I will change it to `in`.

packages/openapi-generator/src/parser/schema.ts
davidkna-sap: I used to have something here, but after reading the spec I noticed that `format: binary` is only legal for multipart, will revert fully.

packages/openapi-generator/src/parser/media-type.ts
davidkna-sap: Will skip this, because changing this leads to type-issues.

davidkna-sap: Will skip this, because changing this leads to type-issues.

packages/openapi-generator/src/parser/media-type.ts
```diff
@@ -11,8 +12,181 @@ const allowedMediaTypes = [
...
+
+/**
+ * Parse encoding object from a media type, extracting contentType for each property.
+ * Also automatically infers content types for properties with binary format.
+ * @param mediaTypeObject - The media type object containing encoding and schema.
+ * @param refs - Object representing cross references throughout the document.
+ * @returns Encoding configuration mapping property names to contentType, or undefined.
+ * @internal
+ */
+function parseEncoding(
```
marikaner: [q] I don't fully understand what the cases are when this is used. I previously suggested to change the names to something with properties, because I thought that this is related to object schemas only, but now I am confused. What does "property" stand for in this context?

marikaner: [pp/req] I believe I now start understanding. It is all related to form data only. It would be really helpful if the function name and the documentation would reflect that. e.g. `parseFormDataEncodings()`.
Also: `buildInferredEncodingsProperties()` => `inferFormDataEncodings()`

davidkna-sap: The properties here are the properties of an object schema, but object is not representing a regular object but the named formdata entires.

marikaner: Got it.

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -25,6 +28,181 @@ import type {
...
+      targetContentType &&
+      (targetContentType.includes('*') ||
+        parsedContentTypes.length > 1 ||
+        allowedTypes.has('any'));
+
+    // If `Blob` has no type, we use value from the specification unless the target content type is complex (multiple choices or wildcards)
+    if (
+      !value.type &&
+      !isFlexibleContentType &&
+      !parsedContentTypes[0].parameters.charset
```
marikaner: [pp] Consider adding a comment why we  would not add a type if there is a charset.

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -25,6 +28,189 @@ import type {
...
+          `Missing encoding metadata for property '${key}'. ` +
+            'This indicates a code generation issue. ' +
+            'Please regenerate your API client.'
+        );
+      }
+
+      const metadata = this.encoding[key];
+      const allowedTypes = new Set(
+        metadata.parsedContentTypes.map(ct => ct.type.toLowerCase())
+      );
```
marikaner, line 78: [pp] I believe you are using a set here to extract unique values. I think often it is a little more straight forward to handle arrays. In this case there might be advantages to a set, however you could consider to use the `unique()` function from the util package, that returns an array, instead. I marked another occurrence of set, where I think this would be more beneficial than here, too.
```suggestion
      const allowedTypes = unique(
        metadata.parsedContentTypes.map(ct => ct.type.toLowerCase())
      );
```

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -25,6 +28,189 @@ import type {
...
+
+  /**
+   * Build and return a FormData object from the body and encoding metadata.
+   * @returns The constructed FormData object.
+   */
+  build(): FormData {
+    const formData = new FormData();
+
+    for (const [key, value] of Object.entries(this.body ?? {})) {
+      if (!this.encoding || !this.encoding[key]) {
```
marikaner:
```suggestion
      if (!this.encoding?.[key]) {
```

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -25,6 +28,189 @@ import type {
...
+    // Only use JSON.stringify for application/json content type - otherwise may unduly escape e.g. stringified XML
+    // To avoid stringifying pre-stringified JSON, users should use `Blob` or raw `FormData`
+    const stringValue = allowedTypes.has('application/json')
+      ? JSON.stringify(value)
+      : String(value);
+
+    // If a charset is specified in the encoding, we encode the string accordingly (if unambiguous)
+    const targetCharset = new Set(
+      metadata.parsedContentTypes.map(ct => ct.parameters.charset)
+    );
```
marikaner, line 203: [pp] This is the other set usage. Changing this to a list would make the usage below much simpler: `targetCharset.values().next().value` => `targetCharset[0]`
```suggestion
    const targetCharset = unique(
      metadata.parsedContentTypes.map(ct => ct.parameters.charset)
    );
```

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -25,6 +28,189 @@ import type {
...
+   * @param key - The field name.
+   * @param value - The Blob value to append.
+   * @param metadata - Encoding metadata for this field.
+   */
+  private appendBlob(
+    formData: FormData,
+    key: string,
+    value: Blob,
+    metadata: EncodingMetadata
+  ): void {
```
marikaner: Just a few ideas:
1. [pp/req] The two `append...` functions have quite a lot of parameters. To reduce them you could return the value to append instead of appending it in this function. That way you wouldn't have to pass the form data at all and the function would be pure. Then it would rather be `sanitizeBlob()` or so.
2. [nth] The key and value somehow belong together, you could group them into one parameter.

```suggestion
  private sanitizeBlob(
    { key, value }: { key: string, value: Blob },
    metadata: EncodingMetadata
  ): Blob {
```

PR #6332, opened 2026-02-16
feat(destination): Add parameter to bypass service retrieval from VCAP

davidkna-sap:
Related https://github.com/SAP/ai-sdk-js-backlog/issues/451


packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -72,19 +72,33 @@ async function retrieveDestination({
...
+  const name = (serviceBinding && service.name) || destinationName;
+
+  return { name, ...destination };
+}
+
+function parseService(
+  service: DestinationFromServiceBindingOptions['service']
+): Service | undefined {
+  if (typeof service === 'string') {
+    return JSON.parse(service) as Service;
```
marikaner: [req] My original thought was, that the string part of service is a replacement for `destinationName`. I even think we could xor service and destination name... Otherwise I think we should not allow passing a JSON string to be parsed - we can expect users to pass in a service as an object.

packages/connectivity/src/scp-cf/destination/destination-from-vcap.ts
```diff
@@ -72,30 +70,42 @@ async function retrieveDestination({
...
-}
+} & Xor<
+  {
+    /**
+     * A service binding to use directly instead of looking it up by name. Mandatory if no destination name is provided.
+     */
+    service: Service;
+  },
+  Pick<DestinationFetchOptions, 'destinationName'>
+>;
```
marikaner: [req] Could you add a type test for this to make sure that it works as expected? I am not sure from merely looking at this...

test-packages/type-tests/test/destination-from-vcap.test-d.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const transformFn: ServiceBindingTransformFunction = undefined as any;
+
+const validWithServiceAndFn: DestinationFromServiceBindingOptions = {
+  service: mockService,
+  serviceBindingTransformFn: transformFn
+};
+expectAssignable<DestinationFromServiceBindingOptions>(validWithServiceAndFn);
+
+// Valid case 4: destinationName with optional serviceBindingTransformFn
```
marikaner: [q] case 4? didn't see 1, 2, 3

davidkna-sap:
```suggestion
```

PR #6335, opened 2026-02-17
feat: Support OpenAPI version 3.1. type null

marikaner:
Closes http://github.com/SAP/ai-sdk-js-backlog/issues/484

To correctly support version 3.1. we would have to go through the code and change all occurrences of `OpenAPIV3.XY` to ``OpenAPIV3.XY | `OpenAPIV3_1.XY` and if needed adjust some code. This effort is not huge, however much bigger than my quick fix. Let me know what you think.


marikaner:
> I understand that this fixes the immediate issue we have in RPT spec where we have
> 
> ```
> {
>    "type": "null"
> }
> ```
> 
> Should we also consider handling type arrays, which can be used instead of `nullable: true` in 3.1 ? For example:
> 
> ```
> {
>     "type": ["string", "null"]  // ← Standard OpenAPI 3.1 for simple nullable fields
> }
> ```
> 
> So,
> 
> ```
> # OpenAPI v3.0
> type: string
> nullable: true
> ```
> 
> now changes to
> 
> ```
> # OpenAPI v3.1
> type:
> - "string"
> - "null" 
> ```
> 
> Referred to: https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

I was wondering the same. Actually there are a few more things that have changed from 3.0 to 3.1, we could generally make it compatible with those. I am not sure about our capacity for this, however @jjtang1985.
Right now my fix only works for usage on CLI, not necessarily for usage in code (it will work with workarounds).

PR #6344, opened 2026-02-19
chore: Replace license-checker with licensee

marikaner:
We currently use license-checker to check our licenses (already using a bit of a workaround with quite a lot of custom code). I found that this tool hasn't been updated in 8 years and depdends on deprecated libraries. The code itself fails in strict mode. Therefore I would like to replace it with a more modern library. As the API is quite simple, I don't see a reason to have an action anymore. We should also adjust it in the SAP Cloud SDK for AI.


marikaner:
> The library seems to be a bit unpopular, even if the code seems very simple.

True, I didn't find anything better, though. I am open for suggestions 😊

davidkna-sap:
Maybe Hyperspace OSC or Black Duck?

davidkna-sap (approved):

PR #6376, opened 2026-03-04
chore: Remove generated API docs and refactor workflow

emincihangeri:
Closes SAP/ai-sdk-js-backlog/issues/453

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,7 +9,7 @@ updates:
       timezone: 'Europe/Berlin'
     open-pull-requests-limit: 10
     cooldown:
-      days: 4
+      default-days: 4
     ignore:
       # We only want to do major node updates on purpose, so don't create dependabot PRs for major versions
       - dependency-name: '@types/node'
@@ -36,4 +36,4 @@ updates:
       time: '01:00'
       timezone: 'Europe/Berlin'
     cooldown:
-      days: 4
+      default-days: 4


.github/workflows/api-docs.yml
```diff
@@ -1,44 +1,73 @@
...
+          
+          # Use the version input parameter
+          VERSION="${{ inputs.version }}"
+          
           cd ..
           git clone --depth 1 git@github.com:SAP/cloud-sdk.git
-          rsync -avz cloud-sdk-js/knowledge-base/api-reference/ cloud-sdk/static/api/$M_VERSION/
+          
+          # Copy generated docs to versioned folder
+          rsync -avz cloud-sdk-js/knowledge-base/api-reference/ cloud-sdk/static/api/${VERSION}/
```
davidkna-sap:
```suggestion
          rsync -avz --delete cloud-sdk-js/knowledge-base/api-reference/ cloud-sdk/static/api/${VERSION}/
```

.github/workflows/bump.yml
```diff
@@ -46,3 +45,11 @@ jobs:
...
         with:
           version: ${{ steps.bump.outputs.version }}
+
+  generate-api-docs:
+    name: Generate and Push API Documentation
+    needs: [bump]
+    uses: ./.github/workflows/api-docs.yml
+    secrets: inherit
+    with:
+      version: v${{ needs.bump.outputs.version }}
```
davidkna-sap, line 55: As far as I can tell this is the full version, whereas the script (now) expects only the major version number.

.github/workflows/api-docs.yml
```diff
@@ -1,44 +1,73 @@
...
+        type: string
 
 jobs:
-  push-to-documentation-branch:
+  generate-and-push-docs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - name: Add key
+        with:
```
davidkna-sap: This checks out the main branch, but I think `ref: main` is wrong in this context, because this job could be part of a bump related to a prior version of the SAP Cloud SDK.

davidkna-sap (dismissed):

PR #6417, opened 2026-03-18
fix: Never cache SAMLAssertion

### LINKED ISSUE #6396: Caching the destination of authentication type SAMLAssertion, results in 401 unauthorized issue when the cached destination details are used next time
### Describe the Bug

Customer has created a destination of authentication type SAMLAssertion (for connecting to s4 hanacloud system from SAP business process automation through communication arrangement (blog for reference - https://community.sap.com/t5/enterprise-resource-planning-blog-posts-by-sap/samlbearerassertion-authentication-with-s-4hana-and-s-4hana-cloud/ba-p/13495308)

When using the destination for the first time, call to the system is successul
From the next call onwards, it results in 401 unauthorized error.

Cloud sdk caches the destination details. Cached destination details will be used for the second request too where the SAML assertion remains same. 
Note: Every SAML assertion has a unique ID. Once S/4HANA verifies and "consumes" an assertion to log you in, it blacklists that ID 
So the same saml assertion saved in the cache cannot be used from second request onwards.

While we can set useCache = false to make cloud sdk not cache the destination, it will apply for every type destination defeats the original purpose of using cache. We have no way to set the useCache = false only for the saml assertion destinaton.

Can the destination details be not cached when the authentication type is saml assertion? 

### Steps to Reproduce

Setup a destination of type authentication type saml assertion

### Expected Behavior

Destination details should not be cached for destination of authentication type saml assertion

### Screenshots

_No response_

### Used Versions

- Node version via `node -v`: ... 22.14.0 
- NPM version via `npm -v: ... 10.8.2
- SAP Cloud SDK version: ... 4.1.2

### Code Examples

_No response_

### Log File

_No response_

### Affected Development Phase

Production

### Impact

Blocked

### Timeline

_No response_

### Additional Context

_No response_

marikaner:
Closes SAP/cloud-sdk-backlog#1293.
Closes #6396.

SAMLAssertions are only valid for one-time use. Therefore it doesn't make sense to ever cache them. This is a proposal to skip caching for authentication type SAMLAssertion, when enabled.

### DIFF
--- a/packages/generator/package.json
+++ b/packages/generator/package.json
@@ -49,7 +49,7 @@
     "@sap-cloud-sdk/odata-v4": "^4.5.1",
     "@sap-cloud-sdk/util": "^4.5.1",
     "@types/fs-extra": "^11.0.4",
-    "fast-xml-parser": "^5.5.7",
+    "fast-xml-parser": "^5.5.8",
     "fs-extra": "^11.3.4",
     "ts-morph": "^27.0.2",
     "typescript": "~5.9.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4565,13 +4565,13 @@ fast-xml-builder@^1.1.4:
   dependencies:
     path-expression-matcher "^1.1.3"
 
-fast-xml-parser@^5.5.7:
-  version "5.5.7"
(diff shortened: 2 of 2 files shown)


manisharma07:
Hi team, 

Can you share if there is already a release version of cloud sdk that has this implemented, or if we have a tentative timeline by when will this be available for consumption

BR,

marikaner:
Hey @manisharma07, we are planning a release this week. In the mean time, feel free to try out the canary version.

PR #6431, opened 2026-03-23
feat: Add `getIasToken()` and `getIasDestination()` convenience functions

davidkna-sap:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/1291
Related https://github.com/SAP/ai-sdk-js-backlog/issues/466

This PR adds the convenience functions `getIasToken()` (returning token, refresh token (if available) and expiration) and `getIasDestination()` (returning a callable destination).

Both accept a service `Service | string | ServiceCredentials`, unlike `transformIasBindingToDestination` they allow bare `ServiceCredentials` (e.g. just `clientid`, `clientsecret` and `url`).

Note: `getIasToken()` only returns the token as a string, rather than a parsed jwt, because it may not be a jwt.

To support this more flexible usage, existing convenience logic (e.g. `app_tid` resolution) has been refactored out of `transformIasBindingToDestination`.

Token caching now relies on the cached token flows added three days ago to `@sap/xssec` 4.13.0 (`getClientCredentialsToken`/`getJwtBearerToken`) instead of the SDK-level IAS client credentials cache to allow for caching closer at lower levels of the implementation.
This means business user tokens are now cached too (https://github.com/SAP/ai-sdk-js-backlog/issues/466).
The `useCache` toggle is still exposed, but it is currently not possible to change the cache size or implementation via SAP Cloud SDK APIs.


packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -88,3 +98,80 @@ export async function jwtBearerToken(
   const resolvedService = resolveServiceBinding(service);
   return getUserToken(resolvedService, jwt);
 }
+
+/**
+ * @internal
+ */
+async function resolveIdentityService(
```
KavithaSiva, line 103: [req] `resolveIdentityService` is unnecessarily async.

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -88,3 +98,80 @@ export async function jwtBearerToken(
...
+  }
+  return 'credentials' in service
+    ? (service as Service)
+    : { name: 'identity', label: 'identity', tags: [], credentials: service };
+}
+
+/**
+ * Returns an IAS token from the Identity Authentication Service.
+ * Supports both technical user (OAuth2ClientCredentials) and business user (OAuth2JWTBearer) flows.
+ * @param service - Service credentials, a service type string (e.g., 'identity'), or a {@link Service} instance.
```
KavithaSiva: [q] Should we also add:
Passing raw `ServiceCredentials`, `Service` directly is only recommended for environments where service bindings are unavailable as they hardcode the credentials? 

`identity` string is the best way as the SDK resolves the credentials from the environment automatically.

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -88,3 +98,80 @@ export async function jwtBearerToken(
...
+ * @param options - Options for IAS token retrieval. See {@link IasTokenOptions}.
+ * @returns An {@link IasTokenResult} containing the access token, expiration, and optional refresh token.
+ */
+export async function getIasToken(
+  service: ServiceCredentials | string | Service,
```
KavithaSiva:
```suggestion
  service: ServiceCredentials | 'identity' | Service,
```
[q] Can't we narrow the type?

packages/connectivity/src/scp-cf/identity-service.ts
```diff
@@ -223,9 +227,13 @@ async function getIasTokenImpl(arg: IasParameters): Promise<IasTokenResponse> {
...
       ? // JWT bearer grant for business user access
-        await identityService.fetchJwtBearerToken(arg.assertion, tokenOptions)
+        arg.useCache !== false
+        ? await identityService.getJwtBearerToken(arg.assertion, tokenOptions)
+        : await identityService.fetchJwtBearerToken(arg.assertion, tokenOptions)
       : // Technical user client credentials grant
-        await identityService.fetchClientCredentialsToken(tokenOptions);
+        arg.useCache !== false
+        ? await identityService.getClientCredentialsToken(tokenOptions)
+        : await identityService.fetchClientCredentialsToken(tokenOptions);
```
KavithaSiva: [req] Extract `arg.useCache !== false` into a variable to simplify the ternary statement.

KavithaSiva (commented):
Initial round of comments.

packages/connectivity/src/scp-cf/token-accessor.ts
davidkna-sap: `resolveServiceBinding` is async, I think this should stay as-is.

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -88,3 +98,86 @@ export async function jwtBearerToken(
...
+ * @remarks
+ * Prefer using `'identity'` (default) for the `service` parameter.
+ * Passing raw ServiceCredentials or Service directly is only recommended for environments where service bindings are unavailable as they hardcode the credentials.
+ * @param service - Service credentials {@link ServiceCredentials}, the service type (always 'identity' for IAS), or a {@link Service} binding.
+ * @param options - Options for IAS token retrieval. See {@link IasTokenOptions}.
+ * @returns An {@link IasTokenResult} containing the access token, expiration, and optional refresh token.
+ */
+export async function getIasToken(
+  service: ServiceCredentials | 'identity' | Service = 'identity',
+  options?: IasTokenOptions
```
davidkna-sap, line 126: What do you think about putting `IasTokenOptions` first with `service` having a sane default?

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -88,3 +98,86 @@ export async function jwtBearerToken(
...
+): Promise<Destination> {
+  const resolvedService = await resolveIdentityService(service);
+  const { token } = await getIasToken(resolvedService, options);
+
+  const iasOptions: IasOptions = {
+    authenticationType: 'OAuth2ClientCredentials',
+    ...options
+  };
+
+  return buildIasDestination(token, resolvedService, iasOptions);
```
marikaner, line 180: [q] Does this include caching?

packages/connectivity/src/scp-cf/token-accessor.ts
marikaner: Hmm, while I see the point, semantically I think it is better as is.

packages/connectivity/src/scp-cf/token-accessor.ts
marikaner: @KavithaSiva I thought the same at first, but I this is useful, so that the second non-async part of the function does not need to be wrapped in a promise.

Example:
Imagine this function (pass in boolean `a`, if it is true return 1 otherwise 0):
```ts
const test = (a: boolean): Promise<any> => (a ? Promise.resolve(1) : 0); # won't work
const test = (a: boolean): Promise<any> => (a ? Promise.resolve(1) : Promise.resolve(0)); # will work
const test = async (a: boolean): Promise<any> => (a ? Promise.resolve(1) : 0); # will work
```

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -8,7 +8,17 @@ import {
...
+import { fetchIasToken, getIasAppTid } from './identity-service';
+/* eslint-disable import/no-internal-modules -- avoid circular imports via destination barrel */
+import { buildIasDestination } from './destination/build-ias-destination';
+import type {
+  IasOptions,
+  IasTokenOptions,
+  IasTokenResult
+} from './destination/ias-types';
+import type { Destination } from './destination/destination-service-types';
+/* eslint-enable import/no-internal-modules */
```
marikaner: [req] Duplicate
```suggestion
```

packages/connectivity/src/scp-cf/token-accessor.ts
```diff
@@ -8,7 +8,17 @@ import {
 import { clientCredentialsTokenCache } from './client-credentials-token-cache';
 import { resolveServiceBinding } from './environment-accessor';
 import { getClientCredentialsToken, getUserToken } from './xsuaa-service';
-import type { Service } from './environment-accessor';
+import { fetchIasToken, getIasAppTid } from './identity-service';
+/* eslint-disable import/no-internal-modules -- avoid circular imports via destination barrel */
```
marikaner: [req] Please consider to only disable this for a specific line. Alternatively we could also generally allow it for specific modules in the (local) eslint config.

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
```diff
@@ -30,53 +29,12 @@ const ClientCredentialsTokenCache = (
     });
   },
 
-  getTokenIas: (
```
marikaner, line 33: [q] Just to make sure, the client credentials token cache is completely internal, therefore it is fine to move the functions, right?

packages/connectivity/src/scp-cf/destination/build-ias-destination.ts
```diff
@@ -0,0 +1,77 @@
...
+  url: string,
+  name: string,
+  authentication: Extract<
+    AuthenticationType,
+    'OAuth2ClientCredentials' | 'OAuth2JWTBearer'
+  > = 'OAuth2ClientCredentials'
+): Destination {
+  const expirationTime = decodeJwt(token).exp;
+  const expiresIn = expirationTime
+    ? Math.floor((expirationTime * 1000 - Date.now()) / 1000).toString(10)
```
marikaner: [pp] Add a comment how long this is and store it in a variable.

marikaner (commented):
I find it a little odd that we have extra functions for the IAS case. I assume you thought about how it could be covered seamlessly in the existing functionality?

packages/connectivity/src/scp-cf/token-accessor.ts
davidkna-sap: I've moved this to the eslint config, the following imports also need this and per-line comments would have required this:
```ts
  // eslint-disable-next-line import/no-internal-modules
} from './destination/ias-types';
```

packages/connectivity/src/scp-cf/client-credentials-token-cache.ts
davidkna-sap: This should only be exported with `@internal`, and I don't see anything related in the typedocs.

packages/connectivity/src/scp-cf/destination/build-ias-destination.ts
davidkna-sap: I restructured it and added a comment.

packages/connectivity/src/scp-cf/token-accessor.ts
davidkna-sap: It does not include caching, but it doesn't do any network requests.

packages/connectivity/src/scp-cf/token-accessor.ts
marikaner: If this helps with circular dependencies, maybe it is better in the config even.

.changeset/wild-kids-poke.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+---
+
+[Compatibility Note] IAS tokens are now cached via `@sap/xssec`.
+`@sap/xssec` uses an LRU cache limited to 100 items, previously this cache was unbounded.
```
marikaner:
```suggestion
`@sap/xssec` uses an LRU cache limited to 100 items, previously this cache was unbound.
```

marikaner (approved):
lgtm

PR #6447, opened 2026-04-01
chore: widen dependency versions

marikaner:
I noticed that this is not configured correctly for a lib. I am pretty sure we already had configured this versioning strategy. If it was removed on purpose, please let me know why.


davidkna-sap:
@marikaner I think generally this is sensible for a library but this also means that more unexpected incompatibilities could sneak in.

PR #6448, opened 2026-04-01
chore: Replace yarn with pnpm

davidkna-sap:
# Replaces Yarn 1 (Classic) with pnpm 10 across the entire monorepo.
       
 Motivation: pnpm natively enforces two security properties that Yarn Classic lacks:                                 
  - Dependency cooldown                                       
  - Build script blocking by default (`better-sqlite3` is allow-listed to make e2e tests works, but building on demand may also be possible)  

## Changes                         
- Adds `.github/actions/setup/action.yml`: reusable composite action that handles pnpm set-up                     
- Replaces `mock-fs` (semi-unmaintained) with `memfs` + `unionfs`                                                             
  - Adds `test-resources/mock-fs.js`: a union overlay filesystem factory (memfs on top of real fs) used across tests                                                                                      
  - `build-packages/check-public-api/index.ts:` removes `mock-fs` usage entirely; now compiles into a temp directory
- `turbo.json`: explicit per-package compile dependency graph added for dependency cycle-resolution
- Replace `licensee` with manual check using `pnpm license list --json`
- Replace `sqlite3` (unmaintained) with `@cap-js/sqlite` (uses `better-sqlite3`)

PR #6467, opened 2026-04-08
chore: Refactor test imports to use @sap-cloud-sdk/test-util-internal pkg

davidkna-sap:
Split from #6448

Required to support different dependency resolution/layout/hoisting when using pnpm instead of yarn.


davidkna-sap:
I think I mostly did this because of pnpm package-hoisting, I will explore both options.

IIRC the second option also requires some ts-jest hacks.

davidkna-sap:
@marikaner I have replace it with a ts-only package.

marikaner (approved):
lgtm

PR #6468, opened 2026-04-08
chore(check-public-api): Use tempdir instead of mockfs

davidkna-sap:
…and use memfs in tests

~~Stacked PR (base: #6467)~~


build-packages/check-public-api/index.ts
```diff
@@ -1,10 +1,10 @@
 /* eslint-disable jsdoc/require-jsdoc */
 import { join, resolve, parse, basename, dirname, posix, sep } from 'node:path';
-import { promises, existsSync } from 'node:fs';
+import { promises, existsSync, mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
```
marikaner: Very cool 😊 Can you please use the async functions when possible?

build-packages/check-public-api/check-public-api.spec.ts
```diff
@@ -1,6 +1,15 @@
...
+import { vol } from 'memfs';
+
+jest.unstable_mockModule('fs', () => import('memfs').then(m => m.fs));
+jest.unstable_mockModule('fs/promises', () =>
+  import('memfs').then(m => m.fs.promises)
+);
+jest.unstable_mockModule('node:fs', () => import('memfs').then(m => m.fs));
+jest.unstable_mockModule('node:fs/promises', () =>
+  import('memfs').then(m => m.fs.promises)
+);
```
marikaner: I don't like this boiler plate code... We should decide whether to keep mock-fs or if we can move this to some reusable test resource when using memfs.

davidkna-sap: I would prefer handling this via https://github.com/SAP/cloud-sdk-js/pull/6467

marikaner (approved):
I assume the memfs part will be reviewed on the other branch then.

PR #6470, opened 2026-04-08
chore: Replace mock-fs with memfs/unionfs for fs mocking

davidkna-sap:
~~Stacked PR (base: https://github.com/SAP/cloud-sdk-js/pull/6468)~~


marikaner (commented):
Did we have an issue with mock-fs? I can see that memfs has more traction, but I liked the tree structure of mock-fs better than the flat JSON in memfs.

davidkna-sap:
@marikaner For mock-fs in conjunction with pnpm there were a lot of issues resolving npm dependencies with pnpm's layout, and mock-fs anyway cannot support more recent file APIs due to node.js not exposing these.

marikaner:
> @marikaner For mock-fs in conjunction with pnpm there were a lot of issues resolving npm dependencies with pnpm's layout, and mock-fs anyway cannot support more recent file APIs due to node.js not exposing these.

Understood.

build-packages/changesets-fixed-version-bump/util.spec.ts
```diff
@@ -1,29 +1,44 @@
...
 import { getNextVersion } from './util.js';
 
+jest.unstable_mockModule('fs', () => import('memfs').then(m => m.fs));
+jest.unstable_mockModule('fs/promises', () =>
+  import('memfs').then(m => m.fs.promises)
+);
+jest.unstable_mockModule('node:fs', () => import('memfs').then(m => m.fs));
+jest.unstable_mockModule('node:fs/promises', () =>
+  import('memfs').then(m => m.fs.promises)
+);
```
marikaner: [pp] I would prefer to not see this everywhere we mock fs. Is it possible to have a util function in the test utils that does this?

build-packages/changesets-fixed-version-bump/util.spec.ts
```diff
@@ -1,29 +1,44 @@
...
-    });
+    vol.fromJSON(
+      {
+        ...sharedFiles,
+        '.changeset/config.json': changesetConfig,
+        '.changeset/alex.md':
+          '---\n' + "'@sap-cloud-sdk/connectivity': patch\n" + '---'
+      },
+      process.cwd()
+    );
```
marikaner, line 35: [pp] This is totally fine, but I honestly liked the tree structure before better.
We have 3 options:
1. Just leave it as is (for now).
2. Contribute this: https://github.com/streamich/memfs/issues/1255
3. Have another util function that takes the tree.

I would leave it at 1. for now. Just wanted to leave my thoughts somewhere 😊

packages/connectivity/src/http-agent/http-agent.spec.ts
```diff
@@ -1,5 +1,11 @@
+jest.mock('fs', () => jest.requireActual('memfs').fs);
+jest.mock('fs/promises', () => jest.requireActual('memfs').fs.promises);
+jest.mock('node:fs', () => jest.requireActual('memfs').fs);
+jest.mock('node:fs/promises', () => jest.requireActual('memfs').fs.promises);
```
marikaner, line 4: [q] Why is this necessary? Did we forget to unmock somewhere else?

packages/connectivity/package.json
```diff
@@ -50,9 +50,12 @@
     "jsonwebtoken": "^9.0.3"
   },
   "devDependencies": {
+    "@jest/globals": "^30.3.0",
     "@sap-cloud-sdk/test-util-shared": "*",
-    "mock-fs": "^5.5.0",
+    "depcheck": "^1.4.7",
+    "memfs": "^4.57.1",
     "nock": "^14.0.11",
+    "prettier": "^3.8.1",
```
marikaner, line 58: [q] Do you know why this wasn't detected by the dependency check? Or why this seemingly worked before? ;)

packages/test-util/src/mock-fs.ts
```diff
@@ -0,0 +1,97 @@
...
+    'mkdir',
+    'appendFile',
+    'rmdir',
+    'unlink',
+    'rename'
+  ] as const) {
+    if (memfs.promises[m]) {
+      (ufs as any).promises[m] = memfs.promises[m].bind(memfs.promises);
+    }
+  }
```
marikaner: [pp] Could you extract this to a separate function, e.g. `overrideWriteOperations()`.

The reason I am asking for this: Even though this is not very long, I had a hard time understanding why this is necessary and that this actually fixes a bug/unexpected behavior. I find these things easier to grasp if they are contained.

packages/test-util/src/mock-fs.ts
```diff
@@ -0,0 +1,97 @@
...
+    try {
+      const result = memfs.lstatSync(p, options);
+      if (result) {
+        return result;
+      }
+    } catch {
+      // Ignore and fall back to real fs
+    }
+    return realFs.lstatSync(p, options);
+  };
```
marikaner: [pp] Same here: please extract to a separate function, e.g. `fixStatSyncFallthrough()`.

packages/test-util/src/mock-fs.ts
```diff
@@ -0,0 +1,97 @@
...
+  // write to memfs
+  (ufs as any).copyFileSync = (src: string, dest: string) => {
+    memfs.writeFileSync(dest, (ufs as any).readFileSync(src));
+  };
+  (ufs as any).promises.copyFile = async (src: string, dest: string) => {
+    await memfs.promises.writeFile(
+      dest,
+      await (ufs as any).promises.readFile(src)
+    );
+  };
```
marikaner, line 125: [pp] If you extract the other two to separate functions, I think this can go into the first one (write).

packages/generator-common/src/file-writer/create-file.spec.ts
```diff
@@ -21,20 +30,22 @@ describe('createFile', () => {
   };
 
   beforeEach(() => {
-    mock({
-      [pathRootNodeModules]: mock.load(pathRootNodeModules),
-      [pathFormattedPackageJson]: mock.load(pathFormattedPackageJson),
+    const realFs = jest.requireActual('fs');
+    const formattedPackageJsonContent = realFs.readFileSync(
```
marikaner: [pp] If possible, use the async functions.
```suggestion
    const realFs = jest.requireActual('fs').promises;
    const formattedPackageJsonContent = await realFs.readFile(
```

packages/generator/src/generator.spec.ts
```diff
@@ -32,8 +42,7 @@ const pathTestService = resolve(
   'API_TEST_SRV',
   'API_TEST_SRV.edmx'
 );
-const pathToGeneratorCommon = resolve(__dirname, '../../generator-common');
-const pathRootNodeModules = resolve(__dirname, '../../../node_modules');
+const pathTestServiceDir = resolve(oDataServiceSpecs, 'v2', 'API_TEST_SRV');
```
marikaner, line 39: [pp] I think the line above can use this one.

marikaner (changes requested):
I have reviewed most files, but only audited the rest of them as the changes seem to be very similar. 

I would like to understand better why the following is ever necessary:

```
jest.mock('fs', () => jest.requireActual('memfs').fs);
jest.mock('fs/promises', () => jest.requireActual('memfs').fs.promises);
jest.mock('node:fs', () => jest.requireActual('memfs').fs);
jest.mock('node:fs/promises', () => jest.requireActual('memfs').fs.promises);
```

and whether we can avoid this. Also, as mentioned, I would appreciate it if, we were able to hide away some of this boiler plate code.

build-packages/changesets-fixed-version-bump/util.spec.ts
davidkna-sap: This is only possible for `jest.unstable_mockModule`, but not for `jest.mock` otherwise issues with `ts-jest`/`babel` appear.

packages/connectivity/src/http-agent/http-agent.spec.ts
davidkna-sap: `jest.requireActual` is needed because otherwise you get endless import loops.

packages/connectivity/package.json
davidkna-sap: I think it worked because `yarn` is less strict about hoisting, and the new packages are (only used as) binaries.

build-packages/changesets-fixed-version-bump/util.spec.ts
davidkna-sap: Will try  fs-monkey instead (same owner as memfs/unionfs)

build-packages/changesets-fixed-version-bump/util.spec.ts
davidkna-sap: I found `vol.fromNestedJSON` which is basically `fromTree` (the non-module version does not need to mock `node:` versions)

build-packages/changesets-fixed-version-bump/util.spec.ts
davidkna-sap: ~~I've simplified this, but decided not to use `fs-monkey`~~

build-packages/changesets-fixed-version-bump/util.spec.ts
marikaner: Apparently the maintainer is not aware of this: https://github.com/streamich/memfs/issues/1255#issuecomment-4273329384 😊

test-resources/test/test-util/fs-mocker.ts
```diff
@@ -0,0 +1,163 @@
+import { createFsFromVolume, vol } from 'memfs';
```
marikaner, line 1: [req] I didn't notice this before, but this package is not intended for our testing, rather for our user's testing. So I think we should move our test code to a different place.

build-packages/changesets-fixed-version-bump/util.spec.ts
marikaner: The docs on this are very sparse. Why did you decide against it?

packages/connectivity/src/http-agent/http-agent.spec.ts
marikaner: I now just realized, that this actually uses memfs. I thought we are trying to unmock the previously mocked memfs here. But the difference is that the other files are ESM this is CJS, right? and fs-monkey doesn't work for this, I assume?

marikaner (changes requested):

marikaner:
I am not so happy about the boilerplate code, but I can and will live with it, if needed ;)
One suggestion we could still try:
Use [manual mocks](https://jestjs.io/docs/manual-mocks) and reduce the imports to the following.

Create these files
```
__mocks__/
  fs.js
  fs/promises.js
  node:fs.js
  node:fs/promises.js
```

Then this becomes the boilerplate code:
```
jest.mock('fs');
jest.mock('fs/promises');
jest.mock('node:fs');
jest.mock('node:fs/promises');
```

It is at least a little shorter, but I am not sure this is much of an improvement. Up to you, if you want to try it.

packages/connectivity/src/http-agent/http-agent.spec.ts
davidkna-sap: For jest the difference is ESM vs CJS. `fs-monkey` doesn't really work at all:
- No types
- OOB-error attempting to override non-mutable constants
- Even if those are excluded some file-accesses leak into the real fs.

davidkna-sap:
@marikaner It doesn't really work with the order `__mocks__` are applied and I imagine that `:` in paths could cause some issues down the line.

test-resources/test/test-util/fs-mocker.ts
davidkna-sap: I moved it to the other test utility package

davidkna-sap:
@marikaner I managed to get rid of all the cjs boilerplate, the direct-ts imports via the other package allow for it. I moved the ESM fs-mocks into their own util package because they are not compatible with that.

marikaner (approved):
Nice, now I am quite happy that I pestered so much 😊 
I like what you did here!

PR #6473, opened 2026-04-09
chore: Add license-checker action for pnpm

davidkna-sap:

Testing via https://github.com/SAP/ai-sdk-js/pull/1728


build-packages/check-license/index.ts
```diff
@@ -0,0 +1,87 @@
+import { execFileSync } from 'node:child_process';
+import { info, setFailed } from '@actions/core';
+
+// Permissive FLOSS licenses are ok, see https://en.wikipedia.org/wiki/Permissive_software_license
```
marikaner: [pp] Consider this instead: https://blueoakcouncil.org/list It comes with an npm package that lists all of those licenses.

build-packages/check-license/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (flagged.length > 0) {
+      setFailed(
+        `Disallowed license "${license}" used by: ${flagged.map(p => `${p.name}@${p.version}`).join(', ')}`
+      );
+      failed = true;
+    }
+  }
+
+  if (failed) {
+    process.exit(1);
```
marikaner: It seems to me that `setFailed()` already sets the status to 1.

build-packages/check-license/index.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // Disallowed license — only flag non-SAP dependencies
+    const flagged = packages.filter(pkg => !isSapDependency(pkg.name));
+    if (flagged.length > 0) {
+      setFailed(
+        `Disallowed license "${license}" used by: ${flagged.map(p => `${p.name}@${p.version}`).join(', ')}`
+      );
+      failed = true;
+    }
+  }
```
marikaner: [req] I think this could be more idiomatic to JS.
E.g. (please check for correctness) :
```suggestion
  const disallowedPackages = Object.entries(licenseMap).reduce(disallowedPackages, ([license, packages]) => {
  const disallowed = packages.filter(pkg => 
    !isSapDependency(pkg.name) ||
    license === 'Unknown' && !ALLOWED_UNKNOWN.includes(pkg.name) ||
    !ALLOWED_UNKNOWN.includes(pkg.name)
  )
  
  if (disallowed.length) {
    return [
      ...flagged,
      disallowed.map(pkg => ({ license, pkg }})
    ];
  }
  return disallowedPackages;
  }, []);
  
  if (disallowedPackages.length) {
     setFailed(
        // ERROR
      );
  } else {
    info('All production dependency licenses are acceptable.');
  }
```

You could further improve this by wrapping this in functions: e.g. `getDisallowedPackages()`, `buildErrorMessage()` and `checkLicenses()`

marikaner (approved):
Very pragmatic approach, love it. I left a few code style suggestions though.

marikaner (changes requested):
wrong review before

build-packages/check-license/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    scope === '@sap' || scope === '@sap-cloud-sdk' || scope === '@sap-ai-sdk'
+  );
+}
+
+const isAllowedPackage = (license: string, pkg: PackageInfo): boolean =>
+  isSapDependency(pkg.name) ||
+  (license === 'Unknown' && ALLOWED_UNKNOWN.includes(pkg.name)) ||
+  ALLOWED_LICENSES.has(license);
+
+const getDisallowedPackages = (
```
marikaner: [req] When not using these functions inline, please make them "real" functions. It reads a little better. => https://github.com/SAP/cloud-sdk-js/blob/main/STYLEGUIDE.md#use-function-declarations-to-reference-functions-by-name

build-packages/check-license/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    maxBuffer: 10 * 1024 * 1024
+  });
+  const licenseMap: Record<string, PackageInfo[]> = JSON.parse(json);
+
+  const disallowedPackages = getDisallowedPackages(licenseMap);
+  if (disallowedPackages.length) {
+    setFailed(buildErrorMessage(disallowedPackages));
+  }
+
+  info('All production dependency licenses are acceptable.');
```
marikaner, line 89: I think you need the else here. If you want you can also use a ternary.
```suggestion
else {
  info('All production dependency licenses are acceptable.');
  }
```

marikaner (approved):
Much better 👍 I just left minor comments.

build-packages/check-license/index.ts
davidkna-sap: I added an early return instead.

PR #6497, opened 2026-04-14
chore: Fix user on bump

marikaner:
The bump on release didn't work, because the GH user was incorrect and declined by the branch protection rules.

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -53,6 +53,8 @@ jobs:
         uses: ./.github/actions/commit-and-tag
         with:
           version: ${{ steps.bump.outputs.version }}
+          user-email: ${{ secrets.SAP_CLOUD_SDK_BOT_EMAIL }}
+          user-name: ${{ secrets.SAP_CLOUD_SDK_BOT_NAME }}
 
   generate-api-docs:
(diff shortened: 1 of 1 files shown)


.github/workflows/bump.yml
```diff
@@ -53,6 +53,8 @@ jobs:
         uses: ./.github/actions/commit-and-tag
         with:
           version: ${{ steps.bump.outputs.version }}
+          user-email: ${{ vars.SAP_CLOUD_SDK_BOT_EMAIL }}
```
davidkna-sap, line 56: I think this is beneficial, but I would think push checks do not decide based on user e-mails as they are user-settable.

marikaner: Very true. I thought maybe the combination of token + user is needed, but I might be wrong.

marikaner:
@KavithaSiva I now changed the branch protection rules into rulesets and it works with that. As @davidkna-sap pointed out, possibly without this change, but it is still an improvement 😊 Feel free to merge.

.github/workflows/bump.yml
marikaner: You are right. The fix is completely unrelated.

davidkna-sap (approved):

Issue #6466, opened 2026-04-08
executeWithAxios method creates a new https.Agent per request — no connection pooling causes unbounded C heap / memory growth


serhatci:
### Describe the Bug

Our CAP application running on SAP BTP Cloud Foundry was crashing repeatedly with out-of-memory errors. Process RSS grew continuously (~420 kB/min under normal traffic) until the CF container killed the process. The leak was completely invisible to standard Node.js monitoring — process.memoryUsage() and V8 heap metrics looked normal throughout. The growth was entirely in the C heap (OpenSSL/glibc allocations).

Root cause: [executeWithAxios (http-client.ts#L434)](https://github.com/SAP/cloud-sdk-js/blob/main/packages/http-client/src/http-client.ts#L434) calls mergeRequestWithAxiosDefaults() on every request, which internally calls [getAxiosConfigWithDefaultsWithoutMethod (http-client.ts#L453)](https://github.com/SAP/cloud-sdk-js/blob/main/packages/http-client/src/http-client.ts#L453) — allocating a brand **new https.Agent with keepAlive: false** on every single outbound call:

Each agent performs a full TLS handshake, allocates an OpenSSL context in the C heap, then closes the socket. The C heap allocation is never returned to the OS — this is glibc's high-water mark behavior by design. The result is permanent, monotonically growing memory that no Node.js monitor will catch.

**How CAP is affected — verified, no alternative path exists:**

CAP's cds.connect.to() remote service implementation (@sap/cds/libx/_runtime/remote/Service.js) has a single HTTP dispatch path — every call regardless of service kind (rest, odata-v4, odata-v2, hcql) goes through:

Service.js:292  →  run(requestConfig)
client.js:57    →  executeHttpRequestWithOrigin(destination, requestConfig, { fetchCsrfToken: false })
http-client.ts  →  execute(executeWithAxios)(...)   **← executeWithAxios hardcoded, no way to override**

executeHttpRequestWithOrigin hardcodes executeWithAxios as the executor. Although execute(executeFn) accepts a custom function, CAP has no way to inject one — meaning every CAP remote service call in production hits new https.Agent() on every request with no escape hatch at the framework level.

Another user also experienced the a similar issue and wrote it in stackoverflow: https://sap.stackenterprise.co/questions/72090

### Steps to Reproduce

1) Configure a cds.connect.to() remote service (kind: rest or OData) pointing to any HTTPS endpoint in you CAP application.
[...]

PR #6498, opened 2026-04-14
feat: Improve handling of HTTP/HTTPS agent sessions

davidkna-sap:
Closes https://github.com/SAP/cloud-sdk-backlog/issues/1298

- For non-custom http/https agent default to node's global http/https agents (default: `keepAlive === true`)
- Cache custom HTTP/HTTPS agents so identical agent configurations are reused instead of creating new agents repeatedly.
- Bound the agent cache with new LRU eviction to limit memory growth due to caching, since unbounded agent creation lead to noticeable memory consumption.
- Add `getOrInsertComputed()` to `Cache<_>` (like `Map.prototype.getOrInsertComputed()`)


.changeset/heavy-foxes-deny.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sap-cloud-sdk/connectivity': minor
+'@sap-cloud-sdk/http-client': minor
+---
+
+[Improvement] Cache custom http and https agents and enable keep-alive by default
```
davidkna-sap:
```suggestion
[Improvement] Cache custom http and https agents and enable the keep-alive option by default.
```

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -283,17 +287,40 @@ function validateFormat(certificate: DestinationCertificate) {
...
-  return getProtocolOrDefault(destination) === 'https'
-    ? { httpsAgent: new https.Agent(options) }
-    : { httpAgent: new http.Agent(options) };
+  const protocol = getProtocolOrDefault(destination);
+  const cacheKey = hashCacheKey({ protocol, options });
+
+  return agentCreateCache.getOrInsertComputed(cacheKey, () => {
+    logger.debug(
+      `Creating new ${protocol.toUpperCase()} agent for destination ${destination.name || ''} with options: ${JSON.stringify(options)}`
+    );
```
davidkna-sap, line 315:
```suggestion
    logger.debug(
      `Creating new ${protocol.toUpperCase()} agent for destination ${destination.name || '<unknown>'}`
    );
```

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -82,10 +102,41 @@ export class Cache<T> implements CacheInterface<T> {
    * @param item - The entry to cache.
    */
   set(key: string | undefined, item: CacheEntry<T>): void {
```
ZhongpinWang, line 111: [q] In general, I wonder why this key can be undefined :)

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -39,21 +45,25 @@ export class Cache<T> implements CacheInterface<T> {
   /**
    * Object that stores all cached entries.
    */
-  private cache: Record<string, CacheEntry<T>>;
+  private cache: Map<string, CacheEntry<T>>;
```
ZhongpinWang, line 49: Nice 👍 Map maintains the insertion order.

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -98,6 +149,17 @@ export class Cache<T> implements CacheInterface<T> {
...
 }
 
+/**
+ * Hashes the given value to create a cache key.
+ * @internal
+ * @param value - The value to hash.
+ * @returns A hash of the given value using a cryptographic hash function.
+ */
+export function hashCacheKey(value: Record<string, unknown>): string {
+  const serialized = JSON.stringify(value);
```
ZhongpinWang: Have you checked if this is sufficient? JSON.stringify cannot stringify many things like Map or Set etc.

Maybe check alternatives https://www.npmjs.com/package/fast-json-stable-stringify or something else.

ZhongpinWang (commented):
Overall very nice implementation. 👍 I love the idea of making our cache a LRU cache. Btw we also have an async cache, not sure if that one needs to be changed as well. Otherwise, some small suggestions.

packages/connectivity/src/scp-cf/cache.ts
davidkna-sap: I switched to `ohash` (v1 instead of v2 for CJS support, v1 is still getting updates), `Map`/`Set` wouldn't be an issue here, but the values can contain functions. 

I can switch to `JSON.stringify`'s `replacer` option to handle functions if you prefer that, as TS is also having some issues with `ohash` in general.

packages/connectivity/src/scp-cf/cache.ts
davidkna-sap: Claude is claiming it's for the benefit of `getDestinationCacheKey`

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -283,17 +287,40 @@ function validateFormat(certificate: DestinationCertificate) {
...
+ * @internal
+ */
+export const agentCreateCache = new Cache<HttpAgentConfig | HttpsAgentConfig>(
+  3600000, // 1 hour
+  100 // max 100 LRU-cached agents
+);
+
 /**
  * @internal
+ * Agents are cache for up to one hour, but can be evicted earlier if more than 100 agents are created.
```
marikaner:
```suggestion
 * Agents are cached for up to one hour, but can be evicted earlier if more than 100 agents are created.
```

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -283,17 +287,40 @@ function validateFormat(certificate: DestinationCertificate) {
   }
 }
 
+/**
+ * Cache for http(s) agents.
+ * Exported for testing purposes only.
+ * @internal
+ */
+export const agentCreateCache = new Cache<HttpAgentConfig | HttpsAgentConfig>(
```
marikaner:
```suggestion
export const agentCache = new Cache<HttpAgentConfig | HttpsAgentConfig>(
```

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -71,9 +81,19 @@ export class Cache<T> implements CacheInterface<T> {
...
+        this.cache.delete(key!);
+        return undefined;
+      }
+      // LRU cache: Move accessed entry to the end of the Map to mark it as recently used
+      if (this.maxSize !== Infinity) {
+        this.cache.delete(key!);
+        this.cache.set(key!, entry);
+      }
+    }
+    return entry?.entry;
```
marikaner: [pp]
```suggestion
    if (key) {
      const entry = this.cache.get(key);
      if (isExpired(entry)) {
        this.cache.delete(key!);
        return undefined;
      }
      // LRU cache: Move accessed entry to the end of the Map to mark it as recently used
      if (this.maxSize !== Infinity) {
        this.cache.delete(key);
        this.cache.set(key, entry);
      }
      return entry.entry;
    }
```

marikaner (commented):
Didn't get through everything yet. Seems like an improvement 😊

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -283,17 +287,40 @@ function validateFormat(certificate: DestinationCertificate) {
...
   destination: HttpDestination,
   options: https.AgentOptions
 ): HttpAgentConfig | HttpsAgentConfig {
-  return getProtocolOrDefault(destination) === 'https'
-    ? { httpsAgent: new https.Agent(options) }
-    : { httpAgent: new http.Agent(options) };
+  const protocol = getProtocolOrDefault(destination);
+  const cacheKey = hashCacheKey({ protocol, options });
+
+  return agentCreateCache.getOrInsertComputed(cacheKey, () => {
```
davidkna-sap:
```suggestion
  return agentCache.getOrInsertComputed(cacheKey, () => {
```

packages/connectivity/src/scp-cf/cache.ts
ZhongpinWang: I also did some search, but tbh none of them seems to be very actively maintained. Maybe they just work without problem. 

Also maybe we can try `require()` now since node 20 is (very soon) EOL

packages/connectivity/src/scp-cf/cache.ts
davidkna-sap: Handled this differently because `this.cache.get(key);` can return `undefined`.

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -71,9 +82,25 @@ export class Cache<T> implements CacheInterface<T> {
...
+
+    if (isExpired(entry)) {
+      this.cache.delete(key!);
+      return undefined;
+    }
+
+    // LRU cache: Move accessed entry to the end of the Map to mark it as recently used
+    if (this.maxSize !== Infinity) {
+      this.cache.delete(key!);
+      this.cache.set(key!, entry);
```
marikaner: [q] is the ! needed? There is a nullish check on the top.

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -39,21 +46,25 @@ export class Cache<T> implements CacheInterface<T> {
...
   /**
    * Creates an instance of Cache.
    * @param defaultValidityTime - The default validity time in milliseconds. Use 0 for unlimited cache duration.
+   * @param maxSize - The maximum number of entries in the cache. Use Infinity for unlimited size. Items are evicted based on a least recently used (LRU) strategy.
    */
-  constructor(private defaultValidityTime: number) {
-    this.cache = {};
+  constructor(
+    private defaultValidityTime: number,
+    private maxSize = Infinity
```
marikaner: [pp] Call this `capacity`.

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -82,10 +109,41 @@ export class Cache<T> implements CacheInterface<T> {
...
   set(key: string | undefined, item: CacheEntry<T>): void {
-    if (key) {
-      const expires = item.expires ?? this.inferDefaultExpirationTime();
-      this.cache[key] = { entry: item.entry, expires };
+    if (!key) {
+      return;
+    }
+    if (
+      this.cache.size >= this.maxSize &&
+      this.cache.size > 0 &&
```
marikaner:
```suggestion
      this.cache.size &&
```

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -82,10 +109,41 @@ export class Cache<T> implements CacheInterface<T> {
...
+      !this.cache.has(key)
+    ) {
+      // Evict the least recently used (LRU) entry
+      const lruKey = this.cache.keys().next().value;
+      this.cache.delete(lruKey!); // SAFETY: size > 0
```
marikaner, line 119: [q] What does the comment tell me?

packages/connectivity/src/scp-cf/cache.spec.ts
```diff
@@ -97,4 +97,138 @@ describe('Cache', () => {
...
+    });
+    it('preserves arrays as arrays (does not coerce to object)', () => {
+      expect(hashCacheKey({ arr: [1, 2, 3] })).not.toEqual(
+        hashCacheKey({ arr: { 0: 1, 1: 2, 2: 3 } })
+      );
+    });
+  });
+
+  describe('getOrInsertComputed', () => {
+    it('calls computeFn on cache miss and stores result', () => {
```
marikaner: [req] The test does not check that the result is stored.

packages/connectivity/src/scp-cf/cache.spec.ts
```diff
@@ -97,4 +97,138 @@ describe('Cache', () => {
...
+      expect(compute).toHaveBeenCalledTimes(1);
+    });
+
+    it('returns cached value on subsequent calls without calling computeFn again', () => {
+      const compute = jest.fn().mockReturnValue({ entry: destinationOne });
+      cacheOne.getOrInsertComputed('cached', compute);
+      const result = cacheOne.getOrInsertComputed('cached', compute);
+      expect(result).toEqual(destinationOne);
+      expect(compute).toHaveBeenCalledTimes(1);
+    });
```
marikaner, line 212: [req] This test does what the test above actually describes. I think we can combine them.

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -82,10 +109,41 @@ export class Cache<T> implements CacheInterface<T> {
...
+    this.cache.set(key, { entry: item.entry, expires });
+  }
+
+  getOrInsertComputed(
+    key: string | undefined,
+    computeFn: () => CacheEntry<T>
+  ): T {
+    if (!key) {
+      return computeFn().entry;
+    }
```
marikaner: [req] I wasn't sure about this, because it seems a little off: It is weird, that this just calls the compute function without caching it. But, when you have a case where the key might not be defined and you want to have a default you don't have to differentiate the cases. So I guess it is better as is. 

But, I am missing signs that this is intended, though, aka. a test and docs.

marikaner (approved):
I left a few comments, please especially take note of the ones with "[req]". But no big blockers.

packages/connectivity/src/scp-cf/cache.ts
davidkna-sap: I've removed this special handling, I added in on purpose but is currently not a useful feature in practice.

packages/connectivity/src/scp-cf/cache.ts
davidkna-sap: The comment explains why the `!` operator is safe to apply here.

packages/connectivity/src/scp-cf/cache.ts
davidkna-sap: It was needed before the refactoring, I was able to remove them now.

serhatci:
https://github.com/SAP/cloud-sdk-js/issues/6466

This PR is going to solve above issue and our CAP project development progress depends on this bug to resolve. Can we help you somehow the accelerate to merge this PR?

PR #6501, opened 2026-04-15
chore: Switch to `client-id`

marikaner:
`app-id` is deprecated.


davidkna-sap (approved):

davidkna-sap:
@marikaner FYI this PR is empty now

PR #6507, opened 2026-04-17
chore: Add composite setup action for pnpm

davidkna-sap:
Add composite setup action for pnpm

Tested via https://github.com/SAP/ai-sdk-js/pull/1749

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -5,16 +5,10 @@ inputs:
     description: 'Node.js version'
     default: '22'
   registry-url:
-    description: 'The registry URL to use for npm packages'
-    default: 'https://registry.npmjs.org'
-  use-pnpm-registry-mirror:
-    description: 'Whether to configure pnpm to use the SAP Artifactory npm mirror'
-    default: 'true'
-  pnpm-registry-mirror-url:
-    description: 'The registry URL to use for pnpm installs when the SAP Artifactory mirror is enabled'
-    default: 'https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/'
-  pnpm-registry-mirror-token:
-    description: 'The auth token for the SAP Artifactory npm mirror'
+    description: 'The registry URL to use for npm packages. Defaults to the SAP Artifactory mirror when npm-token is provided, otherwise npmjs.'
+    default: ''
+  npm-token:
+    description: 'The auth token for the npm registry'
     default: ''
   pnpm-version:
     description: 'PNPM version'
@@ -49,19 +43,10 @@ runs:
     - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
       with:
         node-version: ${{ inputs.node-version }}
-        registry-url: ${{ inputs.registry-url }}
+        registry-url: ${{ inputs.registry-url != '' && inputs.registry-url || (inputs.npm-token != '' && 'https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/' || 'https://registry.npmjs.org') }}
         cache: 'pnpm'
         package-manager-cache: ${{ inputs.package-manager-cache }}
-    - if: ${{ inputs.use-pnpm-registry-mirror == 'true' && inputs.pnpm-registry-mirror-token != '' && (github.repository == 'SAP/ai-sdk-js' || github.repository == 'SAP/cloud-sdk-js') }}
-      shell: bash
-      env:
-        PNPM_REGISTRY_MIRROR_URL: ${{ inputs.pnpm-registry-mirror-url }}
-        PNPM_REGISTRY_MIRROR_TOKEN: ${{ inputs.pnpm-registry-mirror-token }}
-      run: |
-        registry_host="${PNPM_REGISTRY_MIRROR_URL#https://}"
-        registry_host="${registry_host#http://}"
-
-        pnpm config set --location=user registry "$PNPM_REGISTRY_MIRROR_URL"
-        pnpm config set --location=user "//$registry_host:_authToken" "$PNPM_REGISTRY_MIRROR_TOKEN"
     - run: pnpm install ${{ inputs.pnpm-install-args }}
       shell: bash
+      env:
+        NODE_AUTH_TOKEN: ${{ inputs.npm-token }}
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -46,10 +46,7 @@ runs:
         elif [[ -n "$REGISTRY_TOKEN" ]]; then
           echo "url=https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/" >> $GITHUB_OUTPUT
         elif [[ "$REPO" == "SAP/ai-sdk-js" || "$REPO" == "SAP/cloud-sdk-js" ]]; then
-          workflow_file="${WORKFLOW_REF#${REPO}/}"
-          workflow_file="${workflow_file%@*}"
-
-          echo "::error file=$workflow_file::The input 'registry-token' is required when 'registry-url' is not provided."
+          echo "::error::The input 'registry-token' is required when 'registry-url' is not provided."
           exit 1
         else
           echo "url=https://registry.npmjs.org" >> $GITHUB_OUTPUT
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,67 @@
+name: setup
+description: Setup repo, node and pnpm.
+inputs:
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yml
```diff
@@ -0,0 +1,67 @@
...
+    default: 'https://registry.npmjs.org'
+  use-pnpm-registry-mirror:
+    description: 'Whether to configure pnpm to use the SAP Artifactory npm mirror'
+    default: 'true'
+  pnpm-registry-mirror-url:
+    description: 'The registry URL to use for pnpm installs when the SAP Artifactory mirror is enabled'
+    default: 'https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/'
+  pnpm-registry-mirror-token:
+    description: 'The auth token for the SAP Artifactory npm mirror'
+    default: ''
```
marikaner: [q] If there is no default, can it be omitted?

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,67 @@
...
+  pnpm-version:
+    description: 'PNPM version'
+    default: '10'
+  pnpm-install-args:
+    description: 'Arguments to pass to pnpm install'
+    default: '--frozen-lockfile'
+  package-manager-cache:
+    description: 'Whether to enable caching for pnpm. Set to false to disable caching.'
+    default: 'true'
```
marikaner, line 19: [q] Do we ever need this?

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,67 @@
...
+    default: '22'
+  registry-url:
+    description: 'The registry URL to use for npm packages'
+    default: 'https://registry.npmjs.org'
+  use-pnpm-registry-mirror:
+    description: 'Whether to configure pnpm to use the SAP Artifactory npm mirror'
+    default: 'true'
+  pnpm-registry-mirror-url:
+    description: 'The registry URL to use for pnpm installs when the SAP Artifactory mirror is enabled'
+    default: 'https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/'
```
marikaner: [pp] Would it not suffice to keep only the registry-url? And use the sap mirror by default?

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,67 @@
...
+      shell: bash
+      env:
+        PNPM_REGISTRY_MIRROR_URL: ${{ inputs.pnpm-registry-mirror-url }}
+        PNPM_REGISTRY_MIRROR_TOKEN: ${{ inputs.pnpm-registry-mirror-token }}
+      run: |
+        registry_host="${PNPM_REGISTRY_MIRROR_URL#https://}"
+        registry_host="${registry_host#http://}"
+
+        pnpm config set --location=user registry "$PNPM_REGISTRY_MIRROR_URL"
+        pnpm config set --location=user "//$registry_host:_authToken" "$PNPM_REGISTRY_MIRROR_TOKEN"
```
marikaner: [q] Shouldn't this be set through the setup-node action?

marikaner (changes requested):

.github/actions/setup/action.yml
davidkna-sap: This is to allow this sort of thing: https://github.com/SAP/ai-sdk-js/pull/1748

.github/actions/setup/action.yml
davidkna-sap: Done.

.github/actions/setup/action.yml
davidkna-sap: Done.

.github/actions/setup/action.yml
davidkna-sap: Done.

.github/actions/setup/action.yml
marikaner: That it obvious 😊 But why would we (even for releases) not want to use the cache?

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,50 @@
+name: setup
+description: Setup repo, node and pnpm.
+inputs:
+  node-version:
+    description: 'Node.js version'
+    default: '22'
+  registry-url:
+    description: 'The registry URL to use for npm packages. Defaults to the SAP Artifactory mirror when npm-token is provided, otherwise npmjs.'
```
marikaner, line 8: [q] I assume the fallback to npmjs is there for external contributors? If that is the case, why do we even have configuration for it? Apparently it is never used anyways. Also, I fear that it is too easy to forget passing the token and then you never use artifactory at all.

I didn't think this through to the end, but what do you think about checking what repo this is running on aka `if: github.repository == 'SAP/cloud-sdk-js'` then require a token and fail if not passed, otherwise if no token passed use npmjs.org.

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,50 @@
...
+        ref: ${{ inputs.ref }}
+        token: ${{ inputs.token }}
+        persist-credentials: ${{ inputs.persist-credentials }}
+    - uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5.0.0
+      with:
+        version: ${{ inputs.pnpm-version }}
+    - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+      with:
+        node-version: ${{ inputs.node-version }}
+        registry-url: ${{ inputs.registry-url != '' && inputs.registry-url || (inputs.npm-token != '' && 'https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/' || 'https://registry.npmjs.org') }}
```
marikaner: [req] Whatever we decide the logic to be here, let's put this into a separate step with proper outputs for better readability and maintainability.

.github/actions/setup/action.yml
davidkna-sap: Releases are infrequent so the cold-install cost is acceptable.
postinstall scripts (or runtime malicious code that is run in tests) from packages introduced in non-release workflows could poison the pnpm store cache, potentially hijacking a release build via the cache.
Keeping releases on a cold install also places them in a clearly separate trust domain.

See: https://docs.zizmor.sh/audits/#cache-poisoning
In the wild attack: https://snyk.io/blog/cline-supply-chain-attack-prompt-injection-github-actions/

.github/actions/setup/action.yml
davidkna-sap: This is also for moving back to the main npm registry for release builds, see https://github.com/SAP/ai-sdk-js/pull/1749/

.github/actions/setup/action.yml
marikaner: Ok, 👌

.github/actions/setup/action.yml
marikaner: Makes sense. Then, how about the following:
- we leave both inputs `registry-url` and `registry-token`
- by default we check `if: github.repository == 'SAP/cloud-sdk-js'` then require a token for artifactory and fail if not passed, otherwise if no token passed use npmjs.org
- if a `registry-url` was passed on SAP/cloud-sdk-js use it as passed

Also, as this is used in ai-sdk-js as well, maybe we include both in here (now it's getting ugly) or check for the SAP org only (okish, because all projects should use SAP artifactory anyways).
What do you think?

davidkna-sap: @marikaner Doesn't this match the way it is currently implemented?

marikaner: Haha, I just saw, that you already implemented it in a similar way 😊

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,75 @@
...
+    - uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5.0.0
+      with:
+        version: ${{ inputs.pnpm-version }}
+    - id: resolve-registry-url
+      shell: bash
+      run: |
+        if [[ -n "$REGISTRY_URL" ]]; then
+          echo "url=$REGISTRY_URL" >> $GITHUB_OUTPUT
+        elif [[ -n "$REGISTRY_TOKEN" ]]; then
+          echo "url=https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/" >> $GITHUB_OUTPUT
```
marikaner: I like this better now, because you have to set one of the values at least, making it safe against forgetfulness.
[pp] One possible issue though: When both are set, the URL is now overwritten. I would prefer that not to happen.

marikaner (approved):
I like this better. Although I have one small request still, I already approved as it should not be a blocker.

PR #6525, opened 2026-04-23
chore: Disable sap-artifactory registry in dependabot configuration

davidkna-sap:
Dependabot cannot access the required secret despite it being set.

https://github.com/SAP/cloud-sdk-js/actions/runs/24830956002/job/72679044658


KavithaSiva (approved):

ZhongpinWang:
@davidkna-sap can you check if secrets need to be set in the Dependabot tab? 
image

PR #6526, opened 2026-04-24
chore(ci): Migrate repository check to owner check in setup action

davidkna-sap:
[Even for forked repos](https://www.github.com/SAP/ai-sdk-js/pull/1597), `github.repository` evaluates to the parent's repo name, leading to unintended failures.


.github/actions/setup/action.yml
```diff
@@ -47,13 +47,8 @@ runs:
...
           echo "url=$REGISTRY_URL" >> $GITHUB_OUTPUT
-        elif [[ "$REPO" == "SAP/ai-sdk-js" || "$REPO" == "SAP/cloud-sdk-js" ]]; then
-          if [[ -n "$REGISTRY_TOKEN" ]]; then
-            echo "url=https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/" >> $GITHUB_OUTPUT
-          else
-            echo "::error::The input 'registry-token' is required when 'registry-url' is not provided."
-            exit 1
-          fi
+        elif [[ -n "$REGISTRY_TOKEN" ]]; then
+          echo "url=https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/" >> $GITHUB_OUTPUT
```
marikaner: I think it would be better to check `github.event.pull_request.head.repo.fork` => https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request (see Properties of pull_request and navigate the tree > head > repo > fork (boolean))

marikaner (changes requested):

.github/actions/setup/action.yml
davidkna-sap: I've migrated the check, hopefully that condition works better.

.github/actions/setup/action.yml
```diff
@@ -42,12 +42,12 @@ runs:
       env:
         REGISTRY_URL: ${{ inputs.registry-url }}
         REGISTRY_TOKEN: ${{ inputs.registry-token }}
-        REPO: ${{ github.repository }}
+        IS_FORK: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork && 'true' || 'false' }}
```
marikaner: [q] I think we now learned that this might be broken?

marikaner (approved):

.github/actions/setup/action.yml
davidkna-sap: I've moved it to a check on the owner I stole from cap, but it should be fine as long as it works as this is not a security boundary.

PR #6527, opened 2026-04-27
chore: Re-enable sap-artifactory registry in dependabot configuration

davidkna-sap:
This reverts commit 081c446ae3da57370265dfac881f3048cfdfed0d.

I've now added a non-action dependabot secret, which dependabot should be able to access.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ registries:
   sap-artifactory:
     type: npm-registry
     url: https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/
-    token: ${{secrets.NPM_TOKEN_ARTIFACTORY}}
+    token: ${{secrets.DBOT_NPM_TOKEN_ARTIFACTORY}}
     replaces-base: true
 updates:
   - package-ecosystem: npm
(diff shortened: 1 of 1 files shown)


marikaner (approved):

.github/dependabot.yml
```diff
@@ -1,15 +1,15 @@
...
 registries:
-  # sap-artifactory:
-  #   type: npm-registry
-  #   url: https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/
-  #   token: ${{secrets.NPM_TOKEN_ARTIFACTORY}}
-  #   replaces-base: true
+  sap-artifactory:
+    type: npm-registry
+    url: https://common.repositories.cloud.sap/artifactory/api/npm/build.releases.npm/
+    token: ${{secrets.NPM_TOKEN_ARTIFACTORY}}
```
marikaner: [pp] Please consider giving this a different name to avoid confusion, if there is an error, e.g.

marikaner (dismissed):

.github/dependabot.yml
davidkna-sap: @marikaner I've renamed the secret, please add it to the dependabot secrets list for me (I don't have access anymore).
You will also need to keep the existing secret with the name `NPM_TOKEN_ARTIFACTORY ` so that dependabot-triggered CI-workflows can still access it.

KavithaSiva (approved):

PR #6529, opened 2026-04-27
chore(ci/setup): Allow skipping dependency install and name each step

davidkna-sap:
…each step

~~Also add `continue-on-error: true` for rebuild-issues for additional resilience.~~

Fixes https://github.com/SAP/ai-sdk-js/actions/runs/24996752082/job/73195624912?pr=1777

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -66,7 +66,7 @@ runs:
         registry-url: ${{ steps.resolve-registry-url.outputs.url }}
         cache: 'pnpm'
         package-manager-cache: ${{ inputs.package-manager-cache }}
-    - name: Fetch dependencies
+    - name: Install dependencies
       env:
         NODE_AUTH_TOKEN: ${{ inputs.registry-token }}
       shell: bash
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -30,15 +30,18 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+    - name: Checkout repository
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         ref: ${{ inputs.ref }}
         token: ${{ inputs.token }}
         persist-credentials: ${{ inputs.persist-credentials }}
-    - uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5.0.0
+    - name: Setup pnpm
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yml
```diff
@@ -63,11 +66,13 @@ runs:
         registry-url: ${{ steps.resolve-registry-url.outputs.url }}
         cache: 'pnpm'
         package-manager-cache: ${{ inputs.package-manager-cache }}
-    # Install dependencies without lifecycle scripts.
-    - env:
+    - name: Fetch dependencies
```
marikaner: [pp]
```suggestion
    - name: Install dependencies
```

.github/actions/setup/action.yml
```diff
@@ -63,11 +66,13 @@ runs:
...
+      env:
         NODE_AUTH_TOKEN: ${{ inputs.registry-token }}
       shell: bash
       run: pnpm install --ignore-scripts ${{ inputs.pnpm-install-args }}
-    # Run lifecycle scripts after removing the registry token from the environment.
-    - shell: bash
+    - name: Run lifecycle scripts
+      shell: bash
+      if: ${{ !contains('--lockfile-only', inputs.pnpm-install-args) }}
```
marikaner: [q] This honestly seems like a hack to me. I think I would prefer it, if there was an option to skip installation instead, and in the using workflow we just run `pnpm install --ignore-scripts --lockfile-only`. What do you think?

.github/actions/setup/action.yml
```diff
@@ -63,11 +66,13 @@ runs:
...
+      continue-on-error: true
```
marikaner: [q] If our scripts fail, we're good to continue?

.github/actions/setup/action.yml
marikaner: Also, according to this [comment](https://github.com/dependabot/dependabot-core/issues/8186#issuecomment-2727678057), it seems that the dependabot bug should be fixed by now. However, the PR you linked seems to contradict that...

.github/actions/setup/action.yml
davidkna-sap: I forgot that this also includes the TS compilation, otherwise it would be fine in most cases. If it does fail on anything important, everything else would also fail in short order.

.github/actions/setup/action.yml
davidkna-sap: I feel like `'--lockfile-only'` was a decent-enough sentinel, but I replaced it with a flag to make the behaviour more observable.

marikaner (approved):
lgtm

.github/actions/setup/action.yml
marikaner: I also think both are not optimal, but now it is more explicit.

PR #6552, opened 2026-04-29
feat: Bump eslint to v10

davidkna-sap:
Part of #6547

- eslint-plugin-import -> eslint-plugin-import-x (faster and supports eslint 10)
- eslint-plugin-regex (most recent release 4 years ago) -> custom plugin


packages/eslint-config/index.js
```diff
@@ -6,47 +6,29 @@ module.exports = {
...
-      [
-        {
-          id: 'regexLowerCaseInternal',
-          // eslint-disable-next-line regex/invalid
-          regex: '\\@Internal',
-          // eslint-disable-next-line regex/invalid
-          message: 'You are not allowed to use @Internal. Please use @internal.'
-        }
-      ]
-    ],
```
marikaner, line 49: [q/req] There is no replacement for this one?

marikaner (changes requested):
Just one question, not sure if it was forgotten or whether we accept removing this rule for the classic config.

packages/eslint-config/index.js
davidkna-sap: It's limited to the flat configuration now, I will recheck if I can preserve compatibility.

davidkna-sap: I have fully removed it, the jsdoc plugin also denies uppercase `@Internal` via `jsdoc/check-tag-names`.

davidkna-sap: I did not find an external replacement, but the rule is still present in both the project's local and the shared flat eslint configurations. Keeping compatibility would require an additional package.

davidkna-sap:
I have confirmed that the updated package also works with the AI SDK.

packages/eslint-config/index.js
marikaner: Then we live without it.

marikaner (approved):

PR #6564, opened 2026-05-05
chore: Harden Github Actions configuration

davidkna-sap:
Related https://github.com/SAP/ai-sdk-js/pull/1822
Part of https://github.com/SAP/ai-sdk-js-backlog/issues/562

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -42,13 +42,13 @@ jobs:
           GIT_USER: cloud-sdk-js
           BOT_EMAIL: ${{ vars.SAP_CLOUD_SDK_BOT_EMAIL }}
           BOT_NAME: ${{ vars.SAP_CLOUD_SDK_BOT_NAME }}
+          FULL_VERSION: ${{ inputs.version }}
         run: |
           gh auth setup-git
           git config user.email "$BOT_EMAIL"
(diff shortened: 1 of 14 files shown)


.github/workflows/api-docs.yml
```diff
@@ -42,20 +42,21 @@ jobs:
...
+          FULL_VERSION: ${{ inputs.version }}
         run: |
           gh auth setup-git
           git config user.email "$BOT_EMAIL"
           git config user.name "$BOT_NAME"
 
           # Extract major version from input (e.g., v4.5.0 -> v4)
-          FULL_VERSION="${{ inputs.version }}"
-          MAJOR_VERSION=$(echo $FULL_VERSION | sed 's/\(v[0-9]*\).*/\1/')
+          # shellcheck disable=SC2001
```
marikaner, line 52: [q] I forgot to ask on the other PR: Do we use shellcheck? I can't find evidence of that.

marikaner (approved):
lgtm, just one small question, because I just don't know the answer 😊

.github/workflows/api-docs.yml
davidkna-sap: I only used it locally via actionlint, but I did not add the tool to CI because they only have `curl | bash` installers.

PR #6569, opened 2026-05-06
chore: Add links to commits in changelog

davidkna-sap:
Related https://github.com/SAP/ai-sdk/issues/169


.github/workflows/release.yml
```diff
@@ -39,6 +39,8 @@ jobs:
           token: ${{ steps.app-token.outputs.token }}
           fetch-depth: 0 # otherwise, there would be errors pushing refs to the destination repository.
           path: ./cloud-sdk
+      - name: Fetch full commit history (blobless)
+        run: git fetch --filter=blob:none --unshallow || git fetch --filter=blob:none
```
marikaner: [q] Why do we need the `||` ? I just wonder if the latter is sufficient, why not just use that?

scripts/add-changelog.ts
```diff
@@ -1,9 +1,66 @@
...
+const SHORT_SHA_AT_LINE_END_RE = /\(([0-9a-f]{7,12})\)$/;
+
+async function getRepo(): Promise<string> {
+  const { stdout } = await execFileAsync('git', [
+    'remote',
+    'get-url',
+    'origin'
+  ]);
+  // Handles both https://github.com/owner/repo(.git) and git@github.com:owner/repo(.git)
+  const match = stdout.trim().match(/github\.com[:/](.+?)(?:\.git)?$/);
```
marikaner, line 23: [pp] Maybe add a comment somewhere what the expected match[1] is: `owner/repo`.

marikaner (approved):
Nice improvement, I am only wondering if this could already be done earlier, when the short sha is added.

.github/workflows/release.yml
davidkna-sap: I can remove it, `--unshallow` may fail if the repository is not shallow.

PR #6609, opened 2026-05-18
feat: Inherit Content-Type from `Blob` bodies

davidkna-sap:
Related https://github.com/SAP/ai-sdk/pull/464

Should improve batch API UX.

### DIFF
--- a/packages/openapi/src/openapi-request-builder.spec.ts
+++ b/packages/openapi/src/openapi-request-builder.spec.ts
@@ -155,7 +155,9 @@ describe('openapi-request-builder', () => {
     expect(httpClient.executeHttpRequest).toHaveBeenCalledWith(
       sanitizeDestination(destination),
       expect.objectContaining({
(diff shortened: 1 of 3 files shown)


packages/openapi/src/openapi-request-builder.spec.ts
InjunPark-sap, line 1: There is an edge case that I would discuss with you : 

FileApi (in ai-sdk-js) hardcodes `headerParameters: { 'content-type': '*/*' }` in generated code.

So this auto-detection won't kick in for file upload endpoints — the `*/*` will win instead. Should we track a follow-up to clean up that generated default?

InjunPark-sap (commented):
LGTM exceopt 1 comment regarding `packages/openapi/src/openapi-request-builder.spec.ts`

packages/openapi/src/openapi-request-builder.spec.ts
davidkna-sap: Sounds good to me.

PR #6632, opened 2026-05-29
chore: Replace internal.js shim with explicit export

davidkna-sap:
1. Remove the `internal.js` shims which can cause issues with swc and move to dedicated export map.
1. To still support `/nested/imports` where necessary add [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) (`INTERNAL_ONLY/export/name`) tied to a new `workspace` condition and enable in project `tsconfig` and `jest`
1. This required adjusting `jest` mocking
1. This also required moving to more recent `moduleResolution`s
1. As `ncc` does not support conditions, move to `rolldown` (configured via `test-util` package helper)
1. Configure `rolldown` to rewrite `__filename` and `__dirname` to `import.meta.*`, removing the need for excluding these from the bundle with nested imports.
1. Disable `prettier` on generated `.github/actions` to avoid slowdown

Closes https://github.com/SAP/ai-sdk-js-backlog/issues/576


packages/generator/package.json
```diff
@@ -12,6 +12,22 @@
...
+    },
+    "./internal": {
+      "types": "./dist/internal.d.ts",
+      "default": "./dist/internal.js"
+    },
+    "./INTERNAL_ONLY/test/test-util/create-generator-options": {
+      "workspace": {
+        "types": "./test/test-util/create-generator-options.ts",
+        "default": "./test/test-util/create-generator-options.ts"
+      }
```
marikaner: [q] What is `INTERNAL_ONLY`?

davidkna-sap: For any exported files that requires the `workspace` condition (not importable directly by third party packages, afaik even transitively), I added this so it's easier to spot in diffs so it does not end up in any of the non-test production code.

marikaner: I read up on this. Really nice! What I don't like though, is that you have to import it using "INTERNAL_ONLY". Wouldn't it be nicer to export everything like (aka. directly) this for tests?
I would expect something along the lines of:
```json
 ".": {
      "types": "./dist/index.d.ts",
      "default": "./dist/index.js",
      "workspace": {
        "types": "./internal.ts",
        "default": "./internal.ts"
      }
    },
  "./test/test-util/create-generator-options": {
      "workspace": {
        "types": "./test/test-util/create-generator-options.ts",
        "default": "./test/test-util/create-generator-options.ts"
      },
```

I am not even sure you can configure it this way and it would be ok to consider something like this a follow up.

davidkna-sap: It doesn't need to have that prefix, but I wanted to make it very visible as tests will also have the condition enabled; and I wanted to ensure that it's not ever imported in production code.

marikaner: For me personally it is clear that test/test-util should not be imported in production code. Even clearer than INTERNAL_ONLY, because internal does not tell me that it is only for tests. Due to the workspace condition it wouldn't affect users only us, right?

test-packages/test-services-odata-common/generate-common-entity.ts
```diff
@@ -2,12 +2,12 @@
 
 import { promises } from 'fs';
 import { join, resolve } from 'path';
-import { createOptions } from '@sap-cloud-sdk/generator/test/test-util/create-generator-options';
-import { generate } from '@sap-cloud-sdk/generator/src';
+import { createOptions } from '../../packages/generator/test/test-util/create-generator-options';
```
marikaner: [q] Why is this no longer referencing the generator? Or INTERNAL_ONLY?

marikaner (changes requested):

test-packages/test-services-odata-common/generate-common-entity.ts
davidkna-sap: I think this was either to avoid circular dependencies or to allow mocking.

davidkna-sap: Fixed.

packages/generator/package.json
davidkna-sap: I've now removed `INTERNAL_ONLY`

marikaner (dismissed):

packages/odata-common/package.json
```diff
@@ -12,16 +12,24 @@
...
     "access": "public"
   },
   "files": [
     "dist/**/*.js",
     "dist/**/*.js.map",
     "dist/**/*.d.ts",
-    "dist/**/*.d.ts.map",
-    "internal.js",
-    "internal.d.ts"
+    "dist/**/*.d.ts.map"
```
marikaner, line 32: [q] Did you try this? The documentation on [exports](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#exports) reads like this wouldn't work. But it is a bit ambiguous.

marikaner (dismissed):

packages/odata-common/package.json
davidkna-sap: I will do a canary release and try the e2e tests.

davidkna-sap: It appears to work and include files as expected.
https://www.npmjs.com/package/@sap-cloud-sdk/resilience/v/4.8.1-20260730092648.0?activeTab=code

marikaner (approved):

PR #6663, opened 2026-06-10
fix(openapi): Return `Blob` instead of `Buffer` (matching external types)

davidkna-sap:
Related https://github.com/SAP/ai-sdk-js/pull/1923


packages/openapi/src/openapi-request-builder.ts
```diff
@@ -339,6 +339,9 @@ export class OpenApiRequestBuilder<ResponseT = any> {
   ): Promise<ResponseT> {
     const response = await this.executeRaw(destination);
     if (isAxiosResponse(response)) {
+      if (response.data instanceof Buffer) {
+        return new Blob([response.data]) as ResponseT;
```
InjunPark-sap:
```suggestion
        return new Blob([response.data], {
          type: response.headers?.['content-type'] ?? ''
        }) as ResponseT;
```

InjunPark-sap: The `Blob` is created without a MIME type. We need to consider passing the `Content-Type` header.

InjunPark-sap (changes requested):

marikaner (dismissed):

marikaner:
@InjunPark-sap, please re-review 😊

packages/openapi/src/openapi-request-builder.ts
```diff
@@ -339,6 +339,16 @@ export class OpenApiRequestBuilder<ResponseT = any> {
...
     const response = await this.executeRaw(destination);
     if (isAxiosResponse(response)) {
+      if (Buffer.isBuffer(response.data)) {
+        const contentType = pickValueIgnoreCase(
+          response.headers,
+          'content-type'
+        );
+        return new Blob(
+          [response.data],
+          contentType && { type: contentType }
```
InjunPark-sap:
```suggestion
          contentType ? { type: contentType } : undefined
```

packages/openapi/src/openapi-request-builder.spec.ts
```diff
@@ -786,6 +786,83 @@ describe('openapi-request-builder', () => {
...
+      request: {}
+    } as any);
+    const result = await new OpenApiRequestBuilder<Blob>(
+      'get',
+      '/binary'
+    ).execute(destination);
+    expect(result).toBeInstanceOf(Blob);
+    expect(result.type).toBe('');
+  });
+
```
InjunPark-sap, line 865: Missing test case for empty string Content-Type. This is what would catch the bug above.

Suggestion: 
```
  it('does not throw when Content-Type is an empty string', async () => {
    const buffer = Buffer.from('binary content');
    httpSpy.mockResolvedValueOnce({
      data: buffer,
      status: 200,
      headers: { 'content-type': '' },
      config: {} as any,
      statusText: 'OK',
      request: {}
    } as any);
    const result = await new OpenApiRequestBuilder<Blob>('get', '/binary').execute(destination);
    expect(result).toBeInstanceOf(Blob);
    expect(result.type).toBe('');
  });
```

InjunPark-sap (changes requested):

InjunPark-sap:
@marikaner 
Thanks for the update — picking up `Content-Type` via `pickValueIgnoreCase` is actually an improvement over my original suggestion since it handles case-insensitive headers correctly.

One issue I figured out was introduced in the process: `contentType && { type: contentType }` passes an empty string `''` as the second argument to `new Blob()` when the server sends a `content-type: ''` header, which throws a `TypeError`. Please see the inline comments.

Feel free to give me feedback if it is right or not.

marikaner:
> @marikaner Thanks for the update — picking up `Content-Type` via `pickValueIgnoreCase` is actually an improvement over my original suggestion since it handles case-insensitive headers correctly.
> 
> One issue I figured out was introduced in the process: `contentType && { type: contentType }` passes an empty string `''` as the second argument to `new Blob()` when the server sends a `content-type: ''` header, which throws a `TypeError`. Please see the inline comments.
> 
> Feel free to give me feedback if it is right or not.

good catch! seems to be fixed now, right?

InjunPark-sap:
> > @marikaner Thanks for the update — picking up `Content-Type` via `pickValueIgnoreCase` is actually an improvement over my original suggestion since it handles case-insensitive headers correctly.
> > One issue I figured out was introduced in the process: `contentType && { type: contentType }` passes an empty string `''` as the second argument to `new Blob()` when the server sends a `content-type: ''` header, which throws a `TypeError`. Please see the inline comments.
> > Feel free to give me feedback if it is right or not.
> 
> good catch! seems to be fixed now, right?

Yes, all checks pass. Can we close it or is there anything to review?

davidkna-sap:
@InjunPark-sap Please approve if the PR looks good to you or request changes if there are further issues.

InjunPark-sap (approved):

InjunPark-sap:
@davidkna-sap Approved. Could we merge and close it despite of dependabot errror?

PR #6695, opened 2026-06-23
fix: Refactor HTTP agent cache for better runtime compatibility.

davidkna-sap:
Certain environments with older runtimes (openssl versions) like VSCode may not have `blake2s256`. I added some fallbacks.

Also get rid of node 20 compatibility handling.

### DIFF
--- a/packages/connectivity/src/scp-cf/cache.ts
+++ b/packages/connectivity/src/scp-cf/cache.ts
@@ -147,37 +147,27 @@ export class Cache<T> implements CacheInterface<T> {
   }
 }
 
-const hashAlgorithm = getCacheKeyHashAlgorithm();
-
 /**
  * Hashes the given value to create a cache key.
  * @internal
  * @param value - The value to hash.
  * @returns A hash of the given value using a cryptographic hash function.
  */
-export function hashCacheKey(value: Record<string, unknown>): string {
-  const serialized = stringify(value);
-  const algorithm = hashAlgorithm;
+export async function hashCacheKey(
+  value: Record<string, unknown>
+): Promise<string> {
+  const stringifiedValue = stringify(value);
+  const encodedValue = new TextEncoder().encode(stringifiedValue);
+  const hashBuffer = await crypto.subtle.digest('SHA-256', encodedValue);
 
-  return crypto.hash(algorithm, serialized, 'base64url');
-}
-
-function getCacheKeyHashAlgorithm(): string {
-  const availableHashes = crypto.getHashes();
-
-  if (availableHashes.includes('blake2s256')) {
-    return 'blake2s256';
-  }
-  if (availableHashes.includes('sha3-256')) {
-    return 'sha3-256';
-  }
-  if (availableHashes.includes('sha256')) {
-    return 'sha256';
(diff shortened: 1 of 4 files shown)


packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -155,14 +157,27 @@ export class Cache<T> implements CacheInterface<T> {
...
+
+  if (availableHashes.includes('blake2s256')) {
+    return 'blake2s256';
+  }
+  if (availableHashes.includes('sha3-256')) {
+    return 'sha3-256';
+  }
+  if (availableHashes.includes('sha256')) {
+    return 'sha256';
   }
```
davidkna-sap: It would likely be safe to just use `sha256`, but the other options should be considered safe for a longer period of time.

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -153,16 +153,21 @@ export class Cache<T> implements CacheInterface<T> {
...
-  // TODO: crypto.hash is available in Node.js v20.12.0 and later.
-  // Remove the fallback to crypto.createHash once Node.js 22 is the minimum supported version.
-  if (typeof crypto.hash === 'function') {
-    return crypto.hash('blake2s256', serialized, 'base64url');
+export async function hashCacheKey(
+  value: Record<string, unknown>
+): Promise<string> {
+  const stringifiedValue = stringify(value);
+  const encodedValue = new TextEncoder().encode(stringifiedValue);
+  const hashBuffer = await crypto.subtle.digest('SHA-256', encodedValue);
```
davidkna-sap, line 161: Switched to this as this is a browser API which includes a non-dynamic list of algorithms that are supported: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#parameters

packages/connectivity/src/scp-cf/cache.ts
```diff
@@ -153,16 +153,22 @@ export class Cache<T> implements CacheInterface<T> {
...
+
+  // TODO: Supported in Node.js 25 and later + browsers
+  if ((Uint8Array.prototype as any).toHex) {
+    // Use toHex if supported.
+    return (new Uint8Array(hashBuffer) as any).toHex(); // Convert ArrayBuffer to hex string.
   }
```
KavithaSiva: [pp] I personally prefer to add this later when Node 25 is our default.

KavithaSiva (dismissed):
Clean fix! Was this reported by a Cline user hitting the OpenSSL error, or did you discover it another way?

davidkna-sap:
@KavithaSiva I discovered this during Cline nightly testing.

packages/connectivity/src/scp-cf/cache.ts
davidkna-sap:
```suggestion
  // if ((Uint8Array.prototype as any).toHex) {
  //   // Use toHex if supported.
  //   return (new Uint8Array(hashBuffer) as any).toHex(); // Convert ArrayBuffer to hex string.
  // }
```

KavithaSiva (approved):

PR #6701, opened 2026-06-24
fix: destination fetch option XOR typing

davidkna-sap:
Related to https://github.com/SAP/ai-sdk-js-backlog/issues/598.

The `service` key was still hidden due to the way types were composed a higher levels.

### DIFF
--- a/packages/connectivity/src/scp-cf/destination/destination.ts
+++ b/packages/connectivity/src/scp-cf/destination/destination.ts
@@ -8,6 +8,7 @@ import type {
   DestinationCertificate,
   HttpDestination
 } from './destination-service-types';
+import type { WithoutExclusive } from '@sap-cloud-sdk/util';
 
 /**
  * Takes an existing or a parsed destination and returns an SDK compatible destination object.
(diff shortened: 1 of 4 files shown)


InjunPark-sap (dismissed):

packages/connectivity/src/scp-cf/destination/destination.ts
```diff
@@ -468,17 +467,43 @@ export function noDestinationErrorMessage(
...
+ * `service` / `destinationName` discriminator in {@link DestinationFromServiceBindingOptions}),
+ * causing `service` to disappear from completions/assignability — the bug this type fixes.
  */
-export type DestinationOrFetchOptions = Xor<
-  Destination,
-  DestinationFetchOptions & DestinationFromServiceBindingOptions
->;
+type WithoutExclusive<T, U> = {
+  [P in Exclude<keyof T, keyof U>]?: never;
+};
```
KavithaSiva: [req] Consider moving `WithoutExclusive` to `packages/util/src/types.ts` and also updating XOR type to start using this type instead.

Although, updating XOR is a breaking change, I believe the update would enforce correct behaviour of the type, for underlying types with shared keys.

@marikaner WDY think?

test-packages/type-tests/test/destination-or-fetch-options.test-d.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const mockService: Service = undefined as any;
+
+// --- DestinationOrFetchOptions ---
+
+// Destination branch: plain destination is assignable.
+expectAssignable<DestinationOrFetchOptions>({
+  url: 'https://example.com',
+  name: 'my-dest'
+});
```
KavithaSiva: [pp]
```suggestion
expectAssignable<DestinationOrFetchOptions>({
  url: 'https://example.com',
  name: 'my-dest'
} as Destination);
```

packages/connectivity/src/scp-cf/destination/destination.ts
```diff
@@ -468,17 +467,43 @@ export function noDestinationErrorMessage(
...
+/**
+ * Fetch-options side of the XOR.
+ *
+ * `DestinationFetchOptions` requires `destinationName: string`, but when looking up by
+ * `service` (via `DestinationFromServiceBindingOptions`) we don't have a name. Intersecting
+ * the two naively makes `destinationName: string & never = never`, killing the service branch.
+ * Stripping `destinationName` from `DestinationFetchOptions` and letting the inner discriminated
+ * union of `DestinationFromServiceBindingOptions` own that key fixes it.
+ */
+type FetchOptionsLeaf = Omit<DestinationFetchOptions, 'destinationName'> &
```
KavithaSiva, line 479: [q] I find the name cryptic, what about `DestinationFetchOrServiceOptions`  instead?

KavithaSiva (changes requested):
Thanks for fixing the bug, I have added some follow-up questions, suggestions.

InjunPark-sap (dismissed):

KavithaSiva (approved):
LGTM

PR #6716, opened 2026-06-29
fix: Scope cached http(s) agents more strongly

davidkna-sap:
Closes https://github.com/SAP/ai-sdk-js-backlog/issues/604.


packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -306,10 +306,10 @@ function createAgent(
   destination: HttpDestination,
   options: https.AgentOptions
 ): HttpAgentConfig | HttpsAgentConfig {
-  const protocol = getProtocolOrDefault(destination);
-  const cacheKey = hashCacheKey({ protocol, options });
+  const cacheKey = hashCacheKey({ destination, options });
```
InjunPark-sap: [Nit] destination.certificates and options already overlap — the same cert material ends up in the hash twice. Is there any functional impact on this change?

InjunPark-sap: Including the full destination busts the cache on every token refresh for `OAuth2ClientCredentials/OAuth2SAMLBearerFlow` destinations — `authTokens.value changes hourly but is never read by createAgent, so keep-alive reuse breaks entirely for those destinations.

InjunPark-sap: I would like to suggest scoping the key to only what affects agent behavior:

```
  const cacheKey = hashCacheKey({
    protocol,
    options,
    proxyConfiguration: destination.proxyConfiguration,
    cloudConnectorLocationId: destination.cloudConnectorLocationId
  });
```

InjunPark-sap (commented):
nit question and suggest changes on [http-agent.ts:309](https://github.com/SAP/cloud-sdk-js/pull/6716/changes#diff-8b273ba33aa24cb7487eb9f288fe83b1340ec5a135f5b285af79b1f298ee66bbR309)

packages/connectivity/src/http-agent/http-agent.ts
davidkna-sap: I think a 1h cache is acceptable for this use-case. In addition to `cloudConnectorLocationId` for the proxy the agent also needs to be keyed per-user.

InjunPark-sap (commented):
Two open questions:

1. Does `authTokens` actually appear in the destination object passed to `getAgentConfig` in OAuth2 flows? If yes, every token refresh produces a new cache key and keep-alive reuse is lost.

2. Is `proxyConfiguration` sufficient for per-user keying, or are there cases where the user identity affects agent behavior outside of   `proxyConfiguration.headers`?

davidkna-sap:
@InjunPark-sap 
1. I would say that is acceptable.
2. The user keying appears to work for the specific support cases, but in general other destination types appear to be less picky.

InjunPark-sap (dismissed):
Approval with the understanding that OAuth2 destinations will see reduced agent reuse on token refresh. (Worth revisiting if it shows up as a connection overhead issue in production.)

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -306,10 +306,10 @@ async function createAgent(
   destination: HttpDestination,
   options: https.AgentOptions
 ): Promise<HttpAgentConfig | HttpsAgentConfig> {
-  const protocol = getProtocolOrDefault(destination);
-  const cacheKey = await hashCacheKey({ protocol, options });
+  const cacheKey = await hashCacheKey({ destination, options });
```
marikaner: [req] I don't think it is a good idea to hash the whole destination. E.g. auth tokens would be part of the key. Consider using specific values from the destination, like name and auth type. Also, this will expire, right? So if the destination changes it will eventually be re-created?

marikaner (changes requested):

packages/connectivity/src/http-agent/http-agent.ts
davidkna-sap: What do you think about special-casing for connectivity only? So far only for that destination there are additional requirements.

Regarding expiring the http agents should also eventually be evicted from the LRU cache.

I did choose the whole destination because connectivity service seems to be very sensitive to connection re-use: they need to be different both per-user and per-connectivity-location. Including the destination token allowed satisfying the per-user requirement. I will re-evaluate the options, and whether to just default to `keepAlive: false` for the connectivity proxy.

marikaner: > What do you think about special-casing for connectivity only? So far only for that destination there are additional requirements.

That might make it more complex. 

> I did choose the whole destination because connectivity service seems to be very sensitive to connection re-use: they need to be different both per-user and per-connectivity-location. Including the destination token allowed satisfying the per-user requirement. I will re-evaluate the options, and whether to just default to `keepAlive: false` for the connectivity proxy.

Would it work to just include user and connectivity information (without secrets) for the key?

davidkna-sap: I don't have an E2E test setup, but as currently implemented for connectivity this PR extracts user/tenant information and the location Id into the identifier.

davidkna-sap: Currently it has to be connectivity specific (SAP-Connectivity-Authentication), but I could also try to extend this to extract the data from other JWT locations.

davidkna-sap: To expand on that: currently this only looks at `OnPremise`-specific keys, but this PR could also be extended to add tenant and user id data from the generic `.authToken[].value` if they are valid jwt.

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -306,6 +303,71 @@ export const defaultAgentOptions: https.AgentOptions | http.AgentOptions = {
...
+    return undefined;
+  }
+
+  const encoded = authHeader.replace(/^Bearer /i, '');
+  try {
+    const decoded = decodeJwt(encoded);
+    return { userId: userId(decoded), tenantId: getTenantId(decoded) };
+  } catch {
+    // A malformed token must not break agent creation; fall back to location/host scoping.
+    return undefined;
```
marikaner, line 390: [q] Should we then cache it at all?

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -315,14 +377,12 @@ async function createAgent(
...
 ): Promise<HttpAgentConfig | HttpsAgentConfig> {
-  const protocol = getProtocolOrDefault(destination);
-  const cacheKey = await hashCacheKey({
-    protocol,
-    options,
-    agentOptions: destination.agentOptions
-  });
+  const cacheKey = await hashCacheKey(
+    getAgentCacheKeyInput(destination, options)
+  );
```
marikaner: [pp/req] I would prefer to either create a second function `getAgentCacheKey()` and call `hashCacheKey()` in there or rename `getAgentCacheKeyInput()` => `getAgentCacheKey()` and call `hashCacheKey()` in there, so that the API does not require you to know that you have to hash the input and does it for you.

marikaner (changes requested):
generally looks very good to me

packages/connectivity/src/http-agent/http-agent.ts
davidkna-sap: done.

packages/connectivity/src/http-agent/http-agent.ts
davidkna-sap: I've made the checks stronger for the `PrincipalPropagation` case.

marikaner (approved):

packages/connectivity/src/http-agent/http-agent.ts
marikaner: @InjunPark-sap @davidkna-sap this conversation needs to be resolved to merge.

PR #6718, opened 2026-06-30
fix(http-client): Set higher default timeout of 120s for http(s) agents

davidkna-sap:
…eouts

The global HTTP(S) Agent has a 5s timeout ([source](https://nodejs.org/api/http.html#httpglobalagent)).
This is likely the cause of https://github.com/SAP/ai-sdk-js-backlog/issues/603.

### DIFF
--- a/packages/connectivity/src/http-agent/http-agent.ts
+++ b/packages/connectivity/src/http-agent/http-agent.ts
@@ -298,6 +298,15 @@ export const agentCache = new Cache<HttpAgentConfig | HttpsAgentConfig>(
 );
 
 /**
+ * Default options for the http(s) agents.
+ * @internal
+ */
+export const defaultAgentOptions: https.AgentOptions | http.AgentOptions = {
+  keepAlive: true,
+  timeout: 120000
+};
+
+/**
  * @internal
  * Agents are cached for up to one hour, but can be evicted earlier if more than 100 agents are created.
  * See https://nodejs.org/api/https.html#https_https_createserver_options_requestlistener for details on the possible options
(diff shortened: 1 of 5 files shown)


davidkna-sap (commented):
Ok, this does not solve the issue as-is, so I think the fix might belong in `@sap-ai-sdk`.

Nevertheless defaulting to OS-timeouts (presumably the case here) could still be sensible.

davidkna-sap:
Axios (should) set `req.setTimeout(0)` in our case anyway.

davidkna-sap:
Additional information:
- The default agents has timeout of 5s + keepAlive
- If keepAlive is false and timeout is undefined, no timeout happens
- If keepAlive is true and timeout is undefined, timeouts happen at ~5s
- If keepAlive is true and timeout is set, the provided timeout value is used. `Infinity` is possible, but in conjunction with keeping around a pool of sockets I imagine this could lead to socket exhaustion. Thus the alternative approach in  https://github.com/SAP/cloud-sdk-js/pull/6725 disables keepAlive instead.

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -297,6 +297,15 @@ export const agentCache = new Cache<HttpAgentConfig | HttpsAgentConfig>(
   100 // max 100 LRU-cached agents
 );
 
+/**
+ * Default options for the http(s) agents.
+ * @internal
+ */
+export const defaultAgentOptions: https.AgentOptions | http.AgentOptions = {
+  keepAlive: true,
+  timeout: 120000
```
KavithaSiva, line 306: [q] Any reason why you picked 2 mins specifically?

KavithaSiva (changes requested):
I prefer the solution at https://github.com/SAP/cloud-sdk-js/pull/6725 with some changes that I have requested for.
Setting a default of 2 mins for pooled socket timeout feels unecessary.

packages/connectivity/src/http-agent/http-agent.ts
davidkna-sap: I felt it would cover common streaming chat completion use-cases without impacting other styles of HTTP-requests unduly.

KavithaSiva:
Closing in favour of https://github.com/SAP/cloud-sdk-js/pull/6725

PR #6725, opened 2026-07-03
feat: Add agentOptions for configuration

davidkna-sap:
…tions for configuration

Alternative to https://github.com/SAP/cloud-sdk-js/pull/6718
Additional comment: https://github.com/SAP/cloud-sdk-js/pull/6718#issuecomment-4874936692

### DIFF
--- a/.changeset/five-boats-share.md
+++ b/.changeset/five-boats-share.md
@@ -1,0 +1,6 @@
+---
+'@sap-cloud-sdk/openapi-generator': minor
+'@sap-cloud-sdk/generator': minor
+---
+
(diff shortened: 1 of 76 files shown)


KavithaSiva (dismissed):
I propose that we stick to `keepAlive:true` as default as switching it to `false` disables socket pooling which impacts all users of Cloud SDK.
But, also the agent options can still be exposed via `Destination`. 
In the AI SDK then, we can overwrite the `timeout` value for the agent options.

KavithaSiva (approved):
LGTM

PR #6735, opened 2026-07-06
chore: Rewrite resolution dynamically to avoid artifactory tarballs i…

davidkna-sap:
…n public-npm mode

Related https://github.com/SAP/ai-sdk-js-backlog/issues/610.
TODO: use in ai-sdk-js if merged


.github/workflows/api-docs.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       contents: read
     runs-on: ubuntu-latest
     steps:
-      - uses: sap/cloud-sdk-js/.github/actions/setup@main
+      - uses: sap/cloud-sdk-js/.github/actions/setup@davidkna-sap_artif-issues
```
davidkna-sap: To be reverted before merge

BrigittaK307 (dismissed):
LGTM - don't forget to change the action reference

.github/workflows/build.yml
```diff
@@ -105,6 +105,17 @@ jobs:
...
+  npm-install:
+    if: inputs.canary-release-skip-checks == false
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: sap/cloud-sdk-js/.github/actions/setup@main
+        with:
+          registry-url: 'https://registry.npmjs.org'
+          skip-build: 'true'
```
InjunPark-sap, line 117: This job isn't in the `needs:` list of `canary-release-pre-check` or `draft-github-release`.

If public npm install fails, it won't block a release. Is this intentional (informational only), or should we add it to the release gate?

.pnpmfile.mjs
```diff
@@ -0,0 +1,114 @@
...
+  updateConfig(config) {
+    activeDefaultRegistry = normalizeRegistryUrl(
+      config.registries?.default ?? config.registry
+    );
+
+    return config;
+  },
+
+  preResolution({ wantedLockfile, currentLockfile, registries }) {
+    if (!shouldNormalizeTarballs(Object.values(registries ?? {}))) {
```
InjunPark-sap, line 99: `preResolution` skips normalization when Artifactory registry is detected, but `afterAllResolved` always normalizes. 

Is the intent "lockfile should always store public npm URLs even on Artifactory CI"? 

A brief comment would make this non-obvious design decision clearer for future maintainers.

InjunPark-sap (commented):
pnpm-lock.yaml has a conflict. Could you resolve it like rebase on main and regenerate the lockfile?

davidkna-sap:
Anyway, let's wait on the other action merge first for the conflict - pnpm anyway works with merge-conflicted lock files.

.pnpmfile.mjs
davidkna-sap: Done

InjunPark-sap (dismissed):
LGTM!

InjunPark-sap (approved):

PR #6747, opened 2026-07-10
chore: Try to fix dependabot pnpm handling

davidkna-sap:

Make dependabot aware of pnpm version requirement. If the pnpm version is set in package.json, the pnpm action does not support additional version overrides.
In addition, downgrade to pnpm 10 due to limited dependabot compatibility for pnpm 11 and accordingly move to CJS pnpmfile.
If `dependabot` ignores the pnpmfile again, `afterAllResolved` handling in the pnpmfile will have to be adjusted.

Requires same follow-up for ai-sdk.

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -29,7 +29,7 @@ jobs:
       contents: read
     runs-on: ubuntu-latest
     steps:
-      - uses: sap/cloud-sdk-js/.github/actions/setup@main
+      - uses: sap/cloud-sdk-js/.github/actions/setup@davidkna-sap_pnpm-fix-dbot
         with:
           registry-token: ${{ secrets.NPM_TOKEN_ARTIFACTORY }}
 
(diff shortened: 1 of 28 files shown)


.github/workflows/api-docs.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       contents: read
     runs-on: ubuntu-latest
     steps:
-      - uses: sap/cloud-sdk-js/.github/actions/setup@main
+      - uses: sap/cloud-sdk-js/.github/actions/setup@davidkna-sap_pnpm-fix-dbot
```
davidkna-sap: To be reverted

package.json
```diff
@@ -15,6 +15,13 @@
   "tsd": {
     "directory": "test-packages/type-tests/test-new"
   },
+  "devEngines": {
+    "packageManager": {
+      "name": "pnpm",
+      "version": "^11.0",
+      "onFail": "warn"
```
davidkna-sap, line 22: `error` (the default) would make any `npm` commands fail (such as npm publish or npm info)

InjunPark-sap (dismissed):

PR #6760, opened 2026-07-13
chore: Remove devEngines constraint again

davidkna-sap:
Corepack, which dependabot uses, has issues with semver constraints in package.json for package managers.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -10,7 +10,7 @@ inputs:
     description: 'The auth token for the npm registry'
   pnpm-version:
     description: 'PNPM version'
-    default: '11'
+    default: '10'
   pnpm-install-args:
     description: 'Arguments to pass to pnpm install'
     default: '--frozen-lockfile'
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -8,6 +8,9 @@ inputs:
     description: 'The registry URL to use for npm packages. Defaults to the SAP Artifactory mirror when npm-token is provided, otherwise npmjs.'
   registry-token:
     description: 'The auth token for the npm registry'
+  pnpm-version:
+    description: 'PNPM version'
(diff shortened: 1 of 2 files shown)


davidkna-sap (commented):
Alternative approach in https://github.com/SAP/cloud-sdk-js/pull/6761

KavithaSiva (dismissed):
LGTM. Ignoring the failing PR build, as this would work once we merge to main.

marikaner (dismissed):
I think I like it more, when the pnpm version is set explicitly, however we currently have version 11 as a default. I think we should switch to 10.34.5 instead, right?

marikaner:
Also, it is not clear to me which takes precedence, but I would expect the explicit setting in the pnpm setup action.

davidkna-sap:
@marikaner If `devEngines` and the option are both set, the action errors.

marikaner:
> @marikaner If `devEngines` and the option are both set, the action errors.

Makes sense. With that I also wouldn't be able to use pnpm 11 locally right? I guess this restriction is not needed, right?

davidkna-sap:
@marikaner Using pnpm 11 and 10 together in general can lead to issues as pnpm 11 will add checksums to lockfiles, which are not expected by pnpm 10 (unless you are careful about not committing those). `devEngines` at `warn` level will not prevent using a different version, whereas `download` should redirect to a binary that matches the expected requirement - but that is only suitable for pnpm >=11 due to better npm compatibility (providing subcommands without shelling out to npm).

I've opened #6764 to avoid temporary build failures on `main` during changes related to these features.

davidkna-sap (commented):
testing: #6766

davidkna-sap:
For the record: The `packageManager` field also exists, but [pnpm documents it as an legacy option](https://pnpm.io/package_json#devenginespackagemanager).

marikaner (approved):

PR #6761, opened 2026-07-13
chore: Only allow single pnpm version for dependabot (corepack)

davidkna-sap:

Alternative approach to solving #6760

https://www.github.com/dependabot/dependabot-core/issues/13722

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "devEngines": {
     "packageManager": {
       "name": "pnpm",
-      "version": "^10",
+      "version": "10.34.5",
       "onFail": "warn"
     }
   },


package.json
```diff
@@ -18,7 +18,7 @@
   "devEngines": {
     "packageManager": {
       "name": "pnpm",
-      "version": "^10",
+      "version": "10.34.5",
```
KavithaSiva, line 21: [q] Is this version correct? don't we set pnpm 11 in ci workflows?

davidkna-sap: Apparently dependabot still [has various issues around pnpm11 handling](https://www.github.com/dependabot/dependabot-core/issues/14794).

marikaner (approved):

PR #6764, opened 2026-07-14
chore(setup): prioritize version information from package.json

davidkna-sap:

This should give more flexibility for changing this in the future.
Testing via #6765


marikaner (changes requested):
I find this quite complex and unintuitive. Why not set the default version to 10 instead?

davidkna-sap:
@marikaner I wanted to add this to allow changing the pnpm-version globally without moving away from pinning `@main`, and it would make moving to `devEngines` easier once properly supported.

marikaner:
> I find this quite complex and unintuitive. Why not set the default version to 10 instead?

correction: 10 is the default now, but i think this part here is not necessary. wdyt?

davidkna-sap:
Okay, will close this.

PR #6777, opened 2026-07-15
chore: add dev dependency config

marikaner:
I found, that dev dependencies use the same dependabot rules as prod dependencies. I think we should update them to the highest possible version to reduce possible friction in development. We could even consider pinning them.


davidkna-sap (approved):
FYI renovate with best practices enabled defaults to pinning devDeps.

PR #6781, opened 2026-07-16
feat: Support OpenAPI 3.1

ZhongpinWang:
Vibe coded solution to support openapi 3.1

Most changes verified. There might be some generated non-sense though that I missed. But overall this PR shows the possibility and effort to keep maintaining our own generator.

**OAI Spec Releases**

- [3.1.2](https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.2)
- [3.1.1](https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.1)
- [3.1.0](https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0)
- [3.1.0-rc1](https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0-rc1)
- [3.1.0-rc0](https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0-rc0)


ZhongpinWang:
(Generated, but looks pretty legit to me)

# OpenAPI 3.1 Branch — Change Analysis & Coverage Report

## What the Branch Contains

The `support-openapi-3.1` branch adds OpenAPI 3.1.0 support to `@sap-cloud-sdk/openapi-generator`.

---

## Root Cause Fixed

The original generator silently accepted 3.1 documents but parsed them as 3.0, producing broken or `any` output. Two root causes:

1. `swagger2openapi` only converts Swagger 2.0 → OAS 3.0. For any `3.x` document it short-circuits and passes the document through unchanged — including 3.1, which then got cast to `OpenAPIV3.Document` and parsed with 3.0 assumptions.
2. No version detection existed anywhere; `swagger-parser` would accept 3.1 without complaint, hiding the problem.

---

## Feature-by-Feature Changes

### 1. Type System — JSON Schema 2020-12 Alignment

#### New input type `OpenApiV3xSchema` (`src/openapi-types.ts`)

A permissive input interface that accepts both 3.0 and 3.1 schema shapes. It carries an index signature (`[key: string]: any`) so any JSON Schema 2020-12 keyword can flow through even if not explicitly modeled. All parser functions (`parseSchema`, `parseObjectSchema`, etc.) now take this type instead of `OpenAPIV3.SchemaObject`.

#### `type` as array — `["string", "number"]` → union `anyOf`

`getType` in `src/parser/type-mapping.ts` now accepts `string | string[] | undefined`. Array types are mapped element-by-element to TypeScript types and joined as a deduplicated union (e.g. `string | number`). Helper functions `getSchemaTypes()` and `hasType()` in `src/parser/schema.ts` normalize the `type` field across both versions.

#### Nullability via `"null"` in type array (replaces `nullable: true`)

`isNullableSchema()` in `src/parser/schema.ts` checks both `nullable: true` (3.0) and `"null"` in the type array (3.1). `stripNullability()` removes these markers before parsing to prevent emitting `| null` twice. All property and persisted-schema nullable detection updated to use these helpers.

#### `const` keyword

New output type `OpenApiConstSchema` and `parseConstSchema()` function. Literal values are serialized: strings become quoted (`'fixed'`), nulls become `null`, objects are `JSON.stringify`-ed, primitives use `String()`. The file-serializer now emits const schemas as TypeScript literal types.

#### `prefixItems` (tuples)

New output type `OpenApiTupleSchema` and `parseTupleSchema()` function. Positional items from `prefixItems` become a TypeScript tuple `[T1, T2, ...]`. The `items` keyword (governing additional elements beyond the tuple) becomes a rest element: `[T1, T2, ...T3[]]`. Serialized in the file-serializer.

#### `contentEncoding` / `contentMediaType`

Content-encoded strings (e.g. `contentEncoding: "base64"`) are mapped to `Blob`, consistent with the existing `format: binary` → `Blob` mapping. Both keywords are also collected into schema properties for JSDoc emission.

#### `examples` array in schemas

`parseSchemaProperties` now collects the plural `examples` field. JSDoc generation in `src/schema-util.ts` renders each array entry as a separate `@example` tag. When both `example` (deprecated singular) and `examples` are present, a warning is logged and `example` is dropped in favour of `examples`.

#### Numeric `exclusiveMinimum` / `exclusiveMaximum`

These were previously not collected at all. Now collected with an `undefined`-guard (not a falsiness check, so `0` is correctly treated as meaningful). In 3.1 these are numeric bounds; the 3.0 boolean form was not relevant to code generation either, but the type is widened to `number | boolean` to accept both versions.

#### `patternProperties`

Pattern-keyed property schemas are merged into the `Record<string, V>` additional-properties type as a union. If `additionalProperties: false` but `patternProperties` exist, only the pattern schemas govern additional keys. Per-pattern type distinctions are lost when patterns have different value types (see remaining gaps).

#### `$ref` sibling `description`

In 3.1, `$ref` may carry sibling keywords. Previously `description` on a reference object in property position was forced to `undefined`. The branch preserves it.

---

### 2. Document / Top-Level Structure

#### `paths` now optional

An empty `paths: {}` object is injected before `SwaggerParser.parse()` so the underlying validator (which still requires the key) accepts paths-less documents. If a document has no paths, an info log is emitted. `parseApis` already returned an empty array for empty paths — no crash.

#### `webhooks` — warning instead of silent drop

`warnOnUnsupportedFeatures()` in `src/parser/document.ts` lists webhook names in a `logger.warn` call. No client code is generated for webhooks (semantically correct — they model inbound requests to the consumer, not outbound calls).

#### `components/pathItems` — warning

The same `warnOnUnsupportedFeatures()` function emits a warning listing path item names. No code generation for them.

---

### 3. Tests

30 new tests added across 6 spec files:

| File                                 | New tests | What it covers                                                                |
| ------------------------------------ | --------- | ----------------------------------------------------------------------------- |
| `src/parser/schema.spec.ts`          | 16        | 3.1 feature block (type arrays, null, const, tuples, patternProperties, etc.) |
| `src/parser/type-mapping.spec.ts`    | 4         | Type arrays / union mapping                                                   |
| `src/file-serializer/schema.spec.ts` | 4         | Const / tuple serialization                                                   |
| `src/schema-util.spec.ts`            | 1         | 3.1 property documentation                                                    |
| `src/parser/document.spec.ts`        | 4         | Nullable persisted schemas, paths-less docs, webhooks/pathItems warnings      |
| `src/document-converter.spec.ts`     | 1         | 3.1 passthrough / version detection                                           |

---

## OpenAPI 3.1 Release History

| Version     | Date             | Notes                                                                                                                                                                              |
| ----------- | ---------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 3.1.0 RC0   | June 2020        | Introduced webhooks, `jsonSchemaDialect`, `components/pathItems`, `license.identifier`                                                                                             |
| 3.1.0 RC1   | Sept 2020        | Refinements                                                                                                                                                                        |
| 3.1.0 RC2   | Nov 2020         | Further refinements                                                                                                                                                                |
| 3.1.0 final | February 2021    | Full JSON Schema 2020-12 alignment; `nullable` removed; `type` arrays; `const`; `prefixItems`; numeric `exclusiveMin/Max`; `$ref` siblings; `paths` optional; `responses` optional |
| 3.1.1       | October 24, 2024 | Patch only — clarifications on `enum`/`default` (`SHOULD` → `MUST`), OAuth/bearer docs, Security Considerations section. **No breaking changes.**                                  |
| 3.1.2       | In spec repo     | Clarifications only. **No breaking changes.**                                                                                                                                      |

> OpenAPI no longer follows SemVer — future minor increments may introduce breaking changes.

---

## Coverage Matrix

### Fully Covered

| Feature                                       | Handler                                      |
| --------------------------------------------- | -------------------------------------------- |
| `type` as array                               | `getSchemaTypes` / `getType` with union join |
| `nullable` removed; null via type array       | `isNullableSchema` / `stripNullability`      |
| Standalone `type: "null"`                     | Explicit branch in `parseSchema`             |
| `const` keyword                               | `parseConstSchema` + serializer              |
| `prefixItems` (tuples)                        | `parseTupleSchema` with rest element         |
| `contentEncoding`/`contentMediaType` → `Blob` | Schema parsing + property collection         |
| `examples` array                              | Collected + multi-`@example` JSDoc           |
| Numeric `exclusiveMin/Max`                    | Collected with `undefined`-guard             |
| `patternProperties`                           | Merged into `additionalProperties` union     |
| `$ref` sibling `description`                  | Preserved in property parsing                |
| `paths` optional                              | Empty `paths: {}` injected before validation |
| `webhooks`                                    | Explicit named warning                       |
| `components/pathItems`                        | Explicit named warning                       |
| `swagger2openapi` bypass for 3.x docs         | Version string detection in converter        |
| `example` + `examples` mutual exclusivity     | Warning + `example` dropped                  |

---

### Not a Generator Concern (by Design)

| Item                                        | Reason                                                  |
| ------------------------------------------- | ------------------------------------------------------- |
| `responses` optional under Operation        | `parseResponses` already handles `undefined` — no crash |
| `info.version` SemVer dropped               | Generator never validates `info.version`                |
| `info.license.identifier`                   | License data not consumed                               |
| Mutual TLS security scheme                  | Generator does not produce auth/security code           |
| Server variable `enum`/`default` strictness | Generator does not validate server variable constraints |
| `jsonSchemaDialect` / `$schema`             | `swagger-parser` handles document parsing               |
| `$ref` sibling `summary`                    | No TypeScript output for `summary` on a `$ref`          |
| 3.1.1 / 3.1.2 patch releases                | Clarification-only; no code impact                      |

---

### Known Remaining Gaps

| Gap                                                                                                                                                                                                                                                                            | Impact          | Notes                                                                                                                                                    |
| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | --------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Advanced JSON Schema 2020-12 applicators: `if`/`then`/`else`, `$defs`, `unevaluatedProperties`, `unevaluatedItems`, `dependentSchemas`, `dependentRequired`, `propertyNames`, `contains`/`minContains`/`maxContains`, `$dynamicRef`/`$dynamicAnchor`, `$vocabulary`, `$anchor` | Low–Medium      | Silently ignored; schemas using them degrade toward `any` or `Record<string,any>`. Would require a full JSON Schema 2020-12 evaluation layer to close.   |
| `patternProperties` with distinct per-pattern value types                                                                                                                                                                                                                      | Low             | All pattern schemas collapsed into a single `Record<string, V>` union — per-pattern type distinctions are lost when patterns have different value types. |
| `webhooks` code generation                                                                                                                                                                                                                                                     | Design decision | No callable client methods emitted; warning only. Semantically correct for an outbound client SDK, but a gap if webhook handler stubs are ever desired.  |
| `components/pathItems` code generation                                                                                                                                                                                                                                         | Low             | Warning only; path items used as `$ref` targets for webhooks remain unresolved in the generated client.                                                  |

---

## Conclusion

The branch is a solid, backward-compatible first implementation of 3.1 support. All breaking changes from OAS 3.0 → 3.1 that are relevant to TypeScript client codegen are addressed. The remaining gaps are either advanced JSON Schema applicator keywords (requiring a full evaluation engine, out of scope) or structural features (webhooks, pathItems) that do not map naturally to an outbound client SDK.

packages/openapi-generator/src/parser/api.ts
```diff
@@ -6,23 +6,30 @@ import { nameOperations } from './operation-naming';
...
 export function parseApis(
-  document: OpenAPIV3.Document,
+  document: OpenAPIV3.Document | OpenAPIV3_1.Document,
   refs: OpenApiDocumentRefs,
   options: ParserOptions
 ): OpenApiApi[] {
+  if (!document.paths) {
+    throw new Error('document.paths must be set before calling parseApis.');
+  }
+  if (!Object.keys(document.paths).length) {
```
davidkna-sap: [pp] Handle `!document.paths` with `return []` path.

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -75,11 +164,57 @@ export function parseSchema(
...
+  // OpenAPI 3.1 / JSON Schema 2020-12 expresses binary string content via
+  // 'contentEncoding'/'contentMediaType' instead of the 3.0 'format: binary'
+  // idiom. Map such content-encoded strings to 'Blob', consistent with the
+  // binary handling elsewhere.
+  if (
+    hasType(schema, 'string') &&
+    (schema.contentEncoding || schema.contentMediaType)
+  ) {
+    return { type: 'Blob' };
+  }
```
davidkna-sap, line 176: [pp]: Not verified this in depth, local Claude was claiming this is an issue:
```suggestion
  if (schema.contentEncoding || schema.contentMediaType)) {
    const schemaTypes = getSchemaTypes(schema);
    if (!schemaTypes.length || schemaTypes.includes('string')) {
      return { type: 'Blob' };
    }
  }
```

packages/openapi-generator/src/parser/document.ts
```diff
@@ -1,4 +1,5 @@
 import SwaggerParser from '@apidevtools/swagger-parser';
```
davidkna-sap, line 1: [q] Claude was claiming for our use `@apidevtools/json-schema-ref-parser` (or the hey-api fork) would suffice: The most recent versions are ESM, but these may already do some of the compatibility changes (missing paths?), and I expect `@apidevtools/swagger-parser` to do the jump at some point as well.
We should consider `import()`,  vendoring with CJS builds or raising consumer tsconfig requirements.
Apparently the version of `@apidevtools/json-schema-ref-parser` we use also has a 6.x CVE.

packages/openapi-generator/src/parser/document.ts
```diff
@@ -51,6 +63,37 @@ export async function parseOpenApiDocument(
   };
 }
 
+/**
+ * Warn about OpenAPI 3.1 top-level features that are not represented in the
+ * generated client, so their omission is explicit rather than silent.
+ * @param document - The OpenAPI document to inspect.
+ */
+function warnOnUnsupportedFeatures(
```
davidkna-sap, line 71: [pp] Consider warning for 3.1 in general for the time being.

davidkna-sap (commented):
The changes look decent to me, and don't look like they would break much if there is any issue.

I have not run an LLM-review because I assume you have run something of the sort.

packages/openapi-generator/src/parser/api.ts
ZhongpinWang: paths is set to [] before this method got called. But you are right, make sense to me that this method should be able to handle all cases.

packages/openapi-generator/src/parser/schema.ts
ZhongpinWang: I think the original version is correct. What your local claude suggests is that if one of the types is string with content encoding / media type set, then it is blob.

This is incorrect as in line 111 we already checks the type length can recursively parse the schema until it contains only one type. And it makes sense as if multiple types exist, it should be `anyOf` these types.

packages/openapi-generator/src/parser/document.ts
ZhongpinWang: I am not entirely sure what you mean by `@apidevtools/json-schema-ref-parser` will solve the missing `paths` issue? Because it is just resolving the `$Ref` in JSON schema and `paths` being optional comes from openapi 3.1?

Regarding `@apidevtools/swagger-parser`, I know in the readme it does not say it support 3.1. But I found https://github.com/APIDevTools/swagger-parser/issues/169 and https://github.com/APIDevTools/swagger-parser/pull/187. Kind like a grey area as someone also asked why is OAS 3.1 not officially documented as supported https://github.com/APIDevTools/swagger-parser/issues/276

If the parser does not support parsing 3.1 properties, then it is a problem. I am not sure if they have covered all new features / breaking changes though...

packages/openapi-generator/src/parser/document.ts
ZhongpinWang: But when should we remove it then 🤔

packages/openapi-generator/src/parser/document.ts
davidkna-sap: @ZhongpinWang `swagger-parser`'s `parse()` is a fairly thin wrapper around json-schema-ref's parser: https://npmx.dev/package-code/@apidevtools/swagger-parser/v/12.1.0/lib%2Findex.js#L37-L79

packages/openapi-generator/src/parser/document.ts
davidkna-sap: I would mostly prefer it as it hasn't been tested as much yet, and in case there is any breaking change. As for when to remove it, I assume somebody will eventually ask for the label to be removed.

packages/openapi-generator/src/parser/document.ts
ZhongpinWang: Looks to me that it adds some simple logic for oas properties handling https://npmx.dev/package-code/@apidevtools/swagger-parser/v/12.1.0/lib%2Futil.js#L41

But from a first look, it does support some OAS 3.1 properties.

packages/openapi-generator/src/parser/document.ts
ZhongpinWang: Okay I added a warning in document-converter.ts when checking the spec version. Leave it to the team in the future to remove it then.

davidkna-sap (approved):

davidkna-sap (approved):

PR #6782, opened 2026-07-17
chore: Support subpath export in public-api-check

davidkna-sap:

Subpath checks where not supported.
Also tested via https://github.com/SAP/ai-sdk-js/pull/2038


marikaner (approved):
lgtm. fmi: where do we use that? or are we planning to use it?

davidkna-sap:
I tried subpath packages in https://github.com/SAP/ai-sdk-js/pull/2038 (mostly for demonstration purposes). I also used them in https://github.com/SAP/ai-sdk-js/pull/2013, but recently I have felt more like it increases technical complexity too much.

PR #6797, opened 2026-07-22
chore: remove artifactory mirror

davidkna-sap:
Closes SAP/cloud-sdk-backlog#ISSUENUMBER..


.github/actions/setup/action.yml
davidkna-sap, line 1: To avoid breaking CI, this is now just discarding the token unless a specific registry-url is requested.

marikaner: I think we should remove the registry-url option altogether and I think it would be easier to do everything in one PR.

marikaner (changes requested):

davidkna-sap:
Closing in favour of #6799.

PR #6806, opened 2026-07-24
chore: avoid importing prettier for config

davidkna-sap:

Otherwise tries to import prettier in sap/ai-sdk-js#2068, hopefully this fixes the failure.
I assume #6632 could also fix the issue.

### DIFF
--- a/.github/actions/check-public-api/index.js
+++ b/.github/actions/check-public-api/index.js
@@ -1178,6 +1178,2638 @@ module.exports = diagnostics;
 
 /***/ }),
 
+/***/ 4806:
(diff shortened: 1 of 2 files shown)


marikaner (commented):
I don't understand how this would fix the issue in ai-sdk?
I think the problem is, that `transpileDirectory()` uses prettier. `check-public-api` depdends on the `^workspace` version of generator-common. I wonder what this resolves to and whether this is part of the problem?
WDYT?

davidkna-sap:
I tested it, prettier is still causing issues with the oxfmt PR, because it is still unable to import it.
#6632 does work, as the other bundler adds prettier to its bundle.

PR #6808, opened 2026-07-24
chore: replace execa with tinyexec & use plain node for /scripts

davidkna-sap:

Requires `import()` but cmd execution is asynchronous anyway.


test-packages/integration-tests/test/exception-logger.spec.ts
```diff
@@ -1,8 +1,17 @@
 import { resolve } from 'path';
-import execa from 'execa';
 import { createLogger, kibana, local } from '@sap-cloud-sdk/util';
 import type { Logger } from 'winston';
 
+const tinyexec = import('tinyexec');
+
+async function execNodeScript(scriptPath: string): Promise<void> {
+  const { x } = await tinyexec;
+  const result = await x('ts-node', [scriptPath]);
```
marikaner: [q] Shouldn't this work too?
```suggestion
  const result = await x('node', [scriptPath]);
```

test-packages/integration-tests/test/odata-negative-case.spec.ts
```diff
@@ -1,11 +1,20 @@
...
 import {
   testOutputRootDir,
   testResourcesDir
 } from '../../../test-resources/generator';
 
+const tinyexec = import('tinyexec');
+
+async function execGenerator(args: string[], cwd: string): Promise<void> {
+  const { x } = await tinyexec;
+  const result = await x('node', args, { nodeOptions: { cwd } });
```
marikaner: [pp] Given that we always pass `__dirname` here, just add it to the function:
```suggestion
async function execGenerator(args: string[]): Promise<void> {
  const { x } = await tinyexec;
  const result = await x('node', args, { nodeOptions: { cwd: __dirname } });
```

test-packages/integration-tests/test/openapi-negative-case.spec.ts
```diff
@@ -1,11 +1,20 @@
...
 import { existsSync, promises } from 'fs';
-import execa from 'execa';
 import {
   testOutputRootDir,
   testResourcesDir
 } from '../../../test-resources/generator';
 
+const tinyexec = import('tinyexec');
+
+async function execGenerator(args: string[], cwd: string): Promise<void> {
```
marikaner: Same here. Also, consider reusing this function by putting it into a util file.

marikaner (changes requested):

marikaner (dismissed):

marikaner (approved):

PR #6849, opened 2026-08-03
chore: Fix handle content media type correctly in OAS 3.1 OpenAPI generator

ZhongpinWang:
Relates to SAP/ai-sdk-js-backlog#633.

Previously, the logic for checking where the property is a blob was wrong. When `contentMediaType` is not a binary type, then it should not be blob.

### DIFF
--- a/packages/openapi-generator/src/parser/schema.ts
+++ b/packages/openapi-generator/src/parser/schema.ts
@@ -198,15 +198,12 @@ function isBinaryMediaType(mediaType: string | undefined): boolean {
   if (lower.startsWith('text/')) {
     return false;
   }
-  if (
+  return !(
     lower === 'application/json' ||
     lower === 'application/xml' ||
     lower.endsWith('+json') ||
     lower.endsWith('+xml')
-  ) {
-    return false;
-  }
-  return true;
+  );
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/openapi-generator/src/parser/schema.ts
```diff
@@ -180,6 +183,31 @@ export function parseSchema(
   };
 }
 
+/**
+ * Returns true for media types whose content is binary (not human-readable
+ * text). Text-based types — text/*, application/json, application/xml, and
+ * structured-syntax suffixes like +json or +xml — are excluded because their
+ * string values remain plain text and should stay typed as `string`.
+ */
+function isBinaryMediaType(mediaType: string | undefined): boolean {
```
davidkna-sap, line 194: [pp] Use `content-type` package for parsing.

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -180,6 +183,32 @@ export function parseSchema(
...
+  }
+  if (
+    lower === 'application/json' ||
+    lower === 'application/xml' ||
+    lower.endsWith('+json') ||
+    lower.endsWith('+xml')
+  ) {
+    return false;
+  }
+  return true;
```
davidkna-sap: [nth]
```suggestion
  return !(
    lower === 'application/json' ||
    lower === 'application/xml' ||
    lower.endsWith('+json') ||
    lower.endsWith('+xml')
  );
```

packages/openapi-generator/src/parser/schema.ts
ZhongpinWang: Well okay I am now using this package for parsing.

davidkna-sap (approved):

PR #6852, opened 2026-08-04
fix: Extend OnPremise HTTP agent cache key to scope it more strongly

davidkna-sap:

This makes the cache key more encompassing, hopefully this is enough: alternative would be to give up on keepAlive for this case.


packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -326,52 +329,106 @@ function getAgentCacheKeyInput(
...
+ * that keep-alive tunnels are not reused across subaccounts. The connectivity
+ * service JWT is decoded and reduced to the stable, non-secret claims
+ * `tenantId` (subaccount) and `clientId`, because the raw token rotates on
+ * every refresh and must never be part of a cache key.
+ * @param destination - Destination carrying the proxy configuration.
+ * @returns The derivation outcome for the subaccount scope.
+ */
+function getProxyAuthCacheKey(
+  destination: HttpDestination
+): CacheScope<{ tenantId?: string; clientId?: string }> {
```
InjunPark-sap, line 393: [nit] `clientId` here is the `clientid` claim from the connectivity service JWT (the XSUAA service binding identity), not a user/application client ID. Worth a brief clarification in the JSDoc — the two concepts are easy to confuse in SAP context.

packages/connectivity/src/http-agent/http-agent.ts
```diff
@@ -326,52 +329,106 @@ function getAgentCacheKeyInput(
...
 
-    // For all other OnPremise flows the principal is not strictly required, but
-    // we still scope by full available principal information for better cache isolation.
-    if (principal) {
-      keyInput.principal = principal;
-    }
+  keyInput.cloudConnectorLocationId = destination.cloudConnectorLocationId;
+  keyInput.proxyHost = destination.proxyConfiguration?.host;
+  keyInput.proxyPort = destination.proxyConfiguration?.port;
+  keyInput.target = destination.url;
```
InjunPark-sap, line 357: [minor] `destination.url` is used verbatim as the cache key dimension. Is URL normalization intentional to skip here?

If the SDK always sets a consistent URL string per destination this is fine — just worth a note or assertion that normalization is handled upstream.

InjunPark-sap (dismissed):
LGTM. just added 1 nit & 1 minor comment

packages/connectivity/src/http-agent/http-agent.ts
davidkna-sap: Updated the comments.

packages/connectivity/src/http-agent/http-agent.ts
davidkna-sap: Updated the comments.

.changeset/hungry-pens-sell.md
```diff
@@ -0,0 +1,7 @@
+---
+"@sap-cloud-sdk/connectivity": patch
+---
+
+[fix] Extend OnPremise HTTP agent cache key to scope it more strongly.
+If no stable identity context can be derived from the request, a fresh agent without keep-alive is created for each request.
+This matches the behavior of the previous implementation, which did not cache agents in any case.
```
KavithaSiva: [q] Did the previous implementation not cache agents? Wasn't it incorrectly caching agents for technical user flows?

.changeset/hungry-pens-sell.md
```diff
@@ -0,0 +1,7 @@
+---
+"@sap-cloud-sdk/connectivity": patch
+---
+
+[fix] Extend OnPremise HTTP agent cache key to scope it more strongly.
+If no stable identity context can be derived from the request, a fresh agent without keep-alive is created for each request.
```
KavithaSiva, line 7:
```suggestion
[fix] Extend OnPremise HTTP agent cache key to scope it more strongly.
The cache key now includes the subaccount identity ensuring keep-alive sockets are not reused across tenants in multi-tenant technical user flows.
If no stable identity context can be derived from the request, a fresh agent without keep-alive is created for each request.
```

.changeset/hungry-pens-sell.md
davidkna-sap: @KavithaSiva I am referring to v4.6.0 and lower here which created a new agent per-request.

davidkna-sap:
```suggestion
This matches the behavior of the previous implementation in SAP Cloud SDK v4.6.0 and lower, which did not cache agents in any case.
```

KavithaSiva (approved):

KavithaSiva (approved):

InjunPark-sap (approved):

PR #6898, opened 2026-08-17
chore: use node 26 as default

InjunPark-sap:
Closes SAP/ai-sdk-backlog#644

## What

Make Node.js 26 the default across CI.

## Follow-up

- After merge, the temporary `node-version: 26` pin in `SAP/ai-sdk-js` can be removed so it inherits this default.
- `SAP/ai-sdk-js` admins: if the `DEFAULT_NODE_VERSION` var (used by `bump.yml`) is set, update it to `26` — an empty value skips the setup action default.

No changeset — CI-only `chore`, not required by pr-check.


.github/workflows/release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         with:
           ref: 'main'
           persist-credentials: true
-          node-version: 24 # Will install npm 11 needed for trusted publishing
+          node-version: 26 # Will install npm 11 needed for trusted publishing
```
davidkna-sap:
```suggestion
```

.github/workflows/build.yml
```diff
@@ -156,7 +156,7 @@ jobs:
       - uses: sap/cloud-sdk-js/.github/actions/setup@main
         with:
           persist-credentials: true
-          node-version: 24 # Will install npm 11 needed for trusted publishing
+          node-version: 26 # Will install npm 11 needed for trusted publishing
```
davidkna-sap:
```suggestion
```

.github/workflows/build.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [22, 24]
+        node-version: [26, 24, 22]
```
davidkna-sap, line 39: [nth] I think that keeping this in ascending order is fine.

davidkna-sap (changes requested):

.github/workflows/build.yml
InjunPark-sap: in the ai-sdk-js PR @marikaner suggested 26-first, since coverage there is gated on the first matrix entry. No such coupling here, so ascending is fine — but happy to match 26-first if you'd prefer both repos consistent. wdyt?

marikaner: This is not what I meant directly, but I think this is a good argument to have the order descending. No strong opinions from my side, though.

marikaner (approved):

davidkna-sap:
@InjunPark-sap Let's not merge this while CI on main is red.

davidkna-sap (approved):

PR #6911, opened 2026-08-21
feat: support OpenAPI 3.1 contentschemas

marikaner:
This allows reflecting content schemas of strings in types, while serializing them as strings in formdata bodies. Other contexts outside of formdata bodies are not supported.

Related to https://github.com/SAP/ai-sdk-js-backlog/issues/632

### DIFF
--- a/.changeset/support-content-media-type-schema.md
+++ b/.changeset/support-content-media-type-schema.md
@@ -1,5 +1,5 @@
 ---
-'@sap-cloud-sdk/openapi-generator': patch
+'@sap-cloud-sdk/openapi-generator': minor
 ---
 
 [feat] Support `contentSchema` in OpenAPI 3.1 schemas: fields with `contentMediaType: application/json` and a `contentSchema` are now typed as the `contentSchema` type instead of `string`. Serialization to a JSON string is supported for multipart bodies via `FormDataBuilder`; non-multipart bodies are not yet handled.
--- a/.changeset/support-content-media-type-schema.md
+++ b/.changeset/support-content-media-type-schema.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/openapi-generator': patch
+---
+
+[feat] Support `contentSchema` in OpenAPI 3.1 schemas: fields with `contentMediaType: application/json` and a `contentSchema` are now typed as the `contentSchema` type instead of `string`. Serialization to a JSON string is supported for multipart bodies via `FormDataBuilder`; non-multipart bodies are not yet handled.
--- a/.changeset/support-content-media-type-schema.md
+++ b/.changeset/support-content-media-type-schema.md
@@ -3,3 +3,5 @@
 ---
 
 [feat] Support `contentSchema` in OpenAPI 3.1 schemas: fields with `contentMediaType: application/json` and a `contentSchema` are now typed as the `contentSchema` type instead of `string`. Serialization to a JSON string is supported for multipart bodies via `FormDataBuilder`; non-multipart bodies are not yet handled.
+
+[compat] Fields with `contentMediaType: application/json` and a `contentSchema` will now be typed as the `contentSchema` type instead of `string`. Re-generate your client to pick up the new types.
--- a/packages/openapi-generator/src/openapi-types.ts
+++ b/packages/openapi-generator/src/openapi-types.ts
@@ -666,4 +666,10 @@ export interface OpenApiSchemaProperties {
    * 2020-12, 3.1).
    */
   contentMediaType?: string;
+
+  /**
+   * Validates the structure of a string's decoded content when
+   * `contentMediaType` is present (JSON Schema 2020-12, 3.1).
+   */
+  contentSchema?: any;
 }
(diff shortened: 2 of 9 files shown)


.changeset/support-content-media-type-schema.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sap-cloud-sdk/openapi-generator': patch
+---
+
+[feat] Support `contentSchema` in OpenAPI 3.1 schemas: fields with `contentMediaType: application/json` and a `contentSchema` are now typed as the `contentSchema` type instead of `string`. Serialization to a JSON string is supported for multipart bodies via `FormDataBuilder`; non-multipart bodies are not yet handled.
+
+[compat] Fields with `contentMediaType: application/json` and a `contentSchema` will now be typed as the `contentSchema` type instead of `string`. Re-generate your client to pick up the new types.
```
marikaner: This is actually a breaking change. Without this generating RPT 1.5 is a little problematic. We could consider adding a CLI flag.

I would hope this doesn't affect too many users and a compatibility note might suffice.

davidkna-sap: I think it's fine as it matches the specification better.

[req] Split into two changesets, IIRC that causes issues.

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -179,39 +179,67 @@ export function parseSchema(
...
+ * string values remain plain text and should stay typed as `string`.
+ * @internal
+ */
+function isBinaryMediaType(mediaType: string | undefined): boolean {
+  const type = parseMediaTypeString(mediaType);
+  return (
+    !!type &&
+    !(
+      type.startsWith('text/') ||
+      type === 'application/json' ||
```
davidkna-sap: [nth] Use `isJsonMediaType()`.

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -179,39 +179,67 @@ export function parseSchema(
     return { type: 'Blob' };
   }
 
+  // OpenAPI 3.1 / JSON Schema 2020-12: a string with a JSON contentMediaType
+  // and a contentSchema is typed as the contentSchema type. Serialization to a
+  // JSON string is only supported for multipart bodies (via FormDataBuilder);
+  // non-multipart bodies are not yet supported.
+  if (
```
davidkna-sap: [q] Can this be combined with the above check?

packages/openapi-generator/src/openapi-types.ts
```diff
@@ -666,4 +666,10 @@ export interface OpenApiSchemaProperties {
    * 2020-12, 3.1).
    */
   contentMediaType?: string;
+
+  /**
+   * Validates the structure of a string's decoded content when
+   * `contentMediaType` is present (JSON Schema 2020-12, 3.1).
+   */
+  contentSchema?: any;
```
davidkna-sap: [pp]
```suggestion
  contentSchema?: OpenApiSchema | OpenApiReference;
```

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -179,39 +179,67 @@ export function parseSchema(
...
+      type.startsWith('text/') ||
+      type === 'application/json' ||
+      type === 'application/xml' ||
+      type.endsWith('+json') ||
+      type.endsWith('+xml')
+    )
   );
 }
 
+function isJsonMediaType(mediaType: string | undefined): boolean {
```
davidkna-sap: [pp] Use this function in `FormDataBuilder` to check if something should be JSON-stringified.

packages/openapi-generator/src/parser/schema.ts
```diff
@@ -179,39 +179,67 @@ export function parseSchema(
...
+  // OpenAPI 3.1 / JSON Schema 2020-12: a string with a JSON contentMediaType
+  // and a contentSchema is typed as the contentSchema type. Serialization to a
+  // JSON string is only supported for multipart bodies (via FormDataBuilder);
+  // non-multipart bodies are not yet supported.
+  if (
+    schema.contentSchema &&
+    hasType(schema, 'string') &&
+    isJsonMediaType(schema.contentMediaType)
+  ) {
+    return parseSchema(schema.contentSchema, refs, options);
```
davidkna-sap: [pp] Technically, for non-FormData requests this would apparently also have to use `JSON.stringify`.

davidkna-sap (changes requested):

packages/openapi-generator/src/parser/schema.ts
marikaner: Then I would have to add the generator as a dependency to the openapi package. I can however add this part: .endsWith('+json'))

davidkna-sap: That would also work for me.

packages/openapi-generator/src/parser/schema.ts
marikaner: yes they would, but: 1. we currently have no use case for this, 2. it is a lot more effort to implement this, because we need to somehow keep the information that this is a string.

we have three options:
1. keep this as a known gap
2. fix it on this pr
3. fix it on a separate pr

wdyt?

davidkna-sap: I think we can keep it as-is for now, but please add a todo comment.

davidkna-sap (approved):

PR #6948, opened 2026-09-05
fix(connectivity): Handle schemeless issuer URLs in JWT claims

### LINKED ISSUE #6923: Cant read subscriber destination
### Describe the Bug

Scenario:
We are calling const destination = await getDestination({ destinationName, jwt: userJwt }); method. 

jwt token has iss like "iastenant.account.cloud.sap" instead of "https://iastenant.account.cloud.sap". Since https is missing, cloud sdk method fail to read from subscriber account and looks up in the provider account. 

### Steps to Reproduce

Create multi tenant application
Subscribe in subscriber tenant 
Configure IAS tenant for subscriber with not forcing https in open id connection for issuer url
use getDestination method to read a subscriber destination and it will fail 

### Expected Behavior

getDestination  must work and read the destination with token from subscriber 

### Screenshots

_No response_

### Used Versions

- Node version via `node -v`:  ^25
- NPM version via `npm -v: ...
- SAP Cloud SDK version: 4.5
- For CAP users, CAP version: ...

### Code Examples

_No response_

### Log File

_No response_

### Affected Development Phase

Getting Started

### Impact

No Impact

### Timeline

_No response_

### Additional Context

_No response_

IncognitoQuack:
Closes SAP/cloud-sdk-js#6923.

An identity provider can be configured not to force `https` on the issuer URL, in which case the
`iss` (or `ias_iss`) claim arrives as a bare hostname such as `tenant.accounts.ondemand.com`. The
`URL` constructor rejects that, and two places in the connectivity package assume it will not.

### 1. `getIssuerSubdomain` throws where its caller expects a falsy return

`getIdentityServiceInstanceFromCredentials` → `extractSubdomainFromJwt` calls
`getIssuerSubdomain(payload, true)` and then does:

```ts
const subdomain = getIssuerSubdomain(payload, true);
if (!subdomain) {
  logger.warn('Could not extract subdomain from JWT assertion issuer. Falling back to service binding URL.');
}
```

The caller is written for a falsy return, but `getIssuerSubdomain` throws when `isValidUrl(issuer)`
is false, so the intended fallback never runs. The throw propagates to `retrieveServiceToken`, which
catches it and only logs a warning, so `serviceJwt` stays `undefined`, `isSubscriberNeeded()`
returns `false`, and `getDestination({ destinationName, jwt })` silently reads the destination from
the **provider** account instead of the subscriber account — the symptom reported in #6923.

### 2. `isIasToken` misclassifies the same tokens

`isIasToken` parses `decodedJwt.iss` with `new URL(...)` inside a `try` whose `catch` returns
`false`, so a schemeless IAS issuer is classified as *not* an IAS token. `getSubdomain` then takes
the XSUAA branch and calls `getIssuerSubdomain` — which, before this change, throws, and with only
the first fix in place returns a subdomain that is then sent as the `X-tenant` header in
`getExchangeTenant`. Fixing only the first function would turn a loud failure into a silently wrong
tenant header, so both are fixed here.

### Change

Schemeless issuers are prefixed with `https://` before being parsed. Values that are not URLs at all
(`'not a url'`, `''`) are still rejected, and existing `http://` and `https://` issuers, issuers
with paths, and issuers with explicit ports are unaffected.

Note that `URL.canParse` alone is not sufficient: it returns `true` for
`tenant.accounts.ondemand.com:8443`, which parses as a scheme with no host, so the prefix is applied
whenever parsing does not yield a host. There are regression tests for that case.

### Behaviour changes

- `iss: 'localhost'` now throws `Failed to determine hostname: invalid host in "https://localhost/"`
  instead of `Issuer URL in JWT is not a valid URL: "localhost"`.
- `iss: 'https://'` now throws `Failed to determine hostname` instead of `is not a valid URL`.

Both cases still throw; only the messages differ.

### Tests

`subdomain-replacer.ts` had no test coverage before this change. This PR adds
`subdomain-replacer.spec.ts` covering both `getIssuerSubdomain` and `replaceSubdomain`, including
the pre-existing behaviour, and extends `jwt.spec.ts` with `describe('isIasToken()')` and
`describe('getSubdomain()')` blocks.

### DIFF
--- a/packages/connectivity/src/scp-cf/jwt/jwt.ts
+++ b/packages/connectivity/src/scp-cf/jwt/jwt.ts
@@ -83,7 +83,12 @@ export function isIasToken(decodedJwt: JwtPayload): boolean {
     return false;
   }
   try {
-    const issUrl = new URL(decodedJwt.iss);
+    // Identity providers can be configured to issue tokens with a schemeless `iss` claim,
+    // e.g. `tenant.accounts.ondemand.com` instead of `https://tenant.accounts.ondemand.com`.
+    const issWithScheme = URL.canParse(decodedJwt.iss)
+      ? decodedJwt.iss
+      : `https://${decodedJwt.iss}`;
+    const issUrl = new URL(issWithScheme);
     const hostname = issUrl.hostname.toLowerCase();
     return (
       hostname.endsWith('.accounts.ondemand.com') ||
--- a/packages/connectivity/src/scp-cf/subdomain-replacer.ts
+++ b/packages/connectivity/src/scp-cf/subdomain-replacer.ts
@@ -14,10 +14,13 @@ export function getIssuerSubdomain(
     isIasToken && decodedJwt?.ias_iss ? decodedJwt.ias_iss : decodedJwt?.iss;
 
   if (issuer) {
-    if (!isValidUrl(issuer)) {
+    // Identity providers can be configured to issue tokens with a schemeless `iss` claim,
+    // e.g. `tenant.accounts.ondemand.com` instead of `https://tenant.accounts.ondemand.com`.
+    const issWithScheme = URL.canParse(issuer) ? issuer : `https://${issuer}`;
+    if (!URL.canParse(issWithScheme)) {
       throw new Error(`Issuer URL in JWT is not a valid URL: "${issuer}".`);
     }
-    return getHost(new URL(issuer)).split('.')[0];
+    return getHost(new URL(issWithScheme)).split('.')[0];
   }
 }
 
--- a/.changeset/olive-pumas-repeat.md
+++ b/.changeset/olive-pumas-repeat.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': patch
+---
+
+[Fixed Issue] Fix subscriber destination lookup failing when the identity provider issues tokens with a schemeless `iss` claim, e.g. `tenant.accounts.ondemand.com` instead of `https://tenant.accounts.ondemand.com`. Such issuers are now prefixed with `https://` before being parsed, instead of causing a silent fallback to the provider account.
--- a/.changeset/tidy-moons-brake.md
+++ b/.changeset/tidy-moons-brake.md
@@ -1,0 +1,5 @@
+---
+'@sap-cloud-sdk/connectivity': patch
+---
+
+[Fixed Issue] Fix IAS tokens being classified as non-IAS tokens when the identity provider issues them with a schemeless `iss` claim, e.g. `tenant.accounts.ondemand.com` instead of `https://tenant.accounts.ondemand.com`.
--- a/packages/connectivity/src/scp-cf/jwt/jwt.spec.ts
+++ b/packages/connectivity/src/scp-cf/jwt/jwt.spec.ts
@@ -4,7 +4,15 @@ import {
   mockServiceBindings,
   signedJwtForVerification
 } from '@sap-cloud-sdk/test-util-internal';
-import { audiences, decodeJwt, isXsuaaToken, retrieveJwt, userId } from './jwt';
+import {
+  audiences,
+  decodeJwt,
+  getSubdomain,
+  isIasToken,
+  isXsuaaToken,
+  retrieveJwt,
+  userId
+} from './jwt';
 
 describe('jwt', () => {
   describe('userId', () => {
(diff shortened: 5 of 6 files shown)


davidkna-sap (changes requested):
For parsing the URL, please switch to a simpler approach that avoids regular expressions e.g:
```ts
URL.canParse(issuer)
    ? new URL(issuer)
    : new URL(`https://${issuer}`);
```

IncognitoQuack:
Thanks — switched to `URL.canParse`, the regex is gone from both `subdomain-replacer.ts` and
`jwt.ts`, and I've rebased onto latest `main`.

One thing worth flagging before you re-review: the literal form has an edge case I hit while
testing it.

```js
URL.canParse('tenant.accounts.ondemand.com:8443'); // true
```

It returns `true` because the value parses as a scheme (`tenant.accounts.ondemand.com:`) with a path
and **no host** — so the `https://` prefix is not applied, `getHost` finds an empty host, and a
schemeless issuer *with a port* throws instead of resolving. That is the same class of value this PR
is meant to fix, so I didn't want to leave it broken.

The version I pushed keeps your approach and adds a host check:

```ts
const parsedIssuer = URL.canParse(issuer) ? new URL(issuer) : undefined;
const normalizedIssuer = parsedIssuer?.host ? issuer : `https://${issuer}`;
```

I added two regression tests for it (`getIssuerSubdomain` and `isIasToken`); both fail with the
plain `URL.canParse(issuer) ? … : …` form and pass with the above.

Happy to drop back to your exact snippet if you'd rather not carry the extra check — it's your call
on whether a ported issuer is worth covering.

One disclosed behaviour change either way: `iss: "https://"` now throws
`Failed to determine hostname` rather than `is not a valid URL`. Both still reject the value; only
the message differs.

packages/connectivity/src/scp-cf/subdomain-replacer.ts
```diff
@@ -14,10 +14,16 @@ export function getIssuerSubdomain(
     isIasToken && decodedJwt?.ias_iss ? decodedJwt.ias_iss : decodedJwt?.iss;
 
   if (issuer) {
-    if (!isValidUrl(issuer)) {
+    // Identity providers can be configured to issue tokens with a schemeless `iss` claim,
+    // e.g. `tenant.accounts.ondemand.com` instead of `https://tenant.accounts.ondemand.com`.
+    // A schemeless issuer with a port, e.g. `tenant.accounts.ondemand.com:8443`, is parsable
+    // as a scheme without a host, so the default scheme is added whenever there is no host.
+    const parsedIssuer = URL.canParse(issuer) ? new URL(issuer) : undefined;
+    const normalizedIssuer = parsedIssuer?.host ? issuer : `https://${issuer}`;
```
davidkna-sap: Let's not handle this, this adds additional complexity and as far as I can tell, would not occur in practice with IAS:

```suggestion
    const parsedIssuer = URL.canParse(issuer) ? new URL(issuer) :`https://${issuer}`;
```

packages/connectivity/src/scp-cf/subdomain-replacer.ts
```diff
@@ -14,10 +14,16 @@ export function getIssuerSubdomain(
...
+    if (!isValidUrl(normalizedIssuer)) {
```
davidkna-sap: [nth] replace `isValidUrl()` with `URL.canParse()`, but this could also be handled at a later date.

packages/connectivity/src/scp-cf/jwt/jwt.ts
```diff
@@ -83,7 +83,15 @@ export function isIasToken(decodedJwt: JwtPayload): boolean {
...
-    const issUrl = new URL(decodedJwt.iss);
+    // Identity providers can be configured to issue tokens with a schemeless `iss` claim,
+    // e.g. `tenant.accounts.ondemand.com`. A schemeless issuer with a port is parsable as a
+    // scheme without a host, so the default scheme is added whenever there is no host.
+    const parsedIss = URL.canParse(decodedJwt.iss)
+      ? new URL(decodedJwt.iss)
+      : undefined;
+    const issUrl = parsedIss?.host
+      ? parsedIss
+      : new URL(`https://${decodedJwt.iss}`);
```
davidkna-sap:
```suggestion
    const issWithScheme = URL.canParse(decodedJwt.iss)
      ? decodedJwt.iss
      : `https://${decodedJwt.iss}`;
    const issUrl = new URL(issWithScheme);
```

davidkna-sap (changes requested):

IncognitoQuack:
All three applied, thanks:

- Dropped the host check. `jwt.ts` now matches your suggestion verbatim, and
`subdomain-replacer.ts` uses the same shape. Agreed on the ported issuer not occurring with IAS
in practice, so the two regression tests I had added for it are removed.
- Took the `[nth]` as well: `getIssuerSubdomain` now uses `URL.canParse` instead of `isValidUrl`.
I left `replaceSubdomain` on `isValidUrl` since it is outside this change, happy to switch it if
you would rather have it consistent now.

The source diff is down to +3/-1 in each file. Rebased onto `main`, and I have updated the
description, which previously referred to the port case.

For the record, the consequence of dropping the host check: a schemeless issuer with a port throws
rather than resolving, since `URL.canParse` accepts it as a scheme without a host. Noted in the
description under behaviour changes.

Verified locally on the rebased branch: `compile` 21/21, `test:unit` 26/26, `test:type` 17/17,
`lint` 27/27 with 0 errors, connectivity suite 529 passing.

The workflow runs are sitting at `action_required` and need a maintainer to approve them.

davidkna-sap (approved):

davidkna-sap:
Thanks for the contribution @IncognitoQuack!

packages/connectivity/src/scp-cf/jwt/jwt.ts
```diff
@@ -83,7 +83,12 @@ export function isIasToken(decodedJwt: JwtPayload): boolean {
     return false;
   }
   try {
-    const issUrl = new URL(decodedJwt.iss);
+    // Identity providers can be configured to issue tokens with a schemeless `iss` claim,
+    // e.g. `tenant.accounts.ondemand.com` instead of `https://tenant.accounts.ondemand.com`.
+    const issWithScheme = URL.canParse(decodedJwt.iss)
```
mwdb: iss can be in a custom domain, which can cause these issues:
1) Untrusted certificate as this is a customer managed certificate
2) SAP does not own the DNS record when it is using a custom domain.

Solution: check if ias_iss is available (and fallback to iss when not). ias_iss is guaranteed to have https:// and is guaranteed to be using an SAP controlled domain.

